PR #4, opened 2021-06-02
examples/simple: fix constructor and run method

FFY00:
Signed-off-by: Filipe Laíns <lains@riseup.net>


willmcgugan:
Thanks, but this is premature. Code is such an early stage, it will change a lot between updates.

I should have working examples next update!

PR #5, opened 2021-06-12
GitHub Action to lint Python code

cclauss:
Output: https://github.com/cclauss/textual/actions

Future PR could:
* [ ] remove `--ignore=F401,F811,F841` from flake8
* [ ] set flake8 --max-complexity to 10, not 16
* [ ] set flake8 --max-line-length to 88, not 120
* [ ] remove `|| true` from isort, mypy, pytest, pyupgrade


willmcgugan:
Not keen on all these linters. Maybe when Textual is a bit more mature, for now I suspect the signal to noise will be low. Thanks anyway.

cclauss:
The project is in great shape and already makes little “noise” on these tests.

PR #36, opened 2021-07-28
Simple exception handling for code_viewer.py

ethixkr:
...so the program doesn't exit from a misclick. Just having fun 😄


willmcgugan:
I've done it slightly differently in this branch.

PR #43, opened 2021-08-07
Fix typos and grammar in README.md

NoblySP:


README.md
```diff
@@ -112,7 +112,7 @@ The following line is similar:
 await self.view.dock(Placeholder(), Placeholder(), edge="top")
 ```
 
-You will notice that this time we are docking _two_ Placeholder objects on _the_ top edge. We haven't set an explicit size this time so Textual will divide the remaining size amongst the two new widgets.
+You will notice that this time we are docking _two_ Placeholder objects on to the `"top"` edge. We haven't set an explicit size this time so Textual will divide the remaining size amongst the two new widgets.
```
falkben, line 115:
```suggestion
You will notice that this time we are docking _two_ Placeholder objects onto the `"top"` edge. We haven't set an explicit size this time so Textual will divide the remaining size amongst the two new widgets.
```

i think slightly improved with `onto`

I don't see your change as strictly necessary here, though.

willmcgugan (commented):
Thanks!

PR #44, opened 2021-08-07
Splash image - 3D perspective

ewels:
Let me know if you'd like the Photoshop `.psd` file here in the repo as well. That's `13.6 MB` so thought I'd leave it out in case you don't want the bloat.

textual

PSD file: [textual.psd.zip](https://github.com/willmcgugan/textual/files/6949119/textual.psd.zip)


willmcgugan:
Thanks, Phil! After you prompting me I've just about figured out how to do the perspective thing. Going to go with an image I'm working on.

ewels:
Even better! 👏

PR #54, opened 2021-08-11
Tabbed Widget

timfi:
### What?
This PR adds a widget that enables the user to define multiple views as so called tabs and receive a automagically generated bar of tab handles to switch between said views.

### 🖼 Some Pictures
The example code introduced in this PR produces the following:
Screenshot 2021-08-11 at 19 09 04
Screenshot 2021-08-11 at 19 09 08
Screenshot 2021-08-11 at 19 09 11

### Some additional info
- All _4_ modes a tab handle can be in are "user-stylable". Those being default, hover, selected, and hover+selected.
- The type of view a tab uses can be given by the user upon tab creation. Though this is not strictly necessary thanks to `DockView.dock_grid()`.

### DIFF
--- a/src/textual/widgets/_tabs.py
+++ b/src/textual/widgets/_tabs.py
@@ -1,0 +1,165 @@
+from __future__ import annotations
+
+from dataclasses import dataclass, field, InitVar
+from logging import getLogger
+from typing import Generic, Literal, TypeVar, TypedDict, cast
+
+from rich.console import RenderableType
+from rich.style import StyleType
+import rich.repr
+
+from ._button import ButtonRenderable
+from .. import events, views
+from ..widget import Reactive, Widget
+from ..view import View
+
+log = getLogger("rich")
+
+ViewType = TypeVar("ViewType", bound=View)
+
+
+class HandleStyle(TypedDict):
+    default: StyleType
+    default_hover: StyleType
+    selected: StyleType
+    selected_hover: StyleType
+
+
+StyleMode = Literal["default", "default_hover", "selected", "selected_hover"]
+
+
+class TabHandle(Widget):
+    def __init__(
+        self,
+        label: str,
+        name: str | None = None,
+        styles: HandleStyle | None = None,
+    ) -> None:
+        self.label = label
+        self.styles: HandleStyle = styles or cast(
+            HandleStyle,
+            {
+                "default": "default",
+                "default_hover": "bold",
+                "selected": "reverse",
+                "selected_hover": "bold reverse",
+            },
+        )
+        super().__init__(name=name)
+
+    @property
+    def container(self) -> Tabs:
+        if not hasattr(self, "_container"):
+            raise RuntimeError("Tab handle must be bound to container before use.")
+        return self._container
+
+    @container.setter
+    def container(self, new: Tabs):
+        if hasattr(self, "_container"):
+            raise RuntimeError("Can only bind tab handle to one container.")
+        self._container = new
+
+    selected: Reactive[bool] = Reactive(False)
+    current_style: Reactive[StyleMode] = Reactive("default")
+
+    def render(self) -> RenderableType:
+        return ButtonRenderable(self.label, style=self.styles[self.current_style])
+
+    async def on_click(self, event: events.Click) -> None:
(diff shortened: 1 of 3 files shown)


timfi:
@willmcgugan I noticed that the scrollbar in the examples doesn't properly extend down the whole view. For verification I chucked a placeholder in there and it also only extended down to half the size of the available space in the tab. I am reasonably certain that setting a `max_size` on a single row in a `GridView`doesn't actually tell all other rows the proper size that they have to share between each other. Is this intended?

willmcgugan:
That looks great! Impressed you managed to figure so much out without docs. Will check it out soon.

The scrollbars should resize automatically when you call `ScrollView.update`. You might want to check that the ScrollView is receiving the `_window_view.WindowChange` message.

src/textual/widgets/_tabs.py
```diff
@@ -137,7 +137,7 @@ async def on_mount(self, event: events.Mount) -> None:
         max_column = len(self._tabs)
         grid = await self.dock_grid()
         grid.add_column("col", repeat=max_column)
-        grid.add_row("bar", max_size=3)
```
timfi: @willmcgugan The weird siszing came from me only having set a `max_size` it disappeared as soon as I set a proper `size` instead.

timfi: Screenshot 2021-08-12 at 10 03 19

This is what it looked like with a `Placeholder` widget btw.

timfi:
@willmcgugan Having a look the action system I believe it may be possible (and maybe even cleaner) to facilitate tab switching via that. As of now the `TabHandles` require a reference to the `Tabs` object, which feels kinda clunky.

timfi:
@willmcgugan should I upgrade this PR to use the new `invoke` magic? Or do you want to keep this around for future references as is.

I noticed you added this to your project board with some comments on scrolling tab handles in cases there is to little space. I initially thought about the exact same thing. The problem I found when implementing this (maybe I just didn't see how to do it properly) was that if I use a `rich.panel.Panel` object for rendering its size fit to the available space _in the terminal window_ at render time. Making it entirely useless to wrap the handles in a `textual.widgets.ScrollView` since that just passes the "actual terminal size" along instead of reporting an "unlimited size".

tusharsadhwani:
@timfi if you do, can you keep a branch (or a commit SHA) around with the older version, I'd like to read up on what the changes are

timfi:
> @timfi if you do, can you keep a branch (or a commit SHA) around with the older version, I'd like to read up on what the changes are

Sure thing! I'll leave the current branch head here as a marker: f91db6e4686200b1cb073f92916fbd4b19fa2860.
If I get around to apply the changes you can simply diff between the two commits. Or better yet I'll add the commit range in question to this comment (or a new one) later on (if I remember to do so...).

tusharsadhwani:
cool

willmcgugan:
@timfi I appreciate your work here but I have plans around a tabs which will integrate with a CSS system I'm working on. Which means that I wouldn't want merge this PR as is.

I'm afraid this is always the risk with the project in such an early stage. You really grok how to build Textual widgets, so I hope you'll keep contributing.

Maybe factor this out in to a library of its own? So others can use a tabbed widget in the meantime? It might even turn in to something that offers more than the builtin version.

timfi:
> I appreciate your work here but I have plans around a tabs which will integrate with a CSS system I'm working on. Which means that I wouldn't want merge this PR as is.

> I'm afraid this is always the risk with the project in such an early stage.

No problemo! 😄 I expected as much after skimming your project board. 😉
If you ever need help with any busy work or want to bounce some ideas I'd be glad to be of service. 👌  

> Maybe factor this out in to a library of its own? So others can use a tabbed widget in the meantime? It might even turn in to something that offers more than the builtin version.

Sounds like a good idea, I'll see if I can get around to it. 👍

krissemicolon:
When will it be merged?

tusharsadhwani:
@krissemicolon this PR, probably never. But a different method of creating tabs will be added.

willmcgugan:
Looks cool! We have a tabbed widget on the roadmap.

PR #60, opened 2021-08-13
Invoke

willmcgugan:
Adds boilerplate-free callbacks


timfi (changes requested):
This looks very nice! 👌 The only thing that also needs changing as well is the readme. 😉

PR #87, opened 2021-09-04
Documentation and readability suggestions for geometry.py

AlexWaygood:
- Added docstrings.
- Extended & revised existing docstrings.
- Some suggested modifications to type-annotations, mostly for readability reasons.

Hope this was what you were looking for — harsh feedback welcome!

fixes https://github.com/willmcgugan/textual/issues/83

### DIFF
--- a/src/textual/geometry.py
+++ b/src/textual/geometry.py
@@ -3,20 +3,19 @@ from __future__ import annotations
 from typing import Any, NamedTuple, TypeVar, Iterable
 
 
-Number = TypeVar("Number", int, float)
-TwoIntTuple = tuple[int, int]
+T = TypeVar("T", int, float)
 
 
-def clamp(value: Number, minimum: Number, maximum: Number) -> Number:
+def clamp(value: T, minimum: T, maximum: T) -> T:
     """Clamps a value between two other values.
 
     Args:
-        value (Number): A value
-        minimum (Number): Minimum value
-        maximum (Number): maximum value
+        value (T): A value
+        minimum (T): Minimum value
+        maximum (T): maximum value
 
     Returns:
-        Number: New value that is not less than the minimum or greater than the maximum.
+        T: New value that is not less than the minimum or greater than the maximum.
     """
     if minimum > maximum:
         maximum, minimum = minimum, maximum
@@ -36,7 +35,7 @@ class Offset(NamedTuple):
 
     @property
     def is_origin(self) -> bool:
-        """Return True if the point is at the origin (0, 0), else False"""
+        """Check if the point is at the origin (0, 0)."""
         return self == (0, 0)
 
     def __add__(self, other: object) -> Offset:
@@ -75,7 +74,7 @@ class Size(NamedTuple):
     height: int
 
     def __bool__(self) -> bool:
-        """Return True if both width and height are nonzero, else False"""
+        """Return True if both width and height are nonzero."""
         return self.width * self.height != 0
 
     @property
@@ -110,7 +109,7 @@ class Size(NamedTuple):
         width, height = self
         return width > x >= 0 and height > y >= 0
 
-    def contains_point(self, point: TwoIntTuple) -> bool:
+    def contains_point(self, point: tuple[int, int]) -> bool:
         """Check if a point is in the size.
 
         Args:
@@ -142,31 +141,6 @@ class Size(NamedTuple):
         return width > x >= 0 and height > y >= 0
 
     
-C = TypeVar("C", bound="Corners")
-
-
-class Corners(NamedTuple):
-    """A tuple describing the four corners of a region"""
-    
-    x_min: int
-    y_min: int
-    x_max: int
-    y_max: int
-        
-    @classmethod
-    def of_region(cls: type[C], region: Region) -> C:
-        """Get the corners of a region.
-        
-        Args:
-            region (Region): a region to find the corners of.
-            
-        Returns:
-            Corners: The corners of the region.
-        """
-        x, y, width, height = region
-        return cls(x, y, (x + width), (y + height))
-
-    
 R = TypeVar("R", bound="Region")
     
 
@@ -180,7 +154,7 @@ class Region(NamedTuple):
 
     @classmethod
     def from_corners(cls: type[R], x1: int, y1: int, x2: int, y2: int) -> R:
-        """Construct a Region form the top-left and bottom-right corners.
+        """Construct a Region from the top-left and bottom-right corners.
 
         Args:
             x1 (int): Top-left x
@@ -194,7 +168,7 @@ class Region(NamedTuple):
         return cls(x1, y1, x2 - x1, y2 - y1)
 
     @classmethod
-    def from_origin(cls: type[R], origin: TwoIntTuple, size: TwoIntTuple) -> R:
+    def from_origin(cls: type[R], origin: tuple[int, int], size: tuple[int, int]) -> R:
         """Create a new region from `origin` and `size`.
 
         Args:
@@ -210,10 +184,10 @@ class Region(NamedTuple):
 
     def __bool__(self) -> bool:
         """Return True if the width and height of the region are both nonzero."""
-        return self.width and self.height
+        return bool(self.width and self.height)
 
     @property
-    def x_extents(self) -> TwoIntTuple:
+    def x_extents(self) -> tuple[int, int]:
         """Get the starting and ending x-coordinates of the region.
 
         The end value is non-inclusive.
@@ -225,7 +199,7 @@ class Region(NamedTuple):
         return (x, x + self.width)
 
     @property
-    def y_extents(self) -> TwoIntTuple:
+    def y_extents(self) -> tuple[int, int]:
         """Get the starting and ending y-coordinates of the region.
 
         The end value is non-inclusive.
@@ -262,13 +236,14 @@ class Region(NamedTuple):
         return Size(self.width, self.height)
 
     @property
-    def corners(self) -> Corners:
+    def corners(self) -> tuple[int, int, int, int]:
         """Get the maxima and minima of the region.
 
         Returns:
-            Corners: A tuple of (<min x>, <min y>, <max x>, <max y>)
+            tuple[int, int, int, int]: A tuple of (<min x>, <min y>, <max x>, <max y>)
         """
-        return Corners.of_region(self)
+        x, y, width, height = self
+        return x, y, (x + width), (y + height)
 
     @property
     def x_range(self) -> range:
@@ -337,7 +312,7 @@ class Region(NamedTuple):
         self_x, self_y, width, height = self
         return (self_x + width > x >= self_x) and (self_y + height > y >= self_y)
 
-    def contains_point(self, point: TwoIntTuple) -> bool:
+    def contains_point(self, point: tuple[int, int]) -> bool:
         """Check if a point is in the region.
 
         Args:
--- a/src/textual/geometry.py
+++ b/src/textual/geometry.py
@@ -7,7 +7,7 @@ T = TypeVar("T", int, float)
 
 
 def clamp(value: T, minimum: T, maximum: T) -> T:
-    """Clamps a value between two other values.
+    """Clamp a value between two other values.
 
     Args:
         value (T): A value
@@ -91,7 +91,7 @@ class Size(NamedTuple):
         """Get a region of the same size.
         
         Returns:
-            Region: a `Region` object of the same size, with origin (0, 0).
+            Region: a `Region` instance of the same size, with origin (0, 0).
         """
         width, height = self
         return Region(0, 0, width, height)
@@ -122,15 +122,18 @@ class Size(NamedTuple):
         width, height = self
         return width > x >= 0 and height > y >= 0
 
-    def __contains__(self, other: Iterable[int]) -> bool:
+    def __contains__(self, other: Iterable[int]) -> bool:  # type: ignore[override]
         """Determine whether the dimensions of `other` are within the bounds of the size.
         
         Returns:
             bool: True if the dimensions of `other` are within the bounds of the size.
             
         Raises:
-            `TypeError` if `other` cannot be unpacked into a two-item tuple.
+            TypeError if `other` cannot be unpacked into a two-item tuple.
         """
+        # The signature of this method breaks LSP,
+        # but it is useful to keep it as it is.
+
         try:
             x, y = other
         except Exception:
@@ -140,9 +143,9 @@ class Size(NamedTuple):
         width, height = self
         return width > x >= 0 and height > y >= 0
 
-    
+
 R = TypeVar("R", bound="Region")
-    
+
 
 class Region(NamedTuple):
     """Defines a rectangular region."""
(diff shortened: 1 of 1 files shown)


src/textual/geometry.py
```diff
@@ -1,21 +1,22 @@
 from __future__ import annotations
 
-from typing import Any, NamedTuple, TypeVar
+from typing import Any, NamedTuple, TypeVar, Iterable
 
 
-T = TypeVar("T", int, float)
+Number = TypeVar("Number", int, float)
```
willmcgugan, line 6: I applaud the effort to make the code self-documenting, but while it's true these are numbers now, they might not *always* be numbers.

I'd prefer to defer that decision to later.

src/textual/geometry.py
```diff
@@ -35,7 +36,7 @@ class Offset(NamedTuple):
 
     @property
     def is_origin(self) -> bool:
-        """Check if the point is at the origin (0, 0)"""
+        """Return True if the point is at the origin (0, 0), else False"""
```
willmcgugan, line 39: I like to avoid this wording, as its verbose. Every bool return would have the tiresome phrasing "Return True if blah else False".

src/textual/geometry.py
```diff
@@ -56,11 +57,11 @@ def blend(self, destination: Offset, factor: float) -> Offset:
         """Blend (interpolate) to a new point.
 
         Args:
-            destination (Point): Point where progress is 1.0
+            destination (Offset): Offset where progress is 1.0
```
willmcgugan, line 60: Good catch. Offset was originally called Point.

src/textual/geometry.py
```diff
@@ -87,24 +89,28 @@ def area(self) -> int:
...
+            y (int): Y-coordinate (row)
 
         Returns:
             bool: True if the point is within the region.
         """
         width, height = self
         return width > x >= 0 and height > y >= 0
 
-    def contains_point(self, point: tuple[int, int]) -> bool:
+    def contains_point(self, point: TwoIntTuple) -> bool:
```
willmcgugan, line 113: Not overly keen on `TwoIntTuple` as I don't think it is more descriptive than `tuple[int, int]`.

src/textual/geometry.py
```diff
@@ -117,7 +123,15 @@ def contains_point(self, point: tuple[int, int]) -> bool:
         width, height = self
         return width > x >= 0 and height > y >= 0
 
-    def __contains__(self, other: Any) -> bool:
+    def __contains__(self, other: Iterable[int]) -> bool:
```
willmcgugan, line 126: Have you checked this with Mypy? (`make typecheck`). These methods accept any argument, which is why they are typed so broadly.

src/textual/geometry.py
```diff
@@ -127,6 +141,34 @@ def __contains__(self, other: Any) -> bool:
         width, height = self
         return width > x >= 0 and height > y >= 0
 
+    
+C = TypeVar("C", bound="Corners")
+
+
+class Corners(NamedTuple):
```
willmcgugan, line 148: This probably isn't worth doing right now. There may never be a need to refer to the contents by name. 

You'll see expressions like this in the code:

```python
x, y, x2, y2 = self.corners
```

Unpacking to locals like that is considerably faster than attribute lookup. In the case of `corners` it is exclusively used like this.

This *may* change in the future, but for now I think its best left as a 4xtuple.

src/textual/geometry.py
```diff
@@ -137,100 +179,120 @@ class Region(NamedTuple):
...
 
     @property
-    def corners(self) -> tuple[int, int, int, int]:
-        """Get the maxima and minima of region.
+    def corners(self) -> Corners:
+        """Get the maxima and minima of the region.
 
         Returns:
-            tuple[int, int, int, int]: A tuple of (<min x>, <max x>, <min y>, <max y>)
+            Corners: A tuple of (<min x>, <min y>, <max x>, <max y>)
```
willmcgugan, line 269: Good catch

willmcgugan (changes requested):

src/textual/geometry.py
AlexWaygood: Yes, MyPy correctly flags this as an LSP violation -- my bad. Should `__contains__` raise a TypeError, though, if it accepts an argument of any type? To be consistent with the type signature, perhaps it should return `False` rather than raising an exception?

willmcgugan: Good point. A container wouldn't raise TypeError for anything. But you could argue that this class isn't really a container per se, the concept of being "in" something means something different than a dict for example.

My gut feeling is that the TypeError may catch bugs, so maybe it is best left in. Unless you can think of a reason not to?

AlexWaygood: Yes, I think you're right -- strictly speaking it might break LSP, but it's probably better to leave the `TypeError` in. Would you prefer that I change the annotation back to `other: Any` (in the knowledge that it doesn't really accept any type, since there are some types for which it will raise an error), or would you prefer that I add a `#type: ignore[override]` comment to suppress MyPy errors specifically relating to the LSP on that line?

tusharsadhwani: Here's one way to please nypy and satisfy LSP:

```python
    def __contains__(self, other: object) -> bool:
        try:
            assert isinstance(other, Iterable)
            x, y = other
            assert isinstance(x, int)
            assert isinstance(y, int)
        except Exception:
            raise TypeError(
                "Dimensions.__contains__ requires an iterable of two integers"
            )
        width, height = self
        return width > x >= 0 and height > y >= 0
```

tusharsadhwani: Reducing the use of `Any` is always good

AlexWaygood: That still breaks LSP! ["New exceptions cannot be thrown by the methods in the subtype, except if they are subtypes of exceptions thrown by the methods of the supertype." ](https://en.wikipedia.org/wiki/Liskov_substitution_principle). `tuple.__contains__`, the method of the supertype that we're here overriding, would never throw `TypeError` for anything. The only difference is, because we're lying in the type annotation (we're telling MyPy that the function accepts any subtype of `object`, whereas in fact it throws an exception for anything except `Iterable[int]`), MyPy can no longer detect the LSP violation.

tusharsadhwani: Oh, right.

But we can go around that by returning `NotImplemented` instead. That will throw a `TypeError` anyway.

AlexWaygood: I think that would follow the letter of the LSP. But I'm still not sure it would follow the spirit of the LSP, given that we'd still *end up* with an exception being raised where none would be raised in the same method of the supertype. (Are there any situations where `tuple.__contains__` would return `NotImplemented`, leading to a `TypeError` being raised?) And we'd get a generic error message rather than the custom error message we get currently.

I think it's just a bit of a fool's errand tying ourselves in knots to try to conform to the LSP in this instance. (But just my opinion! 🙂)

willmcgugan: You know what, you're right. I think it's worth being pragmatic here. Let's return to an Iterable of ints, as you originally suggested. And keep that TypeError in, because it could cause errors. Silence any errors that come of it.

src/textual/geometry.py
```diff
@@ -127,6 +140,9 @@ def __contains__(self, other: Any) -> bool:
         width, height = self
         return width > x >= 0 and height > y >= 0
 
+    
+R = TypeVar("R", bound="Region")
```
willmcgugan, line 144: Is the typevar here? Is it to type `cls` in `from_corners`? I think Mypy can deduce the type without it.

src/textual/geometry.py
```diff
@@ -137,100 +153,121 @@ class Region(NamedTuple):
...
 
-        The end value is non inclusive.
+        The end value is non-inclusive.
 
         Returns:
-            tuple[int, int]: [description]
+            tuple[int, int]: A tuple describing the minimum and maximum x-coordinates.
         """
-        return (self.x, self.x + self.width)
+        x = self.x
```
willmcgugan, line 198: Why make this two lines? If it's faster, I'll need proof! :-)

src/textual/geometry.py
```diff
@@ -137,100 +153,121 @@ class Region(NamedTuple):
...
 
     @property
     def y_max(self) -> int:
+        """Get the ending (non-inclusive) y-coordinate, an integer."""
         return self.y + self.height
 
     @property
     def area(self) -> int:
-        """Get the area within the region."""
+        """Get the area within the region, an integer."""
```
willmcgugan, line 225: It's probably better to add an `Returns:` section here, and keep the initial sentence brief.

src/textual/geometry.py
```diff
@@ -137,100 +153,121 @@ class Region(NamedTuple):
...
         return self.width * self.height
 
     @property
     def origin(self) -> Offset:
-        """Get the start point of the region."""
+        """Get the start point of the region as an `Offset` NamedTuple."""
```
willmcgugan, line 230: Perhaps "get the start offset of the region."

src/textual/geometry.py
```diff
@@ -307,21 +347,29 @@ def contains_region(self, other: Region) -> bool:
...
+            y (int): Value to add to y-coordinate for the new region. Optional; defaults to 0.
 
         Returns:
-            Region: A new region shifted by x, y
+            Region: A new `Region` object of the same size, shifted by `(x, y)`.
         """
 
         self_x, self_y, width, height = self
-        return Region(self_x + x, self_y + y, width, height)
+        return Region((self_x + x), (self_y + y), width, height)
```
willmcgugan, line 361: Why braces around the clauses here?

src/textual/geometry.py
```diff
@@ -307,21 +347,29 @@ def contains_region(self, other: Region) -> bool:
...
 
     def __contains__(self, other: Any) -> bool:
-        """Check if a point is in this region."""
+        """Check if a point is in this region.
+        
+        Returns:
+            If `other` is of type `Region` or `tuple[(int | float), (int| float)]`:
```
willmcgugan, line 367: This doesn't read like English, we probably don't need the types. Why the colon, is it intended to be psuedo-code?

src/textual/geometry.py
```diff
@@ -331,17 +379,18 @@ def __contains__(self, other: Any) -> bool:
                 return False
 
     def clip(self, width: int, height: int) -> Region:
-        """Clip this region to fit within width, height.
+        """Create a new instance identical to this one, but clipped within (`width`, `height`).
```
willmcgugan, line 382: The new instance is not going to be identical if it is clipped?

src/textual/geometry.py
```diff
@@ -331,17 +379,18 @@ def __contains__(self, other: Any) -> bool:
...
 
         Args:
-            width (int): Width of bounds.
-            height (int): Height of bounds.
+            width (int): Width of the bounds for the new, clipped region.
```
willmcgugan, line 385: Bit verbose TBH. Not sure it adds much.

src/textual/geometry.py
```diff
@@ -352,7 +401,7 @@ def clip(self, width: int, height: int) -> Region:
         return new_region
 
     def intersection(self, region: Region) -> Region:
-        """Get that covers both regions.
+        """Return a new instance describing the intersection between two regions.
```
willmcgugan, line 404: I messed up the original, but the change doesn't suggest what intersection means.

src/textual/geometry.py
```diff
@@ -373,16 +422,16 @@ def intersection(self, region: Region) -> Region:
         rx2 = cx2 if x2 > cx2 else (cx1 if x2 < cx1 else x2)
         ry2 = cy2 if y2 > cy2 else (cy1 if y2 < cy1 else y2)
 
-        return Region(rx1, ry1, rx2 - rx1, ry2 - ry1)
+        return Region(rx1, ry1, (rx2 - rx1), (ry2 - ry1))
```
willmcgugan, line 425: More braces! Do you find these clearer or something?

Did you run black on this?

willmcgugan (changes requested):

src/textual/geometry.py
AlexWaygood: Timings appear to indicate a tiny speed increase, but barely noticeable (weirdly, it seems to be a bigger speed difference when done with fewer repeats -- but I'm guessing that's just noise). I'll revert the change -- sorry for wasting your time.
[geometry.Region.x_extents timings.txt](https://github.com/willmcgugan/textual/files/7243798/geometry.Region.x_extents.timings.txt)

src/textual/geometry.py
AlexWaygood: I added the `R` TypeVar to provide better return-value type hints for `from_corners` and `from_origin`.  In my opinion, it's a bit of an anti-pattern to hard-code the return-type in `classmethod`s that act as alternative constructor-methods. If you ever decide to subclass `Region`, the runtime implementation will still be perfectly valid, but the type hint will break.

For example, in [this gist](https://mypy-play.net/?mypy=latest&python=3.10&flags=show-error-codes%2Cstrict&gist=73e1d45899d7356dfe02578925295d0a), MyPy correctly infers that the type of `region_with_type_var_subclass_instance` is `RegionWithTypeVarSubclass`, but incorrectly assumes that the type of `region_no_type_var_subclass_instance` is `RegionNoTypeVar` (it is in fact `RegionNoTypeVarSubclass`).

I'm obviously happy to revert it if you don't think it's necessary, however!

src/textual/geometry.py
AlexWaygood: Yeah, I do personally find them more readable, but Black takes them right out again. Sorry for wasting your time.

src/textual/geometry.py
```diff
@@ -307,21 +370,28 @@ def contains_region(self, other: Region) -> bool:
...
 
         self_x, self_y, width, height = self
-        return Region(self_x + x, self_y + y, width, height)
+        return Region((self_x + x), (self_y + y), width, height)
 
     def __contains__(self, other: Any) -> bool:
-        """Check if a point is in this region."""
+        """Check if a point is in this region.
+        
+        The return value is always False unless `other`
```
AlexWaygood, line 389: @willmcgugan what do you think of this revised wording? I feel like I'm having trouble describing the relationship between the input type and the return type in the docstring. I tried adding some `overload`s but couldn't find a way to make MyPy happy about them.

src/textual/geometry.py
```diff
@@ -331,17 +401,18 @@ def __contains__(self, other: Any) -> bool:
                 return False
 
     def clip(self, width: int, height: int) -> Region:
-        """Clip this region to fit within width, height.
+        """Return a new region clipped within (width, height) bounds.
```
AlexWaygood, line 404: Wording here altered here, as per request in code review.

src/textual/geometry.py
```diff
@@ -352,13 +423,16 @@ def clip(self, width: int, height: int) -> Region:
         return new_region
 
     def intersection(self, region: Region) -> Region:
-        """Get that covers both regions.
+        """Find the overlap between two regions.
```
AlexWaygood, line 426: Docstring rewritten, as per request in code review.

PR #86, opened 2021-09-04
Add more easing functions

tusharsadhwani:
-- not entirely polished yet, have tried out around 10 of them in animations.py

Resolves #81

PR #88, opened 2021-09-04
Add remaining easing functions

### LINKED ISSUE #81: Add more easing functions
There are a number of 'easing functions' used by the animation system. At the top of _animator.py you will see the following:

```python
EASING = {
    "none": lambda x: 1.0,
    "round": lambda x: 0.0 if x < 0.5 else 1.0,
    "linear": lambda x: x,
    "in_cubic": lambda x: x * x * x,
    "in_out_cubic": lambda x: 4 * x * x * x if x < 0.5 else 1 - pow(-2 * x + 2, 3) / 2,
    "out_cubic": lambda x: 1 - pow(1 - x, 3),
}
```

These function have been copied from https://easings.net/ and translated from Javascript to Python. I would like to add the remaining functions to the EASING dict above.

See the animation.py example as a way of testing the above functions.

RojerGS:
Fixes #81 , does the same work as #86 ; I chose to open the PR either way because the work had already been done.


RojerGS:
@willmcgugan what do you think of putting all the easing functions in a separate file `_easings.py`? All of a sudden, the `_animator.py` grew by quite a bit with these helper functions.

RojerGS:
(For reference, Textual “kind of works” with Windows Subsystem for Linux (WSL) on Windows, but not very well.)

I was also thinking of adding a couple of tests to the easing functions, probably sampling the range [0, 1] in some 20 points, computing the original easing functions at those points, and then comparing with whatever is implemented.

src/textual/_animator.py
```diff
@@ -27,14 +28,128 @@ def blend(self: T, destination: T, factor: float) -> T:
         ...
 
 
+def _in_out_expo(x):
```
willmcgugan, line 31: Could you type these? Seems redundant, but at some point I will switch on strict mode in Mypy and typing will be required.

willmcgugan (changes requested):

willmcgugan:
An _easing.py file makes sense. Tests would be good. Although getting data from all these functions sounds like a lot of effort. If you don't mind doing that, fair enough, but I would be happy with a test that calls every function and checks that it goes from 0 -> 1.

Good to know that it kinda runs under WSL.

tusharsadhwani:
An issue I noticed on WSL was that keyboard/mouse actions were being printed to the console, causing the whole animation to jump 1 row up/down continuously

src/textual/_animator.py
RojerGS: Of course. I would've guess I would be smart enough to look around, see everything is typed, and type those by myself... Turns out I'm not 🙃.

RojerGS:
Added tests to the functions via a TypeScript script where I pasted all the original definitions.

The script was ran online over at TIO.run, and this is the [direct link](https://tio.run/##vVhRb5swEH4Ov8KPJHUoEEjSptGkTX3ow7ROe5z2kFKyoqWQBdgyTf3t2Z2NHRswsGrapHXXw/583539@bzi1z7Oo0OyL06nbZlGRZKlJN7k8V36KUlj@3hN0vL5IT6MhUF@W6NDXJSHlHhkSqIst@0jmZD7uzG5JP54Zb1Ylgb1oSz6sfIk7cO5SwchTW2M6f4OsI5jiNBjeAyujvex3Dx2g2FEx1ZK/XMxPTb@gDAmwjLQGhLJDXGdkLwhPsbE/l6zNfbZT3vKnRfEp5A7I@F35UMSDWBsZD0AQITE@FIyM1EeFIvgHEjOZt6zzkIfioG8uyreB1LjHnSUuz8ewX2pcTfzD7r4J@nf8O/IQS9QLQehOQcDYhI58OaNJJgTEZoTcXvcZ31rrtdr4sKaLsAjNCB6LkMHMXHbBW4orodclLABe8rBTWkaHLE1Ihg0/nMtVuM@T/hENquVfYUVZowPs30ltmrIBQ7pkJXkMEAU8u@w5wXxI@pUey770XQkJGBA40LTi1dlhmfLbsTKthZbQU2TNkTX3zEmzJyut5vomyGgKEvzgkQegeI5C9cLlyvpnIETvgD0ypKhg1c9EFMc0XWA/@Xa@DtbXynDDLmzGLTaGErzymh8Hs0EvoR@qESkF1KrHV7BNsxklZlUlyiky9eKKqfAJ79lGmtR8ID4jCf72XHf3u42eWG85So2AbBhi77fFE8O739mGqlh53vaIlZgyc4K/OhzFiG6o8CgZP8z5JoENqJVguV7r10ih4QctoQcOOErgtY22dRuE9OKx9nnOZ7PmYTN/VbT2QGzq4PYpi0Z/D6gbxdywIebetPziK7cpnhIF044x7NY@R7R50P5MMHJFg/ODSx8CR/GOJPAnyqcVFGs0QuJd3lMxAzfPAPFBbUhrIbw/LlsyQaMHGUE8h2/jnQ1U7E6Zs6bU5dBNdkgfAPK1LyVmkVjWqXtKBRlfZzQM0@9wXmVMLAcCvXZGp0ffVR9tVH96UWVlxNVn0JUf8tQAci6fKo9H2jtDUDVJp1q3TattcwSljWPVOtKaa21pErLR9U@jerN1TlU6BOo2oRQvYegyvVN1fuU6heaBKx0idaklTaUi2qHl@r1o/VdY31ZifrtMyDKCug6LrVwf4IRCsuThnT50pCumTSkK5CGdJ0Nac2lIV0LaUjXUhrSdSWNyuU5LpLaxQX@hwLwYSIN5or5QA6FD0zugwf@Wmg594CWCheY1Uxo0uRUsGGRbXYgNm58km35ARgLKct2sbPLvrKvTrp5jkESRwjzY7MrY5ZnCHPEEFjqEYLXgGGM@DhnX@ZPDISPws4UhEBbgo9EyT2d/gA).
I think this script is valuable but I'm not sure where to put it.

tusharsadhwani:
@RojerGS nice idea, and good work!

willmcgugan:
Nice. You went the extra mile. I did wonder if there were ant floating point differences between Python and Javascript that might throw that out.

tusharsadhwani:
rounding off to 6 decimals would fix that as well :p

RojerGS:
There are cases where it is clear that the "ground truth" results I pasted were affected by floating point inaccuracies, but I decided not to tamper with those at all. I thought `pytest.approx` would be good enough to handle the comparison of the floats.

willmcgugan:
Ah. Missed the `approx`. Good call.

RojerGS:
Hey @willmcgugan, what can I do to help you merge this? I'm afraid you didn't notice that the requested changes have been solved; perhaps _I_ shouldn't have marked the conversation as resolved myself?

willmcgugan:
Looks good! Will merge soon.

willmcgugan:
Do you want to put that script in a top level dir. Call it “reference” maybe.

tusharsadhwani:
There's a `tools/` dir in rich, maybe keep it the same?

RojerGS:
I liked the idea of having the `tools/` directory, parallel to what happens in `rich`, so I went ahead and added the script in `tools/gen_easings_tests.ts`.

willmcgugan:
Great, thanks @RojerGS @tusharsadhwani

RojerGS:
My pleasure.

PR #89, opened 2021-09-04
Add execute_in_thread method to MessagePump

### LINKED ISSUE #85: Method to run code in a thread
I'd like to add a convenient method to execute in job in a thread.

It would *probably* be added to message pump (used by app / widgets).

It should allow a method to be called with arbitrary parameters, and should return an awaitable.

```python
await self.call_threaded(my_code, "foo", "bar")
```

I can see this being used for jobs such as reading from disk (which at the moment blocks the vent loop).

Some considerations: a thread pool may be appropriate. We don't want this to be unbounded and potentially launch an unlimited number of threads. It should probably work with both async and sync functions, which will require a little inspection. If you want to tackle this, please discuss your ideas here first.

adriangb:
Closes #85


src/textual/message_pump.py
```diff
@@ -29,6 +32,32 @@ class MessagePumpClosed(Exception):
...
+T = TypeVar("T")
+
+
+def _is_coroutine_callable(call: Callable) -> bool:
+    if inspect.isroutine(call):
+        return inspect.iscoroutinefunction(call)
+    if inspect.isclass(call):
+        return False
+    call = getattr(call, "__call__", None)
+    return inspect.iscoroutinefunction(call)
```
adriangb, line 44: I haven't added tests for the cases of a callable class and such. But this is [ripped from FastAPI](https://github.com/tiangolo/fastapi/blob/1760da0efa55585c19835d81afa8ca386036c325/fastapi/dependencies/utils.py#L426-L432) so I'm fairly certain we can just go with it for now

src/textual/message_pump.py
```diff
@@ -29,6 +32,32 @@ class MessagePumpClosed(Exception):
...
+
+
+def _callable_in_thread_pool(call: Callable[..., T]) -> Callable[..., Awaitable[T]]:
+    async def inner(*args: Any, **kwargs: Any) -> T:
+        print(call)
+        loop = asyncio.get_event_loop()
+        # Ensure we run in the same context
+        child = functools.partial(call, *args, **kwargs)
+        context = contextvars.copy_context()
+        func = context.run
```
adriangb, line 54: I also haven't (yet) added a test for context copying

src/textual/message_pump.py
```diff
@@ -29,6 +32,32 @@ class MessagePumpClosed(Exception):
     pass
 
 
+T = TypeVar("T")
+
+
+def _is_coroutine_callable(call: Callable) -> bool:
```
adriangb, line 38: @willmcgugan I'm curious what your thoughts are on re-using these methods to allow users to define their `on_key`, etc. methods as sync or async, sorta how Starlette lets you define endpoints as sync or async and automatically runs sync endpoints in a threadpool? It would just be a couple lines in `_callback.py` I think.

tests/test_app.py
```diff
@@ -0,0 +1,49 @@
+import asyncio
+import time
+
+import pytest
+
+from textual.app import App
+
+
+@pytest.mark.asyncio
+async def test_execute_in_thread():
```
adriangb, line 10: I didn't see any tests applied to App or MessagePump directly, so I wasn't sure what the best way to write them is, this is just what I came up with but I'm open to better options

adriangb (commented):
Quick review of the implementation

src/textual/message_pump.py
adriangb: It would be good thought to only inspect the method once, then wrap it and store the wrapped version. This might require a mapping somewhere of `{user_method: possibly_wrapped_method}`.

tests/test_app.py
adriangb: Changed this to test `MessagePump` directly now since it's a public class anyways

src/textual/message_pump.py
adriangb: Added 😄

src/textual/message_pump.py
adriangb: Added a test!

src/textual/message_pump.py
```diff
@@ -29,6 +32,32 @@ class MessagePumpClosed(Exception):
...
+def _callable_in_thread_pool(call: Callable[..., T]) -> Callable[..., Awaitable[T]]:
+    async def inner(*args: Any, **kwargs: Any) -> T:
+        print(call)
+        loop = asyncio.get_event_loop()
+        # Ensure we run in the same context
+        child = functools.partial(call, *args, **kwargs)
+        context = contextvars.copy_context()
+        func = context.run
+        args = (child,)
+        return await loop.run_in_executor(None, func, *args)
```
ksamuel, line 56: Given `run_in_executor` schedule the call and returns an awaitable, there seem to be no need to `await` it if you don't make `inner` and async function. 

This would save a round trip to the event loop

src/textual/message_pump.py
```diff
@@ -29,6 +32,32 @@ class MessagePumpClosed(Exception):
...
+        return inspect.iscoroutinefunction(call)
+    if inspect.isclass(call):
+        return False
+    call = getattr(call, "__call__", None)
+    return inspect.iscoroutinefunction(call)
+
+
+def _callable_in_thread_pool(call: Callable[..., T]) -> Callable[..., Awaitable[T]]:
+    async def inner(*args: Any, **kwargs: Any) -> T:
+        print(call)
```
ksamuel, line 49: Is `print(call)` voluntary? It seems an artifact of a previous debug attempt.

adriangb: yep, mistake, will remove

src/textual/message_pump.py
ksamuel: `*args` is already passed line 52, isn't it redoundant to also pass it line 56? Granted with `partial()` it's not going to matter much, but it shouldn't bring anything either.

src/textual/message_pump.py
adriangb: Removed. Thanks.

src/textual/message_pump.py
adriangb: Not sure I fully understood the latter questions, but I did a bit of simplifying in 9d541a7bcf1389ffec2a9b5e2c0da5a8834f4c77, hopefully that resolves it.

adriangb: > Given `run_in_executor` schedule the call and returns an awaitable, there seem to be no need to `await` it if you don't make `inner` and async function.

Hmm yes, that's a good point! I feel like there might be some cons of returning a regular callable instead of a coroutine function, but I can't immediately think of them, so I'm pushing your suggestion 😄

willmcgugan:
Thanks for this. Just to let you know I haven't forgotten about it, and will review soon.

adriangb:
> Thanks for this. Just to let you know I haven't forgotten about it, and will review soon.

Thanks I appreciate the update 😃

PR #92, opened 2021-09-05
Parser Polishing

### LINKED ISSUE #84: Polish parser
The [parser class](https://github.com/willmcgugan/textual/blob/main/src/textual/_parser.py) parses stdin to extract mouse codes etc. This code could use docstrings and some tests. It could also use a few tests to check edge cases. I don't know of any bugs, but you may come across some when testing.

chelnak:
Closes #84

### DIFF
--- a/src/textual/_parser.py
+++ b/src/textual/_parser.py
@@ -157,7 +157,6 @@ class Parser(Generic[T]):
 
     def parse(self, on_token: Callable[[T], None]) -> Generator[Awaitable, str, None]:
         return
-        yield
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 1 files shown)


chelnak:
Initial PR contains the first pass for the docstrings. I'll start building some tests now that these are in for review.

@willmcgugan What is your motivation behind the test classes in `__main__`? Are these to prove the function of the class or just for an example of how they can be used?

willmcgugan:
@chelnak Code in main is for testing new features, but they will evolve in to demostrations in the future. Almost every module in Rich works this way.

PR #102, opened 2021-09-13
Typo in README?

lllama:
Not sure if this is just your muscle memory typing Reactive rather than React. Obviously JavaScript being JavaScript, there is a library called Reactive but I'm guessing that you mean React. If so, then this comment will be 70 times longer than the actual changes I have proposed.


willmcgugan:
Well spotted. Although I wouldn’t be surprised if there is a JS framework called Reactive out by the next release.

PR #124, opened 2021-09-29
Fix simple.py example when run from another directory

The-Compiler:
Otherwise, when e.g. run via `python examples/simple.py`, it fails with `FileNotFoundError: [Errno 2] No such file or directory: 'richreadme.md'`.

I opted to use pathlib for this - let me know if you prefer keeping the filename as a string and using `os.path` instead.


willmcgugan:

PR #126, opened 2021-09-30
Remove unreachable code

sanders41:
Because there is a `return` preceding the `yield` it isn't possible to reach the `yield`

### DIFF
--- a/src/textual/events.py
+++ b/src/textual/events.py
@@ -21,7 +21,6 @@ if TYPE_CHECKING:
 class Event(Message):
     def __rich_repr__(self) -> rich.repr.Result:
         return
-        yield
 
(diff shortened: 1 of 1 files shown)


willmcgugan:
That needs to be there. Can you think why?

tusharsadhwani:
I suppose because it's meant to be a generator, but it should also actually yield no data, simply close at first go?

willmcgugan:
@tusharsadhwani Bingo. @sanders41  is correct that the `yield` never gets executed, but its presence makes the method a generator.

tusharsadhwani:
would `yield from ()` be a better solution?

willmcgugan:
It might! This pattern pre-dates `yield from` so I never considered it.

tusharsadhwani:
okay I'll test it out

PR #128, opened 2021-10-01
Replace empty generator with empty yield from

tusharsadhwani:


willmcgugan:
@tusharsadhwani is the `yield from` trick just for linters?

tusharsadhwani:
It was aimed as a clearer way to imply it's an empty generator. The current one seemed confusing in general, having code after a return.

PR #131, opened 2021-10-01
Fix unclosed loop error

### LINKED ISSUE #127: Unclosed event loop with easing example
Every time I close the [easing example](https://github.com/willmcgugan/textual/blob/main/examples/easing.py). I get the following error:

```sh
/usr/lib/python3.9/asyncio/base_events.py:681: ResourceWarning: unclosed event loop <_UnixSelectorEventLoop running=False closed=False debug=False>
  _warn(f"unclosed event loop {self!r}", ResourceWarning, source=self)
ResourceWarning: Enable tracemalloc to get the object allocation traceback
```
I am thinking there are still messages processing at shutdown that are keeping the loop running, but I have not been able to track down the root cause yet. 

Possibly a similar issue to #82?

OS: Manjaro Linux 21.1.4

sanders41:
Closes #127

Based on the conference talk @lllama provided I tried this and it cleared up the error. I tried several of the examples with this change in place and they all still worked and closed without error.

@willmcgugan not knowing the architecture of Textual as well as you, can you think of a time that the app would not be able to get a running loop and therefore error because of that?

### DIFF
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -185,7 +185,11 @@ class App(MessagePump):
             app = cls(console=console, screen=screen, driver_class=driver, **kwargs)
             await app.process_messages()
 
-        asyncio.run(run_app())
+        loop = asyncio.get_event_loop()
+        try:
+            asyncio.run(run_app())
+        finally:
+            loop.close()
 
     async def push_view(self, view: ViewType) -> ViewType:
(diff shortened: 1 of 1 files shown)


suraj-git-tech:
Can I also contribute

willmcgugan:

PR #143, opened 2021-10-13
Allow binding of the space key

pvmm:
Allow the space key to be bound to the application.


pvmm:
#410 replaces this fix

PR #149, opened 2021-10-18
Added some information about grid making

Shirobachi:


shirobachi:
that was useful, thank me from the past!

PR #201, opened 2022-01-10
Hover pseudo class

willmcgugan:
Implements psuedo_classes. i.e. allows for `:hover` in css.

### DIFF
--- a/src/textual/css/model.py
+++ b/src/textual/css/model.py
@@ -65,6 +65,11 @@ class Selector:
         }
 
     def _add_pseudo_class(self, pseudo_class: str) -> None:
+        """Adds a pseudo class and updates specificity.
+
+        Args:
+            pseudo_class (str): Name of pseudo class.
+        """
         self.pseudo_classes.append(pseudo_class)
         specificity1, specificity2, specificity3 = self.specificity
         self.specificity = (specificity1, specificity2 + 1, specificity3)
--- a/src/textual/css/model.py
+++ b/src/textual/css/model.py
@@ -64,6 +64,11 @@ class Selector:
             SelectorType.ID: self._check_id,
         }
 
+    def _add_pseudo_class(self, pseudo_class: str) -> None:
+        self.pseudo_classes.append(pseudo_class)
+        specificity1, specificity2, specificity3 = self.specificity
+        self.specificity = (specificity1, specificity2 + 1, specificity3)
+
     def check(self, node: DOMNode) -> bool:
         return self._checks[self.type](node)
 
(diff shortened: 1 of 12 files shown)


src/textual/css/model.py
```diff
@@ -45,15 +45,15 @@ class Selector:
 
     @property
     def css(self) -> str:
-        psuedo_suffix = "".join(f":{name}" for name in self.pseudo_classes)
+        pseudo_suffix = "".join(f":{name}" for name in self.pseudo_classes)
```
willmcgugan, line 48: I made this typo multiple times and did a search and replace.

src/textual/css/stylesheet.py
```diff
@@ -124,10 +124,17 @@ def apply(self, node: DOMNode) -> None:
 
         get_first_item = itemgetter(0)
 
+        log("")
+        log(node)
+        for key, attributes in rule_attributes.items():
```
darrenburns, line 129: Should we maybe add some more context as to what these log messages mean?

src/textual/messages.py
```diff
@@ -41,3 +41,12 @@ class CursorMove(Message):
     def __init__(self, sender: MessagePump, line: int) -> None:
         self.line = line
         super().__init__(sender)
+
+
+@rich.repr.auto
+class RefreshStyles(Message):
```
darrenburns, line 47: One thing I've seen done in message passing systems is to name messages after the "thing that happened" (e.g. `StyleClassesUpdated`), rather than having the sender say what they *want* to happen (`RefreshStyles`). This means a consumer isn't restricted to doing just what they're asked by the sender - they can handle an event however they wish.

At a conceptual level, when you send a message or event, are you saying "here's what happened", or are you saying "whoever is listening, please perform some operation"? I think it's nice to agree on these conventions early.

src/textual/widget.py
```diff
@@ -92,11 +93,21 @@ def __rich_repr__(self) -> rich.repr.Result:
...
 
     def __rich__(self) -> RenderableType:
         renderable = self.render_styled()
         return renderable
 
+    def get_pseudo_classes(self) -> Iterable[str]:
+        """Pseudo classes for a widget"""
+        if self._mouse_over:
+            yield "hover"
+        # TODO: focus
```
darrenburns, line 109: Should this TODO be deleted or fulfilled as part of this PR?

src/textual/css/stylesheet.py
willmcgugan: Those are purely for debugging. Forgive those for the moment. Will clear them in a future update.

src/textual/widget.py
willmcgugan: Still need to implement a mechanism to handle a focus mechansim, like alt-tab. Will look at the pseudo class then.

PR #214, opened 2022-01-18
Splitting out parsing of durations into new token types, avoiding Scalar

darrenburns:


src/textual/_duration.py
```diff
@@ -0,0 +1,43 @@
+import re
+
+_match_duration = re.compile(r"^(-?\d+\.?\d*)(s|ms)$").match
+
+
+class DurationError(Exception):
+    pass
```
willmcgugan, line 7: Could you add a docstring to say what this exception indicates.

src/textual/_duration.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+class DurationParseError(DurationError):
+    pass
```
willmcgugan, line 11: Docstring por favor

src/textual/_duration.py
```diff
@@ -0,0 +1,43 @@
...
+        if unit_name == "ms":
+            duration_secs = value / 1000
+        else:
+            duration_secs = value
+    else:
+        try:
+            duration_secs = float(duration)
+        except ValueError:
+            raise DurationParseError(
+                f"'{duration}' is not a valid duration."
```
willmcgugan, line 40: Rather than surrounding duration in quotes, could you repr it with !r

Might not be applicable here, but in general it guarantees that it will format nicely.

src/textual/css/_styles_builder.py
```diff
@@ -28,6 +21,11 @@
 
 
 class StylesBuilder:
+    """
```
willmcgugan, line 24: 👍

src/textual/css/stylesheet.py
```diff
@@ -86,18 +86,18 @@ def read(self, filename: str) -> None:
                 css = css_file.read()
             path = os.path.abspath(filename)
         except Exception as error:
-            raise StylesheetError(f"unable to read {filename!r}; {error}") from None
+            raise StylesheetError(f"unable to read {filename!r};") from error
```
willmcgugan: This particular error should never render as a traceback, so I'd prefer to keep it with the semi-colon delimited errors.

willmcgugan (approved):
LGTM

PR #219, opened 2022-01-20
Invisible widgets now dont render

### LINKED ISSUE #217: Invisible widgets should allow background to be rendered
Currently when a widget is set as invisible, it draws a blank rectangle in place of the content.

It would be better to not render anything and allow any background to show through.

darrenburns:
Fixes #217

### DIFF
--- a/src/textual/layout.py
+++ b/src/textual/layout.py
@@ -232,7 +232,7 @@ class Layout(ABC):
             widget (Widget): The Widget in this layout you wish to know the Region of.
 
         Raises:
-            KeyError: If the Widget is not contained in this Layout.
+            NoWidget: If the Widget is not contained in this Layout.
 
         Returns:
             Region: The Region of the Widget.
--- a/src/textual/layout.py
+++ b/src/textual/layout.py
@@ -199,6 +199,7 @@ class Layout(ABC):
         raise NoWidget(f"No widget under screen coordinate ({x}, {y})")
 
     def get_style_at(self, x: int, y: int) -> Style:
+        """Get the Style at the given cell or Style.null()"""
         try:
             widget, region = self.get_widget_at(x, y)
         except NoWidget:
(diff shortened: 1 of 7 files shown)


src/textual/layout.py
```diff
@@ -217,6 +218,18 @@ def get_style_at(self, x: int, y: int) -> Style:
         return Style.null()
 
     def get_widget_region(self, widget: Widget) -> Region:
+        """Get the Region of a Widget contained in this Layout.
+
+        Args:
+            widget: The Widget in this layout you wish to know the Region of.
```
willmcgugan, line 224: Could you add the type `(Widget)` to the arg. I know its obvious in context, but we will have auto-docs at some point.

src/textual/layout.py
```diff
@@ -199,6 +199,7 @@ def get_widget_at(self, x: int, y: int) -> tuple[Widget, Region]:
         raise NoWidget(f"No widget under screen coordinate ({x}, {y})")
 
     def get_style_at(self, x: int, y: int) -> Style:
+        """Get the Style at the given cell or Style.null()"""
```
willmcgugan, line 202: Could you also add the full args and return here.

willmcgugan (changes requested):
Looks good. Just some requests re docstrings.

src/textual/layout.py
```diff
@@ -217,6 +226,18 @@ def get_style_at(self, x: int, y: int) -> Style:
         return Style.null()
 
     def get_widget_region(self, widget: Widget) -> Region:
+        """Get the Region of a Widget contained in this Layout.
+
+        Args:
+            widget (Widget): The Widget in this layout you wish to know the Region of.
+
+        Raises:
+            KeyError: If the Widget is not contained in this Layout.
```
willmcgugan, line 235: Should be a `NoWidget` exception

src/textual/view.py
```diff
@@ -1,26 +1,22 @@
...
-from .geometry import Size, Offset, Region
 from .reactive import Reactive, watch
-
-from .widget import Widget, Widget
-
+from .widget import Widget
 
 if TYPE_CHECKING:
-    from .app import App
+    pass
```
willmcgugan, line 19: Guess this block is redundant now.

willmcgugan (approved):
Just a couple of pedantic requests, but LGTM

willmcgugan:
👍

PR #221, opened 2022-01-20
Margin spacing is now invisible

darrenburns:


src/textual/css/_styles_builder.py
```diff
@@ -128,7 +128,7 @@ def _process_space(self, name: str, tokens: list[Token]) -> None:
         append = space.append
         for token in tokens:
             (token_name, value, _, _, location) = token
-            if token_name == "scalar":
+            if token_name in ("number", "scalar"):
```
darrenburns, line 131: Missed this in my PR yesterday which split out tokenising of number/scalar. It broke margin/padding when you specified any of the values in CSS as a number rather than scalar (i.e. no units).

src/textual/geometry.py
```diff
@@ -415,6 +415,24 @@ def clip(self, width: int, height: int) -> Region:
...
+        """Shrink a region by pushing each edge inwards.
+
+        Args:
+            margin (Spacing): Defines how many cells to shrink the Region by at each edge.
+
+        Returns:
+            Region: The new, smaller region.
+        """
+        _clamp = clamp
+        top, right, bottom, left = margin
```
willmcgugan, line 428: Could you also unpack the `self` to replace those `self.width` etc references with a global.

src/textual/view.py
```diff
@@ -114,7 +114,24 @@ def get_arrangement(self, size: Size, scroll: Offset) -> Iterable[WidgetPlacemen
         cached_size, cached_scroll, arrangement = self._cached_arrangement
         if cached_size == size and cached_scroll == scroll:
             return arrangement
-        arrangement = list(self._layout.arrange(self, size, scroll))
+
+        arrangement = []
+        placements = self._layout.arrange(self, size, scroll)
+        for placement in placements:
+            region, widget, order = placement
```
willmcgugan, line 121: Could you do this within the loop, i.e. `for region, widget, order in placements`.

src/textual/view.py
```diff
@@ -114,7 +114,24 @@ def get_arrangement(self, size: Size, scroll: Offset) -> Iterable[WidgetPlacemen
         cached_size, cached_scroll, arrangement = self._cached_arrangement
         if cached_size == size and cached_scroll == scroll:
             return arrangement
-        arrangement = list(self._layout.arrange(self, size, scroll))
+
+        arrangement = []
```
willmcgugan, line 118: I'm thinking this might simplify if we move some of the funcationality to the `WidgetPlacement` object itself.

Let's say we implement a `apply_margin` method on `WidgetPlacement` that returned either a new object or self, dependant on if there is margin or not.

Then this code could be:

```python
placements = [
    placement.apply_margin()
    for placement in self._layout.arrange(self, size, scroll)
]
```

I wouldn't be surprised if we grow other operations on the WidgetPlacement objects, and this may keep it simple.

What do you think?

willmcgugan (changes requested):
Possible simplification

src/textual/view.py
darrenburns: Sounds good to me. I've implemented this change now.

willmcgugan (approved):
LGTM

willmcgugan:
@darrenburns would you mind resolving and merging

PR #225, opened 2022-01-21
Use the `layout` as specified in CSS 

darrenburns:
We no longer create a `DockLayout()` in the code, and instead use the value specified in CSS.

I've added a new rule to the View-level CSS to add `layout: dock` so that it's used by default without you having to specify it in the CSS (and so `basic.css` works without changes).

### DIFF
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -1,49 +1,43 @@
 from __future__ import annotations
-import os
 
+import os
 import asyncio
 
-from typing import Any, Callable, ClassVar, Iterable, Type, TypeVar
 import warnings
-
-from rich.control import Control
+from typing import Any, Callable, Iterable, Type, TypeVar
 
 import rich.repr
-from rich.screen import Screen
 from rich.console import Console, RenderableType
+from rich.control import Control
 from rich.measure import Measurement
-
+from rich.screen import Screen
 from rich.traceback import Traceback
 
-
-from . import events
 from . import actions
-from .dom import DOMNode
-from ._animator import Animator
-from .binding import Bindings, NoBinding
-from .geometry import Offset, Region, Size
+from . import events
 from . import log
+from . import messages
+from ._animator import Animator
 from ._callback import invoke
 from ._context import active_app
-from .css.stylesheet import Stylesheet, StylesheetParseError, StylesheetError
 from ._event_broker import extract_handler_actions, NoHandler
+from ._linux_driver import LinuxDriver
+from ._profile import timer
+from .binding import Bindings, NoBinding
+from .css.stylesheet import Stylesheet, StylesheetParseError, StylesheetError
+from .dom import DOMNode
 from .driver import Driver
 from .file_monitor import FileMonitor
-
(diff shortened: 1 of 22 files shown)


src/textual/app.py
```diff
@@ -510,7 +503,7 @@ async def on_event(self, event: events.Event) -> None:
         # Handle input events that haven't been forwarded
         # If the event has been forwarded it may have bubbled up back to the App
         if isinstance(event, events.Mount):
-            view = View()
+            view = View(id="_root")
```
willmcgugan, line 506: Why the id of _root ?

src/textual/css/_style_properties.py
```diff
@@ -290,6 +290,39 @@ def __set__(self, obj: Styles, spacing: str | None) -> str | None:
...
+    def __set__(self, obj: Styles, layout: LayoutName | Layout):
+        """
+        Args:
+            obj (Styles): The Styles object.
+            layout (LayoutName | Layout): The layout to use. You can supply a ``LayoutName``
+                (a string literal such as ``"dock"``) or a ``Layout`` object.
+        """
+        from ..layouts.factory import get_layout, Layout  # Prevents circular import
+
+        obj.refresh(True)
```
willmcgugan, line 318: Can you set make keyword args explicity, i.e. `layout=True`

src/textual/css/_styles_builder.py
```diff
@@ -288,7 +289,16 @@ def process_layout(self, name: str, tokens: list[Token], important: bool) -> Non
...
-                self.styles._rule_layout = tokens[0].value
+                value = tokens[0].value
+                layout_name = cast(LayoutName, value)
+                try:
+                    self.styles._rule_layout = get_layout(layout_name)
+                except MissingLayout:
+                    self.error(
+                        name,
+                        tokens[0],
+                        f"invalid value for layout (received {value!r}, expected {friendly_list(LAYOUT_MAP.keys())})",
```
willmcgugan, line 300: Nice. Very much in keeping with my helpful error messages policy.

src/textual/css/stylesheet.py
```diff
@@ -108,25 +108,44 @@ def _check_rule(cls, rule: RuleSet, node: DOMNode) -> Iterable[Specificity3]:
                 yield selector_set.specificity
 
     def apply(self, node: DOMNode) -> None:
+        """
```
willmcgugan, line 111: Can you add a brief description in docstrings. Even if it seems a little redundant.

willmcgugan (approved):
Looks good. Just a few requests, and a question.

src/textual/app.py
darrenburns: Think this is just a leftover from debugging. I was printing out the node IDs from the code that walks the DOM tree and was wondering why one of them was printing `None`. Turns out it was the this view, which looks like the root of the tree.

src/textual/css/_style_properties.py
darrenburns: Done (also fixed all other occurrences in this file 😄)

PR #232, opened 2022-01-25
Wrap an open call in the simple example in a try block.

jgerrish:
The simple.py example was failing if it was called from a different directory.

This fix lets it keep working if the richreadme.md file isn't found and logs the file open exception.

You might want to instead fail with a simple error message.

Thanks for the great project.


willmcgugan:
https://github.com/Textualize/textual/wiki/Sorry-we-closed-your-issue

PR #236, opened 2022-01-27
Allow user to supply loop to App.run, ensure it closes

darrenburns:


src/textual/app.py
```diff
@@ -227,13 +221,28 @@ def run(
...
 
         async def run_app() -> None:
             app = cls(screen=screen, driver_class=driver, **kwargs)
             await app.process_messages()
 
-        asyncio.run(run_app())
+        if loop:
+            asyncio.set_event_loop(loop)
+        else:
+            try:
```
willmcgugan, line 234: I think this installs uvloop globally. Perhaps it would be better to create a instance of the UVLoop?

https://uvloop.readthedocs.io/user/index.html

src/textual/app.py
```diff
@@ -227,13 +221,28 @@ def run(
             console (Console, optional): Console object. Defaults to None.
             screen (bool, optional): Enable application mode. Defaults to True.
             driver (Type[Driver], optional): Driver class or None for default. Defaults to None.
+            loop (): Event loop to run the application on. If not specified, uvloop will be used.
```
willmcgugan, line 224: Can you also put the type here.

src/textual/app.py
darrenburns: Probably better yeah, done

willmcgugan (approved):
Great. Small request.

src/textual/app.py
darrenburns: Not sure how that happened 😅

PR #240, opened 2022-01-28
Little refactor

UltraStudioLTD:
Why don't we combine 33rd and 34th lines https://github.com/Textualize/textual/blob/32a3bea47fc15e1f0f1a7e1560abecd150fbe016/src/textual/app.py#L33-L34 into one line using walrus operator? It increases speed and performance (although not quit noticeably but in **Python** every ms counts)

IMG_20220128_153805

### DIFF
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -30,8 +30,7 @@ from .view import View
 from .views import DockView
 from .widget import Widget, Reactive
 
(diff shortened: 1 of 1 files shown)


willmcgugan:
Textual runs on Py3.7 which doesn't support walrus operator.

PR #246, opened 2022-01-31
Underline bar renderable

### LINKED ISSUE #238: Implement a bar renderable for tabbed dialogs
Implement a renderable used to underline tabs in a tabbed dialog.

This renderable should use the same unicode characters as Rich progress bars. It should render a line with a portion in a different color extending from p1 to p2.

Also add the option to not render the highlight and just the background bar.

Suggested interface:

```python
class TabUnderline:
    def __init__(self, highlight: tuple[float, float] | None = None, color1: Color, color2: Color):
        ...
```

darrenburns:
Naming is hard :) more than happy to hear suggestions for the params etc.

Closes #238


willmcgugan (approved):
👍

PR #247, opened 2022-02-01
fix `examples/simple.py` Windows compatibility

mhils:
On Windows, the default encoding used by Python sadly is cp1252. We want to force UTF-8 here (as it is the default on macOS/Linux), otherwise the README cannot be read.


willmcgugan:
Thanks!

PR #250, opened 2022-02-01
Tokenise CSS variables

darrenburns:
Implements tokenisation of SASS-like variables.

Variables can be declared with values at the root scope, and you can refer to these variables inside any rule or variable declaration.


darrenburns:
@willmcgugan have modified the tokenisation as agreed

willmcgugan (approved):
Excellent. And well tested!

PR #258, opened 2022-02-03
Variable support in CSS

darrenburns:
Allows users to use variables in CSS files. See `dev_sandbox.scss` for some examples.


src/textual/css/parse.py
```diff
@@ -205,9 +204,102 @@ def parse_declarations(css: str, path: str) -> Styles:
...
+        token = next(tokens, None)
+        if token is None:
+            break
+        if token.name == "variable_name":
+            variable_name = token.value[1:-1]  # Trim the $ and the :, i.e. "$x:" -> "x"
+            yield token
+
+            # Store the tokens for any variable definitions, and substitute
+            # any variable references we encounter with them.
+            leading_whitespace = True
```
willmcgugan, line 246: I reckon this can be simplified to avoid that `leading_whitespace` variable.

Could you have a loop prior to this which first gets all whitespace from the iterator until it reaches one of the variable tokens.

src/textual/css/parse.py
```diff
@@ -205,9 +204,102 @@ def parse_declarations(css: str, path: str) -> Styles:
...
+                    break
+                # For variables referring to other variables
+                elif token.name == "variable_ref":
+                    ref_name = token.value[1:]
+                    if ref_name in variables:
+                        variable_tokens = variables[variable_name]
+                        reference_tokens = variables[ref_name]
+                        variable_tokens.extend(reference_tokens)
+                        ref_location = token.location
+                        ref_length = cell_len(token.value)
```
willmcgugan, line 265: Is this definitely the cell length you need here? I would have thought it would be the number of characters that is relevant.

src/textual/css/parse.py
```diff
@@ -205,9 +204,102 @@ def parse_declarations(css: str, path: str) -> Styles:
     return styles_builder.styles
 
 
-def parse(css: str, path: str) -> Iterable[RuleSet]:
+def _is_whitespace(token: Token) -> bool:
```
willmcgugan, line 207: Is this used?

src/textual/css/parse.py
```diff
@@ -205,9 +204,102 @@ def parse_declarations(css: str, path: str) -> Styles:
...
+                            length=ref_length,
+                        )
+                    )
+            else:
+                raise _unresolved(variable_name=variable_name, location=token.location)
+        else:
+            yield token
+
+
+def parse(css: str, path: str) -> Iterable[RuleSet]:
```
willmcgugan, line 301: Would you mind giving this a docstring while you're in this code.

willmcgugan (changes requested):
Looks good. Didn't need to run it, think I grok it after your explanation. Just a few requests.

src/textual/css/parse.py
darrenburns: This was making the assumption that variables of "string" type could have non-ASCII values, which admittedly I'm not sure about. If the value is out-with ASCII, don't we want to know how many cells wide it is so that we can highlight all of those cells? If we take the `len`, we may end up highlighting more cells than required.

src/textual/css/parse.py
darrenburns: Nope, good spot

willmcgugan (approved):
LGTM

PR #155, opened 2021-10-27
Fix project description not being shown on PyPI

wasi-master:
Adds the [readme parameter](https://python-poetry.org/docs/pyproject/#readme) to [`pyproject.toml`](https://github.com/willmcgugan/textual/blob/main/pyproject.toml)

The readme parameter is used to provide the full description of the project (i.e. the README). For more info see [this](https://www.python.org/dev/peps/pep-0621/#readme)

Fixes https://github.com/willmcgugan/textual/issues/147

PR #265, opened 2022-02-06
Add README to PyPI

MartinThoma:
Currently, when you go to https://pypi.org/project/textual/ it looks like this:

image

With this PR (+ a new version added to PyPI with those changes), there should be the README.

I'm not 100% certain, but the following indicates it might work: https://github.com/python-poetry/poetry/issues/1979


willmcgugan:
THanks

wasi-master:
I had the same pr (#155) around 5 months ago, Looks like Will forgot to notice. No worries

PR #269, opened 2022-02-07
Sparklines

### LINKED ISSUE #259: Sparkline renderable
Implement a renderable that generates a sparkline.

The input should be a list of floats of any size. The Sparkline renderable should use the same block characters as used in the scrollbar to render a minature graph.

```python
class Sparkline:
    def __init__(self, data: list[float], width:int | None):
        ...
```

Should look something like the following:

https://blog.jonudell.net/2021/08/05/the-tao-of-unicode-sparklines/

I think we should also color each bar by blending two colors.

Note that the width may not match the data length, so you may need to pick the closes data point.

darrenburns:
Closes #259 

image

### DIFF
--- a/src/textual/renderables/sparkline.py
+++ b/src/textual/renderables/sparkline.py
@@ -1,0 +1,120 @@
+from __future__ import annotations
+
+import statistics
+from typing import Sequence, Iterable, Callable, TypeVar
+
+from rich.color import Color
+from rich.console import ConsoleOptions, Console, RenderResult
+from rich.segment import Segment
+from rich.style import Style
+
+T = TypeVar("T", int, float)
+
+
+class Sparkline:
+    """A sparkline representing a series of data.
+
+    Args:
+        data (Sequence[T]): The sequence of data to render.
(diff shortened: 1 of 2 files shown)


src/textual/renderables/sparkline.py
```diff
@@ -0,0 +1,127 @@
...
+    BARS = "▁▂▃▄▅▆▇█"
+
+    def __init__(
+        self,
+        data: Sequence[T],
+        *,
+        width: int | None,
+        min_color: Color = Color.from_rgb(0, 255, 0),
+        max_color: Color = Color.from_rgb(255, 0, 0),
+        summary_func: Callable[[list[T]], float] = max,
```
willmcgugan, line 34: Can we make this `summary_function`, I prefer to avoid abbreviations, except when well established.

willmcgugan (changes requested):
Looks good! Wish I understood _buckets better. Just a couple of minor requests.

willmcgugan (approved):
👍

PR #272, opened 2022-02-08
Render opacity

### LINKED ISSUE #271: Render opacity
Create a renderable which applies an opacity value between 0 and 1.

This value should blend the foreground color to the background color. So an opacity of 0 essentially makes text invisible. And an opacity of 0.5 makes a 50% blend from background to foreground.

Interface should be something like the following:

```python
class OpacityRender:

    def __init__(self, renderable, opacity):
        ...
```

In the `__rich_console__` method, it will call `render_lines` and apply the opacity to the segments.

darrenburns:
Closes #271 

A Panel, and the same Panel passed through the opacity renderable:
image

An example with `Live` because why not? :)

https://user-images.githubusercontent.com/5740731/153045908-f71efb5f-01ee-4c46-87f2-ec35239a9c24.MOV


src/textual/renderables/opacity.py
```diff
@@ -0,0 +1,90 @@
...
+        self, console: Console, options: ConsoleOptions
+    ) -> RenderResult:
+        segments = console.render(self.renderable, options)
+        opacity = self.value
+        for segment in segments:
+            style = segment.style
+            if not style:
+                yield segment
+                continue
+            fg, bg = style.color, style.bgcolor
```
willmcgugan, line 33: I suspect this is slower that doing the assignment on two lines, since you would need to construct a tuple.

Although, if you ran a `timeit` and this proved not the case, I would happily withdraw my objection!

src/textual/renderables/opacity.py
```diff
@@ -0,0 +1,90 @@
...
+            if fg and fg.triplet and bg and bg.triplet:
+                yield Segment(
```
willmcgugan, line 35: Could we replace these with positional args -- I prefer positional args for small atomic objects such as Segments.

willmcgugan (changes requested):
Great. A few small requests.

src/textual/renderables/opacity.py
darrenburns: Missed them all 😄 think I hallucinated doing this. Have fixed now.

PR #279, opened 2022-02-11
Opacity rules

darrenburns:
Error handling for `opacity` property:

image

Also cleaned up `visibility` error output:

image


src/textual/css/_styles_builder.py
```diff
@@ -119,11 +130,46 @@ def process_visibility(
                     self.error(
                         name,
                         token,
-                        f"invalid value for visibility (received {value!r}, expected {friendly_list(VALID_VISIBILITY)})",
+                        f"property 'visibility' has invalid value {value!r}; expected {friendly_list(VALID_VISIBILITY)}",
                     )
             else:
                 self.error(name, token, f"invalid token {value!r} in this context")
 
+    def process_opacity(self, name: str, tokens: list[Token], important: bool) -> None:
```
darrenburns, line 138: Will need to have a think about how the `process_*` methods can be covered by tests. We perhaps don't want to test the StyleBuilder directly, and instead ensure we have solid testing around the parsing functions it's used within.

willmcgugan (approved):
Solid

PR #281, opened 2022-02-12
Fix pyproject.toml syntax

Fyssion:
Currently, some malformed syntax in `pyproject.toml` causes `poetry install` to fail. This PR corrects the syntax.


willmcgugan:
Thanks

PR #282, opened 2022-02-12
fix console not being passed into app

ZeroIntensity:
wasnt being passed in when calling `App.run` (and there was no way to pass it in the constructor)

### DIFF
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -3,7 +3,7 @@ import os
 
 import asyncio
 import platform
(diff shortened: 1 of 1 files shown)


willmcgugan:
I don't want to expose the console here. It is an internal detail, largely not relevant to the user. Why do you need it?

ZeroIntensity:
i was using it to add a custom theme to the app, but i made the pr because run() took a console argument that it didn't do anything with

PR #293, opened 2022-02-17
Tabs

darrenburns:
## Example

The video below shows a few different features:

* Clicking on the underline or anywhere on the bar will activate the nearest tab.
* Using "left" and "right" arrows on the keyboard will cycle through tabs, wrapping around 
* Pressing a number key will cause the tab with that number to activate
* Pressing a letter will "search" through tab labels, so when I press "M" key, it jumps to "March" to "May". It is cycling through tabs that begin with the letter "M".

https://user-images.githubusercontent.com/5740731/154511136-d26e1ff2-ec16-4047-b05f-4ecad4cc90b9.mov

## Demo

See the demo in the `sandbox` dir: `python -m tabs` - or watch [the video](https://www.youtube.com/watch?v=59CVt9gVcvo).

Screenshot of the demo:
image

When there are no tabs, only the "inactive" underline renders:
image

## Outstanding tasks

- [x] Increase tab click target to include padding
- [x] Handle case where user supplies no tabs - just render inactive underline
- [x] Increase tab click target to include underline area
- [x] Widget-level keyboard control
- [x] Minor rounding error fix
- [x] Docstrings


src/textual/widgets/tabs.py
```diff
@@ -0,0 +1,350 @@
...
+            yield from self.get_tab_labels(console, options)
+        yield Segment.line()
+        yield from self.get_underline_bar(console)
+
+    def get_tab_labels(self, console: Console, options: ConsoleOptions) -> RenderResult:
+        """Yields the spaced-out labels that appear above the line for the Tabs widget"""
+        width = self.width or options.max_width
+        tab_values = self.tabs.values()
+
+        space = cast(
```
willmcgugan, line 97: Ah. I so rarely see SimpleNamespace in the wild! I think it may be too dynamic for typing. You may be better off creating a named tuple, so you wouldn't need the cast.

sandbox/tabs.py
```diff
@@ -0,0 +1,147 @@
...
+from rich.padding import Padding
+from rich.rule import Rule
+
+from textual import events
+from textual.app import App
+from textual.widget import Widget
+from textual.widgets.tabs import Tabs, Tab
+
+
+class Hr(Widget):
```
willmcgugan, line 13: Not for this PR, but these should first class widgets at some point. I'm sure we will need them in the future.

src/textual/widgets/tabs.py
```diff
@@ -0,0 +1,350 @@
...
+            automatically be padded such that they fit the  available horizontal space.
+        search_by_first_character (bool): If True, entering a character on your keyboard
+            will activate the next tab (in left-to-right order) with a label starting with
+            that character.
+    """
+
+    DEFAULT_STYLES = "height: 2;"
+
+    _active_tab_name: Reactive[str | None] = Reactive("")
+    _bar_offset: Reactive[float] = Reactive(0.0)
```
willmcgugan, line 203: I *think* you can leave out the type, and Mypy will assume it is the default.

i.e. this works, and Mypy will still assume it is a floar:

```python
_bar_offset = Reactive(0.0)
```

Just letting you know. I don't have a problem with the repetition.

sandbox/tabs.py
```diff
@@ -0,0 +1,147 @@
...
+        self.mount(
+            info=Info(
+                "\n"
+                "• The examples below show customisation options for the [bold #1493FF]Tabs[/] widget.\n"
+                "• Press keys 1-6 on your keyboard to switch tabs, or click on a tab.",
+            )
+        )
+        for example in self.examples:
+            info = Info(example.description)
+            self.mount(Hr())
```
willmcgugan, line 142: You could mount all 3 in a single `mount()` call.

Just letting you know. Doesn't matter for the sandbox.

willmcgugan (commented):
Just comments. Will do a review on Monday. Looks solid.

src/textual/widgets/tabs.py
```diff
@@ -0,0 +1,350 @@
...
+        space = cast(
+            Edge, SimpleNamespace(min_size=1, size=self.tab_padding or None, fraction=1)
+        )
+        edges = []
+        for tab in tab_values:
+            tab = cast(
+                Edge,
+                SimpleNamespace(size=cell_len(tab.label), min_size=1, fraction=None),
+            )
+            edges.extend([space, tab, space])
```
willmcgugan, line 106: I get why there is `[space, tab, space]` to generate space around tabs.

This would give you something like: `[space, tab, space, space, tab, space]` etc. for multiple tabs.

It would be more efficient for layout resolve if you combined those double spaces in to a single space with `fraction=2`, which should be equivalent. So you would generate the following:

`[space, tab, double_space, tab, space]`

willmcgugan (changes requested):
A possible enhancement for the layout resolver.

src/textual/widgets/tabs.py
darrenburns: I've done the thing we spoke about - created a NamedTuple which implements the EdgeProtocol which we can use in general cases.

sandbox/tabs.py
darrenburns: Definitely!

src/textual/widgets/tabs.py
darrenburns: As discussed, have kept it as is in order to have layout_resolve tell us the extents of each tab without further work.

willmcgugan: I see why you might need the repetition to properly define the clickable boundaries. I withdraw my objection!

src/textual/widgets/tabs.py
```diff
@@ -0,0 +1,344 @@
...
+        width = self.width or options.max_width
+        tab_values = self.tabs.values()
+
+        space = Edge(size=self.tab_padding or None, min_size=1, fraction=1)
+        edges = []
+        for tab in tab_values:
+            tab = Edge(size=cell_len(tab.label), min_size=1, fraction=None)
+            edges.extend([space, tab, space])
+
+        spacing = layout_resolve(width, edges=edges)
```
willmcgugan, line 102: Maybe leave it for now, but it occurs to me that is quite cacheable.

willmcgugan (approved):
LGTM

PR #311, opened 2022-02-26
Add py.typed file

sanders41:
I added a py.typed file since types are present so they will be available to editors.

Note: The update to CODE_OF_CONDUCT.md was done by the pre-commit end-of-file-fixer hook.


willmcgugan:
Fixed in https://github.com/Textualize/textual/pull/324 Thanks

PR #319, opened 2022-03-02
Regions

### LINKED ISSUE #315: Overlapping region resolver
We need a container class which manages overlapping regions. Internally it will act like a container of Region objects (probably a simple list).

Additionally there will be a method which generates a sequence of Region objects which cover the same area, but with no overlapping.

Discuss with @willmcgugan as this likely needs further explanation.

darrenburns:
Closes #315


src/textual/_region_group.py
```diff
@@ -0,0 +1,57 @@
...
+
+class InlineRange(NamedTuple):
+    """Represents a region on a single line."""
+
+    line_index: int
+    start: int
+    end: int
+
+
+class RegionGroup:
```
willmcgugan, line 18: On reflection this probably doesn't need to be a class. All we're really interested in is that `inline_ranges` method, which we could extract in to a function that accepts an Iterable of Regions and returns InlineRanges.

We can probably rename it in that case. If it were pixels I'd be tempted to call it `rasterize_regions`. There may be a better term when it comes to characters.

willmcgugan (changes requested):
Great. Simpler that what I had in mind. Only one request re replacing class with a function.

src/textual/_region_group.py
darrenburns: Yeah I had the same thought, agreed!

darrenburns: Coming up empty for better names for the function - let me know if you have anything.

willmcgugan: `resolve_regions_to_ranges` ?

src/textual/_region_group.py
```diff
@@ -0,0 +1,50 @@
...
+
+    inline_ranges: dict[int, list[InlineRange]] = defaultdict(list)
+    for region_x, region_y, width, height in regions:
+        for y in range(region_y, region_y + height):
+            inline_ranges[y].append(
+                InlineRange(line_index=y, start=region_x, end=region_x + width - 1)
+            )
+
+    for line_index, ranges in inline_ranges.items():
+        sorted_ranges = sorted(ranges, key=attrgetter("start"))
```
willmcgugan, line 41: Can we move this `attrgetter` outside of the loop since it never changes.

src/textual/_region_group.py
```diff
@@ -0,0 +1,50 @@
...
+        _, start, end = sorted_ranges[0]
+        for next_line_index, next_start, next_end in sorted_ranges[1:]:
```
willmcgugan, line 43: I think we can probably avoid this slice, and the need to create a new list.

If we have a range iterator, with `iter_ranges = iter(sorted_ranges)` when can get the first one with `next(iter_ranges)` and then iterator over the remaining range without that `[1:]`

src/textual/_region_group.py
```diff
@@ -0,0 +1,50 @@
...
+    results in a shape identical to the combined original regions.
+
+    Args:
+        regions (Iterable[Region]): An iterable of Regions.
+
+    Returns:
+        Iterable[InlineRange]: Yields InlineRange objects representing the content on
+            a single line, with overlaps removed.
+    """
+    if not regions:
```
willmcgugan, line 30: This may always be True for some iterables.

```python
>>> l=[]
>>> i=iter(l)
>>> bool(i)
True
```

I suspect it is fine to remove this check. The caller will know if it is empty, and if its not it is very little work.

src/textual/_region_group.py
```diff
@@ -0,0 +1,50 @@
+from __future__ import annotations
+
+from collections import defaultdict
+from operator import attrgetter
+from typing import NamedTuple, Iterable
+
+from src.textual.geometry import Region
```
willmcgugan, line 7: Should be a relative import. Can you look in to the editor config that added this.

willmcgugan (changes requested):
A few requests

src/textual/_region_group.py
darrenburns: TIL!

willmcgugan (approved):
LGTM

PR #324, opened 2022-03-04
fix: allow downstream projects to use mypy

henryiii:
Downstream projects get a message `error: Skipping analyzing "textual": module is installed, but missing library stubs or py.typed marker` when trying to use mypy. This adds the missing marker, copied from rich. `tool.poetry.include` is not required, though - poetry (now?) picks this up from VCS, like other files.

Tested with:

```console
$ pip run build
...
$ unzip -l dist/textual-0.1.15-py3-none-any.whl | grep py.typed
        0  01-01-1980 00:00   textual/py.typed
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,6 +16,7 @@ classifiers = [
     "Programming Language :: Python :: 3.8",
     "Programming Language :: Python :: 3.9",
     "Programming Language :: Python :: 3.10",
+    "Typing :: Typed",
 ]
 
 


henryiii:
Sorry! Duplicate of #311.

henryiii:
Thought this doesn't have the unneeded "include", and it adds the missing Trove classifier, unlike #311.

willmcgugan:
thanks

henryiii:
Would it be helpful to make a "forward-port" of this to the css branch?

willmcgugan:
I will be merged in to css. Although nobody except us are using the css branch currently.

henryiii:
Great, just saw the note in the readme about work on the css branch, and wanted to make sure there wasn't something I still needed to do. :)

PR #333, opened 2022-03-12
Compositor refactor

willmcgugan:


src/textual/_animator.py
```diff
@@ -165,6 +171,11 @@ def animate(
             easing (EasingFunction | str, optional): An easing function. Defaults to DEFAULT_EASING.
         """
 
+        if not hasattr(obj, attribute):
+            raise AttributeError(
```
willmcgugan, line 175: Catches typos when animating

src/textual/_compositor.py
```diff
@@ -0,0 +1,549 @@
+"""
+
+The compositor handles combining widgets in to a single screen (i.e. compositing).
+
+It also stores the results of that process, so that Textual knows the widgets on
+the screen and their locations. The compositor uses this information to answer
+queries regarding the widget under an office, or the style under an offset.
```
darrenburns, line 7:
```suggestion
queries regarding the widget under an offset, or the style under an offset.
```

src/textual/_compositor.py
```diff
@@ -0,0 +1,549 @@
...
+    """A renderable containing the result of a render for a given region."""
+
+    def __init__(self, lines: Lines, region: Region) -> None:
+        self.lines = lines
+        self.region = region
+
+    def __rich_console__(
+        self, console: Console, options: ConsoleOptions
+    ) -> RenderResult:
+        yield Control.home()
```
darrenburns, line 77: Do we need this? (`move_to` moves to an absolute position anyway)

src/textual/_compositor.py
```diff
@@ -0,0 +1,549 @@
...
+@rich.repr.auto(angular=True)
+class Compositor:
+    """Responsible for storing information regarding the relative positions of Widgets and rendering them."""
+
+    def __init__(self) -> None:
+        # A mapping of Widget on to its "render location" (absolute position / depth)
+        self.map: RenderRegionMap = {}
+
+        # All widgets considered in the arrangement
+        # Not this may be a supperset of self.map.keys() as some widgets may be invisible for various reasons
```
darrenburns, line 104: Supper 😋 
```suggestion
        # Note this may be a superset of self.map.keys() as some widgets may be invisible for various reasons
```

src/textual/_compositor.py
```diff
@@ -0,0 +1,549 @@
...
+
+        # The top level widget
+        self.root: Widget | None = None
+
+        # Dimensions of the arrangement
+        self.size = Size(0, 0)
+
+        # A mapping of Widget on to region, and clip region
+        # The clip region can be considered the window through which a widget is viewed
+        self.regions: dict[Widget, tuple[Region, Region]] = {}
```
darrenburns, line 115: Initial thought on reading this was: if this is a map of Widget -> Region, and self.map is also a mapping of Widget -> Region, do we need both?

sandbox/dev_sandbox.py
```diff
@@ -29,5 +30,8 @@ def on_mount(self):
             sidebar=Widget(),
         )
 
+    def action_dump(self):
```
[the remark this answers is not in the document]
willmcgugan: Think I know why that is. It renders the tree on exit, but on exit the widgets are in the process of being torn down. So it is a race condition of sorts.

It should probably immediately render and print the cached version on exit.

sandbox/dev_sandbox.py
```diff
@@ -29,5 +30,8 @@ def on_mount(self):
...
+        self.panic(self.tree)
```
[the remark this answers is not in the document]
willmcgugan: Updated the docstring...

sandbox/uber.py
```diff
@@ -0,0 +1,43 @@
+from tkinter import Place
```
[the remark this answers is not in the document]
willmcgugan: This is vscode trying to be "helpful"!

src/textual/_animator.py
```diff
@@ -221,8 +232,8 @@ def __call__(self) -> None:
                 animation = self._animations[animation_key]
                 if animation(animation_time):
                     del self._animations[animation_key]
-            self.on_animation_frame()
+        self.on_animation_frame()
```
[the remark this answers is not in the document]
willmcgugan: Yes, fixes a long standing bug! The very last animation frame wasn't being rendered.

src/textual/_compositor.py
willmcgugan: Possibly not!

src/textual/_compositor.py
willmcgugan: You may be right. That information is derived from map anyway. Will see if I can simplify that in a future update.

src/textual/css/styles.py
```diff
@@ -295,6 +333,71 @@ def get_transition(self, key: str) -> Transition | None:
         else:
             return None
 
+    def get_box_model(
+        self, container_size: Size, parent_size: Size
+    ) -> tuple[Size, Spacing]:
+        """Resolve the box model for this Styles.
+
+        Args:
```
darrenburns, line 341: Arg name is off here (and wrong order)

src/textual/css/styles.py
```diff
@@ -295,6 +333,71 @@ def get_transition(self, key: str) -> Transition | None:
...
+        if self.max_height:
+            max_height = self.max_height.resolve_dimension(container_size, parent_size)
+            height = min(width, max_height)
+
+        # TODO: box sizing
+
+        size = Size(width, height)
+        margin = Spacing(0, 0, 0, 0)
+
+        if self.box_sizing == "content-box":
```
darrenburns, line 382: This differs from the CSS box model and feels very surprising to me - "content-box" shouldn't include padding or border in the size, it's only the size of the content box itself. This will cause confusion going forward if we implement CSS rules with different semantics from CSS itself.

src/textual/_compositor.py
willmcgugan: You're right. Removed.

src/textual/css/styles.py
```diff
@@ -295,6 +333,71 @@ def get_transition(self, key: str) -> Transition | None:
...
+            if has_rule("padding"):
+                size += self.padding
+            if has_rule("border"):
+                size += self.border.spacing.totals
+            if has_rule("margin"):
+                margin = self.margin
+
+        else:  # border-box
+            if has_rule("padding"):
+                size -= self.padding
```
darrenburns, line 393: Similar to the above comment, I think my expectation here would be that we would include the padding and border in the overall size rather than remove it, but I could be misunderstanding this code.

src/textual/_compositor.py
```diff
@@ -0,0 +1,549 @@
...
+                if styles_offset
+                else ORIGIN
+            )
+
+            # Container region is minus border
+            container_region = region.shrink(widget.styles.gutter)
+
+            # Containers (widgets with layout) require adding children
+            if widget.layout is not None:
+                scroll = widget.scroll
```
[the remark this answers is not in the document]
willmcgugan: Yeah, I like scroll_offset

src/textual/_compositor.py
```diff
@@ -0,0 +1,549 @@
...
+
+                # The region that contains the content (container region minus scrollbars)
+                child_region = widget._arrange_container(container_region)
```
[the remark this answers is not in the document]
willmcgugan: It's not quite the content region, in that it also includes border and padding.

src/textual/_compositor.py
```diff
@@ -0,0 +1,549 @@
...
+
+                # Adjust the clip region accordingly
+                sub_clip = clip.intersection(child_region)
+
+                arrange_region = child_region
```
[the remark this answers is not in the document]
willmcgugan: Yeah, probably a remenant of an edit.

src/textual/dom.py
```diff
@@ -71,6 +78,19 @@ def parent(self) -> DOMNode:
         assert isinstance(self._parent, DOMNode)
         return self._parent
 
+    @property
+    def screen(self) -> "Screen":
+        """Get the current screen."""
```
darrenburns, line 83:
```suggestion
        """Get the screen that this node is contained within. Note that this may not be the currently active screen within the app."""
```

src/textual/css/styles.py
```diff
@@ -351,6 +454,7 @@ def get_rule(self, rule: str, default: object = None) -> object:
         return self._rules.get(rule, default)
 
     def refresh(self, *, layout: bool = False) -> None:
+        return
```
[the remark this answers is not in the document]
willmcgugan: Ah yes. Fixed.

src/textual/dom.py
```diff
@@ -280,8 +365,8 @@ def walk_children(self, with_self: bool = True) -> Iterable[DOMNode]:
                 pop()
             else:
                 yield node
-                if node.children:
-                    push(iter(node.children))
+                if node.node_list:
```
darrenburns, line 368: NP: Was children perhaps a more semantic name?

src/textual/css/styles.py
```diff
@@ -295,6 +333,71 @@ def get_transition(self, key: str) -> Transition | None:
...
+
+        if self.min_height:
+            min_height = self.min_height.resolve_dimension(container_size, parent_size)
+            height = max(height, min_height)
+
+        if self.max_height:
+            max_height = self.max_height.resolve_dimension(container_size, parent_size)
+            height = min(width, max_height)
+
+        # TODO: box sizing
```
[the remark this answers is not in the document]
willmcgugan: That still is TODO.

src/textual/geometry.py
```diff
@@ -470,11 +484,100 @@ def union(self, region: Region) -> Region:
...
-        union_region = Region.from_corners(
+        union_region = self.from_corners(
             min(x1, ox1), min(y1, oy1), max(x2, ox2), max(y2, oy2)
         )
         return union_region
 
+    def split(self, cut_x: int, cut_y: int) -> tuple[Region, Region, Region, Region]:
+        """Split a region in to 4 from given x and y offsets (cuts).
+
+                   cut_x ↓
```
darrenburns, line 495: Love these diagrams, they help so much.

If docs are generated from these docstrings we'll need to check how to have this diagram be printed inside `<pre>` tags or something

src/textual/css/styles.py
willmcgugan: I think that is correct. The `size` being calculated includes is the space taken up by content + padding + border.

src/textual/css/styles.py
willmcgugan: Yeah, the size being calculated is the size that the Widget will render. The widget itself is responsible for rendering the border and padding.

src/textual/dom.py
willmcgugan: Yes, you're right! At one point I was going to split the nodes in to `children` and a `shadow node list`. I've put it back to `children`

sandbox/dev_sandbox.py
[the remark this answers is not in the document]
darrenburns: Tracked at https://github.com/Textualize/textual/issues/348

willmcgugan:
💯

Issue #348, opened 2022-03-29
App.panic should prerender the renderable.


willmcgugan:
App.panic renders content after the app has exited. If that content contains references to the DOM it may be broken at the point it is rendered.

We should render the renderable within app.panic, so that on exit it displays as expected.

willmcgugan:
Fixed in CSS

PR #349, opened 2022-03-29
WIP New color class

willmcgugan:
- Adds a `Color` class which is conceptually similar to Rich's Color class, but only supports RGB and adds an alpha component.

- Adds a `design` module, with a `ColorSystem` class which builds a theme from some base colors.

- Adds the ability to pre-populate CSS variables, to insert color system concepts.

- Fixes bugs in the compositor related to Z order.


UltraStudioLTD:
Why do you need **alpha**?

willmcgugan:
Same reasons as the web.

UltraStudioLTD:
So do you plan adding transparency? Like [**jexer**](https://jexer.sourceforge.io), [**blessed**](https://github.com/chjj/blessed), [**vtm**](https://github.com/netxs-group/vtm) and [**notcurses**](https://github.com/dankamongmen/notcurses)?

src/textual/_node_list.py
```diff
@@ -19,15 +19,13 @@ class NodeList:
     """
 
     def __init__(self) -> None:
-        self._node_refs: list[ref[DOMNode]] = []
-        self.__nodes: list[DOMNode] | None = []
+        self._nodes: list[DOMNode] = []
```
willmcgugan, line 22: This class has changed in purpose. It was a container that didn't hold on to references. Now it is a simpler type of container.

It could perhaps be a list, and may become one eventually. But it may also present a interface to the developer that prevents them from breaking things.

src/textual/app.py
```diff
@@ -367,16 +395,18 @@ def panic(self, *renderables: RenderableType) -> None:
...
-                    show_locals=True,
-                    width=None,
-                    locals_max_length=5,
-                    suppress=[rich],
+                    show_locals=True, width=None, locals_max_length=5, suppress=[rich]
                 ),
             )
-        self._exit_renderables.extend(renderables)
+
+        prerendered = [
```
willmcgugan, line 404: Fixes https://github.com/Textualize/textual/issues/348

src/textual/design.py
```diff
@@ -0,0 +1,230 @@
...
+        secondary = self.secondary or primary
+        warning = self.warning or primary
+        error = self.error or secondary
+        success = self.success or secondary
+        accent1 = self.accent1 or primary
+        accent2 = self.accent2 or secondary
+
+        background = self.background or (BLACK if dark else Color.parse("#f5f5f5"))
+        surface = self.surface or (
+            Color.parse("#121212") if dark else Color.parse("#efefef")
```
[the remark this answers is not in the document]
willmcgugan: They probably should be constants, and the Color module may be a sensible place to put them!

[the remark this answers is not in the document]
DrBenton: :ok_hand:

src/textual/app.py
```diff
@@ -531,6 +558,19 @@ def refresh(self, *, repaint: bool = True, layout: bool = False) -> None:
             except Exception:
                 self.panic()
 
+    def refresh_css(self, animate: bool = True) -> None:
+        """Refresh CSS.
+
+        Args:
+            animate (bool, optional): Also execute CSS animations. Defaults to True.
+        """
+        # TODO: This doesn't update variables
```
darrenburns, line 567: What does this comment mean?

src/textual/color.py
```diff
@@ -0,0 +1,443 @@
...
+                float(value) for value in rgba.split(",")
+            ]
+            color = cls(
+                clamp(int(float_r), 0, 255),
+                clamp(int(float_g), 0, 255),
+                clamp(int(float_b), 0, 255),
+                clamp(float_a, 0.0, 1.0),
+            )
+        else:
+            raise AssertionError("Can't get here if RE_COLOR matches")
```
darrenburns, line 281: Probably want a better message here for a user-facing class

src/textual/color.py
```diff
@@ -0,0 +1,443 @@
...
+    def lighten(self, amount: float) -> Color:
+        """Lighten the color by a given amount.
+
+        Args:
+            amount (float): Value between 0-1 to increase luminance by.
+
+        Returns:
+            Color: New color.
+        """
+        return self.darken(-amount).clamped
```
darrenburns, line 306: Nice 😄

src/textual/css/tokenize.py
```diff
@@ -152,15 +161,40 @@ class DeclarationTokenizerState(TokenizerState):
...
 tokenize = TokenizerState()
 tokenize_declarations = DeclarationTokenizerState()
+tokenize_value = ValueTokenizerState()
+
+
+def tokenize_values(values: dict[str, str]) -> dict[str, list[Token]]:
+    """Tokens the values in a dict of strings.
+
+    Args:
+        values (dict[str, str]): A mapping of name on to a CSS value.
```
darrenburns, line 177: Could you please clarify what you mean by name here by updating this docstring? Is it CSS property name, or style object attribute name?

src/textual/design.py
```diff
@@ -0,0 +1,239 @@
...
+    surface = ColorProperty()
+    background = ColorProperty()
+
+    @property
+    def shades(self) -> Iterable[str]:
+        """The names of the colors and derived shades."""
+        for color in self.COLOR_NAMES:
+            for shade_number in range(-NUMBER_OF_SHADES, NUMBER_OF_SHADES + 1):
+                if shade_number < 0:
+                    yield f"{color}-darken{abs(shade_number)}"
```
darrenburns, line 105: Take it or leave it, but CSS frameworks usually put a `-` before the number (Bootstrap, Tailwind, Materialize, BassCSS) and I think it reads nicer (e.g. `primary-darken-2`).

src/textual/design.py
```diff
@@ -0,0 +1,239 @@
...
+            return Color.parse(color)
+
+    def __set__(self, obj: ColorSystem, value: Color | str | None) -> None:
+        if isinstance(value, Color):
+            setattr(obj, self._name, value.css)
+        else:
+            setattr(obj, self._name, value)
+
+
+class ColorSystem:
```
darrenburns, line 42: Seems to be untested

darrenburns (changes requested):

examples/basic.css
```diff
@@ -1,43 +1,115 @@
...
 
 App > Screen {
   layout: dock;
   docks: side=left/1;
-  text: on $primary;
-}
-
-Widget:hover {
-  outline: solid green;
+  background: $surface;
```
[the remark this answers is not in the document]
willmcgugan: Yeah, you're right. For some reason they are flipped in my head.

examples/basic.css
```diff
@@ -1,43 +1,115 @@
...
 }
 
 #sidebar.-active {
   offset-x: 0;
 }
 
+#sidebar .title {
+  height: 1;
+  background: $primary-darken2;
+  color: $on-primary-darken2;
```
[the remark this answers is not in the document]
willmcgugan: That's the naming that the Material docs use, but you're right. It conflicts with Rich's conventions where "on something" means that something is the *background*

How about "text-on-primary" or "text-primary"?

src/textual/app.py
willmcgugan: It means I forgot to remote the comment! Previously the new CSS variables wouldn't updated.

src/textual/css/tokenize.py
willmcgugan: It's a css variable name. Updated the docstring to say that.

src/textual/design.py
willmcgugan: Seems reasonable to adopt an existing convention.

PR #351, opened 2022-03-29
Stringify the cast type, for <3.10 support

tusharsadhwani:
Since the types here are an argument, we need to manually stringify them to avoid causing runtime errors.


willmcgugan:
Thanks

PR #360, opened 2022-04-04
Add conversion to/from the CIE-L*ab color space.

rodrigogiraoserrao:
This adds a simple `Lab` class to represent colours in the CIE-L*ab colour space and adds two functions to allow for RGB conversion to and from that space.
Tests are still being written.

Formulas as seen in https://stackoverflow.com/a/8433985/2828287.

Tests use the calculator available in http://www.easyrgb.com/en/convert.php.

### DIFF
--- a/src/textual/color.py
+++ b/src/textual/color.py
@@ -33,6 +33,14 @@ class HSV(NamedTuple):
     v: float
 
 
+class Lab(NamedTuple):
+    """A color in CIE-L*ab format."""
+
+    L: float
+    a: float
+    b: float
+
+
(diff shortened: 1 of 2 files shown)


willmcgugan:
Thanks @rodrigogiraoserrao This looks great!

rodrigogiraoserrao:
Glad to be of help!
Thought about adding a classmethod `Color.from_lab` but forgot to add it 🤷
You'll do it if you think it's worth it.

willmcgugan:
Maybe! Fine for now though.

PR #366, opened 2022-04-06
Logging server

darrenburns:
Adds server which can render/display logs generated by Textual in real time.

Untitled

image

In terms of naming, let me know which direction you want me to take with this.


src/textual/app.py
```diff
@@ -389,6 +414,16 @@ async def process_messages(self) -> None:
         log("---")
         log(f"driver={self.driver_class}")
 
+        if os.getenv("TEXTUAL_DEVTOOLS") == "1":
+            try:
+                await self.devtools.connect()
+                self.log_file.write(f"Connected to devtools ({self.devtools.url})\n")
+                self.log_file.flush()
```
willmcgugan, line 421: Do you need these `flush()` calls? The log file should be line buffered, so will be effectively flushed already.

src/textual/devtools.py
```diff
@@ -0,0 +1,265 @@
...
+from rich.rule import Rule
+from rich.segment import Segments, Segment
+from rich.table import Table
+
+DEFAULT_PORT = 8081
+DEFAULT_SIZE_CHANGE_POLL_DELAY_SECONDS = 2
+QUEUEABLE_TYPES = {"client_log", "client_spillover"}
+
+
+class DevtoolsLogMessage:
```
willmcgugan, line 40: Can we have docstrings please, for Classes and methods.

src/textual/devtools.py
```diff
@@ -0,0 +1,265 @@
...
+            .replace(tzinfo=timezone.utc)
+            .astimezone(tz=tz.tzlocal())
+        )
+        timezone_name = local_time.tzname()
+        table = Table.grid(expand=True)
+        table.add_column()
+        table.add_column()
+        file_link = f"file://{Path(self.path).absolute()}"
+        file_and_line = f"{Path(self.path).name}:{self.line_number}"
+        table.add_row(
```
willmcgugan, line 67: There is a possibility of console markup injection here. e.g. you could have a path with `[blink]` in it. Can you either use Text objects to build longer strings or call markup.escape on the variables in the f strings.

src/textual/devtools.py
```diff
@@ -0,0 +1,265 @@
...
+    return websocket
+
+
+async def _on_shutdown(app: Application) -> None:
+    for task in app["tasks"]:
+        task.cancel()
+        with suppress(CancelledError):
+            await task
+
+    for websocket in set(app["websockets"]):
```
willmcgugan, line 232: Haven't you already killed the task that was processing the websockets by this point? I think this would mean that the websocket isn't closed gracefully.

src/textual/devtools.py
```diff
@@ -0,0 +1,265 @@
...
+        console.print()
+        console.print(
+            DevtoolsInternalMessage(f"Client '{escape(request.remote)}' disconnected")
+        )
+
+    return websocket
+
+
+async def _on_shutdown(app: Application) -> None:
+    for task in app["tasks"]:
```
willmcgugan, line 227: This is kind of the nuclear option ;-)

It's generally better to ask the tasks to shut down and then wait for them.

For the case of a loop with a websocket, you can close the websocket and the loop will end.

For the cases where you have a `while True` consider making that an [event](https://docs.python.org/3/library/asyncio-sync.html#asyncio.Event). i.e. have a single `shutdown_event` and in your tasks do `while not app.shutdown_event.is_set()`. To close all the tasks you can do `self.shutdown_event.set()`

willmcgugan (changes requested):
Great work. I haven't done a deep dive yet. A few requests after a quick run through the code, mostly related to shutdown (always a pain in async).

Also, I think this should probably be a submodule. I can see it growing in features. How about we put it under a submodule called `devtools`?

src/textual/devtools.py
darrenburns: I've already cancelled the task yes, but not the socket. I don't believe task cancellation would close the socket though, so this ensures that after the tasks are cancelled the websockets are also closed.

src/textual/devtools.py
darrenburns: Hadn't heard of asyncio Events before, thanks for pointing me in that direction.

Can I check my understanding of why you're describing this as "nuclear" though?

**cancelling tasks:** When I do `task.cancel()`, I'm basically saying I'm happy that a `CancelledError` can be raised at any `await`, and therefore the coroutine could exit at any `await`. My coroutine would have to be designed to be resilient to that possibility.

**events**: If I use an `Event` and set that, I can control _exactly_ where the coroutine will exit. The coroutine will be able to check this event, and ensure that the task ends exactly where I want it to, and we don't need to worry that any `await` could be the exit-point?

Does that sound right?

willmcgugan: Couldn’t have put it better myself 👍

src/textual/devtools.py
darrenburns: Ah good catch. I caught that elsewhere in the PR but missed it here 😄

src/textual/app.py
darrenburns: Nope, removed them

src/textual/devtools/service.py
```diff
@@ -0,0 +1,260 @@
...
+                    "height": self.console.height,
+                },
+            }
+        )
+
+    async def handle(self, request: Request) -> WebSocketResponse:
+        """Handles a single client connection"""
+        client = ClientHandler(request, service=self)
+        self.clients.append(client)
+        websocket = await client.start()
```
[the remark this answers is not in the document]
darrenburns: Agree! I've renamed to run

willmcgugan:
🥇

PR #372, opened 2022-04-08
Timer fix

willmcgugan:
- Fixes an issue when the CSS was processes while an animation was progresses.
Essentially what was happening is that animations were being restarted, which caused an existing animation to stutter. The solution was that if something is already animating (to the same value) then the subsequent animation is ignored.

- Adds a post_priority_messages to allow Timer events to be processes first.

### DIFF
--- a/sandbox/basic.css
+++ b/sandbox/basic.css
@@ -1,15 +1,17 @@
 /* CSS file for basic.py */
 
 
+/*
 * {
     transition: color 500ms linear, background 500ms linear;
 }
+*/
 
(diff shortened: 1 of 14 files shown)


sandbox/basic.css
```diff
@@ -112,3 +117,57 @@ TweetBody {
   height: 1;
   border-top: hkey $accent2-darken-2;
 }
+
```
willmcgugan, line 120: Just experiments in sandbox, please ignore.

PR #374, opened 2022-04-08
Add Windows and Python3.10 to CI

### LINKED ISSUE #371: Add Windows and Python3.10 to CI

sanders41:
Closes #371


sanders41:
I have not been able to figure out the reason either. What I find weird is the `typed-ast` package is what is causing it to fail, but according to the lock file it shouldn't even try to install this package in 3.8+.

willmcgugan:
We've added 3.10 to the CSS branch now. THanks.

PR #376, opened 2022-04-11
Custom scrollbars

willmcgugan:
- Adds CSS rules to select scrollbar colors.
- Tweaks color system
- Adds doc for color system
- Implements behaviour for nested scrollbars, where if the scrollbar is at the maximum extent, the wheel will scroll the parent


docs/color_system.md
```diff
@@ -0,0 +1,71 @@
...
+
+The _accent_ color should be a contrasting color use in UI elements that should stand out, such as selections, status bars, and underlines.
+
+### Warning, Error, and Success
+
+The _warning_, _error_, and _success_ colors have semantic meaning. While they could be any color, by convention warning should be amber / orange, error should be red, and success should be green.
+
+### System
+
+The system color is used for system controls such as scrollbars. The default is for the system color to be the same as accent, but it is recommended that a different color is chosen to differentiate app controls from those rendered by the Textual system.
```
darrenburns, line 49: > but it is recommended that a different color is chosen to differentiate app controls from those rendered by the Textual system.

IMO this recommendation doesn't solve a real problem. More consistency in colours is more important to me than "differentiating app controls from those rendered by Textual". Knowing whether something on screen is rendered by the app or the app framework probably won't matter to end users.

darrenburns (approved):

docs/color_system.md
willmcgugan: Perhaps. It will default to accent if it is not set. Having a different scrollbar color does help when your status bar is the accent color.

PR #377, opened 2022-04-11
require semi-colons

willmcgugan:
- Require semi-colons to terminate declarations (like the browser), which allows declaration to be on multiple lines
- Display parser related exceptions in a pretty Rich format


src/textual/app.py
```diff
@@ -393,20 +393,12 @@ async def capture_mouse(self, widget: Widget | None) -> None:
             await widget.post_message(events.MouseCapture(self, self.mouse_position))
 
     def panic(self, *renderables: RenderableType) -> None:
-        """Exits the app with a traceback.
+        """Exits the app after displaying a message.
```
darrenburns, line 396:
```suggestion
        """Exits the app and then displays a message.
```

darrenburns (approved):

DrBenton (approved):
I don't understand the full extent of that code yet, but it's good to be a little bit closer to the Web standards in our CSS management! :blush:

src/textual/app.py
willmcgugan: "app" makes me think of the process, rather than just the full-screen terminal.

willmcgugan:
@DrBenton It was a good spot. We want to follow web CSS where it makes sense!

PR #380, opened 2022-04-11
Add lazy imports for Python 3.7+ only

Erotemic:
This PR adds the appropriate scaffolding to support lazy imports at the top level of the textual module.

### Modivation

When using textual I noticed that effectively all imports were done via submodules. This is a great thing from an import time perspective, `import textual` is very quick and  then the user can get exactly the components that they need. However, I find myself wishing I could access core attributes from flat top-level API. Namely:

```
import textual

class MyApp(textual.App):
    pass

```

and then explicitly refer to anything from `textual` when I want it. 

### Description

Lazy init allows for the aforementioned behavior without sacrificing import speed. Using the module level `__getattr__` you can check for a defined top-level name and then route it to the exact appropriate submodule. If you store the attribute in `globals()` any subsequent times that you access the attribute it will be returned as a quick lookup rather than calling `__getattr__` again.

In this PR, I've used [mkinit](https://github.com/Erotemic/mkinit) to autogenerate a basic but explicit lazy initialization structure. It defines `__getattr__`, `__dir__`, and `__all__` based on crawling the entire module directory and looking for submodule names it can expose, and then does so. There are special duner variables and cli flags you can use to modify this behavior. 

Lazy init relies on the module level `__getattr__` function, so it's only possible in Python 3.7. 
For details on how lazy initialization works see: https://scientific-python.org/specs/spec-0001/

There area a few things that probably need tweaking by either setting special mkinit parameters (or perhaps updates to mkinit itself), but I first wanted to add the basic PR and get feedback. It could be that textual doesn't want any sort of top-level API, and everything must be done with submodule imports.


Erotemic:
Note that this need not expose the attributes. Given your design goals, the public API submodules themselves could be exposed.

Currently in IPython (arguably the most important editor/tool to support) the following code:

```
import textual
textual.<tab>
```

where <tab> means then pressing tab, does not give the user any hints as to what functionality is available.

image


```
import textual.<tab>
```

gives a bit more context:
image

but requires the user to import every submodule they plan on using. This adds overhead on the user side. Textual could alleviate this by leveraging lazy imports for an explicit public facing module API (no attributes if that's not you're thing). The mkinit tool makes maintaining such an explicit API fairly easy.

Ultimately, this is a suggestion and you're of course free not to take it, but I do think it would be an improvement for users if they can simply: `import textual`, and then use `textual.<tab>` to navigate the shallow hierarchy. Utilizing lazy imports means there is effectively zero runtime overhead to this, and utilizing mkinit minimizes developer overhead (although what modules are members of the public API do need to be explicitly given in a list).

Even if this idea isn't jiving with your right now, keep it in the back of your mind, maybe for when the API simplification happens. It wouldn't make sense to do this right now if you plan on making significant modifications in the near future.

PR #384, opened 2022-04-12
tweak to log

willmcgugan:
- Adds margin-top, margin-right, padding-top, padding-right etc.
- Modify output of log file to match dev tools

### DIFF
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -250,7 +250,7 @@ class App(DOMNode):
                         objects[0], path=calling_path, lineno=calling_lineno
                     )
             else:
-                output = f" ".join(str(arg) for arg in objects)
+                output = " ".join(str(arg) for arg in objects)
                 if kwargs:
                     key_values = " ".join(
                         f"{key}={value}" for key, value in kwargs.items()
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -234,7 +234,7 @@ class App(DOMNode):
             return
 
         try:
-            if len(objects) == 1:
+            if len(objects) == 1 and not kwargs:
                 if self._log_console is not None:
                     self._log_console.print(objects[0])
                 if self.devtools.is_connected:
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -323,7 +323,7 @@ class App(DOMNode):
             event_loop.close()
 
     async def _on_css_change(self) -> None:
-
+        """Called when the CSS changes (if watch_css is True)."""
         if self.css_file is not None:
             stylesheet = Stylesheet(variables=self.get_css_variables())
             try:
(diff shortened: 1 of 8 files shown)


src/textual/app.py
```diff
@@ -310,11 +327,13 @@ async def _on_css_change(self) -> None:
...
             stylesheet = Stylesheet(variables=self.get_css_variables())
             try:
-                self.log("loading", self.css_file)
+                time = perf_counter()
                 stylesheet.read(self.css_file)
-            except StylesheetError as error:
-                self.log(error)
+                elapsed = (perf_counter() - time) * 1000
+                self.log(f"loaded {self.css_file} in {elapsed:.0f}ms")
+            except Exception as error:
```
willmcgugan, line 334: Forgive the catch all. There are at least 2 errors that could be thrown here. At some point I will give them all a base class.

src/textual/app.py
```diff
@@ -219,30 +234,36 @@ def log(
...
+                    )
+                    output = " ".join((output, key_values))
+                if self._log_console is not None:
+                    self._log_console.print(output, soft_wrap=True)
+                if self.devtools.is_connected:
+                    if not _textual_calling_frame:
+                        _textual_calling_frame = inspect.stack()[1]
+                    calling_path = _textual_calling_frame.filename
+                    calling_lineno = _textual_calling_frame.lineno
+                    self.devtools.log(output, path=calling_path, lineno=calling_lineno)
```
darrenburns, line 266: This is going to conflict quite heavily with my currently open PR I suspect 😄

src/textual/app.py
```diff
@@ -219,30 +234,36 @@ def log(
...
                     self.devtools.log(
-                        *objects, path=calling_path, lineno=calling_lineno
+                        objects[0], path=calling_path, lineno=calling_lineno
                     )
-
-            if self.log_file and verbosity <= self.log_verbosity:
-                self.log_file.write(output + "\n")
-                self.log_file.flush()
+            else:
+                output = f" ".join(str(arg) for arg in objects)
```
DrBenton, line 253: we likely don't need the f-string there?
(pylint detects this by default if I'm not wrong, it could be nice to add such a linter to the project at some point?)

DrBenton (approved):
Sorry my lack of knowledge of the project's code at this point prevent me from doing any relevant comments, so I'm doing trivial ones that don't matter to compensate :sweat_smile:

src/textual/app.py
willmcgugan: Sorry!

src/textual/app.py
willmcgugan: Good catch. I'm dubious about the benefits of such linters, I find the create a lot of busy work that doesn't improve the project. But if it were properly tuned, I could be persuaded!

PR #397, opened 2022-04-14
Textual CLI

### LINKED ISSUE #396: Initial Textual CLI
A CLI tool for textual. It will be built using `click`.

The initial feature exposed by this will be access to the devtools console.

Running `textual console` will run the devtools console server, allowing textual apps to connect to it.

- `textual --help` will display help information
- `textual --version` will display the Textual version

darrenburns:
Closes #396.

- Adds `textual` script entrypoint
- `textual console` runs the development console (aiohttp server)
- The "header" that gets displayed when you run the command has been en***rich***ed (screenshot below)
- `textual --version` returns the textual version (uses click `version_option` feature and `__version__`)
- `textual --help` we get this for free with click.

image

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,6 +23,7 @@ rich = "^12.0.0"
 
 #rich = {git = "git@github.com:willmcgugan/rich", rev = "link-id"}
 typing-extensions = { version = "^3.10.0", python = "<3.8" }
+click = "8.0.4"
 
 [tool.poetry.dev-dependencies]
 pytest = "^6.2.3"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,6 +17,9 @@ classifiers = [
     "Programming Language :: Python :: 3.10",
 ]
 
+[tool.poetry.scripts]
+textual = "textual.cli.cli:run"
+
 [tool.poetry.dependencies]
 python = "^3.7"
 rich = "^12.0.0"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,11 +26,11 @@ rich = "^12.0.0"
 
 #rich = {git = "git@github.com:willmcgugan/rich", rev = "link-id"}
 typing-extensions = { version = "^3.10.0", python = "<3.8" }
-click = "8.0.4"
+click = "8.1.2"
 
 [tool.poetry.dev-dependencies]
 pytest = "^6.2.3"
-black = "^22.1.0"
+black = "^22.3.0"
 mypy = "^0.931"
 pytest-cov = "^2.12.1"
 mkdocs = "^1.2.3"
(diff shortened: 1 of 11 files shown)


src/textual/__init__.py
```diff
@@ -3,6 +3,7 @@
 from rich.console import RenderableType
 
 __all__ = ["log", "panic"]
+__version__ = "0.1.15"
```
darrenburns, line 6: I don't find this particularly clean but it is a commonly used pattern in a bunch of libraries from what I've read. It corresponds to a now rejected PEP: https://peps.python.org/pep-0396/

In 3.8+ they added a nicer API for this:
```python
>>> from importlib.metadata import version
>>> print(version('textual'))
0.1.15
```

pyproject.toml
```diff
@@ -17,16 +17,20 @@ classifiers = [
...
 rich = "^12.0.0"
 
 #rich = {git = "git@github.com:willmcgugan/rich", rev = "link-id"}
 typing-extensions = { version = "^3.10.0", python = "<3.8" }
+click = "8.1.2"
 
 [tool.poetry.dev-dependencies]
 pytest = "^6.2.3"
-black = "^22.1.0"
+black = "^22.3.0"
```
darrenburns, line 33: Had to upgrade black to [avoid it conflicting with click 8.1.2 ](https://github.com/psf/black/issues/2964)

src/textual/__init__.py
willmcgugan: It's going to be a PITA to maintain, and I'd rather not add one of the various hacks to pre-fill this.

How about we use this backport of [importlib-metadata](https://pypi.org/project/importlib-metadata/) ?

src/textual/devtools/renderables.py
```diff
@@ -21,10 +24,31 @@
...
+
+
+class DevConsoleHeader:
+    def __rich_console__(
+        self, console: Console, options: ConsoleOptions
+    ) -> RenderResult:
+        lines = Renderables(
+            [
+                f"[bold]Textual Development Console [#b169dd]v{textual.__version__}",
+                "[#967fa3]Run a Textual app with the environment variable [b]TEXTUAL_DEVTOOLS=1[/] to connect.",
```
willmcgugan, line 37: We probably don't want to use RGB colors here, for weirdos on a light color scheme or who don't want to deviate from their ANSI theme.

willmcgugan (changes requested):

src/textual/devtools/renderables.py
darrenburns: I did initially try using ANSI colours here (and in the devtools log headers where we also use RGB), but honestly, I couldn't get it to look anywhere near as good.

I checked these colours look good on light themes too, and chose them to match the default purple Textual theme/textualize colours. My thinking was if you're building a Textual app we're already completely hijacking your colour scheme there anyway, so some subtle branding colours in the devtools should be no big deal?

image

willmcgugan: Makes sense, but I'm sure we'll get some complaints. Let's for for ANSI here.

willmcgugan (approved):
LGTM

PR #398, opened 2022-04-14
CSS align and box model

willmcgugan:
- Adds `align`, `align-horizontal`, and `align-vertical` to CSS
- Factors our `get_box_model` in to a module of its own, and implements fixes
- Adds tests for get_box_model 
- Vertical and Horizontal layouts now respect margins and box model


src/textual/app.py
```diff
@@ -529,14 +525,15 @@ async def process_messages(self) -> None:
                 self.title = self._title
                 self.refresh()
                 await self.animator.start()
-
```
darrenburns, line 532: The merge of this file looks wrong

willmcgugan: Fixed

tests/test_box_model.py
```diff
@@ -0,0 +1,183 @@
...
+    styles.width = "auto"
+
+    box_model = get_box_model(
+        styles, Size(60, 20), Size(80, 24), get_auto_width, get_auto_height
+    )
+    # Setting width to auto should call get_auto_width
+    assert box_model == BoxModel(Size(10, 20), Spacing(1, 2, 3, 4))
+
+    # Set width to 100 vw which should make it the width of the parent
+    styles.width = "100vw"
```
darrenburns, line 74: This would only be true if the width of the width of the parent is less than the width of the viewport right? If the parent is wider than the viewport then setting the width to 100vw would mean it wouldn't be the width of the parent.

But where does the size of the viewport come into play in these tests? I feel like I'm missing something 🤔

darrenburns (changes requested):
Looks good. A few questions/comments from me

sandbox/align.css
```diff
@@ -0,0 +1,46 @@
...
+
+    width: auto;
+    height: 10;
+    background:magenta;
+    margin: 3;
+    padding: 1; 
+    border: solid white; 
+    box-sizing: border-box;
+    border: solid white; 
+    align-horizontal: center;
```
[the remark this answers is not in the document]
willmcgugan: Yeah, that's right.

src/textual/widget.py
```diff
@@ -105,6 +107,25 @@ def __init__(
...
+        return box_model
+
+    def get_content_width(self, container_size: Size, parent_size: Size) -> int:
+        console = self.app.console
+        renderable = self.render()
+        measurement = Measurement.get(console, console.options, renderable)
+        return measurement.maximum
+
+    def get_content_height(self, container_size: Size, parent_size: Size) -> int:
+        return container_size.height
```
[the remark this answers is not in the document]
willmcgugan: Rich has a way of measuring the renderable via `__rich_measure__` which is what `get_content_width` uses. Most renderables can report their width fairly cheaply.

Getting the height would require rendering it, and it would need to know what width to use before it could report the height. So it is an altogether more complex calculation.

I've made the auto height return the height of the container for now. I would like to make it genuinely fit the content, but I'm going to have to think about how to do that elegantly.

tests/test_box_model.py
```diff
@@ -0,0 +1,183 @@
...
+    styles.width = 10
+    styles.height = 8
+    styles.padding = 1
+    styles.border = ("solid", "red")
+
+    # border-box is default
+    assert styles.box_sizing == "border-box"
+
+    def get_auto_width(container: Size, parent: Size) -> int:
+        assert False, "must not be called"
```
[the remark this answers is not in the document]
willmcgugan: TIL

tests/test_box_model.py
willmcgugan: The parent here is the viewport, which may be the terminal size, or the dimensions of a scrollable region at the next level up.

src/textual/layouts/horizontal.py
```diff
@@ -24,15 +25,30 @@ def arrange(
...
+        ]
+
+        margins = [
+            max((box1.margin.right, box2.margin.left))
+            for box1, box2 in zip(box_models, box_models[1:])
+        ]
+        if box_models:
+            margins.append(box_models[-1].margin.right)
+
+        x = box_models[0].margin.left if box_models else 0
```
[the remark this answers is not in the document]
willmcgugan: Afraid not, because it can still return a single placement on line 54 even with empty children. Also a single box model will produce an empty margins list on L51.

src/textual/layouts/vertical.py
```diff
@@ -26,15 +26,30 @@ def arrange(
...
+            offset_x = widget.styles.align_width(content_width, parent_size.width)
+            region = Region(offset_x, y, content_width, content_height)
+            max_height = max(max_height, content_height)
             add_placement(WidgetPlacement(region, widget, 0))
-            y += region.height + margin.top
-            max_height = y + margin.bottom
+            y += region.height + margin
+            max_height = y
+
+        max_height += margins[-1] if margins else 0
```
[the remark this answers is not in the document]
willmcgugan: Same issue as above. There possibly are some shortcuts for making a layout with no / 1 child, but that is fast anyway.

tests/test_box_model.py
darrenburns: If the parent wasn't the viewport, wouldn't `get_box_size` need to know what the viewport size is so it can return the correct dimensions?

willmcgugan: The parent is size is always the viewport. The naming is inherited from where it is called in the layout classes, but I can see how that would be confusing looking at `get_box_model` in isolation. Perhaps at this point the two sizes should be named `container` and `viewport` for clarity.

darrenburns: Yeah that would be clearer

willmcgugan: Done!

src/textual/screen.py
```diff
@@ -213,6 +218,9 @@ async def forward_event(self, event: events.Event) -> None:
                 return
             scroll_widget = widget
             if scroll_widget is not None:
-                await scroll_widget.forward_event(event)
+                if scroll_widget is self:
+                    await self.post_message(event)
+                else:
+                    await scroll_widget.forward_event(event)
```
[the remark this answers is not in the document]
darrenburns: @willmcgugan ^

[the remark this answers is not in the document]
willmcgugan: Sorry, missed that one.

This is if you're scrolling with wheel / gesture. The event will be forwarded to screen. The check at L221 prevents it from being forwarded again, and handled by the screen itself.

PR #406, opened 2022-04-20
Fix issue with binding space key

3mb3dw0rk5:
This fixes the issue with binding the space key to an action. The use of `key.strip()` will create an empty string when a space is passed as key parameter.


3mb3dw0rk5:
Duplicate to #143

pvmm:
#410  replaces this fix

PR #408, opened 2022-04-21
Style error improvements

darrenburns:
## Goals
- [ ] Extract styling/CSS error related help text into a common location
- [ ] Contextual error messaging - errors for inline vs CSS should be specific to that use case
- [ ] Abstract error messages into an object structure to ensure consistent help text printing

image

### DIFF
--- a/src/textual/css/_error_tools.py
+++ b/src/textual/css/_error_tools.py
@@ -16,7 +16,7 @@ def friendly_list(words: Iterable[str], joiner: str = "or") -> str:
     Returns:
         str: List as prose.
     """
-    words = [repr(word) for word in sorted(words, key=str.lower)]
+    words = [repr(word) for word in sorted(words, key=str.lower) if word]
     if len(words) == 1:
         return words[0]
     elif len(words) == 2:
--- a/src/textual/css/_help_text.py
+++ b/src/textual/css/_help_text.py
@@ -1,0 +1,31 @@
+from textual.css._error_tools import friendly_list
+from textual.css.scalar import SYMBOL_UNIT
+
+
+def _inline_name(property_name: str) -> str:
+    return property_name.replace("-", "_")
+
+
+def spacing_help_text(property_name: str, num_values_supplied: int) -> str:
+    return f"""\
+• You supplied {num_values_supplied} values for the '{property_name}' property
+• Spacing properties like 'margin' and 'padding' require either 1, 2 or 4 integer values
+• In Textual CSS, supply 1, 2 or 4 values separated by a space
+    [dim]e.g. [/][i]{property_name}: 1 2 3 4;[/]
+• In Python, you can set it to a tuple to assign spacing to each edge
+    [dim]e.g. [/][i]widget.styles.{_inline_name(property_name)} = (1, 2, 3, 4)[/]
+• Or to an integer to assign to all edges at once
+    [dim]e.g. [/][i]widget.styles.{_inline_name(property_name)} = 2[/]"""
+
+
+def scalar_help_text(property_name: str) -> str:
+    return f"""\
+• Invalid value for the '{property_name}' property
+• Scalar properties like '{property_name}' require numerical values and an optional unit
+• Valid units are {friendly_list(SYMBOL_UNIT)}
+• Here's an example of how you'd set a scalar property in Textual CSS
+    [dim]e.g. [/][i]{property_name}: 50%;[/]
+• In Python, you can assign a string, int or Scalar object itself
+    [dim]e.g. [/][i]widget.styles.{_inline_name(property_name)} = "50%"[/]
+    [dim]e.g. [/][i]widget.styles.{_inline_name(property_name)} = 10[/]
+    [dim]e.g. [/][i]widget.styles.{_inline_name(property_name)} = Scalar(...)[/]"""
--- a/src/textual/css/_style_properties.py
+++ b/src/textual/css/_style_properties.py
@@ -9,11 +9,13 @@ when setting and getting.
 
 from __future__ import annotations
 
+import inspect
 from typing import Iterable, NamedTuple, TYPE_CHECKING, cast
 
 import rich.repr
 from rich.style import Style
 
+from ._help_text import spacing_help_text, scalar_help_text
 from ..color import Color, ColorPair
 from ._error_tools import friendly_list
 from .constants import NULL_SPACING
@@ -98,7 +100,10 @@ class ScalarProperty:
             try:
                 new_value = Scalar.parse(value)
             except ScalarParseError:
-                raise StyleValueError("unable to parse scalar from {value!r}")
+                raise StyleValueError(
+                    "unable to parse scalar from {value!r}",
+                    help_text=scalar_help_text(property_name=self.name),
+                )
         else:
             raise StyleValueError("expected float, int, Scalar, or None")
         if new_value is not None and new_value.unit not in self.units:
@@ -368,7 +373,16 @@ class SpacingProperty:
             if obj.clear_rule(self.name):
                 obj.refresh(layout=True)
         else:
-            if obj.set_rule(self.name, Spacing.unpack(spacing)):
+            try:
+                unpacked_spacing = Spacing.unpack(spacing)
+            except ValueError as error:
+                raise StyleValueError(
+                    str(error),
+                    help_text=spacing_help_text(
+                        property_name=self.name, num_values_supplied=len(spacing)
+                    ),
+                )
+            if obj.set_rule(self.name, unpacked_spacing):
                 obj.refresh(layout=True)
 
 
--- a/src/textual/css/_style_properties.py
+++ b/src/textual/css/_style_properties.py
@@ -14,7 +14,6 @@ from typing import Iterable, NamedTuple, TYPE_CHECKING, cast
 import rich.repr
 from rich.style import Style
 
-from .. import log
 from ..color import Color, ColorPair
 from ._error_tools import friendly_list
 from .constants import NULL_SPACING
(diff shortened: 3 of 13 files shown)


src/textual/css/_help_text.py
```diff
@@ -0,0 +1,690 @@
...
+
+    Returns:
+        str: The property name converted to the given context.
+    """
+    return _css_name(property_name) if context == "css" else _python_name(property_name)
+
+
+def _spacing_examples(property_name: str) -> ContextSpecificBullets:
+    """Returns examples for spacing properties"""
+    return ContextSpecificBullets(
```
darrenburns, line 104: Both inline and css will be executed on an error occurring since they aren't lazily evaluated. I thought this was fine since it keeps things readable and it's a very minimal overhead that only appears in error cases.

src/textual/geometry.py
```diff
@@ -631,18 +628,58 @@ def css(self) -> str:
...
         if pad_len == 4:
             top, right, bottom, left = cast(Tuple[int, int, int, int], pad)
             return cls(top, right, bottom, left)
-        raise ValueError(f"1, 2 or 4 integers required for spacing; {pad_len} given")
+        raise ValueError(
+            f"1, 2 or 4 integers required for spacing properties; {pad_len} given"
+        )
+
+    @classmethod
+    def vertical(cls, amount: int) -> Spacing:
```
darrenburns, line 647: I found myself reaching for similar utility constructors in other frameworks quite frequently, so thought they'd be a nice addition.

Handy for readable code too: `Spacing.horizontal(2) + Spacing.vertical(4)` is quite a readable way of getting a spacing with 2 on the horizontal edges, and 4 on the vertical.

tests/css/test_help_text.py
```diff
@@ -0,0 +1,123 @@
+import pytest
```
darrenburns, line 1: These tests aren't as precise as they could be - they ensure the error help text code is exercised and the output looks "reasonable". I think they're at a decent balance between value and maintainability overhead.

src/textual/css/_style_properties.py
```diff
@@ -303,8 +337,8 @@ def __set__(
         elif count == 2:
             _border1, _border2 = border
             setattr(obj, top, _border1)
-            setattr(obj, right, _border1)
-            setattr(obj, bottom, _border2)
+            setattr(obj, bottom, _border1)
+            setattr(obj, right, _border2)
```
darrenburns, line 341: These were the wrong way around

src/textual/css/_help_text.py
```diff
@@ -0,0 +1,690 @@
...
+
+if sys.version_info >= (3, 9):
+    from typing import Annotated
+else:
+    from typing_extensions import Annotated
+
+from textual.css._error_tools import friendly_list
+from textual.css.scalar import SYMBOL_UNIT
+
+StylingContext = Annotated[
```
willmcgugan, line 29: What's the `Annotated` doing? I know it attaches meta data to types, but what uses the string? Does Mypy show it on error?

src/textual/css/_help_text.py
```diff
@@ -0,0 +1,690 @@
...
+
+@dataclass
+class ContextSpecificBullets:
+    """
+    Args:
+        inline (Iterable[Bullet]): Information only relevant to users who are using inline styling.
+        css (Iterable[Bullet]): Information only relevant to users who are using CSS.
+    """
+
+    inline: Iterable[Bullet]
```
willmcgugan, line 46: Some Iterables can be exhausted, so you might not get the same result from `get_by_context` on a second call.

Could these be Sequene or List ?

src/textual/css/_help_text.py
```diff
@@ -0,0 +1,690 @@
...
+                ],
+            ),
+            Bullet(
+                "Or to an integer to assign a single value to all edges",
+                examples=[Example(f"widget.styles.{property_name} = 2")],
+            ),
+        ],
+        css=[
+            Bullet(
+                "In Textual CSS, supply 1, 2 or 4 integers separated by a space",
```
willmcgugan, line 124: The "In Textual CSS" may be redundant.

src/textual/css/_help_text.py
```diff
@@ -0,0 +1,690 @@
...
+    """
+    return _css_name(property_name) if context == "css" else _python_name(property_name)
+
+
+def _spacing_examples(property_name: str) -> ContextSpecificBullets:
+    """Returns examples for spacing properties"""
+    return ContextSpecificBullets(
+        inline=[
+            Bullet(
+                "In Python, you can set it to a tuple to assign spacing to each edge",
```
willmcgugan, line 107: The "In Python" feels redundant, given the file is .css and it shows a snippet of code.

The "it" may be unclear. Could you add {property_name} here, so it reads "You can set {property_name} to a tuple..."

src/textual/css/_style_properties.py
```diff
@@ -279,20 +311,22 @@ def __set__(
...
         """
         top, right, bottom, left = self._properties
-        obj.refresh()
         if border is None:
             clear_rule = obj.clear_rule
             clear_rule(top)
             clear_rule(right)
             clear_rule(bottom)
             clear_rule(left)
+            obj.refresh(layout=self._layout)
```
willmcgugan, line 320: I suspect that clearing the rule, or using `setattr` will ultimately call `obj.refresh`, so this may be redundant. Can you check?

src/textual/css/_styles_builder.py
```diff
@@ -181,10 +221,20 @@ def process_box_sizing(self, name: str, tokens: list[Token]) -> None:
                     self.error(
                         name,
                         token,
-                        f"invalid value for box-sizing (received {value!r}, expected {friendly_list(VALID_BOX_SIZING)})",
+                        string_enum_help_text(
+                            "box-sizing",
+                            valid_values=list(VALID_BOX_SIZING),
```
willmcgugan, line 226: Isn't `VALID_BOX_SIZING` a set? Sets have undefined order. Are we sorting the valid values? The should probably be alphabetical.

src/textual/geometry.py
```diff
@@ -631,18 +628,58 @@ def css(self) -> str:
     def unpack(cls, pad: SpacingDimensions) -> Spacing:
         """Unpack padding specified in CSS style."""
         if isinstance(pad, int):
-            return cls(pad, pad, pad, pad)
+            return Spacing.all(pad)
```
willmcgugan, line 631: I'd rather not use this here. It's a tiny but slower, and I don't think it benefits much from readability. I try to keeps functions at this level as atomic as possible.

src/textual/geometry.py
willmcgugan: I'd probably not use that example myself, feels like a roundabout way of doing `Spacing.unpack(4,2)` But I can't argue it is more readable!

src/textual/layouts/grid.py
```diff
@@ -245,9 +245,6 @@ def _align(
         col_align: GridAlign,
         row_align: GridAlign,
     ) -> Region:
-        offset_x = 0
```
willmcgugan, line 248: This file is going to be removed. I've been keeping it around as a reference.

tests/css/test_help_text.py
willmcgugan: I concur. I imagine we will tweak the working of the errors fairly regularly.

willmcgugan (changes requested):
Looks great! A few comments.

src/textual/css/_styles_builder.py
darrenburns: Most of these are ultimately sorted inside `friendly_list`, but there are one or two places they don't get sorted. I'll add that.

src/textual/css/_style_properties.py
darrenburns: Seems to work without. There was a `refresh` call here before without layout (above this if block - see the diff), I didn't understand what that was for? Do you know?

src/textual/css/_help_text.py
darrenburns: I didn't know how to docstring a module level variable and Google yielded absurd amounts of conflicting info. I've changed to """ underneath the variable now. I don't know how or where mypy uses this.

willmcgugan: It's unfortunate there is no docstring equivalent for such things.

src/textual/css/_error_tools.py
```diff
@@ -16,7 +16,7 @@ def friendly_list(words: Iterable[str], joiner: str = "or") -> str:
     Returns:
         str: List as prose.
     """
-    words = [repr(word) for word in sorted(words, key=str.lower)]
+    words = [repr(word) for word in sorted(words, key=str.lower) if word]
```
willmcgugan, line 19: Why the `if word`? Some times an empty string is a valid response.

src/textual/css/_style_properties.py
willmcgugan: Could be a remnant of a previous system. Managing refreshes is tricky.

src/textual/css/_error_tools.py
darrenburns: It just seemed weird to me, this function is only used for CSS units in error messages and it was strange to say a valid unit is `''`.

willmcgugan: Yeah, it's accurate, but doesn't read well.

Maybe we should handle it a different way, perhaps `valid values are "foo", "baz", or an empty string` ?

darrenburns: The error messages all mention "an optional unit" then go on to say "valid units are a, b, c,...". I think saying `valid units are 'vw', 'vh', '%', or an empty string` doesn't sound great either.

willmcgugan: Fair enough. But lets make it optional to omit empty strings. I'd like to keep `friendly_list` generic enough to use in any context.

PR #416, opened 2022-04-25
fix for showing scrollbars

willmcgugan:
Fix for scrollbars not being enabled on first refresh.

The issue was that scrollbars weren't enabled prior to changing the sizes.

### DIFF
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -253,7 +253,7 @@ class App(DOMNode):
                     key_values = " ".join(
                         f"{key}={value}" for key, value in kwargs.items()
                     )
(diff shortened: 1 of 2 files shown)


src/textual/app.py
```diff
@@ -253,7 +253,7 @@ def log(
                     key_values = " ".join(
                         f"{key}={value}" for key, value in kwargs.items()
                     )
-                    output = " ".join((output, key_values))
+                    output = f"{output} {key_values}" if output else key_values
```
willmcgugan, line 256: Previously logs had an extra space if you only used keywords args.

darrenburns (approved):

PR #428, opened 2022-04-27
[widget] Add a test for types of values we handle in style's size

### LINKED ISSUE #392: Setting inline styles width or height to an integer causes a crash
`self.styles.height = 10` doesn't work, but `self.styles.height = 10.0` does.

Traceback:

```
╭─────────────────────────────────────────────────────── Traceback (most recent call last) ───────────────────────────────────────────────────────╮
│                                                                                                                                                 │
│ /Users/darrenburns/code/textual/src/textual/widget.py:539 in get_render_lines                                                                   │
│                                                                                                                                                 │
│   536 │   │   │   Lines: A list of lists of segments.                                         ╭───── locals ─────╮                              │
│   537 │   │   """                                                                             │   end = None     │                              │
│   538 │   │   if self._dirty_regions:                                                         │  self = Button() │                              │
│ ❱ 539 │   │   │   self._render_lines()                                                        │ start = None     │                              │
│   540 │   │   │   if self.is_container:                                                       ╰──────────────────╯                              │
│   541 │   │   │   │   self.horizontal_scrollbar.refresh()                                                                                       │
│   542 │   │   │   │   self.vertical_scrollbar.refresh()                                                                                         │
│ /Users/darrenburns/code/textual/src/textual/widget.py:520 in _render_lines                                                                      │
│                                                                                                                                                 │
│   517 │   def _render_lines(self) -> None:                                                    ╭───── locals ──────╮                             │
│   518 │   │   """Render all lines."""                                                         │ height = 80       │                             │
│   519 │   │   width, height = self.size                                                       │   self = Button() │                             │
│ ❱ 520 │   │   renderable = self.render_styled()                                               │  width = 20       │                             │
│   521 │   │   options = self.console.options.update_dimensions(width, height)                 ╰───────────────────╯                             │
│   522 │   │   lines = self.console.render_lines(renderable, options)                                                                            │
│   523 │   │   self._render_cache = RenderCache(self.size, lines)                                                                                │
│                                                                                                                                                 │
│ /Users/darrenburns/code/textual/src/textual/widget.py:386 in render_styled                                                                      │
│                                                                                                                                                 │
│   383 │   │   │   RenderableType: A new renderable.                                           ╭──── locals ─────╮                               │
│   384 │   │   """                                                                             │ self = Button() │                               │
│   385 │   │                                                                                   ╰─────────────────╯                               │
│ ❱ 386 │   │   renderable = self.render()                                                                                                        │
│   387 │   │   styles = self.styles                                                                                                              │
│   388 │   │   parent_styles = self.parent.styles                                                                                                │
│   389                                                                                                                                           │
│                                                                                                                                                 │
│ /Users/darrenburns/code/textual/src/textual/widgets/_button.py:53 in render                                                                     │
│                                                                                                                                                 │
│   50 │                                                                                       ╭───── locals ─────╮                               │
│   51 │   def render(self) -> RenderableType:                                                 │  self = Button() │                               │
│   52 │   │   width = cell_len(self.label)                                                    │ width = 9        │                               │
│ ❱ 53 │   │   self.styles.width = width                                                       ╰──────────────────╯                               │
│   54 │   │   return ButtonRenderable(                                                                                                           │
│   55 │   │   │   self.label,                                                                                                                    │
│   56 │   │   )                                                                                                                                  │
│                                                                                                                                                 │
│ /Users/darrenburns/code/textual/src/textual/css/_style_properties.py:103 in __set__                                                             │
│                                                                                                                                                 │
│   100 │   │   │   except ScalarParseError:                                                                                                      │
│   101 │   │   │   │   raise StyleValueError("unable to parse scalar from {value!r}")                                                            │
│   102 │   │   else:                                                                                                                             │
│ ❱ 103 │   │   │   raise StyleValueError("expected float, Scalar, or None")                                                                      │
│   104 │   │   if new_value is not None and new_value.unit not in self.units:                                                                    │
│   105 │   │   │   raise StyleValueError(                                                                                                        │
│   106 │   │   │   │   f"{self.name} units must be one of {friendly_list(get_symbols(self.unit                                                   │
│                                                                                                                                                 │
│ ╭───────────────────────────────────────── locals ─────────────────────────────────────────╮                                                    │
│ │   obj = RenderStyles(                                                                    │                                                    │
│ │         │   color=Color(192, 192, 192, a=1.0),                                           │                                                    │
│ │         │   background=Color(0, 0, 128, a=1.0),                                          │                                                    │
│ │         │   border_top=('solid', Color(0, 128, 0, a=1.0)),                               │                                                    │
│ │         │   border_right=('solid', Color(0, 128, 0, a=1.0)),                             │                                                    │
│ │         │   border_bottom=('solid', Color(0, 128, 0, a=1.0)),                            │                                                    │
│ │         │   border_left=('solid', Color(0, 128, 0, a=1.0)),                              │                                                    │
│ │         │   width=Scalar(                                                                │                                                    │
│ │         │   │   value=20.0,                                                              │                                                    │
│ │         │   │   unit=<Unit.CELLS: 1>,                                                    │                                                    │
│ │         │   │   percent_unit=<Unit.WIDTH: 4>                                             │                                                    │
│ │         │   )                                                                            │                                                    │
│ │         )                                                                                │                                                    │
│ │  self = <textual.css._style_properties.ScalarProperty object at 0x101e62860>             │                                                    │
│ │ value = 9                                                                                │                                                    │
│ ╰──────────────────────────────────────────────────────────────────────────────────────────╯                                                    │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
StyleValueError: expected float, Scalar, or None
```

DrBenton:
closes #392 

I started by writing a test that reproduces the bug - but to my surprise when I ran the test that passes an integer as a width I saw that the test was passing right away, without me having to do anything... 
After a quick inspection, it turns out that this bug was already fixed, by Will, the day after it was reported :smile: 
> https://github.com/Textualize/textual/commit/cecbf655abf2a99aeabab4a4d9a64530a882e2fc#diff-0797b08e8ce7e76ca51990c546ef8f392404156a349414b9834530a878070b95R98

However as I'm now here with my test already written, I think we could add it to the tests suite? :slightly_smiling_face:

### DIFF
--- a/src/textual/css/_style_properties.py
+++ b/src/textual/css/_style_properties.py
@@ -115,12 +115,12 @@ class ScalarProperty:
         ):
             raise StyleValueError("'auto' not allowed here")
 
-        if new_value.unit != Unit.AUTO:
-            if new_value is not None and new_value.unit not in self.units:
+        if new_value is not None and new_value.unit != Unit.AUTO:
+            if new_value.unit not in self.units:
                 raise StyleValueError(
                     f"{self.name} units must be one of {friendly_list(get_symbols(self.units))}"
                 )
-            if new_value is not None and new_value.is_percent:
+            if new_value.is_percent:
                 new_value = Scalar(
                     float(new_value.value), self.percent_unit, Unit.WIDTH
                 )
--- a/tests/test_widget.py
+++ b/tests/test_widget.py
@@ -1,16 +1,22 @@
+from contextlib import nullcontext as does_not_raise
+from decimal import Decimal
+
 import pytest
 
 from textual.css.errors import StyleValueError
+from textual.css.scalar import Scalar, Unit
 from textual.widget import Widget
 
 
 @pytest.mark.parametrize(
-    "set_val, get_val, style_str", [
+    "set_val, get_val, style_str",
+    [
         [True, True, "visible"],
         [False, False, "hidden"],
         ["hidden", False, "hidden"],
         ["visible", True, "visible"],
-    ])
+    ],
+)
 def test_widget_set_visible_true(set_val, get_val, style_str):
     widget = Widget()
     widget.visible = set_val
@@ -26,3 +32,29 @@ def test_widget_set_visible_invalid_string():
         widget.visible = "nope! no widget for me!"
 
     assert widget.visible
+
+
+@pytest.mark.parametrize(
+    "size_input, expectation",
+    [
+        (None, does_not_raise()),
+        (10, does_not_raise()),
+        (10.0, does_not_raise()),
+        (10.2, does_not_raise()),
+        (Scalar(100, Unit.CELLS, Unit.WIDTH), does_not_raise()),
+        (Scalar(10.2, Unit.CELLS, Unit.WIDTH), does_not_raise()),
+        ("10", does_not_raise()),
+        # And now for some common types we don't handle...
+        ("a", pytest.raises(StyleValueError)),
+        (list(), pytest.raises(StyleValueError)),
+        (tuple(), pytest.raises(StyleValueError)),
+        (dict(), pytest.raises(StyleValueError)),
+        (3.14j, pytest.raises(StyleValueError)),
+        (Decimal("3.14"), pytest.raises(StyleValueError)),
+    ],
+)
+def test_widget_style_size_can_accept_various_data_types(size_input, expectation):
+    widget = Widget()
+
+    with expectation:
+        widget.styles.width = size_input
--- a/tests/test_widget.py
+++ b/tests/test_widget.py
@@ -32,29 +32,3 @@ def test_widget_set_visible_invalid_string():
         widget.visible = "nope! no widget for me!"
 
     assert widget.visible
-
-
-@pytest.mark.parametrize(
-    "size_input, expectation",
-    [
-        (None, does_not_raise()),
-        (10, does_not_raise()),
-        (10.0, does_not_raise()),
-        (10.2, does_not_raise()),
-        (Scalar(100, Unit.CELLS, Unit.WIDTH), does_not_raise()),
-        (Scalar(10.2, Unit.CELLS, Unit.WIDTH), does_not_raise()),
-        ("10", does_not_raise()),
-        # And now for some common types we don't handle...
-        ("a", pytest.raises(StyleValueError)),
-        (list(), pytest.raises(StyleValueError)),
-        (tuple(), pytest.raises(StyleValueError)),
-        (dict(), pytest.raises(StyleValueError)),
-        (3.14j, pytest.raises(StyleValueError)),
-        (Decimal("3.14"), pytest.raises(StyleValueError)),
-    ],
-)
-def test_widget_style_size_can_accept_various_data_types(size_input, expectation):
-    widget = Widget()
-
-    with expectation:
-        widget.styles.width = size_input
--- a/tests/css/test_styles.py
+++ b/tests/css/test_styles.py
@@ -1,11 +1,15 @@
+from decimal import Decimal
+
 import pytest
 
 from rich.style import Style
 
 from textual.color import Color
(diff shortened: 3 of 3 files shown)


tests/test_widget.py
```diff
@@ -1,16 +1,22 @@
...
 import pytest
 
 from textual.css.errors import StyleValueError
+from textual.css.scalar import Scalar, Unit
 from textual.widget import Widget
 
 
 @pytest.mark.parametrize(
-    "set_val, get_val, style_str", [
+    "set_val, get_val, style_str",
```
DrBenton, line 12: Black formatting was not applied on this file.. It's the case now that I edited it :slightly_smiling_face:

tests/test_widget.py
```diff
@@ -26,3 +32,29 @@ def test_widget_set_visible_invalid_string():
...
+    [
+        (None, does_not_raise()),
+        (10, does_not_raise()),
+        (10.0, does_not_raise()),
+        (10.2, does_not_raise()),
+        (Scalar(100, Unit.CELLS, Unit.WIDTH), does_not_raise()),
+        (Scalar(10.2, Unit.CELLS, Unit.WIDTH), does_not_raise()),
+        ("10", does_not_raise()),
+        # And now for some common types we don't handle...
+        ("a", pytest.raises(StyleValueError)),
```
DrBenton, line 48: Let's not try every possible data type here of course, but just a few ones should allow us to check that the type guard we have in `_style_properties.__set__` does it job? :slightly_smiling_face:

tests/test_widget.py
```diff
@@ -26,3 +32,29 @@ def test_widget_set_visible_invalid_string():
...
+        (list(), pytest.raises(StyleValueError)),
+        (tuple(), pytest.raises(StyleValueError)),
+        (dict(), pytest.raises(StyleValueError)),
+        (3.14j, pytest.raises(StyleValueError)),
+        (Decimal("3.14"), pytest.raises(StyleValueError)),
+    ],
+)
+def test_widget_style_size_can_accept_various_data_types(size_input, expectation):
```
darrenburns, line 56: I'd say we want to do more than just check that this doesn't raise an exception. For each supported datatype, we should assert that the value we read back from the property is what we would expect.

For example, if I do `widget.styles.width = 10` vs `widget.styles.width = 10.0`, then read `widget.styles.width`, will I get the same value back?

Personally I would keep this test for just the common datatypes we don't support, and have separate tests for checking we write and read the expected values for different types.

I'd put this in `test_styles` rather than `test_widget` too, since that file already contains test of that sort.

willmcgugan: Agree that it probably does belong in `test_styles.py`. 

Suggest we flesh out tests like this in the near future, when the Styles object is closer to being finalized.

willmcgugan (approved):

src/textual/css/_style_properties.py
```diff
@@ -115,12 +115,12 @@ def __set__(
         ):
             raise StyleValueError("'auto' not allowed here")
 
-        if new_value.unit != Unit.AUTO:
-            if new_value is not None and new_value.unit not in self.units:
+        if new_value is not None and new_value.unit != Unit.AUTO:
```
DrBenton, line 118: PyCharm was highlighting the fact that `if new_value.unit != Unit.AUTO`  will raise an exception if "new_value" is `None` anyway, so I guess we'd better wrap the whole block under a "is not None" guard?

DrBenton:
Thanks for your comments! :slightly_smiling_face: 
Hopefully this commit addresses them: [f7e72ed](https://github.com/Textualize/textual/pull/428/commits/f7e72ed94583087a2a851d75b81f14529c131e5b) :crossed_fingers:

darrenburns (approved):

willmcgugan (approved):

PR #432, opened 2022-04-27
button widget

willmcgugan:
- Implements a very basic button widget
- Drops DEFAULT_STYLES in favor of a CSS class var
- Changed type of `classes` argument in Widget from a set to a space-separated list, which is often easier to type, especially with the typical case of a single class.
- Adds App.exit which exits the app with a return value 

https://user-images.githubusercontent.com/554369/165757614-a91a690a-1891-45e0-8c56-70a0ca4a4014.mov

### DIFF
--- a/sandbox/basic.css
+++ b/sandbox/basic.css
@@ -41,6 +41,7 @@ App > Screen {
   background: $primary-darken-2;
   color: $text-primary-darken-2 ;
   border-right: outer $primary-darken-3;
+  content-align: center middle;
 }
 
 #sidebar .user {
@@ -48,19 +49,21 @@ App > Screen {
   background: $primary-darken-1;
   color: $text-primary-darken-1;
   border-right: outer $primary-darken-3;
+  content-align: center middle;
 }
 
 #sidebar .content {
   background: $primary;
   color: $text-primary;
   border-right: outer $primary-darken-3;
+  content-align: center middle;
 }
 
 #header {
   color: $text-primary-darken-1;
   background: $primary-darken-1;
-  height: 3
-
+  height: 3;
+  content-align: center middle;
 }
 
 #content {
@@ -84,6 +87,7 @@ Tweet {
   border: wide $panel-darken-2;
   overflow-y: scroll; 
   align-horizontal: center;
+  
 }
 
 .scrollable {
@@ -152,6 +156,7 @@ TweetBody {
   background: $accent;
   height: 1;
   border-top: hkey $accent-darken-2;
+  content-align: center middle;
 }
 
 
@@ -165,6 +170,7 @@ OptionItem {
   transition: background 100ms linear;
   border-right: outer $primary-darken-2;
   border-left: hidden;
+  content-align: center middle;
 }
 
 OptionItem:hover {
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -9,7 +9,7 @@ import warnings
 from asyncio import AbstractEventLoop
 from contextlib import redirect_stdout
 from time import perf_counter
-from typing import Any, Iterable, TextIO, Type, TYPE_CHECKING
+from typing import Any, Generic, Iterable, TextIO, Type, TypeVar, TYPE_CHECKING
 
 import rich
 import rich.repr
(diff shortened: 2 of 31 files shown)


sandbox/basic.css
```diff
@@ -41,26 +41,29 @@ App > Screen {
   background: $primary-darken-2;
   color: $text-primary-darken-2 ;
   border-right: outer $primary-darken-3;
+  content-align: center middle;
```
willmcgugan, line 44: New rule to align content within a widget.

src/textual/app.py
```diff
@@ -76,8 +78,11 @@ class ActionError(Exception):
     pass
 
 
+ReturnType = TypeVar("ReturnType")
+
+
 @rich.repr.auto
-class App(DOMNode):
+class App(Generic[ReturnType], DOMNode):
```
willmcgugan, line 85: This allows you to specify the return type on `run()` So if you want to return an int you might do this:

```python
class MyApp(App[int]):
```

src/textual/color.py
```diff
@@ -253,7 +258,9 @@ def parse(cls, color_text: str | Color) -> Color:
         """
         if isinstance(color_text, Color):
             return color_text
-        ansi_color = ANSI_COLOR_TO_RGB.get(color_text)
+        if color_text == "transparent":
+            return TRANSPARENT
```
willmcgugan, line 262: Kind of a psuedo-color. Might roll this in to the color names mapping at some point.

src/textual/widgets/_button.py
```diff
@@ -1,67 +1,75 @@
...
+    
+    Button {
+        width: auto;
+        height: 3;
+        padding: 0 2;
+        background: $primary;
+        color: $text-primary;
+        content-align: center middle;
+        border: tall $primary-lighten-3;        
+        margin: 1;
```
darrenburns, line 27: IMO Textual widgets should have 0 margin by default, and it's up to the user of the widget if they want to supply margin based on the context they wish to use the widget in.

src/textual/widgets/_button.py
```diff
@@ -1,67 +1,75 @@
...
+        min-width:16;
```
darrenburns, line 28: I'd maybe leave this to the user as well?

If I were to drop a button into my app and wanted it to be small e.g. a button with "Yes" or "No", I would need to know that I need to override the "min-width" property which isn't intuitive to me.

It also means I wouldn't be able to set the `width` property to anything below 16, right? It seems `min-width` takes precedence.

darrenburns (changes requested):
Some comments, questions, suggestions. Also this PR seems to break hot reloading somehow - on making a CSS change on any sandbox app my screen just goes black.

src/textual/app.py
willmcgugan: Came from a Textual issue I can't seem to find now. But a user wan't to be able to launch TUIs within a larger app to get info from the user. The data could be an int, str, or something more complex.

src/textual/widgets/_button.py
willmcgugan: Done. Although without a min-width, the buttons may be all different sizes which may not look good.

src/textual/widgets/_button.py
willmcgugan: Without margin on the vertical, the buttons bump up against each other on the vertical axis, which doesn't look good. But you don't get that on the horizontal axis. So I've set margin to 1 0.

This CSS is a starting point though. Plenty of tweaking to do.

sandbox/basic.css
willmcgugan: align moves the Widget, content-align moves the text inside a widget. The docs will have nice diagrams to show the difference.

I would expect there to be little reason to adjust content-align for a pre-built widgets. For instance a button will likely always have content-align center middle.

DrBenton: It's very subjective, and I'm probably bike-shedding :sweat_smile: , but... wouldn't something like `justify`, `text-justify`, `justify-items` or `justify-content` be more differentiated to `align` than `content-align`? (they all are existing CSS property names)

PR #436, opened 2022-04-27
[layout][bugfix] Horizontal & Vertical layouts shouldn't display children that have `display: none`

### LINKED ISSUE #394: `display: none` is being ignored in vertical and horizontal layouts
Example app:

```python
from textual.app import App

class ButtonsApp(App):

    def on_load(self) -> None:
        self.bind("q", "quit")

    def on_mount(self):
        self.mount(
            Widget(classes={"no-display"})
        )


ButtonsApp.run(css_file="buttons.scss", log="textual.log")
```

Example CSS:
```scss
Widget.no-display {
  display: none;
}
```

The widget is unexpectedly displayed.

DrBenton:
fixes #394

### DIFF
--- a/src/textual/_node_list.py
+++ b/src/textual/_node_list.py
@@ -18,8 +18,8 @@ class NodeList:
 
     """
 
-    def __init__(self, nodes: list[DOMNode] = None) -> None:
-        self._nodes: list[DOMNode] = [] if nodes is None else nodes
+    def __init__(self) -> None:
+        self._nodes: list[DOMNode] = []
 
     def __bool__(self) -> bool:
         return bool(self._nodes)
--- a/src/textual/_node_list.py
+++ b/src/textual/_node_list.py
@@ -18,8 +18,8 @@ class NodeList:
 
     """
 
-    def __init__(self) -> None:
-        self._nodes: list[DOMNode] = []
+    def __init__(self, nodes: list[DOMNode] = None) -> None:
+        self._nodes: list[DOMNode] = [] if nodes is None else nodes
 
     def __bool__(self) -> bool:
         return bool(self._nodes)
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -521,6 +521,7 @@ class App(DOMNode):
                 mount_event = events.Mount(sender=self)
                 await self.dispatch_message(mount_event)
 
+                # TODO: don't override `self.console` here
                 self.console = Console(file=sys.__stdout__)
                 self.title = self._title
                 self.refresh()
--- a/src/textual/css/styles.py
+++ b/src/textual/css/styles.py
@@ -70,7 +70,7 @@ if TYPE_CHECKING:
 class RulesMap(TypedDict, total=False):
     """A typed dict for CSS rules.
 
-    Any key may be absent, indiciating that rule has not been set.
+    Any key may be absent, indicating that rule has not been set.
 
     Does not define composite rules, that is a rule that is made of a combination of other rules.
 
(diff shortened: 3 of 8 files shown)


src/textual/_node_list.py
```diff
@@ -18,8 +18,8 @@ class NodeList:
 
     """
 
-    def __init__(self) -> None:
-        self._nodes: list[DOMNode] = []
+    def __init__(self, nodes: list[DOMNode] = None) -> None:
```
DrBenton, line 21: I was thinking that being able to optionally provide an initial list of nodes to this constructor could be useful?
I can revert this change if you think it's better to not do this of course :slightly_smiling_face: 

(I checked the [NodeList](https://devdocs.io/dom/nodelist) Web API, but there is no constructor on this one anyway so we cannot do any features comparison here :sweat_smile: )

src/textual/app.py
```diff
@@ -521,6 +521,7 @@ async def process_messages(self) -> None:
                 mount_event = events.Mount(sender=self)
                 await self.dispatch_message(mount_event)
 
+                # TODO: don't override `self.console` here
```
DrBenton, line 524: as discussed together :-)

src/textual/css/styles.py
```diff
@@ -70,7 +70,7 @@
 class RulesMap(TypedDict, total=False):
     """A typed dict for CSS rules.
 
-    Any key may be absent, indiciating that rule has not been set.
+    Any key may be absent, indicating that rule has not been set.
```
DrBenton, line 73: [while I was there] As I was digging into the code of Textual I noticed this typo, highlighted by PyCharm

src/textual/layouts/horizontal.py
```diff
@@ -39,7 +39,9 @@ def arrange(
 
         x = box_models[0].margin.left if box_models else 0
 
-        for widget, box_model, margin in zip(parent.children, box_models, margins):
+        displayed_children = parent.displayed_children
```
DrBenton, line 42: Contrary to `parent.children`, `parent.displayed_children` is a Python property getter; with this alias we'll trigger this property getter only once, and also makes the following code a wee bit easier to read

tests/layouts/test_common_layout_features.py
```diff
@@ -0,0 +1,30 @@
...
+
+@pytest.mark.parametrize(
+    "layout,display,expected_in_displayed_children",
+    [
+        ("dock", "block", True),
+        ("horizontal", "block", True),
+        ("vertical", "block", True),
+        ("dock", "none", False),
+        ("horizontal", "none", False),
+        ("vertical", "none", False),
```
DrBenton, line 15: I noticed the existence of a `GridLayout`, but it's not managed in the layouts factory so I reckon it's a Work In Progress API?

darrenburns: It's an older layout that I think we're scrapping

src/textual/dom.py
```diff
@@ -278,6 +278,10 @@ def add_children(tree, node):
         add_children(tree, self)
         return tree
 
+    @property
+    def displayed_children(self) -> NodeList:
```
willmcgugan, line 282: I don't think this needs to be a `NodeList`, which is used to control how the dev adds widgets. Plain old List is fine.

src/textual/_node_list.py
willmcgugan: Not required ATM

willmcgugan (changes requested):

DrBenton:
@willmcgugan Thanks for your feedback! :slightly_smiling_face: 
--> addressed there: [72555f4](https://github.com/Textualize/textual/pull/436/commits/72555f4a77b594c21125b170d4dae75433db8955) (:crossed_fingers:)

willmcgugan (approved):

PR #441, opened 2022-04-28
[css][bugfix] CSS colors can now have digits at the end of their names

### LINKED ISSUE #430: [textual][bug] CSS rule parsing fails when the name of the colour we pass contains a digit
So while this is working correctly:
```css
#my_widget {
  background: dark_cyan;
}
```

...this fails:
```css
#my_widget {
  background: turquoise4;
}
```
...with the following error:
```
 • failed to parse color 'turquoise';                                                                                                                                         
 • failed to parse 'turquoise' as a color;  
```
(maybe just a regex that doesn't take into account the fact that colour names can include numbers?)

DrBenton:
closes #430 
_(me putting the "good first issue" label on this issue was clearly an act of naivety, as it turned out to be more complicated than I thought :sweat_smile:)_

The following CSS declaration is no longer raising an exception:
```css
#my_widget {
  background: turquoise4;
}
```

N.B. Tests are not passing at the moment, but it's only because of this aiohttp-on-Mac temporary bug

### DIFF
--- a/src/textual/css/_styles_builder.py
+++ b/src/textual/css/_styles_builder.py
@@ -461,10 +461,25 @@ class StylesBuilder:
     def process_color(self, name: str, tokens: list[Token]) -> None:
         """Processes a simple color declaration."""
         name = name.replace("-", "_")
-        for token in tokens:
+        token_indexes_to_skip = []
+        for i, token in enumerate(tokens):
+            if i in token_indexes_to_skip:
+                continue
             if token.name in ("color", "token"):
+                value = token.value
+                # Color names can include digits (e.g. "turquoise4"): if the next token is a number
+                # we should consider it part of the color name:
+                next_token = tokens[i + 1] if len(tokens) > (i + 1) else None
+                if (
+                    next_token
+                    and next_token.name == "number"
+                    and next_token.location[1] == token.location[1] + len(value)
+                ):
+                    value = value + next_token.value
+                    # skip next token, as we included it in this one's value:
+                    token_indexes_to_skip.append(i + 1)
                 try:
-                    self.styles._rules[name] = Color.parse(token.value)
+                    self.styles._rules[name] = Color.parse(value)
                 except Exception as error:
                     self.error(
                         name, token, f"failed to parse color {token.value!r}; {error}"
--- a/src/textual/css/_styles_builder.py
+++ b/src/textual/css/_styles_builder.py
@@ -461,25 +461,10 @@ class StylesBuilder:
     def process_color(self, name: str, tokens: list[Token]) -> None:
         """Processes a simple color declaration."""
         name = name.replace("-", "_")
-        token_indexes_to_skip = []
-        for i, token in enumerate(tokens):
-            if i in token_indexes_to_skip:
-                continue
+        for token in tokens:
             if token.name in ("color", "token"):
-                value = token.value
-                # Color names can include digits (e.g. "turquoise4"): if the next token is a number
-                # we should consider it part of the color name:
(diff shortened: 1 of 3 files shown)


tests/css/test_stylesheet.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+@pytest.mark.parametrize(
+    "css_value,expectation,expected_color",
+    [
+        # Valid values:
+        ["red", does_not_raise(), Color(128, 0, 0)],
+        ["dark_cyan", does_not_raise(), Color(0, 175, 135)],
+        ["medium_turquoise", does_not_raise(), Color(95, 215, 215)],
+        ["turquoise4", does_not_raise(), Color(0, 135, 135)],
```
DrBenton, line 16: that's the case that was not initially passing when I wrote this test :-)

src/textual/css/_styles_builder.py
```diff
@@ -461,10 +461,25 @@ def process_layout(self, name: str, tokens: list[Token]) -> None:
...
+        token_indexes_to_skip = []
+        for i, token in enumerate(tokens):
+            if i in token_indexes_to_skip:
+                continue
             if token.name in ("color", "token"):
+                value = token.value
+                # Color names can include digits (e.g. "turquoise4"): if the next token is a number
+                # we should consider it part of the color name:
+                next_token = tokens[i + 1] if len(tokens) > (i + 1) else None
+                if (
```
DrBenton, line 473: I went for a simple and naive implementation, but don't hesitate to tell me if you think of better ways to solve this :-)

tests/css/test_stylesheet.py
```diff
@@ -0,0 +1,48 @@
...
+        ["#ffcc00", does_not_raise(), Color(255, 204, 0)],
+        ["#ffcc0033", does_not_raise(), Color(255, 204, 0, 0.2)],
+        ["rgb(200,90,30)", does_not_raise(), Color(200, 90, 30)],
+        ["rgba(200,90,30,0.3)", does_not_raise(), Color(200, 90, 30, 0.3)],
+        # Some invalid ones:
+        ["coffee", pytest.raises(StylesheetParseError), None],  # invalid color name
+        ["turquoise 4", pytest.raises(StylesheetParseError), None],  # space in it
+        ["1", pytest.raises(StylesheetParseError), None],  # invalid value
+        ["()", pytest.raises(TokenizeError), None],  # invalid tokens
+        # TODO: implement hex colors with 3 chars? @link https://devdocs.io/css/color_value
```
DrBenton, line 26: I noticed some cases of colour values that I was expecting to be able to use as a Web developer, and that are not supported at the minute.
But that is likely not a priority, hence the TODOs

src/textual/css/_styles_builder.py
willmcgugan: This probably is not the fix. In tokenize.py there is a regex which parses a token as a sequence of characters, not including a number.

I think what we should do is change that so that a token must start with a-zA-z but permits digits in further characters.

willmcgugan (changes requested):

src/textual/css/_styles_builder.py
DrBenton: That's much better indeed! I didn't know if that would break existing behaviours or not, which is why I went for a very color-specific local fix. 
But if we can accept that "tokens" can have digits after their first char without breakinf anything, then the fix is much cleaner for sure :blush: 
Fixed in [3ed17cb](https://github.com/Textualize/textual/pull/441/commits/3ed17cb406fb48d93d240f1ef7cc9af5f1b37e5e)

willmcgugan (approved):

PR #443, opened 2022-04-28
[colors] Add management of named Web colors to our CSS processing, prefix ANSI ones with `ansi_`

### LINKED ISSUE #442: [colours] Prefix ANSI colours with `ansi_`, and add management of named Web colours (which would then become the main source of named colours in CSS)

DrBenton:
Submitting this PR as a quick preview, since I realised that it brings an inconsistency issue between Rich and Textual: if we prefix all ANSI colour names with  `ansi_` in Textual we may have unexpected behaviours when we use the underlying Rich primitives in Textual - as Rich still have only ANSI colours, without the prefix in their names.

In the Textual DevTools for example, we use `padding = Segment("▌", Style.parse("bright_magenta"))` : I initially replaced _"bright_magenta"_ with _"ansi_bright_magenta"_ there, but doing so breaks the DevTools because the `Style` class is a Rich component, not a Textual one :grimacing: 

What do you folks think? @willmcgugan @darrenburns 

closes #442


DrBenton:
@willmcgugan As discussed, I merged the first 16 ANSI colors with the named Web ones, and added the `transparent` pseudo-colour [eedec6c](https://github.com/Textualize/textual/pull/443/commits/eedec6c8630af01c83ae905ac6b45d8071faa782) :slightly_smiling_face:

src/textual/_color_constants.py
```diff
@@ -1,239 +1,172 @@
 from __future__ import annotations
 
-ANSI_COLOR_TO_RGB: dict[str, tuple[int, int, int]] = {
+COLOR_NAME_TO_RGB: dict[str, tuple[int, int, int] | tuple[int, int, int, int]] = {
```
willmcgugan, line 3: The alpha component is a float between 0 and 1

willmcgugan (changes requested):
Great. Just a typing request.

src/textual/_color_constants.py
willmcgugan: NM fixed in merge

willmcgugan (approved):

willmcgugan:
👍

PR #445, opened 2022-04-29
[css] Add a "did you mean" suggestion when an unknown CSS property is spotted

### LINKED ISSUE #431: Did you Mean?
I'm a big fan of when software detects typos and offers suggestions.

We should have a function in the project which uses [Levenshtein Word distance](https://en.wikipedia.org/wiki/Levenshtein_distance) to offer suggestions when a symbol is not found.

The first application of this should probably be CSS declarations. i.e. if I use the UK spelling of "colour" it the error would contain a message to the effect of `Did you mean "color" ?`

However I would imagine there are multiple uses for this.

IIRC there is a routine in the standard library that already does this. Might want to google for that.

DrBenton:
So using `bckgroundu: red` in a CSS file reports _"unknown declaration 'bckgroundu'; did you mean 'background'?"_, while using a very wrongly typed CSS property doesn't suggest any.
Screenshot from 2022-04-29 12-14-07

(yeah, I'm afraid italics markups rendering is not as good in the Gnome terminal than in iTerm :sweat_smile: )

closes #431

### DIFF
--- a/src/textual/css/_help_text.py
+++ b/src/textual/css/_help_text.py
@@ -128,7 +128,32 @@ def _spacing_examples(property_name: str) -> ContextSpecificBullets:
     )
 
 
-def spacing_wrong_number_of_values(
+def property_invalid_value_help_text(
+    property_name: str, context: StylingContext, *, suggested_property_name: str = None
+) -> HelpText:
+    """Help text to show when the user supplies an invalid value for CSS property
+    property.
+
+    Args:
+        property_name (str): The name of the property
+        context (StylingContext | None): The context the spacing property is being used in.
+    Keyword Args:
+        suggested_property_name (str | None): A suggested name for the property (e.g. "width" for "wdth"). Defaults to None.
+
+    Returns:
+        HelpText: Renderable for displaying the help text for this property
+    """
+    property_name = _contextualize_property_name(property_name, context)
+    bullets = []
+    if suggested_property_name:
+        suggested_property_name = _contextualize_property_name(
+            suggested_property_name, context
+        )
+        bullets.append(Bullet(f"Did you mean [i]{suggested_property_name}[/]?"))
+    return HelpText(f"Invalid CSS property [i]{property_name}[/]", bullets=bullets)
+
+
+def spacing_wrong_number_of_values_help_text(
     property_name: str,
     num_values_supplied: int,
     context: StylingContext,
@@ -159,7 +184,7 @@ def spacing_wrong_number_of_values(
     )
 
 
-def spacing_invalid_value(
+def spacing_invalid_value_help_text(
     property_name: str,
     context: StylingContext,
 ) -> HelpText:
--- a/src/textual/css/_help_text.py
+++ b/src/textual/css/_help_text.py
@@ -149,7 +149,7 @@ def property_invalid_value_help_text(
         suggested_property_name = _contextualize_property_name(
             suggested_property_name, context
         )
-        bullets.append(Bullet(f"Did you mean [i]{suggested_property_name}[/]?"))
+        bullets.append(Bullet(f'Did you mean "{suggested_property_name}"?'))
     return HelpText(f"Invalid CSS property [i]{property_name}[/]", bullets=bullets)
 
 
--- a/src/textual/css/_styles_builder.py
+++ b/src/textual/css/_styles_builder.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+from difflib import get_close_matches
+from functools import lru_cache
 from typing import cast, Iterable, NoReturn
 
 import rich.repr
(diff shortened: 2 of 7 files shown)


src/textual/css/_help_text.py
```diff
@@ -127,7 +128,31 @@ def _spacing_examples(property_name: str) -> ContextSpecificBullets:
     )
 
 
-def spacing_wrong_number_of_values(
+def property_invalid_value_help_text(
```
DrBenton, line 131: a new HelpText factory, added to the ones already created by Darren :-)

src/textual/css/_help_text.py
```diff
@@ -158,7 +184,7 @@ def spacing_wrong_number_of_values(
     )
 
 
-def spacing_invalid_value(
+def spacing_invalid_value_help_text(
```
DrBenton, line 187: While I was exploring this "HelpText factories" machinery I noticed that contrary to the others the first two ones didn't have the `help_text` suffix - for teh sake of consistency, I added it :slightly_smiling_face:

src/textual/css/_styles_builder.py
```diff
@@ -332,14 +349,20 @@ def _process_space(self, name: str, tokens: list[Token]) -> None:
                 try:
                     append(int(value))
                 except ValueError:
-                    self.error(name, token, spacing_invalid_value(name, context="css"))
+                    self.error(
+                        name,
+                        token,
+                        spacing_invalid_value_help_text(name, context="css"),
```
DrBenton, line 355: just a "Black formatting" consequence of adding the missing `_help_text` suffix to these function names

tests/css/test_help_text.py
```diff
@@ -1,10 +1,22 @@
 import pytest
 
 from tests.utilities.render import render
-from textual.css._help_text import spacing_wrong_number_of_values, spacing_invalid_value, scalar_help_text, \
-    string_enum_help_text, color_property_help_text, border_property_help_text, layout_property_help_text, \
-    docks_property_help_text, dock_property_help_text, fractional_property_help_text, offset_property_help_text, \
-    align_help_text, offset_single_axis_help_text, style_flags_property_help_text
+from textual.css._help_text import (
```
DrBenton, line 4: here again, just the consequences of adding the `_help_text` suffix on these 2 function names

src/textual/css/_help_text.py
```diff
@@ -128,7 +128,32 @@ def _spacing_examples(property_name: str) -> ContextSpecificBullets:
...
+    Returns:
+        HelpText: Renderable for displaying the help text for this property
+    """
+    property_name = _contextualize_property_name(property_name, context)
+    bullets = []
+    if suggested_property_name:
+        suggested_property_name = _contextualize_property_name(
+            suggested_property_name, context
+        )
+        bullets.append(Bullet(f"Did you mean [i]{suggested_property_name}[/]?"))
```
willmcgugan, line 152: I think we might want to drop the italics and use quotes instead. i.e. `Did you mean "background"?`

src/textual/css/_styles_builder.py
```diff
@@ -717,3 +744,9 @@ def process_align_vertical(self, name: str, tokens: list[Token]) -> None:
     process_content_align = process_align
     process_content_align_horizontal = process_align_horizontal
     process_content_align_vertical = process_align_vertical
+
+    def _suggested_property_name_for_rule(self, rule_name: str) -> str | None:
```
willmcgugan, line 748: Docstring por fovor.

src/textual/css/_styles_builder.py
```diff
@@ -717,3 +744,9 @@ def process_align_vertical(self, name: str, tokens: list[Token]) -> None:
...
+        possible_matches = get_close_matches(
```
willmcgugan, line 749: Can we factor this function in to a new module. Probably `suggestions.py`. 

I'm thinking two functions, one for a single suggestion, and another for multiple. With signatures along these lines:

```python
def get_suggestion(word:str, possible_words:list[str]) -> str | None:
```

```python
def get_suggestions(word:str possible_words:list[str], count:int) -> list[str]:
```

src/textual/css/_styles_builder.py
```diff
@@ -84,24 +87,38 @@ def add_declaration(self, declaration: Declaration) -> None:
...
+            self.styles.important.add(rule_name)
+        try:
+            process_method(declaration.name, tokens)
+        except DeclarationError:
+            raise
+        except Exception as error:
+            self.error(declaration.name, declaration.token, str(error))
+
+    @lru_cache(maxsize=None)
+    def _processable_rule_names(self) -> frozenset[str]:
```
willmcgugan, line 118: Best to start with a verb, suggest `_get_processable_rule_names`

Docstring please.

src/textual/css/_styles_builder.py
```diff
@@ -63,27 +68,37 @@ def add_declaration(self, declaration: Declaration) -> None:
...
+        return frozenset(
```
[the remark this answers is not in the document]
willmcgugan: I suspect it doesn't need to be a set. Correct me if I'm wrong but we will only ever iterate through the container, and never check for set membership. In that case, a plain old list is probably better.

willmcgugan (changes requested):
A few requests

src/textual/css/_styles_builder.py
DrBenton: Ah, I was not sure about that one, as the two naming conventions exist in the wild! :smile: 
In Symfony for example we use `$request->getClientIp()` to get the IP address from a HTTP request, whereas it's `$request->ip()` in Laravel.
I'll change to the "getter" style! :+1: :slightly_smiling_face:

src/textual/css/_styles_builder.py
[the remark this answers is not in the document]
DrBenton: It doesn't need to be a set indeed - I chose this data structure only because it seems to be the one with the fastest lookup. But as this is something that is only triggered when fatal errors happen I guess performance is not really a concern there.
I'll change that to a list! :slightly_smiling_face:  :+1:

src/textual/css/_help_text.py
DrBenton: Done!
Screenshot from 2022-05-03 15-22-07

(I've left the first one with italics though, as it's the style used by our other HelpTexts when they display a CSS property name)

DrBenton:
@willmcgugan Thanks for your review! Hopefully addressed your comments in [26f138e](https://github.com/Textualize/textual/pull/445/commits/26f138e69be49f33fe7ff72cebbb51d617a6338f) :crossed_fingers:  :slightly_smiling_face:

src/textual/suggestions.py
```diff
@@ -0,0 +1,24 @@
+from __future__ import annotations
+
+from difflib import get_close_matches
+from typing import Sequence
+
+
+def get_suggestion(word: str, possible_words: Sequence[str]) -> str | None:
+    """
```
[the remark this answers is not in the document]
DrBenton: Ah, true, I forgot that we were using this format sorry! I'll do that! :+1:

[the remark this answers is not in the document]
DrBenton: Done! [a26822b](https://github.com/Textualize/textual/pull/445/commits/a26822b257e187969f9f2dcef0e17af88650995d)

I'll try to better document the stuff I'll add from now on :slightly_smiling_face:

willmcgugan:
très bien!

PR #448, opened 2022-04-29
Add TEXTUAL features env var

willmcgugan:
Adds a TEXTUAL env var which will be used to enable features.

Currently we support two features DEVTOOLS to enable devtools, and DEBUG to enable a debug mode.

Here's how you might use them:

```
TEXTUAL=devtools,debug python basic.py
```

### DIFF
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -97,7 +97,6 @@ class App(Generic[ReturnType], DOMNode):
         css_file: str | None = None,
         css: str | None = None,
         watch_css: bool = True,
-        _features: str | None = None,
     ):
         """Textual application base class
 
@@ -109,7 +108,6 @@ class App(Generic[ReturnType], DOMNode):
             css_file (str | None, optional): Path to CSS or ``None`` for no CSS file. Defaults to None.
             css (str | None, optional): CSS code to parse, or ``None`` for no literal CSS. Defaults to None.
             watch_css (bool, optional): Watch CSS for changes. Defaults to True.
-            _features (set[str] | None, optional): Override env var features for testing. Defaults to None.
         """
         self.console = Console(
             file=sys.__stdout__, markup=False, highlight=False, emoji=False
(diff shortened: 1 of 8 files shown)


DrBenton:
It makes me think of a flag I was thinking of the other day, that we could add: something like `lax_css` 
When enabled, it would make Textual mimic the Web browsers behaviour - which is to fail silently when a CSS declaration is incorrect for whatever reason, rather than crashing. 
I do like strict behaviours as a developer, and wouldn't use such a feature myself, but I reckon that it might not be the case of everyone using Textual? :slightly_smiling_face:

src/textual/app.py
```diff
@@ -160,6 +164,11 @@ def __init__(
         if css is not None:
             self.css = css
 
+        _parse_features = os.getenv("TEXTUAL", "") if _features is None else _features
+        self.features = set(
```
DrBenton, line 165: Maybe we could type `self.features` so that by looking at this code we would instantly see what feature flags are supported by Textual, rather than having to grep `in self.features` on the source code?
i.e. something like:
```py
FeatureFlag = Literal['debug', 'devtools'] # at the module level

...
...
        self.features: set[FeatureFlag] = set(
           ...
        )

**Review by DrBenton — APPROVED** [review]
LGTM! :blush:

File: src/textual/app.py (hunk 1 above)
Comment (willmcgugan) [reply src/textual/app.py:165]: I think that's probably a good idea.

Comment (DrBenton) [reply src/textual/app.py:165]: Ah, don't mind me! `typing.Literal` was added in Python 3.8, and we're targetting 3.7+ :pensive: 
We could use an Enum instead, but that would make the code less "smooth" than using literals imho?

Comment (willmcgugan) [reply src/textual/app.py]: DOne!

Comment (willmcgugan) [reply src/textual/app.py]: You can import it from `typing_extensions`. Makes for ugly imports but it can work in 3.7

File: src/textual/features.py [hunk 2]
```diff
@@ -0,0 +1,34 @@
...
+import sys
+
+from typing import cast
+
+if sys.version_info >= (3, 8):
+    from typing import Final, Literal
+else:
+    from typing_extensions import Final, Literal
+
+FEATURES: Final = {"devtools", "debug"}
```
DrBenton, line 12: We could also do this, which is less elegant but avoids repeating our feature flags: :slightly_smiling_face: 
(because it's not very elegant to rely on this `__arg__` property of Literals, choosing to use it or not is quite subjective ^^)

```py
FeatureFlag = Literal["devtools", "debug"]

FEATURES: Final = set(FeatureFlag.__args__)

willmcgugan: Yeah, I don't like the repetition either. But the dunder attribute in a typing var makes me nervous.

PR #449, opened 2022-04-29
[CI] Check that the "basic.py" sandbox script can be run for a few seconds without crashing

DrBenton:
Right, it took me some time but I finally have code that allows one to run the `basic.py` sandbox script with the 4 major versions of Python we support, with and without `uvloop` installed!
Do I spent the morning switching from one of these venvs to another, and iterate until it no longer crash? Absolutely! :smile: 
Screenshot from 2022-05-03 14-10-17

Thanks to the addition of that "e2e smoke test" script in our CI pipeline, which crashed right away, I realised that Textual apps were genuinely not working on the "css" branch with Python versions older than 3.10, because of asyncio issues where we had 2 loops running in parallel.
The main cause behind this seems to be that creating a `asyncio.PriorityQueue` (which we indirectly do when a `textual.app.App` is instantiated) initiates an event loop - and keeps a reference to it these Queues - in versions of Python older than 3.10.
We can see the difference here for example:
 - 3.9: https://github.com/python/cpython/blob/3.9/Lib/asyncio/queues.py#L36
 - 3.10: the running loop is now "lazy-getted" when the queues need it: https://github.com/python/cpython/blob/3.10/Lib/asyncio/mixins.py#L22

Also, in the compositor we were using the `start` param of the built-in _sum()_ function as a kwarg, but it turns out that can only be specified with a keyword starting from Python 3.8 - as a result this was causing a crash with Python 3.7.

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -39,6 +39,10 @@ jobs:
         run: |
           source $VENV
           pytest tests -v --cov=./src/textual --cov-report=xml:./coverage.xml --cov-report term-missing
+      - name: Quick e2e smoke test
+        run: |
+          source $VENV
+          python tools/sandbox_e2e_smoke_test.py
       - name: Upload code coverage
         uses: codecov/codecov-action@v1.0.10
         with:
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -42,7 +42,7 @@ jobs:
       - name: Quick e2e smoke test
         run: |
           source $VENV
-          python e2e_tests/sandbox_basic_test.py
+          python e2e_tests/sandbox_basic_test.py basic 2.0
       - name: Upload code coverage
         uses: codecov/codecov-action@v1.0.10
         with:
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -42,7 +42,7 @@ jobs:
       - name: Quick e2e smoke test
         run: |
           source $VENV
-          python tools/sandbox_e2e_smoke_test.py
+          python e2e_tests/sandbox_basic_test.py
       - name: Upload code coverage
         uses: codecov/codecov-action@v1.0.10
         with:
--- a/e2e_tests/sandbox_basic_test.py
+++ b/e2e_tests/sandbox_basic_test.py
@@ -14,6 +14,11 @@ def launch_sandbox_script():
     basic_app.run()
 
 
+# The following line seems to be required in order to have this work in the MacOS part
+# of our CI - despite the fact that Python docs say that this function
+# "has no effect when invoked on any operating system other than Windows" 🤔
+multiprocessing.freeze_support()
+
 process = multiprocessing.Process(target=launch_sandbox_script, daemon=True)
 process.start()
 
--- a/e2e_tests/sandbox_basic_test.py
+++ b/e2e_tests/sandbox_basic_test.py
@@ -1,32 +1,55 @@
 from __future__ import annotations
+
+import shlex
 import sys
-import multiprocessing
+import subprocess
+import threading
 from pathlib import Path
 
-project_root = Path(__file__).parent.parent
-for python_path in [str(project_root), str(project_root / "src")]:
-    sys.path.append(python_path)
+target_script_name = "basic"
+script_time_to_live = 2.0  # in seconds
+
+if len(sys.argv) > 1:
+    target_script_name = sys.argv[1]
+if len(sys.argv) > 2:
+    script_time_to_live = float(sys.argv[2])
+
+e2e_root = Path(__file__).parent
+
+completed_process = None
 
 
-def launch_sandbox_script():
-    from sandbox.basic import app as basic_app
+def launch_sandbox_script(python_file_name: str) -> None:
+    global completed_process
 
-    basic_app.run()
+    command = f"{sys.executable} ./test_apps/{shlex.quote(python_file_name)}.py"
+    print(f"Launching command '{command}'...")
+    try:
+        completed_process = subprocess.run(
+            command, shell=True, check=True, capture_output=True, cwd=str(e2e_root)
+        )
+    except subprocess.CalledProcessError as err:
+        print(f"Process error: {err.stderr}")
+        raise
 
 
-# The following line seems to be required in order to have this work in the MacOS part
-# of our CI - despite the fact that Python docs say that this function
-# "has no effect when invoked on any operating system other than Windows" 🤔
-multiprocessing.freeze_support()
+thread = threading.Thread(
+    target=launch_sandbox_script, args=(target_script_name,), daemon=True
+)
+thread.start()
 
-process = multiprocessing.Process(target=launch_sandbox_script, daemon=True)
-process.start()
+print(
+    f"Launching Python script in a sub-thread; we'll wait for it for {script_time_to_live} seconds..."
+)
+thread.join(timeout=script_time_to_live)
+print("The wait is over.")
 
-process.join(timeout=2)
-
-process_still_running = process.exitcode is None
+process_still_running = completed_process is None
 process_was_able_to_run_without_errors = process_still_running
-if process_still_running:
-    process.kill()
+
+if process_was_able_to_run_without_errors:
+    print("Python script is still running :-)")
+else:
+    print("Python script is no longer running :-/")
 
 sys.exit(0 if process_was_able_to_run_without_errors else 1)
--- a/e2e_tests/test_apps/basic.css
+++ b/e2e_tests/test_apps/basic.css
@@ -1,0 +1,227 @@
+/* CSS file for basic.py */
+
+
+
+* {
+    transition: color 300ms linear, background 300ms linear;
+}
+
+
+* {
+  scrollbar-background: $panel-darken-2;
+  scrollbar-background-hover: $panel-darken-3;
+  scrollbar-color: $system;
+  scrollbar-color-active: $accent-darken-1;
+}
+
+App > Screen {
+  layout: dock;
+  docks: side=left/1;
+  background: $surface;
+  color: $text-surface;
+}
+
+
+#sidebar {
+  color: $text-primary;
+  background: $primary;
+  dock: side;
+  width: 30;
+  offset-x: -100%;
+  layout: dock;
+  transition: offset 500ms in_out_cubic;
+}
+
+#sidebar.-active {
+  offset-x: 0;
+}
+
+#sidebar .title {
+  height: 3;
+  background: $primary-darken-2;
+  color: $text-primary-darken-2 ;
+  border-right: outer $primary-darken-3;
+  content-align: center middle;
+}
+
+#sidebar .user {
+  height: 8;
+  background: $primary-darken-1;
+  color: $text-primary-darken-1;
+  border-right: outer $primary-darken-3;
+  content-align: center middle;
+}
+
+#sidebar .content {
+  background: $primary;
+  color: $text-primary;
+  border-right: outer $primary-darken-3;
+  content-align: center middle;
+}
+
+#header {
+  color: $text-primary-darken-1;
+  background: $primary-darken-1;
+  height: 3;
+  content-align: center middle;
+}
+
+#content {
+  color: $text-background;
+  background: $background;
+  layout: vertical;
+  overflow-y: scroll;  
+}
+
+
+Tweet {
+  height: 12;
+  width: 80;
+ 
+  margin: 1 3;
+  background: $panel;
+  color: $text-panel;
+  layout: vertical;
+  /* border: outer $primary; */
+  padding: 1;
+  border: wide $panel-darken-2;
+  overflow-y: scroll; 
+  align-horizontal: center;
+  
+}
+
+.scrollable {
+  width: 80;
+  overflow-y: scroll;
+  max-width:80;
+  height: 20;
+  align-horizontal: center;
+  layout: vertical;
+}
+
+.code {
+
+  height: 34;
+  width: 100%;
+  
+}
+
+
+TweetHeader {
+  height:1;
+  background: $accent;
+  color: $text-accent
+}
+
+TweetBody {  
+  width: 100%;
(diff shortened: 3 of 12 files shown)


DrBenton:
@willmcgugan it seems that this new CI step is already doing a good job :slightly_smiling_face:  : the error it is failing on (_"got Future <Future pending> attached to a 
different loop"_) appears to be a genuine one, that I also have locally if I run the `basic.py` sandbox script with Python 3.7!
Screenshot from 2022-04-29 15-03-33

I can reproduce this instant crash of "basic.py" with Python versions `3.7.13` and `3.9.12`. It works however with `3.10.4`, which is the one I usually activate.

GitHub Action result:
``` 
╭───────────────────── Traceback (most recent call last) ──────────────────────╮
│                                                                              │
│ /home/runner/work/textual/textual/src/textual/message_pump.py:118 in         │
│ get_message                                                                  │
│                                                                              │
│   115 │   │   │   │   return self._pending_message                           │
│   116 │   │   │   finally:                                                   │
│   117 │   │   │   │   self._pending_message = None                           │
│ ❱ 118 │   │   message = (await self._message_queue.get()).message            │
│   119 │   │   if message is None:                                            │
│   120 │   │   │   self._closed = True                                        │
│   121 │   │   │   raise MessagePumpClosed("The message pump is now closed")  │
│                                                                              │
│ ╭─────────────────── locals ───────────────────╮                             │
│ │ self = BasicApp(title='Textual Application') │                             │
│ ╰──────────────────────────────────────────────╯                             │
│ /opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/asyncio/queues.py:159   │
│ in get                                                                       │
│                                                                              │
│   156 │   │   │   getter = self._loop.create_future()                        │
│   157 │   │   │   self._getters.append(getter)                               │
│   158 │   │   │   try:                                                       │
│ ❱ 159 │   │   │   │   await getter                                           │
│   160 │   │   │   except:                                                    │
│   161 │   │   │   │   getter.cancel()  # Just in case getter is not done yet │
│   162 │   │   │   │   try:                                                   │
│                                                                              │
│ ╭───────────────────────────────── locals ─────────────────────────────────╮ │
│ │ getter = <Future finished result=None>                                   │ │
│ │   self = <PriorityQueue at 0x7f8eaecab4d0 maxsize=0                      │ │
│ │          _queue=[<textual.message_pump.MessagePriority object at         │ │
│ │          0x7f8eb076e050>, <textual.message_pump.MessagePriority object   │ │
│ │          at 0x7f8eadc40f10>] tasks=2>                                    │ │
│ ╰──────────────────────────────────────────────────────────────────────────╯ │
╰──────────────────────────────────────────────────────────────────────────────╯
RuntimeError: Task <Task pending coro=<App.run.<locals>.run_app() running at 
/home/runner/work/textual/textual/src/textual/app.py:310> 
cb=[_run_until_complete_cb() at /opt/hostedtoolcache/Python/3.7.12/x64/lib/pytho
n3.7/asyncio/base_events.py:157]> got Future <Future pending> attached to a 
different loop

File: sandbox/basic.py [hunk 1]
```diff
@@ -139,7 +141,8 @@ def key_x(self):
         self.panic(self.tree)
 
 
-app = BasicApp(css_file="basic.css", watch_css=True, log="textual.log")
+css_file = Path(__file__).parent / "basic.css"
```
DrBenton, line 144: Just using an absolute path rather than a relative one here, so this App can be used from any other folder :-)

tools/sandbox_e2e_smoke_test.py
```diff
@@ -0,0 +1,27 @@
+from __future__ import annotations
+import sys
+import multiprocessing
+from pathlib import Path
+
+project_root = Path(__file__).parent.parent
+for python_path in [str(project_root), str(project_root / "src")]:
+    sys.path.append(python_path)
```
DrBenton, line 8: Let's start with a quick'n'dirty `sys.path` update, just to be sure that both the git repo's root (we need this to be able to do `from sandbox.basic import app`) and the "src/" folders are part of the Python path :-)

tools/sandbox_e2e_smoke_test.py
```diff
@@ -0,0 +1,27 @@
...
+def launch_sandbox_script():
+    from sandbox.basic import app as basic_app
+
+    basic_app.run()
+
+
+process = multiprocessing.Process(target=launch_sandbox_script, daemon=True)
+process.start()
+
+process.join(timeout=2)
```
DrBenton, line 20: A very simple script, that could probably be done in Bash too but why not using Python? ^^ 
 - Start the "basic" sandbox script in a Python subprocess, thanks to the nice [multiprocessing](https://devdocs.io/python~3.10/library/multiprocessing) module 
 - Let it do its things for at most 2 seconds
 - Check if it crashed. 
    - If yes, we exit with an error code.
    - If not, we kill it (sorry, BasicApp! :bow: )  and exit with a success `0` code

e2e_tests/sandbox_basic_test.py
```diff
@@ -0,0 +1,55 @@
...
+
+process_still_running = completed_process is None
+process_was_able_to_run_without_errors = process_still_running
+
+if process_was_able_to_run_without_errors:
+    print("Python script is still running :-)")
+else:
+    print("Python script is no longer running :-/")
+
+sys.exit(0 if process_was_able_to_run_without_errors else 1)
```
DrBenton, line 55: the job of this quick-n-dirty script is basically to launch a Textual app, such as `basic.py`, let it leave for at most 2 seconds, and return a successful exit code if the app is still alive after that time - or an error code if the app crashed before this

.github/workflows/pythonpackage.yml
```diff
@@ -39,6 +39,10 @@ jobs:
         run: |
           source $VENV
           pytest tests -v --cov=./src/textual --cov-report=xml:./coverage.xml --cov-report term-missing
+      - name: Quick e2e smoke test
+        run: |
+          source $VENV
+          python e2e_tests/sandbox_basic_test.py basic 2.0
```
DrBenton, line 45: = will run the `basic.py` script for 2 seconds :slightly_smiling_face:

e2e_tests/test_apps/basic.css
```diff
@@ -0,0 +1,227 @@
+/* CSS file for basic.py */
```
DrBenton, line 1: copy-paste of the sandbox's one

e2e_tests/test_apps/basic.py
```diff
@@ -0,0 +1,150 @@
+from pathlib import Path
```
DrBenton, line 1: ditto

src/textual/_compositor.py
```diff
@@ -448,7 +448,7 @@ def _assemble_chops(
         segment_lines: list[list[Segment]] = [
             sum(
                 [line for line in bucket.values() if line is not None],
-                start=[],
+                [],
```
DrBenton, line 451: the `start` param of the sum function can only be specified with a keyword starting from Python 3.8 - as a result this was causing a crash with Python 3.7

src/textual/app.py
```diff
@@ -106,6 +109,8 @@ def __init__(
             driver_class (Type[Driver], optional): Driver class, or None to use default. Defaults to None.
             title (str, optional): Title of the application. Defaults to "Textual Application".
         """
+        App._init_uvloop()
```
DrBenton, line 112: We have to initialise `uvloop` as soon as possible, as it will replace the event loop if it's installed - hence this initialisation code I had to add in the constructor of our Textual apps :confused: 

That's because this class is a child of the MessagePump one, which does `self._message_queue: PriorityQueue[MessagePriority] = PriorityQueue()` in its constructor.
...And this class inherits its constructor from its parent Queue, which contains the following code in Python versions 3.7, 3.8 and 3.9 - but no longer in 3.10:
```py
        if loop is None:
            self._loop = events.get_event_loop()
        else:
            self._loop = loop
```
( https://github.com/python/cpython/blob/3.9/Lib/asyncio/queues.py )

So as soon as our App is instantiated, these queues are created and they have a reference to a asyncio loop.
Which make them crash with Python versions older than 3.10.

src/textual/app.py
```diff
@@ -305,28 +310,33 @@ def bind(
...
-                import uvloop
-            except ImportError:
-                pass
-            else:
-                asyncio.set_event_loop(uvloop.new_event_loop())
-
-        event_loop = asyncio.get_event_loop()
-        try:
+        if _ASYNCIO_GET_EVENT_LOOP_IS_DEPRECATED:
+            # N.B. This doesn't work with Python<3.10, as we end up with 2 event loops:
```
DrBenton, line 318: took me a while to find something that make this function not crash for the 4 major versions of Python we support, with and without uvloop installed :sweat_smile:

src/textual/driver.py
```diff
@@ -18,7 +18,7 @@ class Driver(ABC):
     def __init__(self, console: "Console", target: "MessageTarget") -> None:
         self.console = console
         self._target = target
-        self._loop = asyncio.get_event_loop()
+        self._loop = asyncio.get_running_loop()
```
DrBenton, line 21: now that `get_event_loop()` is deprecated we should replace it with `get_running_loop()` when possible

src/textual/drivers/linux_driver.py
```diff
@@ -71,7 +71,7 @@ def _disable_mouse_support(self) -> None:
 
     def start_application_mode(self):
 
-        loop = asyncio.get_event_loop()
+        loop = asyncio.get_running_loop()
```
DrBenton, line 74: ditto

src/textual/drivers/linux_driver.py
```diff
@@ -119,7 +119,7 @@ def on_terminal_resize(signum, stack) -> None:
         self.console.file.write("\033[?1003h\n")
         self.console.file.flush()
         self._key_thread = Thread(
-            target=self.run_input_thread, args=(asyncio.get_event_loop(),)
+            target=self.run_input_thread, args=(asyncio.get_running_loop(),)
```
DrBenton, line 122: ditto (again)

src/textual/drivers/windows_driver.py
```diff
@@ -44,7 +44,7 @@ def _disable_mouse_support(self) -> None:
 
     def start_application_mode(self) -> None:
 
-        loop = asyncio.get_event_loop()
+        loop = asyncio.get_running_loop()
```
DrBenton, line 47: ..and... ditto! :smile:

src/textual/app.py
```diff
@@ -508,6 +518,7 @@ async def process_messages(self) -> None:
         active_app.set(self)
         log("---")
         log(f"driver={self.driver_class}")
+        log(f"uvloop installed: {self.__class__.__uvloop_installed!r}")
```
DrBenton, line 521: can be useful to know whether we're using uvloop or not?

src/textual/app.py
```diff
@@ -319,28 +324,33 @@ def bind(
...
+        else:
+            # However, this works with Python<3.10:
+            event_loop = asyncio.get_event_loop()
+            event_loop.run_until_complete(run_app())
 
         return self._return_value
 
+    @classmethod
+    def _init_uvloop(cls) -> None:
+        if hasattr(cls, "__uvloop_installed"):
```
willmcgugan, line 343: No sure this is necessary. I think that `uvloop.install` may be a no-op when called twice. Can you check?

src/textual/app.py
willmcgugan: How about instead of this class attribute, we print the running loop? I would think that the repr would indicate if it is uvloop or not.

src/textual/_timer.py
```diff
@@ -114,7 +115,12 @@ async def _run(self) -> None:
                     continue
                 wait_time = max(0, next_timer - monotonic())
                 if wait_time:
-                    await sleep(wait_time)
+                    try:
+                        await sleep(wait_time)
+                    except asyncio.CancelledError:
+                        # Likely our program terminating: this is fine, we just have to
+                        # shut down out asyncio Task properly:
+                        await self.stop()
```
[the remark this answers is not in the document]
willmcgugan: This is confusing me. `Timer.stop` cancels the task (i.e. this function), and I would expect that to throw another CancelledError which would be caught by L140.

Can you check that this is really neccesary?

willmcgugan (changes requested):
Great work. A few requests and questions.

src/textual/app.py
DrBenton: I checked before writing this (ugly :sweat_smile: ) code, yes, and it seems that calling it twice "installs" it twice.
> https://github.com/MagicStack/uvloop/blob/master/uvloop/__init__.py#L23-L25

I reckon that executing this `__asyncio.set_event_loop_policy(EventLoopPolicy())` statement twice doesn't harm, but I don't know how costly replacing the event loop policy multiple times is for Python in terms of performance?

src/textual/app.py
DrBenton: It would be better indeed! :+1: 
The repr is not super easy to read in the case of a "native" asyncio loop, but it's useful for sure :slightly_smiling_face: 
 - native loop, for my Unix/Linux system: `<_UnixSelectorEventLoop running=True closed=False debug=False>`
 - uvloop's repr: `<uvloop.Loop running=True closed=False debug=False>`

DrBenton:
@willmcgugan Done! :slightly_smiling_face:  [3198d28](https://github.com/Textualize/textual/pull/449/commits/3198d28588a591ff7262f77a1a756cadf598503b) (while I was there I added a quick docstring to the `run()` method ^^)

src/textual/app.py
willmcgugan: Fair enough. Still I think we should store that boolean somewhere else, where it can be typed and doesn't need `hasattr`.

How about a classvar?

DrBenton: As discussed, I moved this logic to the module scope : [e324de7](https://github.com/Textualize/textual/pull/449/commits/e324de761306eed5252b9189aa8a132ac7378dac) :slightly_smiling_face:

willmcgugan (approved):

willmcgugan:
Great

PR #454, opened 2022-05-01
Add links to CHANGELOG and fix date for v0.1.18

bbugyi200:
The CHANGELOG now conforms to conventions set by [Keep a Changelog](http://keepachangelog.com/). Many of these new links are broken ATM, but that is only because git tags have only been added for some (not all) releases. These tags would need to be added by a maintainer (not something I can do in a PR).

I would also recommend adding an "Unreleased" section to the top (so contributors can add changelog bullets as they work), but that is not something I have done in this PR.

Awesome project btw @willmcgugan. I'm super excited to see what this looks like once it has matured a bit!


willmcgugan:
Thanks

willmcgugan (approved):

PR #457, opened 2022-05-03
Omit style property setter frames from tracebacks

darrenburns:
Also upgrades Rich version to make use of this new feature. Had to major bump typing-extensions too in order to stay compatible with Rich.

image


willmcgugan (approved):
Great

willmcgugan:
Unrelated to the PR, but I think "Values for border type" might look better if the values carry on the same line.

PR #458, opened 2022-05-03
Tab focus

willmcgugan:
- Tab shows focused widget, tab again moves to next widget, shift+tab moves to previous widget
- Drops css parameter on App, in favor of CSS class attribute
- Adds Tint renderable and CSS declaration
- Fixes issue where universal selector didn't respect pseudo selectors


sandbox/basic.py
```diff
@@ -91,7 +91,7 @@ class BasicApp(App):
 
     def on_load(self):
         """Bind keys here."""
-        self.bind("tab", "toggle_class('#sidebar', '-active')")
+        self.bind("s", "toggle_class('#sidebar', '-active')")
```
willmcgugan, line 94: The app now uses tab. Which means we can't use tab elsewhere. Is this a reasonable sacrifice?

DrBenton: The only use case of letting <tab> be used in the userland space of Textual would be for widgets that let people edit some code? In which case, the equivalent of a browser's textarea could take priority over the behaviour of "pressing tab" - but I reckon this would break accessibility. 
I don't know what the best practices are in terms of managing a11y and code editing?

src/textual/app.py
```diff
@@ -401,33 +507,52 @@ def mount(self, *anon_widgets: Widget, **widgets: Widget) -> None:
...
+                    # Blur currently focused widget
+                    self.focused.post_message_no_wait(events.Blur(self))
+                # Change focus
                 self.focused = widget
-                await widget.post_message(events.Focus(self))
+                # Send focus event
+                widget.post_message_no_wait(events.Focus(self))
+
+    async def _set_mouse_over(self, widget: Widget | None) -> None:
+        """Called when the mouse is over a nother widget.
```
DrBenton, line 551: _"a nother widget"_: typo :-)

src/textual/app.py
```diff
@@ -202,6 +209,105 @@ def exit(self, result: ReturnType | None = None) -> None:
         self._return_value = result
         self.close_messages_no_wait()
 
+    @property
+    def focus_chain(self) -> list[Widget]:
+        """Get widgets that may receive focus, in focus order."""
+        widgets: list[Widget] = []
+        add_widget = widgets.append
+        root = self.screen
+        stack: list[Iterator[Widget]] = [iter(root.children)]
```
DrBenton, line 218: shouldn't we use `root.displayed_children` , so children that have `display: none` are not included in the focus chain?

src/textual/app.py
```diff
@@ -87,7 +98,9 @@ class ActionError(Exception):
 class App(Generic[ReturnType], DOMNode):
     """The base class for Textual Applications"""
 
-    css = ""
+    CSS = """
```
DrBenton, line 101: Just to be sure I understand correctly... The only reason we can move this to a class-level variable that can no longer be overridden in the constructor is because we expect to never have more than one instance of App instantiated in a Python program, right? :slightly_smiling_face: 
(so customising this CSS at the class level can only be be done before instantiating an App for the first time)

DrBenton (approved):
LGTM!
I only have superficial comments that you're totally free to ignore :-)

src/textual/app.py
willmcgugan: fixed

src/textual/app.py
willmcgugan: Yep

src/textual/app.py
willmcgugan: You can still define a subclass with different CSS.

I wanted to remove it from the constructor so that there is only one place to included the CSS if its a string literal. A classvar constant seems like the best place.

src/textual/message_pump.py
```diff
@@ -233,7 +240,7 @@ async def _process_messages(self) -> None:
             except CancelledError:
                 raise
             except Exception as error:
-                self.app.panic(error)
+                self.app.on_exception(error)
```
willmcgugan, line 243: @darrenburns pretty sure this is the cause of the "can't render blah..." error you saw.

src/textual/app.py
```diff
@@ -202,6 +209,105 @@ def exit(self, result: ReturnType | None = None) -> None:
...
+        pop = stack.pop
+        push = stack.append
+
+        while stack:
+            node = next(stack[-1], None)
+            if node is None:
+                pop()
+            else:
+                if node.is_container and node.can_focus:
+                    push(iter(node.displayed_children))
```
darrenburns, line 228: Are we alright with invisible widgets getting focus? Can "focusable" be set dynamically on a Widget by users of it?

darrenburns (approved):

src/textual/app.py
willmcgugan: It should probably filter `visible: hidden` out as well. Good point.

At the moment, its not designed to be dynamic. But maybe it should be, to skip over fields that have been disabled or something. I'll think about that for a future update.

willmcgugan: Added a `focusable_children` property.

darrenburns:
Only other thing I want to mention here is that we might need to revisit the accessibility of this (either by default or via config) since it's just overlaying a colour. We may need an option to display some kind of marker too.

darrenburns (approved):

willmcgugan:
Yeah, we need something other than commit to show focus. Underline in addition to colour makes sense, but we have that issue that it underlines everything.

PR #459, opened 2022-05-03
Support 3 and 4 character hex strings for colours

### LINKED ISSUE #456: Add 3 and 8 digit hex colors to Color and CSS parser
For 3 digits colors, each digit is repeat. Thus #17a is the same as #1177aa

For 8 digit colors, the last two hex digits translate to the alpha component in a color, which should be 0 - 1, for hex 0 - FF.

darrenburns:
Adds support for 3 and 4 character CSS hex strings

Closes #456


src/textual/color.py
```diff
@@ -54,6 +54,8 @@ class Lab(NamedTuple):
 
 RE_COLOR = re.compile(
     r"""^
+\#([0-9a-fA-F]{3})$|
```
DrBenton, line 57: Shouldn't we start naming these capture groups, in order to make the parsing a bit clearer and rely less on the unpacking of a tuple with 6 "position-only" items?

tests/css/test_stylesheet.py
```diff
@@ -29,8 +29,6 @@
         ["red 4", pytest.raises(StylesheetParseError), None],  # space in it
         ["1", pytest.raises(StylesheetParseError), None],  # invalid value
         ["()", pytest.raises(TokenizeError), None],  # invalid tokens
-        # TODO: implement hex colors with 3 chars? @link https://devdocs.io/css/color_value
-        ["#09f", pytest.raises(TokenizeError), None],
```
DrBenton, line 33: :tada:

DrBenton (approved):
Nice one! :blush: 
I only have a trivial comment, very subjective so feel free to ignore it :-)

src/textual/color.py
willmcgugan: I suspect it's a little faster to parse with positional arguments. Unpacking in to locals is very quick compared to a number of dict lookups.

willmcgugan (approved):

src/textual/color.py
darrenburns: I agree it would probably be overall beneficial in some cases we have. In this case though the capture group is so short I actually think the regex is clearer without it.

PR #461, opened 2022-05-04
Skip CI-based MacOS devtools tests

darrenburns:
Devtools tests fail on MacOS sometimes (although very rare, maybe less than 5% of the time?).

The tracebacks suggest the event loop is being shutdown before our async fixture teardown code has complete (a CancellationError is sometimes raised during the shutdown process which occurs inside this teardown).

I can't find the source of this cancellation (i.e. where the event loop is being shut down from), and it's extremely hard to reproduce since it occurs so rarely.

Given my investigation didn't yield a solution, I've created this PR to skip devtools tests on CI on MacOS only.

This is a short-term "solution" to work around the annoyance of arbitrarily failing CI builds - we can track this issue on our project board and revisit it in the future. Hopefully with this in place we will be able to avoid these failures, but still have confidence in the devtools tests given they're run on Ubuntu on CI, and ran very frequently on MacOS machines locally.

### DIFF
--- a/tests/devtools/__init__.py
+++ b/tests/devtools/__init__.py
@@ -9,5 +9,5 @@ import pytest
 # teardown code has finished running. These are very rare, but are much more of an issue on
 # CI since they can delay builds that have passed locally.
 pytestmark = pytest.mark.skipif(
-    sys.platform == "darwin" and os.getenv("CI"), reason="Issue #411"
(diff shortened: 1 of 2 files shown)


DrBenton (approved):
Makes sense! Getting rid of these flaky tests

willmcgugan (approved):

PR #463, opened 2022-05-04
[API] Start accepting PathLike objects here and there

DrBenton:
These objects are managed by more and more components of Python's own stdlib: the builtin `open()` function, the `os.path.expanduser()`, `tempfile.mkstemp()` , etc, all accept not only strings for files or directories paths, but also path-like objects.
As a big fan myself of the `pathlib.Path` class, for its handiness but also for the semantics it adds to the signature of type-annotated code, I gave a quick shot at starting to accept such objects in the Textual API .

My original reason for this was that the `log: string = ""` arg of the App's constructor was not obvious to me: just by quickly checking the signature of the method I didn't get that the nature of this param was a file path.
I find the nature of this arg more obvious with `log: str | PurePath = ""` ? :slightly_smiling_face: 

### In search of the best type hint for PathLike objects

As PEP-519 [intentionally didn't want to introduce a dedicated type annotation for PathLike objects,](https://peps.python.org/pep-0519/#provide-specific-type-hinting-support) we have to find one that suits our needs and taste :slightly_smiling_face: 
 
We could use the ABC `os.PathLike` as a type hint instead of `pathlib.PurePath`, but I'm not sure it's very much used in the Python ecosystem?
Sanic for example uses PurePath for its path-like objects type annotations:
https://github.com/sanic-org/sanic/blob/main/sanic/response.py#L307

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ class ColorChanger(App):
             self.background = f"on color({event.key})"
 
 
-ColorChanger.run(log="textual.log")
+ColorChanger.run(log_path="textual.log")
 ```
 
 You'll notice that the `on_key` method above contains an additional `event` parameter which wasn't present on the beeper example. If the `event` argument is present, Textual will call the handler with an event object. Every event has an associated handler object, in this case it is a KeyEvent which contains additional information regarding which key was pressed.
@@ -114,7 +114,7 @@ class SimpleApp(App):
         await self.view.dock(Placeholder(), Placeholder(), edge="top")
 
 
-SimpleApp.run(log="textual.log")
+SimpleApp.run(log_path="textual.log")
 ```
 
 This app contains a single event handler `on_mount`. The mount event is sent when the app or widget is ready to start processing events, and is typically used for initialization. You may have noticed that `on_mount` is an `async` function. Since Textual is an asynchronous framework we will need this if we need to call most other methods.
@@ -137,7 +137,7 @@ await self.view.dock(Placeholder(), Placeholder(), edge="top")
 
 You will notice that this time we are docking _two_ Placeholder objects onto the `"top"` edge. We haven't set an explicit size this time so Textual will divide the remaining size amongst the two new widgets.
 
-The last line calls the `run` class method in the usual way, but with an argument we haven't seen before: `log="textual.log"` tells Textual to write log information to the given file. You can tail textual.log to see events being processed and other debug information.
+The last line calls the `run` class method in the usual way, but with an argument we haven't seen before: `log_path="textual.log"` tells Textual to write log information to the given file. You can tail textual.log to see events being processed and other debug information.
 
 If you run the above example, you will see something like the following:
 
@@ -183,7 +183,7 @@ class HoverApp(App):
         await self.view.dock(*hovers, edge="top")
 
 
-HoverApp.run(log="textual.log")
+HoverApp.run(log_path="textual.log")
 ```
 
 The `Hover` class is a custom widget which displays a panel containing the classic text "Hello World". The first line in the Hover class may seem a little mysterious at this point:
--- a/sandbox/basic.py
+++ b/sandbox/basic.py
@@ -141,8 +141,12 @@ class BasicApp(App):
         self.panic(self.tree)
 
 
-css_file = Path(__file__).parent / "basic.css"
-app = BasicApp(css_file=str(css_file), watch_css=True, log="textual.log")
+sandbox_folder = Path(__file__).parent
+app = BasicApp(
+    css_file=sandbox_folder / "basic.css",
+    watch_css=True,
+    log=sandbox_folder / "textual.log",
+)
 
 if __name__ == "__main__":
     app.run()
--- a/sandbox/basic.py
+++ b/sandbox/basic.py
@@ -1,6 +1,3 @@
-from pathlib import Path
-
-from rich.align import Align
 from rich.console import RenderableType
 from rich.syntax import Syntax
 from rich.text import Text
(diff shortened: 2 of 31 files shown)


src/textual/css/parse.py
```diff
@@ -305,7 +306,7 @@ def substitute_references(
 
 
 def parse(
-    css: str, path: str, variables: dict[str, str] | None = None
+    css: str, path: str | PurePath, variables: dict[str, str] | None = None
```
DrBenton, line 309: nothing else to do here, as we only pass this `path` to the TokenizerState class

src/textual/app.py
```diff
@@ -92,21 +93,21 @@ class App(Generic[ReturnType], DOMNode):
     def __init__(
         self,
         driver_class: Type[Driver] | None = None,
-        log: str = "",
+        log: str | PurePath = "",
```
DrBenton, line 96: nothing else to do here, as:
 -  `log` is only used with the builtin function `open()` - which supports PathLike objects out of the box
 - `css_file`is only passed to a `FileMonitor`, which in turn will use it with the `os.stat()` function - which also supports PathLike objects out of the box

src/textual/css/stylesheet.py
```diff
@@ -188,15 +188,16 @@ def read(self, filename: str) -> None:
             path = os.path.abspath(filename)
         except Exception as error:
             raise StylesheetError(f"unable to read {filename!r}; {error}")
-        self.source[path] = css
+        self.source[str(path)] = css
```
DrBenton, line 191: to avoid disruption I simply cast the path to a `str` - so `self.source` is still a `dict[str, str]` :slightly_smiling_face:

src/textual/css/stylesheet.py
```diff
@@ -205,6 +206,8 @@ def add_source(self, css: str, path: str | None = None) -> None:
 
         if path is None:
             path = str(hash(css))
+        elif isinstance(path, PurePath):
+            path = str(css)
```
DrBenton, line 210: same here

src/textual/css/tokenizer.py
```diff
@@ -148,8 +149,8 @@ def __rich_repr__(self) -> rich.repr.Result:
 
 
 class Tokenizer:
-    def __init__(self, text: str, path: str = "") -> None:
-        self.path = path
+    def __init__(self, text: str, path: str | PurePath = "") -> None:
+        self.path = str(path)
```
DrBenton, line 153: in order to avoid too much changes with the introduction of PathLike objects, I cast this `path` to a string so the behaviour of the Tokenizer is still the same

darrenburns:
Isn't there a PathLike typehint? I thought I'd seen that when using Pathlib in the past

DrBenton:
> Isn't there a PathLike typehint? I thought I'd seen that when using Pathlib in the past

I thought that too :slightly_smiling_face:  , but it seems to intentionally not be the case:
> https://peps.python.org/pep-0519/#provide-specific-type-hinting-support
>
> This also removed any desire to have a class in the typing module which represented the union of all acceptable path-representing types as that can be represented with `typing.Union[str, bytes, os.PathLike]` easily enough and the hope is users will slowly gravitate to path objects only.

It seems that the intent is to progressively push developers to exclusively use Path objects for anything that has some filesystem path semantics?

We could use the ABC `os.PathLike` as a type hint instead of `pathlib.PurePath` , but I'm not sure it's very much used in the Python ecosystem?
Sanic for example uses PurePath for its path-like objects type annotations:
https://github.com/sanic-org/sanic/blob/main/sanic/response.py#L307

sandbox/basic.py
```diff
@@ -141,8 +141,12 @@ def key_x(self):
         self.panic(self.tree)
 
 
-css_file = Path(__file__).parent / "basic.css"
-app = BasicApp(css_file=str(css_file), watch_css=True, log="textual.log")
+sandbox_folder = Path(__file__).parent
```
willmcgugan, line 144: I'm not keen on using `__file__` because I don't want it to emerge as a standard. There's no `__file__` when running interactively.

In the case of the sandbox, do we need it at all? I run the scripts with sandbox as the CWD.

src/textual/app.py
willmcgugan: On reflection it should probably be `log_path` and `css_path`, which I think makes it clear. But I pro the PurePath for typing purposes.

sandbox/basic.py
DrBenton: True! I'll remove the use of `Path` here (but I'll keep it in the e2e copy :-)

src/textual/app.py
DrBenton: I agree that `_path`  suffixes should make the nature of these params clearer! :slightly_smiling_face: 

I started to rename them, however I ran into an unexpected issue: doing `self.css_path = css_path` in the App constructor is not possible, because it's a subclass of DOMNode and this one already has a `css_path` read-only property :pensive: 
This `DOMNode.css_path` property seems to be used only once in the Textual codebase, so maybe we can rename it to something like `nodes_css_path` for example, as it returns a `list[DOMNode]` ?

willmcgugan: Good point. Suggest `css_path_nodes`

README.md
```diff
@@ -87,7 +87,7 @@ class ColorChanger(App):
             self.background = f"on color({event.key})"
 
 
-ColorChanger.run(log="textual.log")
+ColorChanger.run(log_path="textual.log")
```
DrBenton, line 90: just the result of the renaming of `log`to `log_path` (and `css_file` to `css_path`) in the App constructor :slightly_smiling_face:

willmcgugan (approved):

PR #464, opened 2022-05-04
Windows CI

darrenburns:
Adds Windows support to CI.

Devtools tests are currently skipped because they fail with connection errors (they make use of networking).


DrBenton (approved):
Makes sense! :-)

willmcgugan (approved):

PR #467, opened 2022-05-04
Fix broken `overflow` CSS rule

darrenburns:
Seems like it was just a typo, but there was no test for this. I've added one which catches this issue.


DrBenton (approved):
:+1: !

PR #471, opened 2022-05-04
[e2e] Add a "easing" e2e test, so we can test that easing works well on the `css` branch

DrBenton:
Just running this simple app for 2 seconds and checking that it doesn't crash.

Ideally, in the longer term I reckon we would have a Cypress/Selenium-like machinery that would allow a test to interact with a Textual app, sending it clicks, mouseovers and keyboard events, and check the visual result - but that will likely require quite a lot of work, so let's just do "smoke testing" for the moment :slightly_smiling_face: 

https://user-images.githubusercontent.com/722388/166701615-e2f6f290-b0d9-4d34-b04a-c1aad58d221a.mp4


e2e_tests/test_apps/easing.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def on_load(self):
+        """Bind keys here."""
+        self.bind("q", "quit")
+
+    def compose(self) -> ComposeResult:
+        """Build layout here."""
+        yield Placeholder(name="Moving!", id="moving_placeholder")
+
+    def on_mount(self):
```
DrBenton, line 27: I don't understand why this is working to be honest :sweat_smile: 
In the App class the implementation of `on_mount` is the following:
```py
    def on_mount(self) -> None:
        widgets = list(self.compose())
        if widgets:
            self.mount(*widgets)
            self.screen.refresh()
```

Fine, so it takes the widgets yielded by `self.compose()` and mount them.

However, if I override this `on_mount()` method like I do in this app, without calling `super().on_mount()`, I would expect to see this behaviour disappearing - and I shouldn't see the widgets defined in the `compose()` method, since it seems that the only place in the code where we're using `compose()` is in the `on_mount()` method?

And if I add `super.on_mount()` - which would have been why I was expecting to have to do on top of adding my own behaviours with this call to `self.set_timer()`, then the placeholder is added twice :thinking:

src/textual/app.py
```diff
@@ -755,7 +755,7 @@ async def _disconnect_devtools(self):
         await self.devtools.disconnect()
 
     def start_widget(self, parent: Widget, widget: Widget) -> None:
-        """Start a widget (run it's task) so that it can receive messages.
+        """Start a widget (run its task) so that it can receive messages.
```
DrBenton, line 758: micro typo fix while I was there

e2e_tests/test_apps/easing.py
DrBenton: Solved by asking Will: it turns out that Textual replicates a DOM-like events bubbling machinery, so even when we override the implementation of an event handler such as `on_mount` the parent implementation will still be called if the event's `prevent_default()` method is is not called :slightly_smiling_face:

PR #476, opened 2022-05-05
added auto height

willmcgugan:
- Implemented auto height, so "height: auto" will set the content height to the height of the renderable
- Sets watch_css default to False,
- Enables watch_css if debug features is set

### DIFF
--- a/src/textual/geometry.py
+++ b/src/textual/geometry.py
@@ -376,8 +376,11 @@ class Region(NamedTuple):
         """
         x1, y1, x2, y2 = self.corners
         ox, oy, ox2, oy2 = other.corners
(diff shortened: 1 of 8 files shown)


src/textual/geometry.py
```diff
@@ -376,8 +376,11 @@ def contains_region(self, other: Region) -> bool:
         """
         x1, y1, x2, y2 = self.corners
         ox, oy, ox2, oy2 = other.corners
-        return (x2 >= ox >= x1 and y2 >= oy >= y1) and (
-            x2 >= ox2 >= x1 and y2 >= oy2 >= y1
+        return (
```
willmcgugan, line 379: Just formatting.

darrenburns (approved):

PR #479, opened 2022-05-05
Add style param to Widget render method

darrenburns:
Does two things:

1. The Rich `Style` object is now passed into a Textual Widget's `render` method.
2. Only foreground and background colours are applied in advance inside `render_styled`. Previously the entire style was being applied in here, which led to the issue of all segments being underlined if the underline style was applied.

<hr>

However... I don't know that I like this:

Passing the style into render feels like verbosity for little gain. We're using it to "advertise" the fact that the style object is available. However, it's already available via `self` and really it's a tiny overhead to learn that you can access it via `self`.

<hr>

This also isn't really solving the original problem because the `Pretty` renderable doesn't support `style`, and `style` doesn't seem to work on `Panel` headers as well if you pass in a `Text` for them.

### DIFF
--- a/src/textual/scrollbar.py
+++ b/src/textual/scrollbar.py
@@ -8,6 +8,7 @@ from rich.console import ConsoleOptions, RenderResult, RenderableType
 from rich.segment import Segment, Segments
 from rich.style import Style, StyleType
 
+from textual.css.styles import Styles
 from textual.reactive import Reactive
 from . import events
 from ._types import MessageTarget
@@ -205,7 +206,7 @@ class ScrollBar(Widget):
         yield "window_size", self.window_size
         yield "position", self.position
 
-    def render(self) -> RenderableType:
+    def render(self, styles: Styles) -> RenderableType:
         styles = self.parent.styles
         style = Style(
             bgcolor=(
--- a/src/textual/scrollbar.py
+++ b/src/textual/scrollbar.py
@@ -8,7 +8,6 @@ from rich.console import ConsoleOptions, RenderResult, RenderableType
 from rich.segment import Segment, Segments
 from rich.style import Style, StyleType
 
-from textual.css.styles import Styles
 from textual.reactive import Reactive
 from . import events
 from ._types import MessageTarget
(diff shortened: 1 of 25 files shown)


src/textual/widget.py
```diff
@@ -479,9 +480,11 @@ def render_styled(self) -> RenderableType:
 
         renderable_text_style = parent_text_style + text_style
         if renderable_text_style:
-            renderable = Styled(renderable, renderable_text_style)
+            color = text_style.color
+            bgcolor = text_style.bgcolor
+            renderable = Styled(renderable, Style(color=color, bgcolor=bgcolor))
```
willmcgugan, line 485: There is a shortcut for creating a style from just colors which avoid some rendundant work.

```python
Style.from_colors(text_style.color, text_style.bgcolor)
```

src/textual/scrollbar.py
```diff
@@ -205,18 +205,18 @@ def __rich_repr__(self) -> rich.repr.Result:
         yield "window_size", self.window_size
         yield "position", self.position
 
-    def render(self) -> RenderableType:
-        styles = self.parent.styles
+    def render(self, style: Style) -> RenderableType:
+        style = self.parent.styles
```
willmcgugan, line 209: This is overriding the incoming `style` object. It should perhaps be renamed `styles`.

willmcgugan (changes requested):
Couple of small requests.

I see your point about the style arg. You may be right, but let's trial it for a bit.

src/textual/scrollbar.py
darrenburns: Very good spot

src/textual/widget.py
darrenburns: TIL, thanks

src/textual/widgets/_button.py
```diff
@@ -70,8 +75,8 @@ def validate_label(self, label: RenderableType) -> RenderableType:
             return Text.from_markup(label)
         return label
 
-    def render(self) -> RenderableType:
-        return self.label
+    def render(self, style: Style) -> RenderableType:
+        return Text.styled(self.label.plain, style)
```
willmcgugan, line 79: This will replace all the styles in the Text object.

Imagine a Text object created with `Text.from_markup("hello [b]World")`. The bold will be removed.

Suggest will apply the style on top of the label. So 

`label = self.label.copy().stylize(style)`

That way if the CSS styles it to italic, it will still have the additional bold style.

willmcgugan (changes requested):
Couple of requests

src/textual/widgets/_button.py
```diff
@@ -57,7 +58,11 @@ def __init__(
     ):
         super().__init__(name=name, id=id, classes=classes)
 
-        self.label = self.css_identifier_styled if label is None else label
+        if label is None:
+            label = self.css_identifier_styled
+
+        self.label: Text = self.validate_label(label)
```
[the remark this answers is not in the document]
darrenburns: Cheers. Just to check I'm following the code correctly, the reactive attribute gets set to the return value of the validate function, right?

src/textual/widgets/_button.py
darrenburns: Ah sorry, think I misunderstood our chat yesterday. Will fix this today.

src/textual/widgets/_button.py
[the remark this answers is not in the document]
willmcgugan: Yup!

willmcgugan (approved):

PR #480, opened 2022-05-05
diplay event method

willmcgugan:
In the logs messages when an event is dispatched to handlers in base classes it generates the same log messages -- which appears as though Textual is sending the same event multiple times.

This change adds the method that the event is being dispatched to for clarity.

Before:

Screenshot 2022-05-05 at 15 06 36

After:

Screenshot 2022-05-05 at 16 18 51


willmcgugan:
Taking the liberty of merging without review. It's trivial.

PR #482, opened 2022-05-06
[tests] Add an integration test for the vertical layout, fix the bug that messes up the layout when the children have `align: center top;`

### LINKED ISSUE #422: Align center middle not working correctly with scrollbars
With the CSS below on the uber.py example, the content gets pushed behind the scrollbar. There is an unexpected space on the left hand side.
```
#uber1 {
  layout: vertical;

  background: dark_green;
  overflow: hidden auto;
  border: heavy white;
}

.list-item {
  height: 8;
  align: center middle;
  background: dark_blue;
}
```

image

### LINKED ISSUE #472: [textual][tests] set up a TestDriver, that allows us to set arbitrary console sizes and not rely on the terminal that is used while the test is running

DrBenton:
In the end fixing the bug seems to just require changing a `parent_size` to `size`... but it took me almost 500 new lines of code to figure this out :smile: 

#### Before:

Having children centred in their parent (with `align: center top;`  for example) creates the following glitch, where they have an unexpected offset that moves them too much to the right:

Screenshot from 2022-05-06 16-41-25

#### After:

Screenshot from 2022-05-06 16-40-53

### First integration test in pytest

This PR also brings a first attempt to have integration tests in Textual, during which we actually start a Textual app, leave it some time to initialise its layout, and then assert things about its content.
I did a quick check on a few "examples"  and "sandbox" scripts and they look to be working ok after my one line fix, but as we don't have other integration tests yet I cannot be super confident that this fix doesn't break anything. Hopefully that's not the case! :crossed_fingers: 

close #472
close #422

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,6 +46,7 @@ includes = "src"
 
 [tool.pytest.ini_options]
 asyncio_mode = "auto"
+testpaths = ["tests"]
 
 [build-system]
 requires = ["poetry-core>=1.0.0"]
--- a/sandbox/color_names.py
+++ b/sandbox/color_names.py
@@ -1,0 +1,46 @@
+import rich.repr
+from rich.align import Align
+from rich.console import RenderableType
+from rich.panel import Panel
+from rich.pretty import Pretty
+
+from textual._color_constants import COLOR_NAME_TO_RGB
+from textual.app import App, ComposeResult
+from textual.widget import Widget
+from textual.widgets import Placeholder
+
+
+@rich.repr.auto(angular=False)
+class ColorDisplay(Widget, can_focus=True):
+    def render(self) -> RenderableType:
+        return Panel(
+            Align.center(
+                Pretty(self, no_wrap=True, overflow="ellipsis"), vertical="middle"
+            ),
+            title=self.name,
+            border_style="none",
+        )
+
+
+class ColorNames(App):
+    CSS = """
+    ColorDisplay {
+        height: 1;
+    }
+    """
+
+    def on_mount(self):
+        self.bind("q", "quit")
+
+    def compose(self) -> ComposeResult:
+        for color_name, color in COLOR_NAME_TO_RGB.items():
+            color_placeholder = ColorDisplay(name=color_name)
+            is_dark_color = sum(color) < 400
+            color_placeholder.styles.color = "white" if is_dark_color else "black"
+            color_placeholder.styles.background = color_name
+            yield color_placeholder
+
+
+if __name__ == "__main__":
+    color_name_app = ColorNames()
+    color_name_app.run()
--- a/sandbox/vertical_container.py
+++ b/sandbox/vertical_container.py
@@ -1,0 +1,94 @@
+from rich.console import RenderableType
+from rich.text import Text
+
+from textual.app import App, ComposeResult
+from textual.widget import Widget
+from textual.widgets import Placeholder
+
+root_container_style = "border: solid white;"
+initial_placeholders_count = 4
+
+
+class VerticalContainer(Widget):
+    CSS = """
+    VerticalContainer {
+        layout: vertical;
+        overflow: hidden auto;
+        background: darkblue;
+        ${root_container_style}
+    }
+
+    VerticalContainer Placeholder {
+        margin: 1 0;
+        height: 3;
+        border: solid lime;
+        align: center top;
+    }
+    """.replace(
+        "${root_container_style}", root_container_style
+    )
+
+
+class Introduction(Widget):
+    CSS = """
+    Introduction {
+        background: indigo;
+        color: white;
+        height: 3;
+        padding: 1 0;
+    }
+    """
+
+    def render(self) -> RenderableType:
+        return Text(
+            "Press '-' and '+' to add or remove placeholders.", justify="center"
+        )
+
+
+class MyTestApp(App):
+    def compose(self) -> ComposeResult:
+        # yield Introduction()
+
+        placeholders = [
+            Placeholder(id=f"placeholder_{i}", name=f"Placeholder #{i}")
+            for i in range(initial_placeholders_count)
+        ]
+
+        yield VerticalContainer(Introduction(), *placeholders, id="root")
+
+    def on_mount(self):
+        self.bind("q", "quit")
+        self.bind("t", "tree")
+        self.bind("-", "remove_placeholder")
+        self.bind("+", "add_placeholder")
+
+    def action_tree(self):
+        self.log(self.tree)
+
+    async def action_remove_placeholder(self):
+        placeholders = self.query("Placeholder")
+        placeholders_count = len(placeholders)
+        for i, placeholder in enumerate(placeholders):
+            if i == placeholders_count - 1:
+                await self.remove(placeholder)
+                placeholder.parent.children._nodes.remove(placeholder)
+        self.refresh(repaint=True, layout=True)
+        self.refresh_css()
+
+    async def action_add_placeholder(self):
+        placeholders = self.query("Placeholder")
+        placeholders_count = len(placeholders)
+        placeholder = Placeholder(
+            id=f"placeholder_{placeholders_count+1}",
+            name=f"Placeholder #{placeholders_count+1}",
+        )
+        root = self.query_one("#root")
+        root.mount(placeholder)
+        self.refresh(repaint=True, layout=True)
+        self.refresh_css()
+
+
+app = MyTestApp()
+
+if __name__ == "__main__":
+    app.run()
--- a/src/textual/_border.py
+++ b/src/textual/_border.py
@@ -15,6 +15,7 @@ INNER = 1
 OUTER = 2
 
 BORDER_CHARS: dict[EdgeType, tuple[str, str, str]] = {
+    # TODO: in "browsers' CSS" `none` and `hidden` both set the border width to zero. Should we do the same?
     "": ("   ", "   ", "   "),
     "none": ("   ", "   ", "   "),
     "hidden": ("   ", "   ", "   "),
--- a/src/textual/_compositor.py
+++ b/src/textual/_compositor.py
@@ -15,7 +15,7 @@ from __future__ import annotations
 
 from operator import attrgetter, itemgetter
 import sys
-from typing import cast, Callable, Iterator, Iterable, NamedTuple, TYPE_CHECKING
+from typing import cast, Iterator, Iterable, NamedTuple, TYPE_CHECKING
 
 import rich.repr
 from rich.console import Console, ConsoleOptions, RenderResult
@@ -23,15 +23,13 @@ from rich.control import Control
 from rich.segment import Segment, SegmentLines
 from rich.style import Style
 
-from . import errors, log
+from . import errors
 from .geometry import Region, Offset, Size
 
 
 from ._loop import loop_last
-from ._profile import timer
 from ._segment_tools import line_crop
 from ._types import Lines
-from .widget import Widget
 
 if sys.version_info >= (3, 10):
     from typing import TypeAlias
@@ -462,7 +460,6 @@ class Compositor:
         """Render a layout.
 
         Args:
-            console (Console): Console instance.
             clip (Optional[Region]): Region to clip to.
 
         Returns:
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -494,18 +494,6 @@ class App(Generic[ReturnType], DOMNode):
 
         return DOMQuery(self.screen, selector)
 
-    def query_one(self, selector: str) -> DOMNode | None:
-        """Retrieves a single node via a DOM query in the current screen.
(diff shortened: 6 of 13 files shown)


sandbox/color_names.py
```diff
@@ -0,0 +1,46 @@
...
+from rich.pretty import Pretty
+
+from textual._color_constants import COLOR_NAME_TO_RGB
+from textual.app import App, ComposeResult
+from textual.widget import Widget
+from textual.widgets import Placeholder
+
+
+@rich.repr.auto(angular=False)
+class ColorDisplay(Widget, can_focus=True):
```
DrBenton, line 14: just a quick "sandbox" app I added, just because I wanted to pick a colour for my new "vertical" sandbox so I wanted to be able to visualise our available colours :smile: 

https://user-images.githubusercontent.com/722388/167121946-7ade143f-672c-4fda-b066-b8c18e385284.mp4

sandbox/vertical_container.py
```diff
@@ -0,0 +1,94 @@
...
+
+from textual.app import App, ComposeResult
+from textual.widget import Widget
+from textual.widgets import Placeholder
+
+root_container_style = "border: solid white;"
+initial_placeholders_count = 4
+
+
+class VerticalContainer(Widget):
```
DrBenton, line 12: a quick "vertical layout" sandbox I created to reproduce the bug I was working on

https://user-images.githubusercontent.com/722388/167122123-8ddfe75f-4588-4c4f-97ea-e891143d94ae.mp4

src/textual/_border.py
```diff
@@ -15,6 +15,7 @@
 OUTER = 2
 
 BORDER_CHARS: dict[EdgeType, tuple[str, str, str]] = {
+    # TODO: in "browsers' CSS" `none` and `hidden` both set the border width to zero. Should we do the same?
```
DrBenton, line 18: Just because as a Web developer I was surprised to see that in Textual borders that are set to `hidden` or `none` still have a width, which contrasts to the _"browsers' CSS"_ behaviour.
> Like the hidden keyword, `none` displays no border. Unless a background-image is set, the computed value of the same side's border-width will be 0, even if the specified value is something else. 

But I know that our primary target is _not_ Web developers like myself, so we might want to keep that behaviour as is of course :slightly_smiling_face:

src/textual/_compositor.py
```diff
@@ -15,23 +15,21 @@
 
 from operator import attrgetter, itemgetter
 import sys
-from typing import cast, Callable, Iterator, Iterable, NamedTuple, TYPE_CHECKING
+from typing import cast, Iterator, Iterable, NamedTuple, TYPE_CHECKING
```
DrBenton, line 18: removing a few unused imports while I'm there

src/textual/_compositor.py
```diff
@@ -462,7 +460,6 @@ def render(
         """Render a layout.
 
         Args:
-            console (Console): Console instance.
```
DrBenton, line 465: this is no longer an arg of this method

src/textual/app.py
```diff
@@ -484,6 +494,18 @@ def query(self, selector: str | None = None) -> DOMQuery:
 
         return DOMQuery(self.screen, selector)
 
+    def query_one(self, selector: str) -> DOMNode | None:
```
DrBenton, line 497: I added this `query_one` method because I found it handy in the tests I was doing to chase that bug, but if you folks think that it's an unnecessary addition to the App's API I'll remove it :-)

src/textual/drivers/linux_driver.py
```diff
@@ -120,9 +120,7 @@ def on_terminal_resize(signum, stack) -> None:
         self.console.show_cursor(False)
         self.console.file.write("\033[?1003h\n")
         self.console.file.flush()
-        self._key_thread = Thread(
-            target=self.run_input_thread, args=(asyncio.get_running_loop(),)
-        )
+        self._key_thread = Thread(target=self.run_input_thread, args=(loop,))
```
DrBenton, line 123: [while I was setting up the integration framework] I noticed that we already have a reference to the event loop in this method, so let's re-use it :-)

src/textual/layouts/vertical.py
```diff
@@ -23,7 +21,7 @@ def arrange(
         placements: list[WidgetPlacement] = []
         add_placement = placements.append
 
-        y = max_width = max_height = 0
```
DrBenton, line 26: PyCharm was highlighting that this initialisation of `y` is useless since we're re-initialising it later in the same scope, so I removed it from there :-)

src/textual/layouts/vertical.py
```diff
@@ -40,14 +38,11 @@ def arrange(
 
         y = box_models[0].margin.top if box_models else 0
 
-        displayed_children = parent.displayed_children
-
+        displayed_children = cast("list[Widget]", parent.displayed_children)
```
DrBenton, line 41: just adding a cast, so displayed_children is not considered as just a list of DOMNodes

src/textual/layouts/vertical.py
```diff
@@ -40,14 +38,11 @@ def arrange(
...
         for widget, box_model, margin in zip(displayed_children, box_models, margins):
             content_width, content_height = box_model.size
-            offset_x = widget.styles.align_width(content_width, parent_size.width)
+            offset_x = widget.styles.align_width(content_width, size.width)
```
DrBenton, line 44: THAT'S IT, THAT'S THE BUGFIX! :sweat_smile: 
+482 lines of code in this PR's diff, just for this one-line bugfix ^^

src/textual/screen.py
```diff
@@ -20,8 +20,8 @@ class Screen(Widget):
     CSS = """
     
     Screen {
-        layout: dock;
-        docks: _default=top;
+        layout: vertical;
+        overflow-y: auto;
```
DrBenton, line 24: as discussed with Will: having a vertical layout by default in our Screens is a bit more intuitive than the Dock layout

src/textual/widget.py
```diff
@@ -82,6 +83,7 @@ def __init__(
         self._virtual_size = Size(0, 0)
         self._container_size = Size(0, 0)
         self._layout_required = False
+        self._default_layout = VerticalLayout()
```
DrBenton, line 86: done while pair-programming with Will: it allows us to always have a layout for Widgets that have children, even if no layout was specified
(without this we have Widgets that don't display their children by default, which is a bit counter-intuitive)

tests/test_integration_layout.py
```diff
@@ -0,0 +1,171 @@
...
+# Let's allow ourselves some abbreviated names for those tests,
+# in order to make the test cases a bit easier to read :-)
+SCREEN_W = 100  # width of our Screens
+SCREEN_H = 8  # height of our Screens
+SCREEN_SIZE = Size(SCREEN_W, SCREEN_H)
+PLACEHOLDERS_DEFAULT_H = 3  # the default height for our Placeholder widgets
+
+
+@pytest.mark.asyncio
+@pytest.mark.integration_test  # this is a slow test, we may want to skip them in some contexts
```
DrBenton, line 22: here come our first integration test, that starts a Textual app in pytest and assert that stuff is happening we way we expect it to be! :slightly_smiling_face:

tests/test_integration_layout.py
```diff
@@ -0,0 +1,171 @@
...
+            SCREEN_SIZE,
+            1,
+            "border: solid white;",  # #root has a visible border
+            "align: center top;",  # placeholders are centered horizontally
+            # #root's virtual size=screen size
+            (SCREEN_W, SCREEN_H),
+            # placeholders width=same than screen, minus 2 borders :: height=default height
+            (SCREEN_W - 2, PLACEHOLDERS_DEFAULT_H),
+            # placeholders should be at offset 1 because of #root's border
+            1,
```
DrBenton, line 93: before the bugfix we had a offset_x=2 for Placeholders that are horizontally centred, which creates the visual glitch

tests/test_integration_layout.py
```diff
@@ -0,0 +1,171 @@
...
+            placeholders = [
+                Placeholder(id=f"placeholder_{i}", name=f"Placeholder #{i}")
+                for i in range(placeholders_count)
+            ]
+
+            yield VerticalContainer(*placeholders, id="root")
+
+    app = MyTestApp(size=screen_size, test_name="compositor")
+
+    async with app.in_running_state():
```
DrBenton, line 150: this context manager starts the App, let it run for a little while so things are initialised (it takes a few event cycles to have the initial layout properly painted for example), yields it, and then shuts it down

tests/utilities/test_app.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def compose(self) -> ComposeResult:
+        raise NotImplementedError(
+            "Create a subclass of TestApp and override its `compose()` method, rather than using TestApp directly"
+        )
+
+    def in_running_state(
+        self,
+        *,
+        initialisation_timeout: float = 0.1,
```
DrBenton, line 52: by default we let the App run for 100ms before yielding it

tests/utilities/test_app.py
```diff
@@ -0,0 +1,130 @@
...
+            file=file,
+            width=width,
+            height=height,
+            force_terminal=True,
+        )
+
+
+class DriverTest(Driver):
+    def start_application_mode(self) -> None:
+        size = Size(self.console.size.width, self.console.size.height)
```
DrBenton, line 119: simplified version of what the LinuxDriver does: when the app starts we simply dispatch a Resize event

pyproject.toml
```diff
@@ -46,6 +46,7 @@ includes = "src"
 
 [tool.pytest.ini_options]
 asyncio_mode = "auto"
+testpaths = ["tests"]
```
DrBenton, line 49: I noticed that Pytest was trying to discover tests in folders such as src/ or sandbox/, so let's limit its search scope 🙂

src/textual/layouts/vertical.py
```diff
@@ -40,14 +38,11 @@ def arrange(
...
             region = Region(offset_x, y, content_width, content_height)
-            # TODO: it seems that `max_height` is not used?
-            max_height = max(max_height, content_height)
```
DrBenton, line 50: removing dead code (highlighted as such by PyCharm)

tests/test_integration_layout.py
```diff
@@ -0,0 +1,188 @@
...
+import pytest
+
+from tests.utilities.test_app import AppTest
+from textual.app import ComposeResult
+from textual.geometry import Size
+from textual.widget import Widget
+from textual.widgets import Placeholder
+
+PLATFORM = platform.system()
+WINDOWS = PLATFORM == "Windows"
```
DrBenton, line 15: we have to remove 1 pixel to the expected widths when the test runs on Windows - which is a hard-coded behaviour in the Rich `Console` iirc

src/textual/_border.py
willmcgugan: No, you're probably right. It should match the browser. I would like to have something that reserved space but didn't render anything. Perhaps "blank"?

src/textual/app.py
willmcgugan: Yeah, not keen on this, because it classes with `DOMQuery.first`. Can discuss on Tuesday.

src/textual/layouts/vertical.py
willmcgugan: I suspect DOMNodes and Widgets are going to merge at some point. The distinction was never clear.

src/textual/widget.py
```diff
@@ -401,6 +403,7 @@ def _arrange_container(self, region: Region) -> Region:
         Returns:
             Region: The widget region minus scrollbars.
         """
+        # return region
```
willmcgugan, line 406: Debug comment

tests/test_integration_layout.py
```diff
@@ -0,0 +1,188 @@
...
+
+            yield VerticalContainer(*placeholders, id="root")
+
+    app = MyTestApp(size=screen_size, test_name="compositor")
+
+    expected_screen_size = Size(*screen_size)
+
+    # On Windows the screen size is always 1 cell smaller than the requested dimensions.
+    # Let's take this into account into our "expected_*" measurements:
+    if WINDOWS:
```
willmcgugan, line 158: This is only the case on legacy windows, which we won't be supporting. So we can set `legacy_windows=False` on Windows, and we won't need this conditional.

willmcgugan (changes requested):
Fantastic. Happy we have a way of doing integration tests. Kudos for the persistance!

DrBenton:
@willmcgugan Thanks for your review! Hopefully addressed there: 72c9622b82 
(I haven't changed the "none"/"hidden" border rules yet, I guess such a change could deserve its own PR later on? :slightly_smiling_face: )

willmcgugan:
Yep, https://github.com/Textualize/textual/issues/484

PR #483, opened 2022-05-06
Added scroll_to_widget

willmcgugan:
- Adds a `scroll_to_widget` method which scrolls a container to make a widget visible.
- Renamed `RenderRegion` to `MapGeometry` which is more descriptive.
- Added a few properties on Region class

### DIFF
--- a/src/textual/_compositor.py
+++ b/src/textual/_compositor.py
@@ -350,7 +350,7 @@ class Compositor:
             NoWidget: If the Widget is not contained in this Layout.
 
         Returns:
-            RenderRegion: Widget information.
+            MapGeometry: Widget's composition information.
 
         """
         try:
--- a/src/textual/_compositor.py
+++ b/src/textual/_compositor.py
@@ -52,7 +52,7 @@ class ReflowResult(NamedTuple):
     resized: set[Widget]  # Widgets that have been resized
 
 
-class RegionGeometry(NamedTuple):
+class MapGeometry(NamedTuple):
     """Defines the absolute location of a Widget."""
 
     region: Region  # The region occupied by the widget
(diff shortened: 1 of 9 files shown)


src/textual/_compositor.py
```diff
@@ -52,17 +52,17 @@ class ReflowResult(NamedTuple):
     resized: set[Widget]  # Widgets that have been resized
 
 
-class RenderRegion(NamedTuple):
+class RegionGeometry(NamedTuple):
```
willmcgugan, line 55: A better name...

willmcgugan:
Yeah, that would be great. Hard to unit test some of that logic.

src/textual/widget.py
```diff
@@ -327,54 +329,115 @@ def scroll_to(
...
+    ) -> bool:
+        """Scroll relative to current position.
+
+        Args:
+            x (int | None, optional): X coordinate (column) to scroll to, or ``None`` for no change. Defaults to None.
+            x (int | None, optional): Y coordinate (row) to scroll to, or ``None`` for no change. Defaults to None.
+            animate (bool, optional): Animate to new scroll position. Defaults to False.
+
+        Returns:
+            bool: True if the scroll position changed, otherwise False.
```
darrenburns, line 359:
```suggestion
            x (int | None, optional): X distance (columns) to scroll, or ``None`` for no change. Defaults to None.
            y (int | None, optional): Y distance (rows) to scroll, or ``None`` for no change. Defaults to None.
            animate (bool, optional): Animate to new scroll position. Defaults to False.

        Returns:
            bool: True if the scroll position changed, otherwise False.
```

darrenburns (commented):
Just one comment. Also noticed the scroll animation is a constant speed (it feels very slow as well), so if you wrap focus from the bottom to the top it's taking multiple seconds.

willmcgugan:
Good point. It should be a fixed time.

willmcgugan:
I've exposed `duration` on scroll. The scroll_to_widget will use a duration of 200ms.

PR #486, opened 2022-05-10
[css] Add a "Did you mean" suggestion when the value of a color is wrong but we can find a close one

DrBenton:
[WIP]


sandbox/basic.css
```diff
@@ -224,4 +224,4 @@ Success {
 
 .horizontal {
   layout: horizontal
-}
\ No newline at end of file
+}
```
DrBenton, line 227: just briefly touched this file for my tests, and my IDE added this missing line return at the end of it :slightly_smiling_face:

src/textual/css/_style_properties.py
```diff
@@ -782,11 +782,13 @@ def __set__(self, obj: StylesBase, color: Color | str | None):
...
-            except ColorParseError:
+            except ColorParseError as error:
                 raise StyleValueError(
                     f"Invalid color value '{color}'",
-                    help_text=color_property_help_text(self.name, context="inline"),
-                )
+                    help_text=color_property_help_text(
+                        self.name, context="inline", error=error
+                    ),
+                ) from error
```
DrBenton, line 791: added this ` from error` while I was there; it might be useful one day to be able to get the parent error simply by reading its `__cause__` attribute?
(to be honest I was mostly trained to that discipline of specifying "cause" errors by years of Pylint usage, with its [raise-missing-from](https://pylint.pycqa.org/en/latest/messages/warning/raise-missing-from.html) rule enabled ^^)

sandbox/basic.css
darrenburns: The pre-commit hooks add missing newlines too

DrBenton: Ah, true! Well that's many tools trying hard to add this newline then, we should be safe :smile: 

(just checked my own PyCharm settings, and I do have this option enabled on top of pre-commit :-)
Screenshot from 2022-05-10 11-54-34

src/textual/css/_style_properties.py
willmcgugan: I think in this instance, the previous ColorParseError doesn't add anything. The full stack-trace will never be displayed to the user.

tests/css/test_stylesheet.py
```diff
@@ -101,3 +101,47 @@ def test_did_you_mean_for_css_property_names(
...
+        ["tint", "ansi_whi", "ansi_white"],
+        ["scrollbar-color", "transprnt", "transparent"],
+        ["color", "xkcd", None],
+    ],
+)
+def test_did_you_mean_for_color_names(
+    css_property_name: str, css_property_value: str, expected_color_suggestion
+):
+    stylesheet = Stylesheet()
+    css = """
```
willmcgugan, line 120: Could this be an f-string?

src/textual/color.py
```diff
@@ -77,6 +77,17 @@ class Lab(NamedTuple):
...
 
+    def __init__(self, message: str, suggested_color: str | None = None):
+        """
+        Creates a new ColorParseError
+
+        Args:
+            message (str): the error message
+            suggested_color (str | None): a close color we can suggest. Defaults to None.
+        """
+        super().__init__(*[message])
```
willmcgugan, line 88: Could the `*[message]` simply be `message` ?

tests/css/test_stylesheet.py
DrBenton: It could! Using `.replace()` this way with placeholders is not great for sure, but using f-strings would require escaping all the curly braces in the "CSS" string, which is not great either :grimacing:

DrBenton:
Feedback addressed there: [aa7b19a](https://github.com/Textualize/textual/pull/486/commits/aa7b19a19352e91af0ef4ce37b46e66e9e342f2a) :crossed_fingers: 

Regarding how to display "Did you mean" suggestions... I simply went for an addition to the base error message, as thanks to the Rich-colourisation of it the value of the suggestion will be highlighted anyway which might be enough to put the emphasis on it?

I also updated the previous "Did you mean" (for CSS property names) for consistency.

The result:
Screenshot from 2022-05-11 10-36-31
Screenshot from 2022-05-11 10-33-54

...or, in a terminal with less capacities by default, such as the `tmux` one:
Screenshot from 2022-05-11 10-33-28

Do you folks think we can go for this in the short term? :slightly_smiling_face:  @willmcgugan  @darrenburns

willmcgugan (approved):

PR #494, opened 2022-05-10
[tests][e2e] Add a test for `Widget#scroll_to_widget()`

### LINKED ISSUE #489: [tests] See if we can test the "scroll to widget" behaviour in a e2e test

DrBenton:
In a nutshell the test does the following:
- Creating 10 Placeholders in a vertical container, in a viewport which has a size that can only display 5 of them
- Scrolling to one in particular, with and without animation

closes #489 

### TIL: the transition timing is not so deterministic ^^

In one of the test cases I wanted to check that there is well and truly a transition, by checking what the screen is looking like after approximately half the duration of the transition.
My reasoning what that after having waited for approximately half of the scrolling duration, we should see the middle Placeholders as we're scrolling towards the last of them.
However, it seems that the state of the screen at this "halfway there" timing is not really deterministic, as some placeholders in the middle were visible or not depending on the environment.

Here for example the environments all saw the same placeholders at half the transition time - excepted Python 3.7 on macOS :shrug: 
Screenshot from 2022-05-11 09-38-31

As a result I had to make this test case's assertions a bit looser - which seem to do the job! :slightly_smiling_face:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,6 +47,9 @@ includes = "src"
 [tool.pytest.ini_options]
 asyncio_mode = "auto"
 testpaths = ["tests"]
+markers = [
+    "integration_test: marks tests as slow integration tests(deselect with '-m \"not integration_test\"')",
+]
 
 [build-system]
 requires = ["poetry-core>=1.0.0"]
--- a/sandbox/vertical_container.py
+++ b/sandbox/vertical_container.py
@@ -20,7 +20,7 @@ class VerticalContainer(Widget):
 
     VerticalContainer Placeholder {
         margin: 1 0;
-        height: 3;
+        height: 5;
         border: solid lime;
         align: center top;
     }
@@ -79,10 +79,10 @@ class MyTestApp(App):
         placeholders = self.query("Placeholder")
         placeholders_count = len(placeholders)
         placeholder = Placeholder(
-            id=f"placeholder_{placeholders_count+1}",
-            name=f"Placeholder #{placeholders_count+1}",
+            id=f"placeholder_{placeholders_count}",
+            name=f"Placeholder #{placeholders_count}",
         )
-        root = self.query_one("#root")
+        root = self.get_child("root")
         root.mount(placeholder)
         self.refresh(repaint=True, layout=True)
         self.refresh_css()
--- a/tests/utilities/test_app.py
+++ b/tests/utilities/test_app.py
@@ -4,9 +4,10 @@ import asyncio
 import contextlib
 import io
 from pathlib import Path
-from typing import AsyncContextManager
+from typing import AsyncContextManager, cast
 
-from rich.console import Console, Capture
+from rich.console import Console
+
 from textual import events
 from textual.app import App, ReturnType, ComposeResult
 from textual.driver import Driver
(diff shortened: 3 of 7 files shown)


tests/utilities/test_app.py
```diff
@@ -49,16 +58,18 @@ def compose(self) -> ComposeResult:
     def in_running_state(
         self,
         *,
-        initialisation_timeout: float = 0.1,
-    ) -> AsyncContextManager[Capture]:
+        waiting_duration_after_initialisation: float = 0.001,
```
DrBenton, line 61: Good news! :star2: 
I was (arbitrarily) waiting for 100ms before yielding the app in a "running state", but it turns out that 1ms is enough :rocket:  - which will make most of our integration tests much faster to run! :slightly_smiling_face: 

(expected those that use animation, like some of the test cases of this new test I'm adding in this PR)

pyproject.toml
```diff
@@ -47,6 +47,9 @@ includes = "src"
 [tool.pytest.ini_options]
 asyncio_mode = "auto"
 testpaths = ["tests"]
+markers = [
```
DrBenton, line 50: cleaner to see our custom markers properly registered amongst other ones: :slightly_smiling_face: 
Screenshot from 2022-05-11 10-12-01

tests/utilities/test_app.py
```diff
@@ -110,10 +134,18 @@ def __init__(self, *, width: int, height: int):
             file=file,
             width=width,
             height=height,
-            force_terminal=True,
+            force_terminal=False,
             legacy_windows=False,
         )
 
+    @property
+    def file(self) -> io.StringIO:
```
DrBenton, line 142: just to make autocompletion better when working with these test classes :-)

tests/utilities/test_app.py
```diff
@@ -110,10 +134,18 @@ def __init__(self, *, width: int, height: int):
...
+        return cast(io.StringIO, self._file)
+
+    @property
+    def is_dumb_terminal(self) -> bool:
```
DrBenton, line 146: let's force this test Console to always be seen as "not dumb"

sandbox/vertical_container.py
```diff
@@ -79,10 +79,10 @@ async def action_add_placeholder(self):
         placeholders = self.query("Placeholder")
         placeholders_count = len(placeholders)
         placeholder = Placeholder(
-            id=f"placeholder_{placeholders_count+1}",
-            name=f"Placeholder #{placeholders_count+1}",
+            id=f"placeholder_{placeholders_count}",
+            name=f"Placeholder #{placeholders_count}",
         )
-        root = self.query_one("#root")
+        root = self.get_child("root")
```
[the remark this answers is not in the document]
darrenburns: I think the thinking behind this name as far as I can recall was that the signature is something like `get_child(id: str)`, so the `by_id` would be redundant.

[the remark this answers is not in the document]
willmcgugan: Perhaps at some point we can make it keyword arg only, to enforce the `id="foo"`.

willmcgugan (approved):

willmcgugan:
Great. Will add a ticket to make testing anything with time more deterministic.

PR #495, opened 2022-05-11
Basic text input

darrenburns:


src/textual/widgets/text_input.py
```diff
@@ -0,0 +1,172 @@
...
+from textual import events
+from textual._types import MessageTarget
+from textual.geometry import Size
+from textual.keys import Keys
+from textual.message import Message
+from textual.reactive import Reactive
+from textual.widget import Widget
+
+
+class TextInputBase(Widget):
```
darrenburns, line 15: Trying to think ahead a little here by extracting some common keybinding/text navigation stuff into a base class. Not sure if it will work out, but that's my thinking.

src/textual/widgets/text_input.py
```diff
@@ -0,0 +1,172 @@
...
+    ALLOW_PROPAGATE = {"tab", "shift+tab"}
```
darrenburns, line 16: This probably belongs in the existing child class since for more advanced editors we don't want these to propagate.

src/textual/widgets/text_input.py
```diff
@@ -0,0 +1,172 @@
...
+
+    current_text = Reactive("", layout=True)
+    cursor_index = Reactive(0)
+
+    class Changed(Message, bubble=True):
```
darrenburns, line 21: I'm thinking of potentially adding a `validation_result` field or something like that to this message, so that for each change, the message contains the value and some object which describes that value's validity.

src/textual/widgets/text_input.py
```diff
@@ -0,0 +1,172 @@
...
+                )
+            return display_text
+        else:
+            # The user has not entered text - show the placeholder
+            display_text = Text(self.placeholder, "dim")
+            if show_cursor:
+                display_text = self._apply_cursor_to_text(display_text, 0)
+            return display_text
+
+    def _apply_cursor_to_text(self, display_text: Text, index: int):
```
darrenburns, line 154: Cursor implementation applies `reverse` style to the cursor location. If the cursor is at the very end (character append position), then we use the Unicode block character to represent it.

src/textual/widgets/text_input.py
```diff
@@ -0,0 +1,172 @@
...
+            display_text += "█"
+        else:
+            display_text.stylize(
+                "reverse",
+                start=index,
+                end=index + 1,
+            )
+        return display_text
+
+    def on_focus(self, event: events.Focus) -> None:
```
darrenburns, line 166: I can't remember why I added this or if it's needed 🤷

src/textual/widgets/text_input.py
```diff
@@ -0,0 +1,172 @@
...
+        self.refresh(layout=True)
+
+    def on_key(self, event: events.Key) -> None:
+        key = event.key
+        if key == "enter" and self.current_text:
```
darrenburns, line 171: We post events on every change, but if the user presses enter on a TextInput then a "Submitted" message is sent.

src/textual/widgets/text_input.py
```diff
@@ -0,0 +1,172 @@
...
+            self.current_text = new_text
+            self.cursor_index = min(len(self.current_text), self.cursor_index + 1)
+            changed = True
+
+        if changed:
+            self.post_message_no_wait(self.Changed(self, value=self.current_text))
+
+
+class TextInput(TextInputBase, can_focus=True):
+    CSS = """
```
darrenburns, line 81: CSS is very much subject to change - I won't spend too much time tweaking things unless they're causing issues for now.

src/textual/widgets/text_input.py
darrenburns: Seems like the focus styles aren't being applied when the widget gets focus unless I manually refresh like this.

src/textual/widgets/text_input.py
```diff
@@ -0,0 +1,263 @@
...
+        if show_cursor:
+            display_text = self._apply_cursor_to_text(
+                display_text, self._editor.cursor_index
+            )
+        return Padding(display_text, pad=1)
+
+    def get_content_height(
+        self, container_size: Size, viewport_size: Size, width: int
+    ) -> int:
+        return self._editor.content.count("\n") + 1 + 2
```
DrBenton, line 249: could be nice to add a quick comment that explains why we add 1 and 2 ? :-)

DrBenton (approved):
Amazing work! :open_mouth: 
I just have a few very picky comments on things that don't matter, that you're totally free to ignore :-)

willmcgugan (approved):
I like the direction this is going.

Haven't been too pedantic re docstrings etc. since I know this is the first part of a feature branch.

src/textual/widgets/text_input.py
```diff
@@ -0,0 +1,263 @@
...
+class TextWidgetBase(Widget):
+    STOP_PROPAGATE = set()
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._editor = TextEditorBackend()
+
+    def on_key(self, event: events.Key) -> None:
+        key = event.key
+        if key == "\x1b":
```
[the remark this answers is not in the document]
willmcgugan: Would you mind adding that as a bug?

src/textual/keys.py
```diff
@@ -200,6 +200,9 @@ class Keys(str, Enum):
     ShiftControlEnd = ControlShiftEnd
 
 
+KEY_BINDINGS = Keys._value2member_map_.values()
```
DrBenton, line 203: :ok_hand:

PR #507, opened 2022-05-13
Mock time during integration tests

DrBenton:
[WIP, NOT READY FOR REVIEW YET]


tests/utilities/test_app.py
```diff
@@ -161,3 +222,92 @@ def disable_input(self) -> None:
...
+            current_time += 1.1e-06
+
+            return current_time
+
+        async def move_clock_forward(*, seconds: float) -> tuple[float, int]:
+            nonlocal current_time, start_time
+
+            ticks_count = ceil(seconds * ticks_granularity_fps)
+            activated_timers_count_total = 0
+            for tick_counter in range(ticks_count):
```
DrBenton, line 274: naive implementation for the moment: for each simulated second we loop 60 times, and for each tick we check if we have some mocked calls to `asyncio.wait([duration])` to trigger

tests/utilities/test_app.py
```diff
@@ -177,3 +222,92 @@ def disable_input(self) -> None:
...
+                    continue  # not time for you yet, dear awaiter...
+                # Right, let's release this waiting event!
+                event.set()
+                activated_timers_count += 1
+                # ...and remove it from our pending sleep events list:
+                del pending_sleep_events[i]
+
+            return activated_timers_count
+
+        with mock.patch("textual._timer._TIMERS_CAN_SKIP", new=False), mock.patch(
```
DrBenton, line 304: we end up with 3 things to mock in our tests to simulate time in Textual; not too bad! :slightly_smiling_face:

willmcgugan (changes requested):
After a discussion, I think what we need is a Clock class with two methods; `get_time` and `sleep` which by default would return the monotonic time and call asyncio.sleep. This would essentially replicate the current behaviour.

The `get_time` method should probably be a coroutine.

When testing we can replace this with a idealize clock which can be mocked. An `update_time` method would advance the time, and the sleep method would wake coroutines according to the test time.

This should help us avoid some ackward logic in tests.

DrBenton:
@willmcgugan As discussed, I added a new `_clock` module, exposing 3 functions that should now be the single source of truth for anything related to time in Textual: [f5969fb](https://github.com/Textualize/textual/pull/507/commits/f5969fb5f68dccea14ac49b66368216670cf0d83)   :slightly_smiling_face: 

My initial implementation was made of a `Clock` class in this module, and a singleton of it exposed as `_clock.clock` .
However, the issue with this is that mocking the singleton was not effective, because other Python modules were using `from ._clock import clock`, according to the existing code style where we use direct imports from modules. 
Because of this it was too late for the patch, as a direct reference to this non-patched singleton were already created in all these modules here and there - that would not be impacted by the mocking of the singleton later on.

To solve this I opted for a slightly different approach, where I expose only 3 plain functions in this `_clock` module, which are simple proxys for calls to such a _clock singleton - the difference being that the singleton and its underlying class are hidden APIs, so we can mock the singleton and be sure that other Python modules didn't have any direct reference to it before the patch. 
What do you think?

tests/utilities/test_app.py
```diff
@@ -177,3 +221,107 @@ def disable_input(self) -> None:
...
+
+    async def sleep(self, seconds: float) -> None:
+        event = asyncio.Event()
+        target_event_monotonic_time = self._current_time + seconds
+        self._pending_sleep_events.append((target_event_monotonic_time, event))
+        # Ok, let's wait for this Event
+        # (which can only be "unlocked" by calls to `move_clock_forward()`)
+        await event.wait()
+
+    async def move_clock_forward(self, *, seconds: float) -> tuple[float, int]:
```
willmcgugan, line 268: Suggest `"advance_clock"`

Does it need to be a keyword only argument? We generally exclusively use seconds for time.

tests/utilities/test_app.py
```diff
@@ -177,3 +221,107 @@ def disable_input(self) -> None:
...
+        self._start_time = self._current_time = None
+        self._pending_sleep_events: list[tuple[float, asyncio.Event]] = []
+
+    def get_time(self) -> float:
+        if self._current_time is None:
+            self._start_clock()
+
+        # let's make the time advance _very_ slightly between 2 consecutive calls of this function,
+        # within the same order of magnitude than 2 consecutive calls to ` timer.monotonic()`:
+        self._current_time += 1.1e-06
```
willmcgugan, line 256: Not keen on this. I think it may create spin loops and it may create non-deterministic output.

The `get_time` should bean idealized version of time that advances in predictable increments.

tests/utilities/test_app.py
```diff
@@ -102,15 +119,42 @@ async def boot_and_shutdown(
...
+    async def force_full_screen_update(
+        self, *, repaint: bool = True, layout: bool = True
+    ) -> None:
+        try:
+            screen = self.screen
+        except IndexError:
+            return  # the app may not have a screen yet
+
+        # We artificially tell the Compositor that the whole area should be refreshed
+        screen._compositor._dirty_regions = {
```
willmcgugan, line 135: I think I might expose this on the Screen object to avoid digging in to the internals.

tests/utilities/test_app.py
```diff
@@ -177,3 +221,107 @@ def disable_input(self) -> None:
...
+
+        return self._current_time
+
+    async def sleep(self, seconds: float) -> None:
+        event = asyncio.Event()
+        target_event_monotonic_time = self._current_time + seconds
+        self._pending_sleep_events.append((target_event_monotonic_time, event))
+        # Ok, let's wait for this Event
+        # (which can only be "unlocked" by calls to `move_clock_forward()`)
+        await event.wait()
```
willmcgugan, line 266: Nice

src/textual/_clock.py
```diff
@@ -0,0 +1,58 @@
...
+    """
+    Get the current wall clock time.
+
+    Returns:
+        float: the value (in fractional seconds) of a monotonic clock, i.e. a clock that cannot go backwards.
+    """
+    return _clock.get_time()
+
+
+async def aget_time() -> float:
```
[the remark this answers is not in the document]
DrBenton: I'm not a fan at all either to be honest :sweat_smile:  - I just opted for that `a` prefix because it seems to become the idiomatic way to make things "asyncio-related" in Python?

i.e. `object.__enter__` and `object.__exit__`  become `object.__aenter__` and `object.__aexit__` for async contexts in the Python data model.

And we find the same pattern in Django now: `aget_or_create`, `acount`, `aaggregate`, etc .
> https://github.com/django/django/pull/14843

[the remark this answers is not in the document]
tusharsadhwani: Consider using the "await me maybe" pattern 👀

https://simonwillison.net/2020/Sep/2/await-me-maybe/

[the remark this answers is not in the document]
DrBenton: > Consider using the "await me maybe" pattern eyes
> 
> https://simonwillison.net/2020/Sep/2/await-me-maybe/

This pattern is good, and we use it already in Textual (for event listeners for example) :slightly_smiling_face: , but the problem with it is that the caller must be in an async context itself to use it - which is not the case for some calls we need to make to these clock-related functions, as they can be called in some classes' constructors for example

DrBenton:
@willmcgugan here you go, feedback hopefully addressed! :slightly_smiling_face:  [ca6d7be](https://github.com/Textualize/textual/pull/507/commits/ca6d7bea6082f657841940a2bd6d7630a64b28a1)

tests/utilities/test_app.py
```diff
@@ -177,3 +217,103 @@ def disable_input(self) -> None:
...
+        for i, (target_event_monotonic_time, event) in enumerate(
+            self._pending_sleep_events
+        ):
+            if self._current_time < target_event_monotonic_time:
+                continue  # not time for you yet, dear awaiter...
+            # Right, let's release this waiting event!
+            event.set()
+            activated_timers_count += 1
+            # ...and remove it from our pending sleep events list:
+            del self._pending_sleep_events[i]
```
willmcgugan, line 304: This is mutating a list being iterated over, which may cause unexpected side-effects. Deleting indexes from a list can be problematic. Suggest a different structure, dict or set...

willmcgugan (changes requested):

tests/utilities/test_app.py
DrBenton: I converted the list into a dict, which is also a much more elegant design :ok_hand: [0ee9c57](https://github.com/Textualize/textual/pull/507/commits/0ee9c57f597b26088c076e934f7f4492c869ae6d)

Oh and while I was there I also (arbitrarily) increased the non-Windows duration of the "atomic waiting time to let asyncio do its stuff after we advanced the clock", hopefully this will make the test suite pass on your local own environment! :crossed_fingers:

DrBenton:
@willmcgugan could that be the last commit of this PR before its merging? Time will tell! :smile: [631f7a5](https://github.com/Textualize/textual/pull/507/commits/631f7a5af8a2af17954167449b0c75b0778c163a)

willmcgugan:
Thanks for sticking with it!

PR #512, opened 2022-05-15
Compositor deltas

willmcgugan:
- Compares the previous layout against the new layout and refreshes only those regions which have changed.

This makes many operations more efficient. Previously Textual would repaint the entire screen for many operations, particularly where there was animation or scrolling. Now only the changed portions of the screen will be updated.

It is faster because the terminal is doing less work. It will also reduce the amount of data sent over the wire when we run it over the web service.


src/textual/_compositor.py
```diff
@@ -198,11 +212,12 @@ def reflow(self, parent: Widget, size: Size) -> ReflowResult:
         self.root = parent
         self.size = size
 
+        old_map = self.map.copy()
         # TODO: Handle virtual size
         map, widgets = self._arrange_root(parent)
 
-        old_widgets = set(self.map.keys())
-        new_widgets = set(map.keys())
+        old_widgets = self.map.keys()
```
willmcgugan, line 219: KeysView objects are already set-like

sandbox/basic.css
```diff
@@ -6,7 +6,6 @@
     transition: color 300ms linear, background 300ms linear;
 }
 
-
```
DrBenton, line 9: it's the little things... :smile:

src/textual/_animator.py
```diff
@@ -246,8 +246,3 @@ def __call__(self) -> None:
                 animation = self._animations[animation_key]
                 if animation(animation_time):
                     del self._animations[animation_key]
-        self.on_animation_frame()
-
-    def on_animation_frame(self) -> None:
-        # TODO: We should be able to do animation without refreshing everything
```
DrBenton, line 252: looks like your past yourself was right! :blush:

src/textual/app.py
```diff
@@ -875,10 +865,15 @@ def display(self, renderable: RenderableType) -> None:
             return
         if not self._closed:
             console = self.console
+            if self._sync_available:
+                console.file.write("\x1bP=1s\x1b\\")
```
DrBenton, line 869: it could be nice to extract these 2 special ANSI sequences (`\x1bP=1s\x1b\` and `\x1bP=2s\x1b\`) into module-scoped constants, so their name would carry a semantic?
(I'm not even sure what they represent to be honest - are these "Clear To Left" and "Clear All"?)

DrBenton (approved):
I'm afraid I'm still not aware enough of how the internals of the composition work to be able to make any relevant comment, but it's awesome to start having such "smart rendering"! :balloon:

src/textual/app.py
willmcgugan: You're right. They should really go in to Rich so I could do `console.print(SyncStart())`

They tell the terminal that an update is about to start / end, so it will try and do it in a single frame. In theory it reduces flickering.

PR #522, opened 2022-05-18
[css] add "blank" border type

### LINKED ISSUE #484: Add "blank" border type, and change "none" and "hidden" borders
Textual has a "none" border which reserves space with no characters, which conflicts with Browser CSS.

We should have "none" to mean the same as no border. And deprecate "hidden".

If we want to toggle a border on and off we can set the border-color to transparent.

DrBenton:
This PR brings the following changes:
 - `border: none;` now displays no border at all (same than `border: ;`
 - same for `border: hidden`
 - we cam now use `border: blank;` to reproduce the previous behaviour of `border: none;` - i.e. no visible border but the border space is reserved
 
It also brings:
 - some additional type annotations here and there (which I added to help me understand some parts of the code - I guess it could be useful for other developers too in the future :slightly_smiling_face: )
 - a new `get_char_at(x, y)->str` method on the AppTest class
 - a test that check the impact of each border edge type on the layout

To help me write the test that checks the impact of the border edge type on the layout I wrote one more "sandbox"; it can always be useful to have a visual reference of the rendering of these types so I ended up versioning it too:
Screenshot from 2022-05-18 11-17-09
(my colour choices are great, I know :art: :ok_hand:  )

closes #484

### DIFF
--- a/sandbox/borders.py
+++ b/sandbox/borders.py
@@ -1,0 +1,65 @@
+from rich.console import RenderableType
+from rich.text import Text
+
+from textual.app import App, ComposeResult
+from textual.css.types import EdgeType
+from textual.widget import Widget
+from textual.widgets import Placeholder
+
+
+class VerticalContainer(Widget):
+    CSS = """
+    VerticalContainer {
+        layout: vertical;
+        overflow: hidden auto;
+        background: darkblue;
+    }
+
+    VerticalContainer Placeholder {
+        margin: 1 0;
+        height: 5;
+        align: center top;
+    }
+    """
+
+
+class Introduction(Widget):
+    CSS = """
+    Introduction {
+        background: indigo;
+        color: white;
+        height: 3;
+        padding: 1 0;
+    }
+    """
+
+    def render(self, styles) -> RenderableType:
+        return Text("Here are the color edge types we support.", justify="center")
+
+
+class MyTestApp(App):
+    def compose(self) -> ComposeResult:
+        placeholders = []
+        for border_edge_type in EdgeType.__args__:
+            border_placeholder = Placeholder(
+                id=f"placeholder_{border_edge_type}",
+                title=(border_edge_type or " ").upper(),
+                name=f"border: {border_edge_type} white",
+            )
+            border_placeholder.styles.border = (border_edge_type, "white")
+            placeholders.append(border_placeholder)
+
+        yield VerticalContainer(Introduction(), *placeholders, id="root")
+
+    def on_mount(self):
+        self.bind("q", "quit")
+        self.bind("t", "tree")
+
+    def action_tree(self):
+        self.log(self.tree)
+
+
+app = MyTestApp()
+
+if __name__ == "__main__":
+    app.run()
--- a/sandbox/borders.py
+++ b/sandbox/borders.py
@@ -17,7 +17,7 @@ class VerticalContainer(Widget):
 
     VerticalContainer Placeholder {
         margin: 1 0;
-        height: 5;
+        height: auto;
         align: center top;
     }
     """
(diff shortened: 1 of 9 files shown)


sandbox/borders.py
```diff
@@ -0,0 +1,67 @@
+from rich.console import RenderableType
```
DrBenton, line 1: the sandbox I used as a visual reference while I was working on this feature

src/textual/_border.py
```diff
@@ -5,20 +5,23 @@
...
-from .color import Color
 from .css.types import EdgeStyle, EdgeType
 
 
 INNER = 1
 OUTER = 2
 
 BORDER_CHARS: dict[EdgeType, tuple[str, str, str]] = {
-    # TODO: in "browsers' CSS" `none` and `hidden` both set the border width to zero. Should we do the same?
+    # Each string of the tuple represents a sub-tuple itself:
```
DrBenton, line 17: As I was removing this "TODO" (which is now covered by this PR) I also added a bit of internal documentation in order to understand this data structure :-)

src/textual/_border.py
```diff
@@ -135,7 +141,12 @@ def __init__(
             (bottom, bottom_color),
             (left, left_color),
         ) = edge_styles
-        self._sides = (top or "none", right or "none", bottom or "none", left or "none")
+        self._sides: tuple[EdgeType, ...] = (
```
DrBenton, line 144: Added a few type hints here and there to help me understand what some of our variables represent... 
And I guess it will also be useful to have MyPy/PyRight double check these things too :slightly_smiling_face:

src/textual/widgets/_placeholder.py
```diff
@@ -19,6 +19,19 @@ class Placeholder(Widget, can_focus=True):
...
 
+    def __init__(
+        # parent class constructor signature:
+        self,
+        *children: Widget,
+        name: str | None = None,
+        id: str | None = None,
+        classes: str | None = None,
+        # ...and now for our own class specific params:
+        title: str | None = None,
```
DrBenton, line 30: While I was working on this I had a test case where it was useful to be able to specify a custom title for our Placeholders, rather than seeing them always use their class name for this. 
It can be useful as these Placeholders are the bread and butter of our quick tests, so I kept the modification :-)

tests/test_integration_layout.py
```diff
@@ -169,3 +162,75 @@ def compose(self) -> ComposeResult:
...
         assert app.screen.get_offset(placeholder).x == expected_placeholders_offset_x
+
+
+@pytest.mark.asyncio
+@pytest.mark.integration_test
+@pytest.mark.parametrize(
+    "edge_type,expected_box_inner_size,expected_box_size,expected_top_left_edge_color,expects_visible_char_at_top_left_edge",
+    (
+        # These first 3 types of border edge types are synonyms, and display no borders:
+        ["", Size(SCREEN_W, 1), Size(SCREEN_W, 1), "black", False],
```
DrBenton, line 173: this test checks the visual result of different border styles - that's pretty much the pytest equivalent of visually checking the layout properties of this:
Screenshot from 2022-05-18 11-17-09

tests/utilities/test_app.py
```diff
@@ -123,6 +123,36 @@ def last_display_capture(self) -> str | None:
         last_display_start_index = total_capture.rindex(CLEAR_SCREEN_SEQUENCE)
         return total_capture[last_display_start_index:]
 
+    def get_char_at(self, x: int, y: int) -> str:
```
DrBenton, line 126: a new helper method for our AppTest class :slightly_smiling_face:

src/textual/_border.py
willmcgugan: Not sure this is the most precise type, `tuple[EdgeType, ...]` means a tuple of any number of `EdgeType` instances, where I think that it may be `tuple[EdgeType, EdgeType, EdgeType, EdgeType]`, i.e. a tuple of exactly 4.

willmcgugan (approved):
LGTM

PR #523, opened 2022-05-18
Text input completion suggestions

darrenburns:
# What it is

* You can supply an "autocompleter" function which returns the text to autocomplete.
* The autocompleter function will run when the content of the text input changes
* The text will be displayed as dimmed
* If you press the right arrow key when the cursor is at the end, the text will autocomplete and the cursor will move to the end of the newly inserted text (we also scroll if necessary).

https://user-images.githubusercontent.com/5740731/169021463-4642c326-9947-46df-bad4-e11161170b17.mov

### DIFF
--- a/src/textual/widgets/text_input.py
+++ b/src/textual/widgets/text_input.py
@@ -56,7 +56,6 @@ class TextWidgetBase(Widget):
             self.post_message_no_wait(self.Changed(self, value=self._editor.content))
 
         self.refresh(layout=True)
-        print("at end of TextWidgetBase.on_key")
 
     def _apply_cursor_to_text(self, display_text: Text, index: int) -> Text:
         # Either write a cursor character or apply reverse style to cursor location
@@ -160,6 +159,9 @@ class TextInput(TextWidgetBase, can_focus=True):
 
     def on_resize(self, event: events.Resize) -> None:
         # Ensure the cursor remains visible when the widget is resized
+        self._reset_visible_range()
+
+    def _reset_visible_range(self):
         new_visible_range_end = max(
             self._editor.cursor_index + 1, self.content_region.width
         )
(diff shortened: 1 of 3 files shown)


src/textual/widgets/text_input.py
```diff
@@ -222,11 +253,22 @@ def on_key(self, event: events.Key) -> None:
...
         start, end = self.visible_range
         self.visible_range = (max(0, start), end)
 
+    def _update_suggestion(self, event: events.Key) -> None:
+        """Run the autocompleter function, updating the suggestion if necessary"""
+        if self.autocompleter is not None:
+            event.prevent_default()
+            super().on_key(event)
+            if self.value:
+                self._suggestion = self.autocompleter(self.value)
```
DrBenton, line 268: I guess we could have something like this here, so we can handle both sync and async autocompleters? :-)
```py
if self.value:
    if isawaitable(self.autocompleter):
        self._suggestion = await self.autocompleter(self.value)
    else:       
        self._suggestion = self.autocompleter(self.value)
``` 

...which implies that `_update_suggestion` must become async, and so must `on_key` as it seems to be the function that calls it? 
But if I'm not wrong the Textual events management already handles async event managers, using the same `if isawaitable(...)` trick, so it shouldn't be an issue? :slightly_smiling_face:

src/textual/widgets/text_input.py
```diff
@@ -115,6 +128,7 @@ def __init__(
         *,
         placeholder: str = "",
         initial: str = "",
+        autocompleter: Callable[[str], str | None] | None = None,
```
DrBenton, line 131: If we want to support async autocomplter (see my comment below) I guess we should extend this type hint to
`Callable[[str], str | None] | Awaitable[[str], str | None] | None = None` ? :-)
(as it starts getting quite long though it might be worth creating a type alias for it :sweat_smile: )

DrBenton (commented):
Nice addition! :balloon:

PR #527, opened 2022-05-18
Auto dimensions

willmcgugan:
- Fixes for nested containers and auto dimensions
- Deferred layout to Screen.on_idle to avoid unnecessary layouts
- Fix for log events arriving after app shutdown

There's a lack of tests here for the box model / layout system here. Please let that slide just now. I have a feeling there will be more changes required, and I don't yet have a good enough understanding of the various CSS rule interactions.


src/textual/_layout.py
```diff
@@ -28,16 +36,55 @@ def __repr__(self) -> str:
         return f"<{self.name}>"
 
     @abstractmethod
-    def arrange(
-        self, parent: Widget, size: Size, scroll: Offset
-    ) -> tuple[list[WidgetPlacement], set[Widget]]:
+    def arrange(self, parent: Widget, size: Size) -> ArrangeResult:
```
willmcgugan, line 39: Dropped `scroll` parameter which wasn't used. Added a type alias for the return type.

tests/test_box_model.py
```diff
@@ -59,7 +59,7 @@ def get_auto_height(container: Size, parent: Size) -> int:
     box_model = get_box_model(
         styles, Size(60, 20), Size(80, 24), get_auto_width, get_auto_height
     )
-    assert box_model == BoxModel(Size(60, 20), Spacing(1, 2, 3, 4))
+    assert box_model == BoxModel(Size(54, 16), Spacing(1, 2, 3, 4))
```
DrBenton, line 62: Just to be sure that I understand this correctly... The returned box model now shrinks the available space of a widget according to its margin, right? :slightly_smiling_face:

DrBenton (dismissed):
I'm afraid the layout and geometry updates themselves are still a bit too obscure for me to do any relevant feedback at this point, so as often I was only able to leave some pretty picky-and-useless ones that's you're totally free to ignore, sorry :sweat_smile:

tests/test_box_model.py
willmcgugan: yep

PR #528, opened 2022-05-18
Text input improvements

darrenburns:
Another pass through text input which adds:

* Click to move cursor
* Cursor blinking
* Scrolling support
* Autocomplete via completion function
* Support for CJK and single-codepoint double width emoji
* Adds a new sandbox example "file search" demo (see video below)

https://user-images.githubusercontent.com/5740731/169546232-c6e8f136-efe5-4ea3-91a6-a8d707d4b4e5.mov


src/textual/widgets/text_input.py
```diff
@@ -123,6 +136,17 @@ def __init__(
...
         self._editor = TextEditorBackend(initial, 0)
         self.visible_range: tuple[int, int] | None = None
+        self.autocompleter = autocompleter
+        self._suggestion_suffix = ""
+
+        self._cursor_blink_visible = True
+        self._cursor_blink_timer: Timer | None = None
+        self._last_keypress_time: float = 0.0
+        if self.cursor_blink_enabled:
+            self._last_keypress_time = time.monotonic()
```
DrBenton, line 146: Now that we have [this PR](https://github.com/Textualize/textual/pull/507) merged we should avoid direct calls to `time.monotonic()`, and use the new `textual._clock` module as our internal single source of truth for everything related to time :slightly_smiling_face: 
So here it would be `self._last_keypress_time = _clock.get_time_no_wait()` 

(`_clock` exposes 3 functions: `get_time_no_wait() -> float`, `async get_time() -> float` (same, but asynchronous - useful in some cases) and `async sleep(seconds: float) -> None`

DrBenton (commented):
Awesome! :star_struck: 

I only have a very little comment regarding the call to `time.monotonic()` :-)

src/textual/widgets/text_input.py
```diff
@@ -31,31 +40,34 @@ def __init__(
...
             changed = self._editor.delete_forward()
         elif key == "left":
             self._editor.cursor_left()
         elif key == "right":
             self._editor.cursor_right()
-        elif key == "home":
+        elif key == "home" or key == "ctrl+a":
             self._editor.cursor_text_start()
-        elif key == "end":
+        elif key == "end" or key == "ctrl+e":
```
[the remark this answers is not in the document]
willmcgugan: Could this be solved with bindings and actions? Bindings adds this layer of indirection, so the dev can decide what happens.

But this does feel like something we can figure out later.

src/textual/widgets/text_input.py
willmcgugan: Would you mind making that change @darrenburns

src/textual/widgets/text_input.py
```diff
@@ -143,11 +167,56 @@ def value(self, value: str) -> None:
...
+
+        self._cursor_blink_visible = True
+        start_index, end_index = self.visible_range
+
+        click_x = event.screen_x - self.content_region.x
+        new_cursor_index = start_index + click_x
+
+        # Convert click offset to cursor index accounting for varying cell lengths
+        cell_len_accumulated = 0
+        for i, character in enumerate(self._editor.get_range(start_index, end_index)):
```
willmcgugan, line 189: No single character vars please

src/textual/widgets/text_input.py
```diff
@@ -180,18 +253,68 @@ def on_key(self, event: events.Key) -> None:
         if key in self.STOP_PROPAGATE:
             event.stop()
 
+        self._last_keypress_time = time.monotonic()
```
willmcgugan, line 256: Another montonic to be replaced with `_clock.get_time_no_wait`

src/textual/widgets/text_input.py
```diff
@@ -200,33 +323,63 @@ def on_key(self, event: events.Key) -> None:
...
+                    self._suggestion_suffix = suffix
+                else:
+                    self._suggestion_suffix = None
+            else:
+                self._suggestion_suffix = None
+
+    def _toggle_cursor_visible(self):
+        """Manages the blinking of the cursor - ensuring blinking only starts when the
+        user hasn't pressed a key in some time"""
+        if time.monotonic() - self._last_keypress_time > self.cursor_blink_period:
```
willmcgugan, line 379: Another monotonic

src/textual/widgets/text_input.py
```diff
@@ -200,33 +323,63 @@ def on_key(self, event: events.Key) -> None:
...
+            self._cursor_blink_visible = not self._cursor_blink_visible
```
willmcgugan, line 380: Is this, or could it be, a Reactive value? In that case you wouldn't need the explicit refresh.

willmcgugan (approved):
Looks great. A few comments.

src/textual/widgets/text_input.py
willmcgugan: The refreshes are always done on idle, so if you set reactive values and call `refresh()` explicitly, Textual will only repaint the screen once.

PR #529, opened 2022-05-19
[css] add `scrollbar-size` properties

### LINKED ISSUE #513: Set scrollbar size in CSS
The scrollbars can be a little hard to click when cells are quite small, particularly a vertical scrollbar.

We should add CSS rules to set the size of scrollbars.

```
scrollbar-size: 2 1;
scrollbar-size-vertical: 2;
scorllbar-size-horizontal: 1;
```

We could make this responsive as the size of the screen changes. For now we just need the rules.

DrBenton:
In a nutshell: we can now do things like this: :slightly_smiling_face: 
Screenshot from 2022-05-19 12-50-06

closes #513

### DIFF
--- a/src/textual/css/_styles_builder.py
+++ b/src/textual/css/_styles_builder.py
@@ -802,7 +802,12 @@ class StylesBuilder:
             if token2.name != "number" or not token2.value.isdigit():
                 scrollbar_size_error(name, token2)
 
-            horizontal, vertical = int(token1.value), int(token2.value)
+            horizontal = int(token1.value)
+            if horizontal == 0:
+                scrollbar_size_error(name, token1)
+            vertical = int(token2.value)
+            if vertical == 0:
+                scrollbar_size_error(name, token2)
             self.styles._rules["scrollbar_size_horizontal"] = horizontal
             self.styles._rules["scrollbar_size_vertical"] = vertical
 
@@ -815,7 +820,10 @@ class StylesBuilder:
             token = tokens[0]
             if token.name != "number" or not token.value.isdigit():
                 self.error(name, token, scrollbar_size_single_axis_help_text(name))
-            self.styles._rules["scrollbar_size_vertical"] = int(token.value)
+            value = int(token.value)
+            if value == 0:
+                self.error(name, token, scrollbar_size_single_axis_help_text(name))
+            self.styles._rules["scrollbar_size_vertical"] = value
 
     def process_scrollbar_size_horizontal(self, name: str, tokens: list[Token]) -> None:
         if not tokens:
@@ -826,7 +834,10 @@ class StylesBuilder:
             token = tokens[0]
             if token.name != "number" or not token.value.isdigit():
                 self.error(name, token, scrollbar_size_single_axis_help_text(name))
-            self.styles._rules["scrollbar_size_horizontal"] = int(token.value)
+            value = int(token.value)
+            if value == 0:
+                self.error(name, token, scrollbar_size_single_axis_help_text(name))
+            self.styles._rules["scrollbar_size_horizontal"] = value
 
     def _get_suggested_property_name_for_rule(self, rule_name: str) -> str | None:
         """
--- a/src/textual/css/_styles_builder.py
+++ b/src/textual/css/_styles_builder.py
@@ -23,6 +23,8 @@ from ._help_text import (
     offset_single_axis_help_text,
     style_flags_property_help_text,
     property_invalid_value_help_text,
+    scrollbar_size_property_help_text,
+    scrollbar_size_single_axis_help_text,
 )
 from .constants import (
     VALID_ALIGN_HORIZONTAL,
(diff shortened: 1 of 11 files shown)


src/textual/css/_styles_builder.py
```diff
@@ -783,6 +785,48 @@ def process_scrollbar_gutter(self, name: str, tokens: list[Token]) -> None:
...
+            scrollbar_size_error(name, tokens[0])
+        else:
+            token1, token2 = tokens
+
+            if token1.name != "number" or not token1.value.isdigit():
+                scrollbar_size_error(name, token1)
+            if token2.name != "number" or not token2.value.isdigit():
+                scrollbar_size_error(name, token2)
+
+            horizontal, vertical = int(token1.value), int(token2.value)
```
DrBenton, line 804: I followed the convention we've been using so far for this kind of CS properties, with the 1st number being `<horizontal>` and the second one the `<vertical>`.

A funny aspect of this though is that they both actually impact the layout of the "other" axis, which was counter intuitive for me while I was writing the test for it :smile: 
--> setting a _horizontal_ scrollbar size of `3` will reduce the _vertical_ available space, and vice-versa

src/textual/css/stylesheet.py
```diff
@@ -8,17 +8,13 @@
 
 import rich.repr
 from rich.console import RenderableType, RenderResult, Console, ConsoleOptions
-from rich.highlighter import ReprHighlighter
```
DrBenton, line 11: Removing some unused imports while I'm there

(I think it could definitely be nice to add Pylint to the project, so it could spot those across the whole codebase automatically - even if it's to disable all its default rules and enable only this one?)

src/textual/widget.py
```diff
@@ -263,6 +276,8 @@ def _refresh_scrollbars(self) -> None:
             show_horizontal = True
         elif overflow_x == "auto":
             show_horizontal = self.virtual_size.width > width
+        if scrollbar_size_horizontal == 0:
```
DrBenton: new specific case: a scrollbar size of zero should probably not result in any scrollbar display, right? :slightly_smiling_face:

src/textual/widget.py
```diff
@@ -263,6 +276,11 @@ def _refresh_scrollbars(self) -> None:
             show_horizontal = True
         elif overflow_x == "auto":
             show_horizontal = self.virtual_size.width > width
+        if (
+            scrollbar_size_horizontal is not None
+            and scrollbar_size_horizontal.value == 0
+        ):
+            show_horizontal = False
```
DrBenton, line 283: new specific case: a scrollbar size of zero should probably not result in any scrollbar display, right? :slightly_smiling_face:

tests/test_integration_layout.py
```diff
@@ -22,7 +22,6 @@
 @pytest.mark.integration_test  # this is a slow test, we may want to skip them in some contexts
 @pytest.mark.parametrize(
     (
-        "screen_size",
```
DrBenton, line 25: I removed this from our test cases' params while I was there - since it doesn't vary we can just hard-code it in the test itself :slightly_smiling_face:

src/textual/css/_styles_builder.py
willmcgugan: Yes, that's been a constant source of pain when working with layouts!

BTW, this construct is slower than it needs to be.

When you do the following:

`a, b = foo, bar`

You are creating a tuple (foo, bar) then unpacking it. You can avoid that by doing the following:

```
a = foo
b = bar
```

willmcgugan (changes requested):
Correct me if I'm wrong, but this has introduced logic in a few places that handles the scrollbar size being 0 (i.e. invisible).

I think we can avoid that by ensuring that the scrollbar size defaults to 1 in the Styles property, and raising an error if the CSS has <= 0 for scrollbar size.

My thinking is that we already have a way of disabling the scrollbar, and we don't want to muddy the waters by having 2 ways of doing the same thing.

src/textual/css/_styles_builder.py
DrBenton: Alright, I'll replace the unpacking expression by a proper multi-statements one :slightly_smiling_face:

DrBenton:
@willmcgugan ready for a last (:crossed_fingers:  ^_^) review: [bd232ac](https://github.com/Textualize/textual/pull/529/commits/bd232ac60a19220a27562ddcad07b4ed771f319f)

willmcgugan:
@DrBenton This looks great, but I have probably broken things by adding some scrollbar related code. See https://github.com/Textualize/textual/blob/css/src/textual/widget.py#L347

Can you merge CSS and check that everything renders as expected?

DrBenton:
> @DrBenton This looks great, but I have probably broken things by adding some scrollbar related code. See https://github.com/Textualize/textual/blob/css/src/textual/widget.py#L347
> 
> Can you merge CSS and check that everything renders as expected?

It seems to be working! I only had to reverse a few recent testing updates in `basic.css`, so the "basic" sandbox script still looks the same as before. And I also added some custom classes for individual instances of Twet and TweetBody so we can see the various scrollbar types at work. Overall, it seems to be working :slightly_smiling_face: 

We can see on this video:
- first "tweet" has the same default vertical scrollbar than before
- second one has a `scrollbar-size-vertical: 2;`
- third one has a horizontal scrolling, as well as `scrollbar-size-horizontal: 2;`


https://user-images.githubusercontent.com/722388/170017568-854471d2-6b66-4552-bcd5-bf313b022476.mp4

DrBenton:
@willmcgugan there were indeed some last (:crossed_fingers: ) missing parts of the code where we were assuming (rightly, at the time) that the size of the scrollbars were always `0` or `1`. 
These parts were a bit harder to spot, as we were generating these numbers with `int([a boolean value])` :slightly_smiling_face: 

I fixed this now that the scrollbars can have any thickness, and added some test cases to make sure the potential bug it could cause is fixed - which should also prevent future regressions on that ofc.
Diff there: https://github.com/Textualize/textual/compare/a55b122...159d52

willmcgugan (approved):
LGTM

PR #533, opened 2022-05-20
[terminal buffering] Add support for the "mode 2026" - aka SynchronizedOutput 

### LINKED ISSUE #517: Implement synchronised output
We use an iTerm extension for this, but there is a new emerging standard.

https://gist.github.com/christianparpart/d8a62cc1ab659194337d73e399004036

The idea is you tell the terminal that an update is about to begin, which defers rendering. You then write some output, and issue the command to end the update. The terminal will then update the screen. This should reduce flickering on terminals where it is supported.

The link above describes how to auto-detect this feature.

Here's the current (deprecated) spec, which we are currently using:

https://gitlab.com/gnachman/iterm2/-/wikis/synchronized-updates-spec

DrBenton:
That's the first time I have go so deep into the low layers of Textual to implement a feature - and it was not easy :sweat_smile: 

But thanks to this I now have a better understanding of how the input is processed and how the events machinery works, which is good :slightly_smiling_face: 

We have three cases of support on this "mode 2026", depending on the terminal emulator we're using:
 - The terminal doesn't recognise the "mode request": it will not reply at all to our request (tmux for example), so `mode2026_synchronized_update` will keep its default value (`False`).
   No specific Textual logs are made.
 - The terminal understands the "mode request", but doesn't support the "2026" one: it will reply with `2026;0$y`. Our `mode2026_synchronized_update` is still `False` (that's the case of the Gnome terminal for example).
   We'll see this specific Textual logs: 
Screenshot from 2022-05-24 09-44-44

 - The terminal understands the "mode request", and supports the "2026" one: it will reply with one of the different ways it can tell us the SynchronizedOutput is now set (Set, Reset, Permanently set, Permanently reset), and we'll set `mode2026_synchronized_update` to `True`. Kitty for example responds with `2026;2$y`.
    We'll see this specific Textual logs:  
Screenshot from 2022-05-24 09-43-56

closes #517


sandbox/vertical_container.py
```diff
@@ -39,7 +39,7 @@ class Introduction(Widget):
     }
     """
 
-    def render(self) -> RenderableType:
+    def render(self, styles) -> RenderableType:
```
DrBenton, line 42: [while I was there] fixed that sandbox script ^^

src/textual/_ansi_sequences.py
```diff
@@ -1,9 +1,14 @@
...
+from ._terminal_modes import (
+    get__mode_report_sequence,
+    Mode,
+    ModeReportParameter,
+)
 from .keys import Keys
 
 # Mapping of vt100 escape codes to Keys.
-ANSI_SEQUENCES: Dict[str, Tuple[Keys, ...]] = {
+ANSI_SEQUENCES_KEYS: Dict[str, Tuple[Keys, ...]] = {
```
DrBenton, line 11: Apart form keyboards' keys there are other ANSI sequences we might be interested, like the brand new "modes"-related ones introduced in this PR.
That's why I opted for a slightly more specific renaming of this constant

src/textual/_terminal_features.py
```diff
@@ -0,0 +1,84 @@
+from __future__ import annotations
+import os
+import platform
+from dataclasses import dataclass
+
+
+@dataclass
+class TerminalSupportedFeatures:
```
DrBenton, line 8: This is the class that replaces the `this._sync_available: bool` we currently have on the App class.
I introduced because we now have 2 "sync" modes, each with their own set of ANSI sequences and their own machinery.
But also because we might have other optional terminal emulators features that we could use in the future, so adding flags on this class will likely scale better than adding them on the App class :slightly_smiling_face:

src/textual/_xterm_parser.py
```diff
@@ -108,21 +108,29 @@ def parse(self, on_token: TokenCallback) -> Generator[Awaitable, str, None]:
...
                         break
-                    else:
-                        mouse_match = _re_mouse_event.match(sequence)
-                        if mouse_match is not None:
-                            mouse_code = mouse_match.group(0)
-                            event = self.parse_mouse_code(mouse_code, self.sender)
-                            if event:
-                                on_token(event)
-                            break
+                    # Or a mouse event?
```
DrBenton, line 117: I removed the `else` level of indentation here, because keeping it would have introduced the need to also add a new level of indent to manage the "mode reports" and that would have been quite ugly :sweat_smile: 
The `break` statements here acts as _"early returns"_ so we should be good without this `else:` level :slightly_smiling_face:

src/textual/driver.py
```diff
@@ -2,12 +2,10 @@
 
 import asyncio
 from time import time
-import platform
```
DrBenton, line 5: [while I was there] the usual "unused imports" manual cleaning ^^

src/textual/_terminal_features.py
```diff
@@ -0,0 +1,84 @@
...
+
+        Returns:
+            TerminalSupportedFeatures: a new TerminalSupportedFeatures
+        """
+
+        # Using macOS, but not using the default terminal: let's assume we're on iTerm2
+        iterm2_synchronized_update = (
+            platform.system() == "Darwin"
+            and os.environ.get("TERM_PROGRAM", "") != "Apple_Terminal"
+        )
```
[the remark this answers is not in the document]
DrBenton: @darrenburns Following our discussion I ended up removing the iTerm2-specific code :slightly_smiling_face:  [996f25e](https://github.com/Textualize/textual/pull/533/commits/996f25effef8a850db658eae8db1fac7b5ce77a4)

src/textual/drivers/linux_driver.py
```diff
@@ -123,6 +124,11 @@ def on_terminal_resize(signum, stack) -> None:
         self._key_thread = Thread(target=self.run_input_thread, args=(loop,))
         send_size_event()
         self._key_thread.start()
+        self._request_terminal_mode_support(Mode.SynchronizedOutput)
+
+    def _request_terminal_mode_support(self, mode: Mode):
+        self.console.file.write(get_mode_request_sequence(mode) + "\n")
+        self.console.file.flush()
```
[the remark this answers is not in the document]
DrBenton: You're right, it works if I remove the newline character! :ok_hand: [2f0ae58](https://github.com/Textualize/textual/pull/533/commits/2f0ae58c23509f83d8c8763665f4d5df4c6ee2e0)

src/textual/_ansi_sequences.py
```diff
@@ -299,3 +304,16 @@
     "\x1b[1;8x": (Keys.Escape, Keys.ControlShift8),
     "\x1b[1;8y": (Keys.Escape, Keys.ControlShift9),
 }
+
+
+# Mapping of escape codes to report whether they support a "mode" we requested.
+ANSI_SEQUENCES_MODE_REPORTS: Dict[str, Tuple[Mode, ModeReportParameter]] = {
```
willmcgugan, line 310: Are we capturing all possible (now and future) responses? I have a feeling that a single regex would be more future safe option rather than enumerating all possible combinations.

src/textual/_ansi_sequences.py
```diff
@@ -299,3 +304,16 @@
...
+    get_mode_report_sequence(mode, parameter): (mode, parameter)
+    for mode, parameter in [
+        (mode, parameter) for parameter in ModeReportParameter for mode in Mode
+    ]
+}
+
+TERMINAL_MODES_ANSI_SEQUENCES: Dict[Mode, dict] = {
```
willmcgugan, line 317: Typing is imprecise here. I'm assuming the second dict could be `dict[str, str]` ?

src/textual/_terminal_features.py
```diff
@@ -0,0 +1,74 @@
...
+from textual._ansi_sequences import TERMINAL_MODES_ANSI_SEQUENCES
+from textual._terminal_modes import Mode
+
+
+class TerminalSupportedFeatures(NamedTuple):
+    """
+    Handles information about the features the current terminal emulator seems to support.
+    """
+
+    mode2026_synchronized_update: bool = False
```
willmcgugan, line 13: Correct me if I'm wrong, but it's not officially know as "mode 2026" which is rather cryptic.

Perhaps "synchronised_output" would be lest cryptic?

src/textual/_terminal_features.py
```diff
@@ -0,0 +1,74 @@
...
+    """@link https://gist.github.com/christianparpart/d8a62cc1ab659194337d73e399004036"""
+
+    @classmethod
+    def from_autodetect(cls) -> TerminalSupportedFeatures:
```
willmcgugan, line 17: This doesn't seem to do anything (currently). And I don't think we would be able to autodetect much outside of the driver.

Perhaps this should be a simple structure to record the supported features, without much functionality.

src/textual/events.py
```diff
@@ -411,3 +416,37 @@ class DescendantFocus(Event, verbosity=2, bubble=True):
 
 class DescendantBlur(Event, verbosity=2, bubble=True):
     pass
+
+
+@rich.repr.auto
+class ModeReport(InputEvent):
```
willmcgugan, line 422: This isn't really an InputEvent per se (i.e. key / mouse). It's also not really an event, as it is not relevant to any widgets.

This information is only relevant to the App, so it is probably best as a simple Message sent directly to the App class.

src/textual/app.py
```diff
@@ -858,14 +860,18 @@ def _display(self, renderable: RenderableType) -> None:
         """
         if self._running and not self._closed:
             console = self.console
-            if self._sync_available:
-                console.file.write("\x1bP=1s\x1b\\")
+            (
```
willmcgugan, line 863: This should probably be factored out in to two methods. Suggest `_begin_update` and `_end_update`

src/textual/app.py
```diff
@@ -939,6 +945,17 @@ async def on_event(self, event: events.Event) -> None:
             else:
                 # Forward the event to the view
                 await self.screen.forward_event(event)
+
+        elif isinstance(event, ModeReport):
```
willmcgugan, line 949: Probably shouldn't be an event. Makes more sense to be a `Message` which can be caught with `def handle_mode_report`.

src/textual/_terminal_features.py
```diff
@@ -0,0 +1,74 @@
...
+        # machinery with the terminal emulator - for now we should just assume it's not supported.
+        # See the use of the Mode and ModeReportParameter classes in the Textual code to check this machinery.
+        mode2026_synchronized_update = False
+
+        return cls(
+            mode2026_synchronized_update=mode2026_synchronized_update,
+        )
+
+    @property
+    def supports_synchronized_update(self) -> bool:
```
willmcgugan, line 40: Doesn't seem to be much need for an alias, when we have an attribute...

willmcgugan (changes requested):
I think there is some opportunity to simplify here.

Let's have a chat about it when you're ready.

src/textual/_terminal_features.py
DrBenton: Indeed, that's a remnant of the stage where this class was able to manage both the "mode 2026" and the iTerm2 own sync mode - in which case it was able to set the mode for iTerm2 by using env vars.
I can totally remove that method now that its implementation is empty :slightly_smiling_face:

src/textual/_terminal_features.py
DrBenton: yeah, that's also a remnant of when this class was supported the 2 "sync" modes (2026 and iTerm2) - I can drop it now that the behaviour is indeed linked to one single attribute :-)

DrBenton:
@willmcgugan Hopefully this new commit will address your feedback:  [7f27e70](https://github.com/Textualize/textual/pull/533/commits/7f27e70440c177b2a047b7f74a78ed5cd5b4b596) :slightly_smiling_face:

willmcgugan (changes requested):
As discussed, I think we can simplify to support just the one flag for now. And resurrect the code for the more general case should we need it.

DrBenton:
@willmcgugan Code simplified! :slightly_smiling_face: [43ce3e8](https://github.com/Textualize/textual/pull/533/commits/43ce3e8363b1a9b9889ea5be6039c702a7ecb230)

willmcgugan:
👍

PR #534, opened 2022-05-20
Fix escape key

willmcgugan:
Fixes the escape key event sent as a single char, and replaces it with `"escape"`

### DIFF
--- a/src/textual/keys.py
+++ b/src/textual/keys.py
@@ -200,11 +200,4 @@ class Keys(str, Enum):
     ShiftControlEnd = ControlShiftEnd
 
 
-KEY_BINDINGS = Keys._value2member_map_.values()
-
-
-@dataclass
-class Binding:
-    action: str
-    description: str
-    show: bool = False
+KEY_VALUES = frozenset(Keys._value2member_map_.values())
--- a/src/textual/widget.py
+++ b/src/textual/widget.py
@@ -814,6 +814,8 @@ class Widget(DOMNode):
         if layout:
             self._layout_required = True
         if repaint:
+            self._content_width_cache = (None, 0)
+            self._content_height_cache = (None, 0)
             self.set_dirty()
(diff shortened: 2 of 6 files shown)


src/textual/keys.py
```diff
@@ -200,11 +200,4 @@ class Keys(str, Enum):
     ShiftControlEnd = ControlShiftEnd
 
 
-KEY_BINDINGS = Keys._value2member_map_.values()
-
-
-@dataclass
```
willmcgugan, line 206: There is a Binding dataclass defined elsewhere.

src/textual/widget.py
```diff
@@ -814,6 +814,8 @@ def refresh(self, *, repaint: bool = True, layout: bool = False) -> None:
         if layout:
             self._layout_required = True
         if repaint:
+            self._content_width_cache = (None, 0)
```
willmcgugan, line 817: Clear the content size cache on repaint.

src/textual/keys.py
```diff
@@ -200,11 +200,4 @@ class Keys(str, Enum):
...
-class Binding:
-    action: str
-    description: str
-    show: bool = False
+KEY_VALUES = frozenset(Keys._value2member_map_.values())
```
willmcgugan, line 203: Values is sequence-like, a frozenset should have faster lookups

DrBenton, line 203: [picky mode] I noticed that this dunder property of Enums is not officially supported by Python, contrary to `__members__` for example - wouldn't it be a bit safer to resolve this via a simple list comprehension?
> https://docs.python.org/3/library/enum.html#supported-dunder-names

DrBenton (approved):
:+1: ! :-)

src/textual/keys.py
willmcgugan: Yes, you're absolutely correct.

PR #535, opened 2022-05-22
compensated in box mode for scrollbars

willmcgugan:


src/textual/widget.py
```diff
@@ -915,8 +944,6 @@ def on_blur(self, event: events.Blur) -> None:
 
     def on_descendant_focus(self, event: events.DescendantFocus) -> None:
         self.descendant_has_focus = True
-        if self.is_container and isinstance(event.sender, Widget):
```
willmcgugan, line 918: This was a naive solution to ensuring widgets were visible when focus. Never worked properly, and we need to revisit it.

PR #536, opened 2022-05-23
docs plugin

willmcgugan:
- Added a command to textual cli to run a textual app.

You can now run an app like this:

```
textual run basic.py
```

Where you may previously have done this:

```
python basic.py
```

Why? Well it allows setting things from the CLI which previously you could only set via env vars. Not everyone knows how to set env vars, particularly on Windows, and I don't want to have to educate people.

It also allows for a little less boilerplate in examples in docs. i.e. we won't have to repeat this:

```python
  app = MyApp(css_path="app.css")
  if __name__ == "__main__:
      app.run()
```

- Added a mkdocs extension to insert SVG screenshots in docs.

Screenshot 2022-05-23 at 17 33 13

https://user-images.githubusercontent.com/554369/169871817-2182b4c0-9f54-437c-992f-6d31f33c4ee7.mov

- Added a "headless" mode for the above, so we don't need to take over the terminal to get screenshots


src/textual/app.py
```diff
@@ -754,6 +762,28 @@ async def process_messages(self) -> None:
...
                 self._log_console = None
 
+    async def _ready(self) -> None:
+        """Called immediately prior to processing messages.
+
+        May be used as a hook for any operations that should run first.
+
+        """
+        try:
+            screenshot_timer = float(os.environ.get("TEXTUAL_SCREENSHOT", "0"))
```
DrBenton, line 772: I know this is made at build time and we're only waiting for 0.1 second which is quite short, but we may even use a fake clock to make it even faster if we want to, now that we have this machinery in place :slightly_smiling_face:

DrBenton (approved):
These docs are going to be so good! :heart_eyes:

src/textual/cli/cli.py
```diff
@@ -60,10 +59,41 @@ def import_app(import_name: str) -> App:
         global_vars: dict[str, object] = {}
         exec(py_code, global_vars)
 
-        try:
-            app = global_vars[name]
-        except KeyError:
-            raise AppFail(f"App {name!r} not found in {lib!r}")
+        if name:
+            # User has give a name, use that
```
DrBenton, line 63: "given"? (but I might be wrong :fr:  :sweat_smile: )

src/textual/app.py
willmcgugan: I'll revisit that at some point. I think we could avoid the delay entirely. If we have dozens of these, it might add up!

PR #538, opened 2022-05-24
Scrolling nested containers

darrenburns:
In the video below you can see me pressing a keybind to scroll to a nested (grandchild) widget.

https://user-images.githubusercontent.com/5740731/170022543-33d3cabb-daf4-4d3f-bc60-950848353d47.mov


sandbox/basic.css
```diff
@@ -69,14 +69,14 @@ App > Screen {
...
   background: $background;
   layout: vertical;
-  overflow-y: scroll; 
+  overflow-y: scroll;
 }
 
 
 Tweet {
-  height: auto;
+  height: 10;
```
DrBenton, line 77: this poor CSS is constantly updated by us in every direction :sweat_smile:

src/textual/widget.py
```diff
@@ -514,38 +514,50 @@ def scroll_to_widget(self, widget: Widget, *, animate: bool = True) -> bool:
...
-            delta_y = min(top_delta.y, bottom_delta.y, key=abs)
+            if node == self:
+                break
+            child = node
+            node = node.parent
 
-        return self.scroll_relative(
-            delta_x or None, delta_y or None, animate=animate, duration=0.2
-        )
+        # TODO: Return boolean value
```
DrBenton, line 560: shouldn't this TODO be addressed before we merge the PR?
(unless it's a complicated value to compute with this new version of the method, of course :-)

DrBenton (commented):
Wow, that was quick! Nice one :blush:

darrenburns:
@willmcgugan Could you have a look at this when you get a chance? It's essentially just the same code as before but inside a loop, applying going up the DOM.

It also updates it such that it's the whole region of the widget that scrolls into view, rather than just the content region.

willmcgugan (approved):
LGTM. I think you said there were some awkward edge cases not yet handled?

PR #539, opened 2022-05-24
Fix Option+Backspace hanging

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/488

### DIFF
--- a/src/textual/_ansi_sequences.py
+++ b/src/textual/_ansi_sequences.py
@@ -47,7 +47,7 @@ ANSI_SEQUENCES: Dict[str, Tuple[Keys, ...]] = {
     # support it. (Most terminals send ControlH when backspace is pressed.)
     # See: http://www.ibb.net/~anne/keyboard.html
     "\x7f": (Keys.ControlH,),
-    # --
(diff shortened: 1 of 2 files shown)


src/textual/_ansi_sequences.py
```diff
@@ -47,7 +47,7 @@
     # support it. (Most terminals send ControlH when backspace is pressed.)
     # See: http://www.ibb.net/~anne/keyboard.html
     "\x7f": (Keys.ControlH,),
-    # --
+    "\x1b\x7f": (Keys.ControlW,),
```
willmcgugan, line 50: Ctrl+W does the same thing as Option+Backspace so I assume they are the same

src/textual/cli/cli.py
```diff
@@ -70,7 +70,7 @@ def import_app(import_name: str) -> App:
             if "app" in global_vars:
                 # App exists, lets use that
                 try:
-                    app = global_vars[name]
+                    app = global_vars["app"]
```
willmcgugan, line 73: fix for `textual.run`

DrBenton (approved):
:ok_hand:

PR #540, opened 2022-05-25
Docs Refresh

willmcgugan:
Please don't proof read the docs just yet. Will request that in a later updated

- Updated mkdocs and associated libs
- Defines a structure in the docs
- Adds a few docs, and a mkdocs plugin to render a screenshot to SVG
- Removes `style` parameter from render method
- Fixes issue with simple case of widget on an App


docs/introduction.md
```diff
@@ -0,0 +1,147 @@
...
+
+## Pre-requisites
+
+- Python 3.7 or later. If you have a choice, pick the most recent version.
+- Installed `textual` from Pypi.
+- Basic Python skills.
+
+## A Simple App
+
+Lets looks at the simplest possible Textual app. It doesn't do much, but will demonstrate the basic steps you will need to create any application.
```
DrBenton, line 15:
```suggestion
Let's looks at the simplest possible Textual app. It doesn't do much, but will demonstrate the basic steps you will need to create any application.
```

DrBenton (approved):
Awesome! :heart_eyes: 

As we were discussing, it could be nice to get rid of that odd spacing in the API documentation headers:
Screenshot from 2022-05-27 11-32-59

It seems that just setting `line-height: 1;` to a `h2.doc.doc-heading` selector does the job?
Screenshot from 2022-05-27 11-33-06

(it seems that we can simply add a `extra.css` file in the docs folder make it be picked automatically by MKDocs? - [doc](https://www.mkdocs.org/user-guide/customizing-your-theme/))

willmcgugan:
There is definitely room for tweaking those API reference pages. They don't look great. Will come back to it.

pawamoy:
There are some options you could use to tweak the rendering of API reference, given you are using the new handler and not the legacy one:

```yaml
- mkdocstrings:
    handlers:
      python:
        selection:
          docstring_style: google
          docstring_options:
            ignore_init_summary: yes
        rendering:
          merge_init_into_class: yes
          separate_signature: yes
          line_length: 80
```

If Black is installed, the signatures are formatted, respecting the specified line length.

willmcgugan:
Thanks Timothée!

PR #545, opened 2022-05-27
Refresh scrollbars on style changes.

darrenburns:
Previously scrollbar visibility was only refreshed when the size of a widget changes.

However, when you modify the CSS with "watch_css" enabled, scrollbar visibility may change despite the size staying the same.

I suspect (but didn't confirm) that scrollbars wouldn't appear if you toggled overflow_x/y using Python at runtime too.

- Refresh scrollbars on style changes.
- Instantiate text input visible range to (0,0) by default.
- Remove debug code.

### DIFF
--- a/src/textual/widgets/text_input.py
+++ b/src/textual/widgets/text_input.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import time
 from typing import Callable
 
 from rich.cells import cell_len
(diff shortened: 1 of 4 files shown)


src/textual/widgets/text_input.py
```diff
@@ -282,18 +281,6 @@ def on_key(self, event: events.Key) -> None:
...
-        ):  # TODO: This allows us to query and print the text input state
-            self.log(start=start)
-            self.log(end=end)
-            self.log(visible_content=visible_content)
-            self.log(visible_content_cell_len=visible_content_cell_len)
-            self.log(
-                visible_content_to_cursor_cell_len=visible_content_to_cursor_cell_len
-            )
-            self.log(available_width=available_width)
-            self.log(cursor_index=self._editor.cursor_index)
```
DrBenton, line 296: I like the way that code tells a story... A dreadful story of a complex feature development.... :sweat_smile:

DrBenton (approved):

willmcgugan (approved):

PR #546, opened 2022-05-27
Widget focus keybindings

### LINKED ISSUE #504: Add built-in action handler for switching focus (e.g. `self.bind("f1", "focus(widget_id)")`)

darrenburns:
You can create a keybinding to immediately focus a widget based on the widget ID. For example:

```python
    def on_load(self) -> None:
        self.bind("1", "focus('widget1')")
```

This will focus the widget with the ID `widget1` when you press `1` on your keyboard.

<hr>

Here's an example video showing how we can use the keyboard to quickly shift our focus across multiple widgets:

https://user-images.githubusercontent.com/5740731/170708435-523003e3-6307-4345-b369-ab1932f6f037.mov

This demo is available in `sandbox/focus_keybindings.py`.

<hr>

This PR also fixes an issue with descendant_focus, although the approach is quite inefficient.

I would like to write an integration test for this, but it depends on some of the in-flux stuff inside #543, so I will wait til that is merged before doing so.

Closes #504

### DIFF
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -1085,10 +1085,11 @@ class App(Generic[ReturnType], DOMNode):
     async def action_focus(self, widget_id: str) -> None:
         try:
             node = self.query(f"#{widget_id}").first()
-            if isinstance(node, Widget):
-                self.set_focus(node)
         except NoMatchingNodesError:
             pass
+        else:
+            if isinstance(node, Widget):
+                self.set_focus(node)
 
     async def action_add_class_(self, selector: str, class_name: str) -> None:
         self.screen.query(selector).add_class(class_name)


sandbox/focus_keybindings.py
```diff
@@ -0,0 +1,54 @@
+from textual.app import App
+from textual.widget import Widget
+from textual.widgets import Static
+
+
+class FocusKeybindsApp(App):
+    dark = True
+
+    def on_load(self) -> None:
+        self.bind("1", "focus('widget1')")
```
DrBenton, line 10: That's indeed a typical use case of PR #543 ! :slightly_smiling_face: 
Once it's merged this code could be rewritten into something like this:
```py
    def on_load(self) -> None:
        target_keys = ("1", "2", "3", "4", "q", "w", "e", "r")
        self.bind(",".join(target_keys), "widget_key('$event.key')")

    def action_widget_key(self, key: str) -> None:
        self.focus(f"widget{key}")

File: src/textual/app.py [hunk 2]
```diff
@@ -1081,6 +1082,14 @@ async def action_bang(self) -> None:
     async def action_bell(self) -> None:
         self.bell()
 
+    async def action_focus(self, widget_id: str) -> None:
+        try:
+            node = self.query(f"#{widget_id}").first()
+            if isinstance(node, Widget):
+                self.set_focus(node)
+        except NoMatchingNodesError:
+            pass
```
DrBenton, line 1091: We could maybe shrink the scope of the `try / except` block, and replace it with an _early return_?
(but that's very subjective of course, so feel free to ignore this comment :smile: )
```suggestion
        try:
            node = self.query(f"#{widget_id}").first()
        except NoMatchingNodesError:
            return
        if isinstance(node, Widget):
            self.set_focus(node)
```

src/textual/widget.py
```diff
@@ -993,9 +993,21 @@ def on_blur(self, event: events.Blur) -> None:
 
     def on_descendant_focus(self, event: events.DescendantFocus) -> None:
         self.descendant_has_focus = True
+        if "focus-within" in self.pseudo_classes:
```
DrBenton, line 996: Maybe this `if` block can be factorised to a private method, since it seems to be exactly the same in the 2 methods?

But we might also want to wait for the hypothetic 3rd occurrence of it to appear before doing so :slightly_smiling_face: 
> The Rule of Three
> 
>     The first time you do something, you just do it. The second time you do
>     something similar, you wince at the duplication, but you do the duplicate
>     thing anyway. The third time you do something similar, you refactor.

sandbox/focus_keybindings.py
DrBenton: it seems that #543 won't be merged before a while, though, so I guess we'll have to merge this one without it as a first iteration :slightly_smiling_face:

src/textual/app.py
willmcgugan: I'd use the else: clause here. I too like the smaller scope for the except block

```suggestion
        try:
            node = self.query(f"#{widget_id}").first()
        except NoMatchingNodesError:
            pass
        else:
            if isinstance(node, Widget):
                self.set_focus(node)
```

src/textual/widget.py
willmcgugan: Good point. If there is a third, we probably want a `refresh_children`.

willmcgugan (approved):
Looks good. I'll leave that suggested change up to you. Merge when you're happy.

PR #552, opened 2022-05-29
Box model fractional

willmcgugan:
- Fixes rounding issues in layout, i.e. Two widgets with width: 50% will consume the full width, even if it is odd
-  Implements fr units for horizontal and vertical layouts

### DIFF
--- a/src/textual/css/scalar.py
+++ b/src/textual/css/scalar.py
@@ -1,13 +1,14 @@
 from __future__ import annotations
 
 from enum import Enum, unique
+from fractions import Fraction
 from functools import lru_cache
 import re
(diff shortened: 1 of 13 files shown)


src/textual/css/scalar.py
```diff
@@ -48,19 +51,118 @@ class Unit(Enum):
 
 _MATCH_SCALAR = re.compile(r"^(-?\d+\.?\d*)(fr|%|w|h|vw|vh)?$").match
 
+
```
willmcgugan, line 54: Converted from lambdas

willmcgugan:

PR #553, opened 2022-05-29
simplify scrollbar dimensions

willmcgugan:
The logic to get / set scrollbar size was unnecessarily convoluted

### DIFF
--- a/src/textual/widget.py
+++ b/src/textual/widget.py
@@ -192,7 +192,7 @@ class Widget(DOMNode):
             assert self.layout is not None
             return (
                 self.layout.get_content_width(self, container, viewport)
-                + self.scrollbar_width
+                + self.scrollbar_size_vertical
             )
 
         cache_key = container.width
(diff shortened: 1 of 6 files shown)


src/textual/widget.py
```diff
@@ -368,20 +368,17 @@ def scrollbars_enabled(self) -> tuple[bool, bool]:
...
         """Get the size of any scrollbars on the widget"""
-        return (
-            self._get_scrollbar_thickness_horizontal(),
-            self._get_scrollbar_thickness_vertical(),
-        )
+        return (self.scrollbar_size_horizontal, self.scrollbar_size_vertical)
 
     @property
-    def scrollbar_width(self) -> int:
+    def scrollbar_size_vertical(self) -> int:
```
willmcgugan, line 374: In retrospect the original was a poor name. It is the "width" of the "vertical" scrollbar, which is accurate, but muddles a few concepts.

willmcgugan:
Taking the liberty of merging #executivedecision

PR #555, opened 2022-05-30
Cache arrangements

willmcgugan:
- Cache layout operation
- Do layout operations on idle

### DIFF
--- a/src/textual/message_pump.py
+++ b/src/textual/message_pump.py
@@ -274,7 +274,11 @@ class MessagePump:
                         for _cls, method in self._get_dispatch_methods(
                             "on_idle", event
                         ):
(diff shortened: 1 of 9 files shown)


src/textual/message_pump.py
```diff
@@ -274,7 +274,11 @@ async def _process_messages(self) -> None:
                         for _cls, method in self._get_dispatch_methods(
                             "on_idle", event
                         ):
-                            await invoke(method, event)
+                            try:
+                                await invoke(method, event)
```
willmcgugan, line 278: This was being silently ignored. PITA to track down.

willmcgugan:
Merging #executivedecision

PR #566, opened 2022-06-07
Support HSL colour space, allow spaces in RGB/HSL values

### LINKED ISSUE #565: Support HSL and HSLA colors in CSS
Suspect this is fairly easy todo, and a nice feature.

darrenburns:
- Allows you to specify `hsl` and `hsla` colours in CSS.
- Updates `Color.parse` to accept these too.
- You can now use spaces in the colour strings in CSS and args to `Color.parse` (e.g. `rgb( 255, 26, 3 )`)
- Improves test coverage around color parsing

Closes #565

### DIFF
--- a/src/textual/color.py
+++ b/src/textual/color.py
@@ -288,11 +288,7 @@ class Color(NamedTuple):
         if color_match is None:
             error_message = f"failed to parse {color_text!r} as a color"
             suggested_color = None
-            if (
-                not color_text.startswith("#")
-                and not color_text.startswith("rgb")
-                and not color_text.startswith("hsl")
-            ):
+            if not color_text.startswith(('#', 'rgb', 'hsl')):
                 # Seems like we tried to use a color name: let's try to find one that is close enough:
                 suggested_color = get_suggestion(color_text, COLOR_NAME_TO_RGB.keys())
                 if suggested_color:
--- a/src/textual/color.py
+++ b/src/textual/color.py
@@ -127,9 +127,7 @@ class Color(NamedTuple):
         Returns:
             Color: A new color.
         """
-        print("A")
         r, g, b = hls_to_rgb(h, l, s)
-        print("B")
         return cls(int(r * 255 + 0.5), int(g * 255 + 0.5), int(b * 255 + 0.5))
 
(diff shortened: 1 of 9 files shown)


src/textual/color.py
```diff
@@ -329,6 +339,19 @@ def parse(cls, color_text: str | Color) -> Color:
                 clamp(int(float_b), 0, 255),
                 clamp(float_a, 0.0, 1.0),
             )
+        elif hsl is not None:
+            h, s, l = hsl.split(",")
+            h = clamp(float(h), 0, 360) / 360
```
willmcgugan, line 344: I'm not sure that the clamping is the right thing to do here. An `h` of -10 shouldn't be 0, it should be 350.

```suggestion
            h = float(h) % 360
```

Suggest using different names for the strs and the floats. I suspect Mypy might complain about this.

src/textual/css/scalar.py
```diff
@@ -326,6 +326,17 @@ def resolve(self, size: Size, viewport: Size) -> Offset:
 
 NULL_SCALAR = ScalarOffset(Scalar.from_number(0), Scalar.from_number(0))
 
+
+def percentage_string_to_float(string: str) -> float:
```
willmcgugan, line 330: Docstring por favor

src/textual/color.py
darrenburns: Good point!

darrenburns: Still needed to `/ 360` at the end

tests/test_color.py
```diff
@@ -108,12 +108,34 @@ def test_color_blend():
         ("rgb(2,3,4)", Color(2, 3, 4, 1.0)),
         ("rgba(2,3,4,1.0)", Color(2, 3, 4, 1.0)),
         ("rgba(2,3,4,0.058823529411764705)", Color(2, 3, 4, 0.058823529411764705)),
+        ("hsl(45,25%,25%)", Color(80, 72, 48)),
```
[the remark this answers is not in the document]
darrenburns: Done

willmcgugan (approved):
LGTM

PR #567, opened 2022-06-08
Support for bracketed paste mode

### LINKED ISSUE #564: Implement bracketed paste mode
We need to implement "bracketed paste" mode to allow pasting in to text input.

I think they way it works is that you get a sequence that say that bracket paste mode has started, then a bunch of keys, then something to say it has ended. This should be translated in to a `events.Paste` containing the string of characters just pasted.

darrenburns:
- Adds support for bracketed paste mode.
- When text is pasted, it gets forwarded to the focused widget as a new `Paste` event, which contains a `text` attribute.
- Adds support in the TextInput widget for pasting text. If the pasted text contains newline characters, they are stripped.
- This mode is enabled by sending a sequence to the terminal emulator when the application starts.
- It's then disabled when the application shuts down by sending another sequence.
- While this mode is active, the terminal will wrap any pasted text in special escape sequences, which allows us to differentiate it from common input.

Closes #564.


src/textual/driver.py
```diff
@@ -41,6 +41,14 @@ def process_event(self, event: events.Event) -> None:
             click_event = events.Click.from_event(event)
             self.send_event(click_event)
 
+    def enable_bracketed_paste(self) -> None:
```
willmcgugan, line 44: docstrings お願いします

src/textual/app.py
```diff
@@ -758,6 +758,7 @@ async def process_messages(self) -> None:
 
             driver = self._driver = self.driver_class(self.console, self)
             driver.start_application_mode()
+            driver.enable_bracketed_paste()
```
willmcgugan, line 761: Would it make sense to do enable the bracketed paste mode as part of `start_application_mode` ?

willmcgugan (approved):
LGTM!

Couple of comments to consider. But good to merge.

src/textual/app.py
darrenburns: I think if I would do that, I'd probably want to rename that method as they're distinct things in my head. I'll leave it like this for now

PR #570, opened 2022-06-09
XTerm parsing improvements

darrenburns:
- Added a unit test suite covering the XTermParser code
- Fixes freeze/hanging issue: When a candidate escape sequence grows in length to exceed a threshold, or when another \x1b is found, it'll backtrack and treat the characters until that point as keypresses (and escape codes that follow the non-matched sequence won't be lost).

### DIFF
--- a/src/textual/_xterm_parser.py
+++ b/src/textual/_xterm_parser.py
@@ -94,6 +94,7 @@ class XTermParser(Parser[events.Event]):
         more_data = self.more_data
         paste_buffer: list[str] = []
         bracketed_paste = False
+        reuse_escape = False
 
         while not self.is_eof:
             if not bracketed_paste and paste_buffer:
@@ -107,7 +108,8 @@ class XTermParser(Parser[events.Event]):
                 on_token(events.Paste(self.sender, text=pasted_text))
                 paste_buffer.clear()
 
-            character = yield read1()
+            character = ESC if reuse_escape else (yield read1())
+            reuse_escape = False
 
             if bracketed_paste:
                 paste_buffer.append(character)
@@ -137,22 +139,35 @@ class XTermParser(Parser[events.Event]):
                     # If we run into another ESC at this point, then we've failed
                     # to find a match, and should issue everything we've seen within
                     # the suspected sequence as Key events instead.
-                    buffer = yield self.peek_buffer()
+                    sequence_character = yield read1()
+                    next_sequence = sequence + sequence_character
                     if (
-                        buffer
-                        and buffer[0] == ESC
+                        sequence_character
+                        and sequence_character == ESC
                         or len(sequence) > _MAX_SEQUENCE_SEARCH_THRESHOLD
                     ):
-                        for character in sequence:
+                        if sequence_character == ESC:
+                            # We've hit an escape, so we need to issue all the keys
+                            # up to but not including it, since this escape could be
+                            # part of an upcoming control sequence.
+                            reuse_escape = True
+                            reissue_sequence = sequence
+                        else:
+                            # We exceeded the sequence length threshold, so reissue all the
+                            # characters in that sequence as key-presses.
+                            reissue_sequence = next_sequence
+
+                        for character in reissue_sequence:
                             key_events = sequence_to_key_events(character)
                             for event in key_events:
                                 if event.key == "escape":
                                     event = events.Key(event.sender, key="^")
                                 on_token(event)
+
+                        # We're done looking for escape sequences
                         break
 
-                    sequence_character = yield read1()
-                    sequence += sequence_character
+                    sequence = next_sequence
 
                     self.debug_log(f"sequence={sequence!r}")
 
--- a/src/textual/_xterm_parser.py
+++ b/src/textual/_xterm_parser.py
@@ -1,18 +1,19 @@
 from __future__ import annotations
 
-
 import re
 from typing import Any, Callable, Generator, Iterable
 
-from . import messages
 from . import events
-from ._types import MessageTarget
-from ._parser import Awaitable, Parser, TokenCallback
+from . import messages
 from ._ansi_sequences import ANSI_SEQUENCES_KEYS
+from ._parser import Awaitable, Parser, TokenCallback
+from ._types import MessageTarget
 
 # When trying to determine whether the current sequence is a supported/valid
 # escape sequence, at which length should we give up and consider our search
 # to be unsuccessful?
+from .keys import Keys
+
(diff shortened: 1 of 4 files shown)


src/textual/_xterm_parser.py
```diff
@@ -30,7 +36,7 @@ def __init__(
         self.last_x = 0
         self.last_y = 0
 
-        self._debug_log_file = open("keys.log", "wt") if debug else None
+        self._debug_log_file = open("keys.log", "wt")
```
darrenburns, line 39: Note to self @darrenburns - revert this line

src/textual/_xterm_parser.py
```diff
@@ -10,6 +10,11 @@
 from ._parser import Awaitable, Parser, TokenCallback
 from ._ansi_sequences import ANSI_SEQUENCES_KEYS
 
+# When trying to determine whether the current sequence is a supported/valid
+# escape sequence, at which length should we give up and consider our search
+# to be unsuccessful?
+_MAX_SEQUENCE_SEARCH_THRESHOLD = 20
+
 _re_mouse_event = re.compile("^" + re.escape("\x1b[") + r"(<?[\d;]+[mM]|M...)\Z")
```
darrenburns, line 18: There are two regexes for mouse events that run.

This one suggests a mouse event can have any number of parameters, but when we parse the mouse event it assumes there are a fixed number.

Is this one definitely required? Could we possibly get away with running just the 1 regex for mouse events rather than 2?

src/textual/_xterm_parser.py
```diff
@@ -10,6 +10,11 @@
 from ._parser import Awaitable, Parser, TokenCallback
 from ._ansi_sequences import ANSI_SEQUENCES_KEYS
 
+# When trying to determine whether the current sequence is a supported/valid
+# escape sequence, at which length should we give up and consider our search
+# to be unsuccessful?
+_MAX_SEQUENCE_SEARCH_THRESHOLD = 20
```
willmcgugan, line 16: Why was 20 chosen?

src/textual/_xterm_parser.py
willmcgugan: Quite possibly. IIRC there are at least two mouse codes that can be generated. At one point I thought I would have to support both, but everything supports the newer method. That might explain the dual regexes.

src/textual/_xterm_parser.py
```diff
@@ -34,7 +39,7 @@ def __init__(
 
         super().__init__()
 
-    def debug_log(self, *args: Any) -> None:
+    def debug_log(self, *args: Any) -> None:  # pragma: no cover
```
willmcgugan, line 42: What was the typing error?

src/textual/_xterm_parser.py
```diff
@@ -129,25 +131,42 @@ def parse(self, on_token: TokenCallback) -> Generator[Awaitable, str, None]:
...
                             continue
 
+                # Look ahead through the suspected escape sequence for a match
                 while True:
-                    sequence += yield read1()
+
+                    # If we run into another ESC at this point, then we've failed
+                    # to find a match, and should issue everything we've seen within
+                    # the suspected sequence as Key events instead.
+                    buffer = yield self.peek_buffer()
```
willmcgugan, line 140: This doesn't look right. `peek_buffer` may return empty string if there is nothing in the buffer, but it could still read an `ESC` on L155

src/textual/_xterm_parser.py
```diff
@@ -129,25 +131,42 @@ def parse(self, on_token: TokenCallback) -> Generator[Awaitable, str, None]:
...
+                    if (
+                        buffer
+                        and buffer[0] == ESC
+                        or len(sequence) > _MAX_SEQUENCE_SEARCH_THRESHOLD
+                    ):
+                        for character in sequence:
+                            keys = get_key_ansi_sequence(character, None)
```
willmcgugan, line 147: I suspect that if we have an `ESC` to be sent from an unrecognised sequence, we should send a `^`  since other wise it would be seen as the user hitting the literal escape key.

src/textual/_xterm_parser.py
```diff
@@ -129,25 +131,42 @@ def parse(self, on_token: TokenCallback) -> Generator[Awaitable, str, None]:
...
+                    # If we run into another ESC at this point, then we've failed
+                    # to find a match, and should issue everything we've seen within
+                    # the suspected sequence as Key events instead.
+                    buffer = yield self.peek_buffer()
+                    if (
+                        buffer
+                        and buffer[0] == ESC
+                        or len(sequence) > _MAX_SEQUENCE_SEARCH_THRESHOLD
+                    ):
+                        for character in sequence:
```
willmcgugan, line 146: There's some very similar code further down, I wonder if it could be hoisted in to a closure?

willmcgugan (commented):
Mostly questions

src/textual/_xterm_parser.py
darrenburns: The largest that I think we could expect to receive (that I'm aware of) is length 14. The existence of the `re_sgr_mouse` regex suggested to me that longer sequences could exist, so I added on some leeway for defensiveness.

src/textual/_xterm_parser.py
darrenburns: There was no typing error. I was trying to ensure we had full test coverage of this module, and thought this code was temporary "for us" code so didn't think coverage.py should measure it. I can remove this and add a test case around it if you'd prefer.

src/textual/_xterm_parser.py
darrenburns: Yeah, you might be right. I was thinking that we might not know if it was literally the user pressing an escape key followed by some data that doesn't match an ANSI sequence.

src/textual/_xterm_parser.py
darrenburns: I considered this, but will do it now that I know I'm not the only one with that thought :)

darrenburns: Oh, you said closure...

I've factored it out into a separate method

src/textual/_xterm_parser.py
darrenburns: Have made it such that when we perform the backtracking, `\x1b` gets converted to `^` rather than `escape`.

src/textual/_xterm_parser.py
darrenburns: I'll think about how I can get a test to cover this scenario and fix it on Tuesday.

PR #575, opened 2022-06-10
Add success/warning/error button variants

darrenburns:
I kept the primary button as is on dark mode. I think I'd prefer if it was the blue/primary background colour on both light and dark mode, but not too fussed either way.

One kind of annoying thing is the "success" button is the only one where, on hover, the text colour changes.

https://user-images.githubusercontent.com/5740731/173611944-6ff07367-60a8-4c45-98ae-a83a2e8148fc.mov

### DIFF
--- a/src/textual/widgets/_button.py
+++ b/src/textual/widgets/_button.py
@@ -69,7 +69,7 @@ class Button(Widget, can_focus=True):
     Button.-success:hover {
         background: $success-darken-1;
         color: $text-success-darken-1;
-        border: tall $success-lighten-2;  /* TODO: This shouldn't be necessary?? */
+        border: tall $success-lighten-2;
     }
 
     .-dark-mode Button.-success {
@@ -109,6 +109,7 @@ class Button(Widget, can_focus=True):
         border: tall $warning-lighten-3;
     }
 
+    /* Error variant */
     Button.-error {
         background: $error;
         color: $text-error;
--- a/src/textual/widgets/_button.py
+++ b/src/textual/widgets/_button.py
@@ -1,6 +1,12 @@
 from __future__ import annotations
 
-from typing import cast, Literal
+import sys
+from typing import cast
+
+if sys.version_info >= (3, 8):
+    from typing import Literal
+else:
+    from typing_extensions import Literal  # pragma: no cover
 
 from rich.console import RenderableType
 from rich.text import Text, TextType
@@ -130,7 +136,6 @@ class Button(Widget, can_focus=True):
     App.-show-focus Button:focus {
         tint: $accent 20%;
     }
-
     """
 
     class Pressed(Message, bubble=True):
--- a/src/textual/widgets/_button.py
+++ b/src/textual/widgets/_button.py
@@ -1,51 +1,130 @@
 from __future__ import annotations
 
-from typing import cast
+from typing import cast, Literal
 
 from rich.console import RenderableType
-from rich.style import Style
 from rich.text import Text, TextType
 
 from .. import events
+from ..css._error_tools import friendly_list
 from ..message import Message
 from ..reactive import Reactive
 from ..widget import Widget
+
+ButtonVariant = Literal["default", "success", "warning", "error"]
+_VALID_BUTTON_VARIANTS = {"default", "success", "warning", "error"}
+
+
+class InvalidButtonVariant(Exception):
+    pass
 
 
 class Button(Widget, can_focus=True):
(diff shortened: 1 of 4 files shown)


src/textual/widgets/_button.py
```diff
@@ -1,57 +1,142 @@
...
 from rich.text import Text, TextType
 
 from .. import events
+from ..css._error_tools import friendly_list
 from ..message import Message
 from ..reactive import Reactive
 from ..widget import Widget
 
+ButtonVariant = Literal["default", "success", "warning", "error"]
+_VALID_BUTTON_VARIANTS = {"default", "success", "warning", "error"}
```
darrenburns, line 21: I hate this :(

willmcgugan: The repetition? Yeah, feels so awkward.

src/textual/widgets/_button.py
```diff
@@ -100,3 +205,93 @@ async def on_click(self, event: events.Click) -> None:
     async def on_key(self, event: events.Key) -> None:
         if event.key == "enter" and not self.disabled:
             await self.emit(Button.Pressed(self))
+
+    @staticmethod
+    def success(
```
darrenburns, line 210: I've added both the `variant` constructor arg and these utility constructors. I'm not sure which is best. The utility constructors feel like they may be a bit of a maintenance burden with keeping the signatures in sync (unless we do some `getattr` craziness).

Happy to go with either or both, or none if anyone has other thoughts.

src/textual/widgets/_button.py
darrenburns: Yep!

willmcgugan:
Can you post a video of the hover states?

src/textual/widgets/_button.py
DrBenton: `_VALID_BUTTON_VARIANTS = ButtonVariant.__args__` would work - and the signature of the Literal class is unlikely to change (it would probably break types checkers).
Not very satisfying for sure, as it feels a bit hacky, but it avoids the repetition :slightly_smiling_face:

src/textual/widgets/_button.py
```diff
@@ -1,57 +1,142 @@
...
+from ..css._error_tools import friendly_list
 from ..message import Message
 from ..reactive import Reactive
 from ..widget import Widget
 
+ButtonVariant = Literal["default", "success", "warning", "error"]
+_VALID_BUTTON_VARIANTS = {"default", "success", "warning", "error"}
+
+
+class InvalidButtonVariant(Exception):
```
DrBenton, line 24: I wonder if our custom exceptions shouldn't at least be sub-classes of a common "Textual" exception, rather than "root" Python ones? So people could do something like `except TextualException: ...` ? :thinking: 
But maybe that's unnecessary additional intra-dependencies :-)

src/textual/widgets/_button.py
```diff
@@ -100,3 +205,93 @@ async def on_click(self, event: events.Click) -> None:
     async def on_key(self, event: events.Key) -> None:
         if event.key == "enter" and not self.disabled:
             await self.emit(Button.Pressed(self))
+
+    @staticmethod
```
DrBenton, line 209: shouldn't this be a `@classmethod`, so sub-classes of this `Button` class could re-use these helper constructors? :slightly_smiling_face:

src/textual/widgets/_button.py
willmcgugan: I think you would get a less efficient tuple there, and I'm not too keen on the use of the dunder attributes.

src/textual/widgets/_button.py
willmcgugan: I'm happy with both. It is a maintenance burden, but we suffer so Textual devs can save a second or two.

willmcgugan:
@darrenburns is there a style for the disabled state?

src/textual/widgets/_button.py
darrenburns: Changed

willmcgugan:
@darrenburns Can you add a disabled state please. `opacity: 70%` maybe.

src/textual/widgets/_button.py
willmcgugan: There should definitely be an exception hierarchy, but a `TextualException` is too broad. I wouldn't want to encourage people to catch anything that Textual can throw.

willmcgugan (approved):
Legolas Goes to Mordor

PR #579, opened 2022-06-14
Make devtools deps extras, add note to internal ref docs, raise excep…

darrenburns:
…tion when deps missing

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -25,7 +25,7 @@ jobs:
           version: 1.1.6
           virtualenvs-in-project: true
       - name: Install dependencies
-        run: poetry install
+        run: poetry install --extras "dev"
         if: steps.cached-poetry-dependencies.outputs.cache-hit != 'true'
       - name: Format check with black
         run: |
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,7 +27,10 @@ rich = "^12.4.3"
 click = "8.1.2"
 importlib-metadata = "^4.11.3"
 typing-extensions = { version = "^4.0.0", python = "<3.8" }
-msgpack = "^1.0.3"
+
+# Dependencies below are required for devtools only
+aiohttp = { version = "^3.8.1", optional = true }
+msgpack = { version = "^1.0.3", optional = true }
 
 [tool.poetry.dev-dependencies]
 pytest = "^6.2.3"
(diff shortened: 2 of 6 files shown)


.github/workflows/pythonpackage.yml
```diff
@@ -25,7 +25,7 @@ jobs:
           version: 1.1.6
           virtualenvs-in-project: true
       - name: Install dependencies
-        run: poetry install
+        run: poetry install --extras "dev"
```
darrenburns, line 28: https://python-poetry.org/docs/pyproject/#extras

pyproject.toml
```diff
@@ -27,7 +27,10 @@ rich = "^12.4.3"
 click = "8.1.2"
 importlib-metadata = "^4.11.3"
 typing-extensions = { version = "^4.0.0", python = "<3.8" }
-msgpack = "^1.0.3"
+
+# Dependencies below are required for devtools only
+aiohttp = { version = "^3.8.1", optional = true }
```
darrenburns, line 32: @willmcgugan Do we move `click` into the `dev` extras too?

reference/_devtools.md
```diff
@@ -0,0 +1,12 @@
+# Devtools
```
darrenburns, line 1: These are more just notes to remind us what to include in the docs when the structure is in place. They're not the final docs.

DrBenton (approved):
LGTM! :ok_hand:

pyproject.toml
willmcgugan: I think so. Is it possible to install the `textual` command only with `dev` extras?

src/textual/devtools/client.py
```diff
@@ -3,19 +3,34 @@
...
+    import aiohttp
+    from aiohttp import (
+        ClientResponseError,
+        ClientConnectorError,
+        ClientWebSocketResponse,
+    )
+    import msgpack
+except ImportError:
+    raise DevtoolsDependenciesMissingError(
+        "Textual Devtools requires installation of the 'dev' extra dependencies. "
```
willmcgugan, line 31: Can you add the command required to install the dev extras? We should always assume novice devs.

willmcgugan: A link to the docs would work. Maybe add a TODO then...

pyproject.toml
darrenburns: I'm not sure I understand the question, but I don't believe you can install only the extras.

willmcgugan: We should only install the `textual` CLI when you `pip install textual[dev]`. Because we don't need it just to distribute apps. Is that possible?

darrenburns: I don't believe so. Think we'd have to distribute the devtools as a separate package.

willmcgugan (approved):

PR #582, opened 2022-06-16
Tests around CSS specificity, and fix ordering in case of specificity clash

darrenburns:
When 2 rules have matching specificity, take the final one, not the first.

### DIFF
--- a/src/textual/css/stylesheet.py
+++ b/src/textual/css/stylesheet.py
@@ -288,7 +288,7 @@ class Stylesheet:
             rule_attributes[key].append((default_specificity, value))
 
         # Collect the rules defined in the stylesheet
-        for rule in self.rules:
+        for rule in reversed(self.rules):
             for specificity in _check_rule(rule, node):
                 for key, rule_specificity, value in rule.styles.extract_rules(
                     specificity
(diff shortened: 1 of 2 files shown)


tests/css/test_stylesheet.py
```diff
@@ -7,6 +7,98 @@
...
+
+def test_stylesheet_apply_highest_specificity_wins_multiple_classes():
+    """When we use two selectors containing only classes, then the selector
+    `.b.c` has greater specificity than the selector `.a`"""
+    css = ".b.c {background: blue;} .a {background: red; color: lime;}"
+    stylesheet = _make_stylesheet(css)
+    node = DOMNode(classes="a b c")
+    stylesheet.apply(node)
+
+    assert node.styles.background == Color(0, 0, 255)
```
DrBenton, line 49: maybe using `COLOR_NAME_TO_RGB` would make these tests a bit easier to read? :slightly_smiling_face: 
```suggestion
    assert node.styles.background == COLOR_NAME_TO_RGB["blue"]
```

DrBenton (approved):
Nice to see these CSS selector specificity tests ^_^

tests/css/test_stylesheet.py
darrenburns: I did have `Color.parse("ansi_blue")` before but didn't think it added enough as the RGB tuple is pretty clear :D

src/textual/css/stylesheet.py
```diff
@@ -297,13 +297,16 @@ def apply(self, node: DOMNode, animate: bool = False) -> None:
...
         # For each rule declared for this node, keep only the most specific one
         get_first_item = itemgetter(0)
-        node_rules: RulesMap = cast(
-            RulesMap,
-            {
-                name: max(specificity_rules, key=get_first_item)[1]
-                for name, specificity_rules in rule_attributes.items()
-            },
-        )
+        node_rules: RulesMap = cast(RulesMap, {})
```
willmcgugan, line 300: Humor me. I suspect this is more work that required.

Can you restore this code, and change L291 to:

```python
    for rule in reversed(self.rules):
```

darrenburns: That works, good spot!

willmcgugan (approved):

PR #584, opened 2022-06-17
[perf] Widget classes in `textual.widget.__init__.py` are now lazy-loaded

### LINKED ISSUE #578: Lazy load Widgets
We have a `widgets` module that imports all the widgets. THis allows for easy imports like:

```python
from textual.widgets import DataTable, Button
```

The downside is that all widgets are imported even if you only need one, increasing startup time.

I think we should be able to do this lazily some how, possibly by implementing a module level `__getattr__`.

DrBenton:
Thanks to the `textual.widget.__init__.pyi` file we keep the benefits of having the assistance of code editors and type checkers despite the lazy-loading:
Screenshot from 2022-06-17 10-57-09

fixes #578

### DIFF
--- a/src/textual/widgets/__init__.py
+++ b/src/textual/widgets/__init__.py
@@ -1,21 +1,45 @@
-from ._footer import Footer
-from ._header import Header
-from ._button import Button
-from ._placeholder import Placeholder
-from ._static import Static
-from ._tree_control import TreeControl, TreeClick, TreeNode, NodeID
-from ._directory_tree import DirectoryTree, FileClick
+from __future__ import annotations
+from importlib import import_module
+import typing
 
-__all__ = [
-    "Button",
-    "DirectoryTree",
-    "FileClick",
-    "Footer",
-    "Header",
-    "Placeholder",
-    "Static",
-    "TreeClick",
-    "TreeControl",
-    "TreeNode",
-    "NodeID",
-]
+if typing.TYPE_CHECKING:
+    from ..widget import Widget
+
+# Let's decrease startup time by lazy loading our Widgets:
+# ⚠ For any new built-in Widget we create, not only we have to add them to the following dictionary, but also to the
+# `__init__.pyi` file in this same folder - otherwise text editors and type checkers won't be able to "see" them.
+_WIDGETS_LAZY_LOADING_MAPPING: dict[str, str] = {
+    "Footer": "._footer:Footer",
+    "Header": "._header:Header",
+    "Button": "._button:Button",
+    "Placeholder": "._placeholder:Placeholder",
+    "Static": "._static:Static",
+    "TreeControl": "._tree_control:TreeControl",
+    "TreeClick": "._tree_control:TreeClick",
+    "TreeNode": "._tree_control:TreeNode",
+    "NodeID": "._tree_control:NodeID",
+    "DirectoryTree": "._directory_tree:DirectoryTree",
+    "FileClick": "._directory_tree:FileClick",
+}
+
+_WIDGETS_LAZY_LOADING_CACHE: dict[str, type[Widget]] = {}
+
+
+def __getattr__(widget_name: str) -> type[Widget]:
+    try:
+        return _WIDGETS_LAZY_LOADING_CACHE[widget_name]
+    except KeyError:
+        pass
+
+    if widget_name not in _WIDGETS_LAZY_LOADING_MAPPING:
+        raise ImportError(f"Widget {widget_name} not found")
+
+    module_name, class_name = _WIDGETS_LAZY_LOADING_MAPPING[widget_name].split(":")
+    module = import_module(module_name, package="textual.widgets")
+    class_ = getattr(module, class_name)
+
+    _WIDGETS_LAZY_LOADING_CACHE[widget_name] = class_
+    return class_
+
+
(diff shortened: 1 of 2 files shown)


src/textual/widgets/__init__.pyi
```diff
@@ -0,0 +1,12 @@
+from ._footer import Footer as Footer
```
DrBenton, line 1: According to [PEP 484](https://peps.python.org/pep-0484/#stub-files):
> Modules and variables imported into the stub are not considered exported from the stub unless the import uses the `import ... as ...` form or the equivalent `from ... import ... as ...` form.
> To clarify, the intention here is that only names imported using the form `X as X` will be exported, i.e. the name before and after as must be the same.

src/textual/widgets/__init__.py
```diff
@@ -1,21 +1,45 @@
...
+import typing
+
+if typing.TYPE_CHECKING:
+    from ..widget import Widget
+
+# Let's decrease startup time by lazy loading our Widgets:
+# ⚠ For any new built-in Widget we create, not only we have to add them to the following dictionary, but also to the
+# `__init__.pyi` file in this same folder - otherwise text editors and type checkers won't be able to "see" them.
+_WIDGETS_LAZY_LOADING_MAPPING: dict[str, str] = {
+    "Footer": "._footer:Footer",
```
DrBenton, line 12: I opted for a simple-and-stupid approach; combined to the fact that the lazy-loaded Widget classes are kept in a (simple) cache this implementation should do the job? :slightly_smiling_face: :crossed_fingers:

willmcgugan: I think we can eliminate this dict because the imports can al be deduced from the name. i.e. if you know the name `Button`, you can deduce the import is `_button.Button`

I'd suggest using `__all__` to keep a list of the widgets, and deducing the imports from there.

src/textual/widgets/__init__.pyi
willmcgugan: Love the typing magic here. It's a little burden to maintain, but worth it.

Suggest putting a comment in `__init__.py` to remind people to update the `.pyi` when adding a new widget.

willmcgugan (changes requested):
Clever! We still have all the benefits of typing and auto-complete.

src/textual/widgets/__init__.py
DrBenton: My first implementation was using such guesses indeed, rather than having to maintain that mapping; however it doesn't work with the various types Widgets can also carry alongside their main class?
i.e. it's easy to infer that `TreeControl` is the result of `from ._tree_control import TreeControl`, but without such a mapping how do we determine that `NodeID`is the result of `from ._tree_control import NodeID`, since there is no links between the string _"tree_control"_ and the string _"NodeID"_ ?

src/textual/widgets/__init__.pyi
DrBenton: I agree, and that's what I was trying to do with this comment there: :slightly_smiling_face: 
Screenshot from 2022-06-17 12-14-59
Are you thinking of another type of comment?

willmcgugan: That’s cool. Missed that!

src/textual/widgets/__init__.py
willmcgugan: True. But we can adopt this convention going forward. The TreeControl hasn’t been updated from the old API, and we will probably not expose symbols other than the Widgets themselves.

DrBenton: Alright, let's do that! And on top of the automatic inference, which simplifier the maintenance of our built-in Widgets, I guess we can also fall back to such a manual mapping if we need to, for Widgets that would require "companion classes"? :slightly_smiling_face:

willmcgugan: I'm thinking that we can probably mostly avoid companion classes. Small objects can go in the class itself. For instance `TreeControl.NodeID`. Will see how that works out.

DrBenton:
@willmcgugan hard-coded `__all__` list is back, as well as the automatic inference of import paths from Widget class names :slightly_smiling_face: 
> https://github.com/Textualize/textual/pull/584/files

src/textual/widgets/__init__.py
```diff
@@ -1,21 +1,41 @@
...
+from importlib import import_module
+import typing
 
+from ..case import camel_to_snake
+
+if typing.TYPE_CHECKING:
+    from ..widget import Widget
+
+
+_WIDGETS_LAZY_LOADING_CACHE: dict[str, type[Widget]] = {}
```
willmcgugan, line 11: We might not need this. I suspect the import machinery will cache for us.

Could you move `__all__` to the top of the file, just to make it clear what we need to update.

willmcgugan (changes requested):

src/textual/widgets/__init__.py
DrBenton: After having checked, it seems that each module importing a given widget will activate the `__get__` method - after that it's kept in the caller module's cache, but if 3 modules do `from textual.widgets import Placeholder` it will still be called 3 times.
So we'd better keep that cache. I moved the `__all__` to the top of the file, though  :slightly_smiling_face: [0bdd2be](https://github.com/Textualize/textual/pull/584/commits/0bdd2be141c6227af26bdb4d30bdce193ae4623b)

Issue #581, opened 2022-06-15
Default versus User CSS


willmcgugan:
We have CSS specified within a widget (default CSS) and CSS specified in an external file. They are currently considered the same as far as the Textual CSS parser is concerned.

We should probably treat default CSS as always having lower specificity than user specificity. This should make it easier to apply changes to builtin widgets, and is analogous to the browsers default CSS.

We are also probably processing rules in the wrong order. When rules have equal specificity the last rule in the file should take precedence.

PR #587, opened 2022-06-24
[tests] Restore layout integration tests

DrBenton:
These were broken-and-disabled in a previous PR because of 2 things:
 - The Widgets' default vertical scrollbar size is now 2 cells: 
    Screenshot from 2022-06-24 11-59-52
    This has to be reflected in these tests, in order to benefit from their "regressions spotting" role :slightly_smiling_face: 
 - As long as #581 is not solved, the Widget's default CSS overrides the custom Widgets' own CSS.
   --> The test now temporarily uses a more specific CSS selector in order to bypass this bug.

### DIFF
--- a/tests/test_integration_layout.py
+++ b/tests/test_integration_layout.py
@@ -19,13 +19,12 @@ SCREEN_H = 8  # height of our Screens
 SCREEN_SIZE = Size(SCREEN_W, SCREEN_H)
 PLACEHOLDERS_DEFAULT_H = 3  # the default height for our Placeholder widgets
 
-# TODO: Brittle test
-# These tests are currently way to brittle due to the CSS layout not being final
-# They are also very hard to follow, if they break its not clear *what* went wrong
-# Going to leave them marked as "skip" for now.
+# As per Widget's CSS property, by default Widgets have a horizontal scrollbar of size 1
+# and a vertical scrollbar of size 2:
+SCROLL_H_SIZE = 1
+SCROLL_V_SIZE = 2
 
 
-@pytest.mark.skip("brittle")
 @pytest.mark.asyncio
 @pytest.mark.integration_test  # this is a slow test, we may want to skip them in some contexts
 @pytest.mark.parametrize(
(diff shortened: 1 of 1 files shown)


tests/test_integration_layout.py
```diff
@@ -70,7 +69,7 @@
             # #root's virtual height should be as high as its stacked content
             (SCREEN_W - 2 - 1, PLACEHOLDERS_DEFAULT_H * 4),
             # placeholders width=same than screen, minus 2 borders, minus scrollbar :: height=default height minus 2 borders
-            (SCREEN_W - 2 - 1, PLACEHOLDERS_DEFAULT_H),
+            (SCREEN_W - 2 - SCROLL_V_SIZE, PLACEHOLDERS_DEFAULT_H),
```
DrBenton, line 72: this test was assuming that scrollbar sizes was always `1` by default. Now that it's no longer the case we'd better use constants to store the default scrollbar sizes of our Widgets :slightly_smiling_face:

tests/test_integration_layout.py
```diff
@@ -19,13 +19,12 @@
 SCREEN_SIZE = Size(SCREEN_W, SCREEN_H)
 PLACEHOLDERS_DEFAULT_H = 3  # the default height for our Placeholder widgets
 
-# TODO: Brittle test
```
DrBenton, line 22: Not really brittle :smile: : these integration tests have one job, which is to spot regressions. 
They just did their duty when a PR recently brought a breaking change (new scrollbar default size) :slightly_smiling_face:

tests/test_integration_layout.py
```diff
@@ -285,22 +297,26 @@ def on_mount(self):
...
 
+    container_style_rules = (
+        [container_style] if isinstance(container_style, str) else container_style
+    )
+
     class LargeWidgetContainer(Widget):
+        # TODO: Once textual#581 ("Default versus User CSS") is solved the following CSS should just use the
+        #  "LargeWidgetContainer" selector, without having to use a more specific one to be able to override Widget's CSS:
         CSS = """
-        LargeWidgetContainer {
```
DrBenton, line 290: It seems that Widget's `CSS` declarations take precedence over children classes' ones - i.e. setting a custom `scrollbar-size-vertical` here is always overridden by the Widget class' own one. Which is why this test broke when the breaking change was introduced.
A temporary workaround (preceded by with a `TODO` comment so we don't forget to remove it later on) is to use a more specific selector in this test :slightly_smiling_face:

tests/test_integration_layout.py
```diff
@@ -246,32 +247,43 @@ def compose(self) -> ComposeResult:
...
             ("overflow: auto", "scrollbar-size: 3 3"),
-            Size(17, 20),
+            Size(20 - 3, 20),
         ],
         # scrollbar needed only horizontally, custom scrollbar size
         [
             Size(30, 20),
             ("overflow: auto", "scrollbar-size: 3 3"),
-            Size(20, 17),
+            Size(20, 20 - 3),
```
DrBenton, line 286: While I was there... `20 - 3` might be a bit clearer regarding what this "expected size" number is: it's the size of the container (20, as explained at the beginning of the test' parameters list), minus the custom scrollbar size we define on the line above :slightly_smiling_face:

PR #588, opened 2022-06-24
[button] Add a brief "active" effect when a button a clicked

### LINKED ISSUE #576: Activated button animations
The Button widget should have a visual indicator that it was clicked. Material buttons do this. The reasoning is that if an action doesn't immediately update the UI, the user tends to click again to make sure it works.

I'm thinking a brief pulse animation, or 500ms flash. Something subtly enough to not be irritating, but still convey the active state.

this may be achieved by adding an `-active` class to the Button to allow a dev to define the active animation.

DrBenton:
This effect is achieved by adding a `-active` class on the button instance when it's clicked, and removing it a few hundreds milliseconds later.
Our 4 default buttons types have default styles (background color is lightened) for this effect, with a transition to make it smoother.

closes #576 

#### How users can customise it:
 - :art: The visual impact of the `-active` class is pure CSS, and thus can simply be overridden with custom userland CSS
 - :watch: The duration for which the `-active` class is added can be customised on each subclass of Button, or even at the Button instance level, by overriding the `ACTIVE_EFFECT_DURATION` property

#### Demo time!

https://user-images.githubusercontent.com/722388/175553711-ff3eb4df-3125-400b-937b-e68338f101da.mp4

### DIFF
--- a/src/textual/widgets/_button.py
+++ b/src/textual/widgets/_button.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import sys
+from functools import partial
 from typing import cast
 
 if sys.version_info >= (3, 8):
@@ -41,12 +42,18 @@ class Button(Widget, can_focus=True):
         margin: 1 0;
         align: center middle;
         text-style: bold;
+
+        transition: background 0.1;/* for "active" effect */
     }
 
     Button:hover {
         background: $primary-darken-2;
         color: $text-primary-darken-2;
         border: tall $primary-lighten-1;
+    }
+
+    Button.-active {
+        background: $primary-lighten-1;
     }
 
     .-dark-mode Button {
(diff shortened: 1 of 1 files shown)


src/textual/widgets/_button.py
```diff
@@ -49,6 +52,10 @@ class Button(Widget, can_focus=True):
         border: tall $primary-lighten-1;
     }
 
+    Button.-active {
+        background: $primary-lighten-1;
```
DrBenton, line 56: the CSS part of the implementation... :slightly_smiling_face:

src/textual/widgets/_button.py
```diff
@@ -199,8 +237,15 @@ def render(self) -> RenderableType:
 
     async def on_click(self, event: events.Click) -> None:
         event.stop()
-        if not self.disabled:
-            await self.emit(Button.Pressed(self))
+        if self.disabled:
+            return
+        # Manage the "active" effect:
+        self.add_class("-active")
+        self.set_timer(
```
DrBenton, line 244: ...and its Python side :slightly_smiling_face:

willmcgugan:
Nice. I think I'm going to tweak the colors a bit. But LGTM

PR #591, opened 2022-06-28
Add a cursor to table

willmcgugan:


src/textual/_cells.py
```diff
@@ -0,0 +1,6 @@
+__all__ = ["cell_len"]
+
+try:
```
willmcgugan, line 3: This may be temporary until the next version of Rich.

src/textual/_compositor.py
```diff
@@ -98,37 +99,75 @@ class ChopsUpdate:
...
         self, console: Console, options: ConsoleOptions
     ) -> RenderResult:
         move_to = Control.move_to
         new_line = Segment.line()
         chops = self.chops
-        crop = self.crop
-        last_y = crop.y_max - 1
-        x1, x2 = crop.x_extents
-        for y in crop.y_range:
+        chop_ends = self.chop_ends
```
willmcgugan, line 124: This has gotten more complicated because it renders a partial update.

src/textual/widget.py
```diff
@@ -428,10 +427,33 @@ def scrollbar_size_horizontal(self) -> int:
             else 0
         )
 
-    def set_dirty(self) -> None:
-        """Set the Widget as 'dirty' (requiring re-render)."""
-        self._dirty_regions.clear()
-        self._dirty_regions.append(self.size.region)
+    def _set_dirty(self, *regions: Region) -> None:
```
willmcgugan, line 430: Private because the dev should always use `self.refresh`

darrenburns (approved):
One comment

src/textual/widgets/_data_table.py
```diff
@@ -179,15 +286,25 @@ def add_row(self, *cells: CellType, height: int = 1) -> None:
...
+        self._line_no += height
         self._update_dimensions()
         self.refresh()
 
+    def refresh_cell(self, row_index: int, column_index: int) -> None:
+        if row_index < 0 or column_index < 0:
+            return
+        region = self._get_cell_region(row_index, column_index)
+        if not self.window_region.overlaps(region):
+            return
```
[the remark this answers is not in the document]
willmcgugan: Both `window_region` and `region` at that point are in *virtual* coordinates (i.e. scrollable range). Subtracting the scroll position converts the region to *widget* coordinates (i.e. top left of Widget is (0,0)).

PR #595, opened 2022-06-28
User CSS should always take precedence over Widget CSS

darrenburns:
Adds an additional (greater) level of specificity (`Specificity5`), which means user-defined CSS (i.e. that inside a `.css` file) _always_ takes precedence over CSS defined inside the `CSS` class var.

### DIFF
--- a/src/textual/css/styles.py
+++ b/src/textual/css/styles.py
@@ -514,13 +514,15 @@ class Styles(StylesBase):
     def extract_rules(
         self,
         specificity: Specificity3,
-        is_widget_rule: bool = False,
+        is_default_rules: bool = False,
     ) -> list[tuple[str, Specificity5, Any]]:
         """Extract rules from Styles object, and apply !important css specificity as
         well as higher specificity of user CSS vs widget CSS.
 
         Args:
             specificity (Specificity3): A node specificity.
+            is_default_rules (bool): True if the rules we're extracting are
+                default (i.e. in Widget.CSS) rules. False if they're from user defined CSS.
 
         Returns:
             list[tuple[str, Specificity5, Any]]]: A list containing a tuple of <RULE NAME>, <SPECIFICITY> <RULE VALUE>.
@@ -531,7 +533,7 @@ class Styles(StylesBase):
             (
                 rule_name,
                 (
-                    0 if is_widget_rule else 1,
+                    0 if is_default_rules else 1,
                     1 if is_important(rule_name) else 0,
                     *specificity,
                 ),
--- a/src/textual/css/stylesheet.py
+++ b/src/textual/css/stylesheet.py
@@ -4,7 +4,7 @@ import os
 from collections import defaultdict
 from operator import itemgetter
 from pathlib import Path, PurePath
-from typing import cast, Iterable
+from typing import cast, Iterable, NamedTuple
 
 import rich.repr
 from rich.console import RenderableType, RenderResult, Console, ConsoleOptions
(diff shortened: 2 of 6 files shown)


src/textual/css/stylesheet.py
```diff
@@ -177,6 +179,11 @@ def _parse_rules(self, css: str, path: str | PurePath) -> list[RuleSet]:
             raise
         except Exception as error:
             raise StylesheetError(f"failed to parse css; {error}")
+
+        if path in self._widget_css_paths:
```
willmcgugan, line 183: What if `_parse_rules` could take a boolean indicating that it was Widget CSS, and pass that in to `parse` which would store it in the rule? I think that would negate this check and loop.

src/textual/css/stylesheet.py
```diff
@@ -199,13 +206,17 @@ def read(self, filename: str | PurePath) -> None:
         self.source[str(path)] = css
         self._require_parse = True
 
-    def add_source(self, css: str, path: str | PurePath | None = None) -> None:
+    def add_source(
+        self, css: str, path: str | PurePath | None = None, is_widget_css: bool = False
```
willmcgugan, line 210: An `App` can have CSS, which isn't technically a widget.

Suggest a name which doesn't refer to widgets / nodes. Something like `is_defaults`

tests/css/test_stylesheet.py
```diff
@@ -101,6 +104,27 @@ def test_stylesheet_apply_empty_rulesets():
     stylesheet.apply(node)
 
 
+def test_stylesheet_apply_user_css_over_widget_css():
```
willmcgugan, line 107: Could we have a `Widget.CSS` with a clearly more specific rule. Something like `App .a:hover` and see if that overrides properly.

Also a rule that has `!important` within Widget.CSS, just to illustrate that can be overriden.

willmcgugan (changes requested):
Great. Not a big change at all. A change to the data structure to consider.

src/textual/css/stylesheet.py
```diff
@@ -122,6 +122,8 @@ def __init__(self, *, variables: dict[str, str] | None = None) -> None:
         self._rules: list[RuleSet] = []
         self.variables = variables or {}
         self.source: dict[str, str] = {}
+        # Records which of the source keys represent CSS defined at the widget level
+        self._widget_css_paths: set[str] = set()
```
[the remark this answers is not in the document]
darrenburns: I was actually on the fence about doing something like this but opted against it as it made the lookups into source a bit uglier and there were a good few of them. It's a trade-off though - happy to make the change though if you prefer it.

src/textual/css/stylesheet.py
darrenburns: Good point

src/textual/css/styles.py
```diff
@@ -511,20 +512,27 @@ def merge_rules(self, rules: RulesMap) -> None:
...
-            list[tuple[str, Specificity4, Any]]]: A list containing a tuple of <RULE NAME>, <SPECIFICITY> <RULE VALUE>.
+            list[tuple[str, Specificity5, Any]]]: A list containing a tuple of <RULE NAME>, <SPECIFICITY> <RULE VALUE>.
         """
         is_important = self.important.__contains__
 
         rules = [
-            (rule_name, (int(is_important(rule_name)), *specificity), rule_value)
+            (
+                rule_name,
+                (int(not is_widget_rule), int(is_important(rule_name)), *specificity),
```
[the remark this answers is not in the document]
darrenburns: Done :)

src/textual/css/stylesheet.py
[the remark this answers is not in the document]
willmcgugan: Let’s give it a shot 👍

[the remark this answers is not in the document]
darrenburns: No objections to a NamedTuple instead?

src/textual/css/stylesheet.py
darrenburns: Not sure about is_defaults though, sounds easy to conflate with the CSS default values from the style property descriptors? Wonder if we could have a different name for "the browsers CSS".

src/textual/css/stylesheet.py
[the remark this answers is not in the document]
willmcgugan: I try not to use too many namedtuples and dataclasses, because they do add a slight penalty to startup. But if you find it makes the code more readable, go for it...

src/textual/css/stylesheet.py
willmcgugan: "is_default_rules" maybe? We can always change it later if we come up with a better name.

willmcgugan (approved):
LGTM

PR #596, opened 2022-06-29
[colors] Add a "auto" color

olivierphi:
When text has `color: auto [percentage]` it will be displayed with the contrasting colour of its background

As a result we can have widgets that have `color: auto;` in their CSS, and see their text colour automatically adapted to their background:
Screenshot from 2022-06-29 14-33-56

### DIFF
--- a/sandbox/color_names.py
+++ b/sandbox/color_names.py
@@ -7,7 +7,6 @@ from rich.pretty import Pretty
 from textual._color_constants import COLOR_NAME_TO_RGB
 from textual.app import App, ComposeResult
 from textual.widget import Widget
-from textual.widgets import Placeholder
 
 
 @rich.repr.auto(angular=False)
@@ -35,8 +34,7 @@ class ColorNames(App):
     def compose(self) -> ComposeResult:
         for color_name, color in COLOR_NAME_TO_RGB.items():
             color_placeholder = ColorDisplay(name=color_name)
-            is_dark_color = sum(color) < 400
-            color_placeholder.styles.color = "white" if is_dark_color else "black"
+            color_placeholder.styles.color = "auto 90%"
             color_placeholder.styles.background = color_name
             yield color_placeholder
 
--- a/src/textual/color.py
+++ b/src/textual/color.py
@@ -63,8 +63,7 @@ RE_COLOR = re.compile(
 rgb{OPEN_BRACE}({DECIMAL}{COMMA}{DECIMAL}{COMMA}{DECIMAL}){CLOSE_BRACE}$|
 rgba{OPEN_BRACE}({DECIMAL}{COMMA}{DECIMAL}{COMMA}{DECIMAL}{COMMA}{DECIMAL}){CLOSE_BRACE}$|
 hsl{OPEN_BRACE}({DECIMAL}{COMMA}{PERCENT}{COMMA}{PERCENT}){CLOSE_BRACE}$|
-hsla{OPEN_BRACE}({DECIMAL}{COMMA}{PERCENT}{COMMA}{PERCENT}{COMMA}{DECIMAL}){CLOSE_BRACE}$|
-(auto)(?:\s+({DECIMAL})%)?$
+hsla{OPEN_BRACE}({DECIMAL}{COMMA}{PERCENT}{COMMA}{PERCENT}{COMMA}{DECIMAL}){CLOSE_BRACE}$
 """,
     re.VERBOSE,
 )
@@ -145,11 +144,6 @@ class Color(NamedTuple):
     def is_transparent(self) -> bool:
         """Check if the color is transparent, i.e. has 0 alpha."""
         return self.a == 0
-
-    @property
-    def is_auto(self) -> bool:
-        """Check if the color is "auto"."""
-        return self.r == AUTO.r and self.g == AUTO.g and self.b == AUTO.b
 
     @property
     def clamped(self) -> Color:
@@ -308,8 +302,6 @@ class Color(NamedTuple):
             rgba,
             hsl,
             hsla,
-            auto,
-            auto_percentage,
         ) = color_match.groups()
 
         if rgb_hex_triple is not None:
@@ -355,13 +347,6 @@ class Color(NamedTuple):
             l = percentage_string_to_float(l)
             a = clamp(float(a), 0.0, 1.0)
             color = Color.from_hls(h, l, s).with_alpha(a)
-        elif auto == "auto":
-            auto_percentage_float = (
-                clamp(float(auto_percentage) / 100.0, 0.0, 1.0)
-                if auto_percentage
-                else 1.0
-            )
-            color = Color(AUTO.r, AUTO.g, AUTO.b, auto_percentage_float)
         else:
             raise AssertionError("Can't get here if RE_COLOR matches")
         return color
@@ -414,11 +399,6 @@ class Color(NamedTuple):
 WHITE = Color(255, 255, 255)
 BLACK = Color(0, 0, 0)
 TRANSPARENT = Color(0, 0, 0, 0)
-# We cannot express "auto" using Color primitives, so for this one we'll use some sentinel values for each R,G,B value.
-# (because "auto" is resolved at runtime, depending on the background color of the Node and its parents)
-# A color is "auto" if its R, G and B values are exactly equal to the following ones:
-# (we could have "userland colors" that accidentally have these exact same values, but it's probably unlikely enough?)
-AUTO = Color(0.123456789, 1.234567890, 2.345678901)
 
 
 class ColorPair(NamedTuple):
--- a/src/textual/color.py
+++ b/src/textual/color.py
@@ -63,7 +63,8 @@ RE_COLOR = re.compile(
 rgb{OPEN_BRACE}({DECIMAL}{COMMA}{DECIMAL}{COMMA}{DECIMAL}){CLOSE_BRACE}$|
 rgba{OPEN_BRACE}({DECIMAL}{COMMA}{DECIMAL}{COMMA}{DECIMAL}{COMMA}{DECIMAL}){CLOSE_BRACE}$|
 hsl{OPEN_BRACE}({DECIMAL}{COMMA}{PERCENT}{COMMA}{PERCENT}){CLOSE_BRACE}$|
-hsla{OPEN_BRACE}({DECIMAL}{COMMA}{PERCENT}{COMMA}{PERCENT}{COMMA}{DECIMAL}){CLOSE_BRACE}$
+hsla{OPEN_BRACE}({DECIMAL}{COMMA}{PERCENT}{COMMA}{PERCENT}{COMMA}{DECIMAL}){CLOSE_BRACE}$|
+(auto)(?:\s+({DECIMAL})%)?$
 """,
     re.VERBOSE,
 )
@@ -144,6 +145,11 @@ class Color(NamedTuple):
     def is_transparent(self) -> bool:
         """Check if the color is transparent, i.e. has 0 alpha."""
         return self.a == 0
+
+    @property
+    def is_auto(self) -> bool:
+        """Check if the color is "auto"."""
+        return self.r == AUTO.r and self.g == AUTO.g and self.b == AUTO.b
 
     @property
     def clamped(self) -> Color:
@@ -302,6 +308,8 @@ class Color(NamedTuple):
             rgba,
             hsl,
             hsla,
+            auto,
+            auto_percentage,
         ) = color_match.groups()
 
         if rgb_hex_triple is not None:
@@ -347,6 +355,13 @@ class Color(NamedTuple):
             l = percentage_string_to_float(l)
             a = clamp(float(a), 0.0, 1.0)
             color = Color.from_hls(h, l, s).with_alpha(a)
+        elif auto == "auto":
+            auto_percentage_float = (
+                clamp(float(auto_percentage) / 100.0, 0.0, 1.0)
+                if auto_percentage
+                else 1.0
+            )
+            color = Color(AUTO.r, AUTO.g, AUTO.b, auto_percentage_float)
         else:
             raise AssertionError("Can't get here if RE_COLOR matches")
         return color
@@ -399,6 +414,11 @@ class Color(NamedTuple):
 WHITE = Color(255, 255, 255)
 BLACK = Color(0, 0, 0)
 TRANSPARENT = Color(0, 0, 0, 0)
+# We cannot express "auto" using Color primitives, so for this one we'll use some sentinel values for each R,G,B value.
+# (because "auto" is resolved at runtime, depending on the background color of the Node and its parents)
+# A color is "auto" if its R, G and B values are exactly equal to the following ones:
+# (we could have "userland colors" that accidentally have these exact same values, but it's probably unlikely enough?)
+AUTO = Color(0.123456789, 1.234567890, 2.345678901)
 
 
 class ColorPair(NamedTuple):
--- a/src/textual/css/_color.py
+++ b/src/textual/css/_color.py
@@ -3,55 +3,75 @@ Helper functions to work with Textual Colors.
 """
 
 from __future__ import annotations
-import re
 
 from textual.color import Color
-from textual.css.scalar import Scalar
-from textual.css.tokenize import DECIMAL
+from textual.css._help_text import color_property_help_text
+from textual.css.errors import DeclarationError, StyleValueError
+from textual.css.scalar import Scalar, Unit
+from textual.css.tokenizer import Token
 
 
-RE_ALPHA = (
-    # Color expressions themselves can contain percentage values (e.g. `hsl( 180, 50%, 50% )`), so we'll only
-    # consider a percentage expression to be an alpha value if it's at the start or the end of the color expression.
-    # (i.e. `color: red 50%` or `color: 50% red`)
-    re.compile(rf"^({DECIMAL})%"),  # starting with a percentage
-    re.compile(rf"({DECIMAL})%$"),  # ending with a percentage
-)
-
-
-def parse_color_expression(color_text: str) -> tuple[Color, bool]:
+def parse_color_tokens(property_name: str, tokens: list[Token]) -> tuple[Color, bool]:
     """
-    Parses a color expression, such as "red", "#FFCC00", "rgb(1,255,50)", "lime 50%", "95% auto"...
+    Parses the tokens of a color expression, such as "red", "#FFCC00", "rgb(1,255,50)", "lime 50%", "95% auto"...
 
     Args:
-        color_text: a color expression supported by ``Color.parse``, optionally prefixed or suffixed with a
-            percentage value token.
+        property_name (str): The name of the CSS property. Only used to raise DeclarationErrors that include this name.
+        tokens (list[Token]): CSS tokens describing a color - i.e. an expression supported by ``Color.parse``,
+            optionally prefixed or suffixed with a percentage value token.
 
     Returns:
         tuple[Color, bool]: a Color instance, as well as a boolean telling us if it's a pseudo-color "auto" one.
+
+    Raises:
+        DeclarationError: if we come across an unexpected value
     """
+    property_name = property_name.replace("-", "_")
+
     alpha = 1.0
+    color = None
+    is_auto = False
 
-    color_text = color_text.strip()
-
-    for alpha_pattern in RE_ALPHA:
-        alpha_match = alpha_pattern.search(color_text)
-        if alpha_match:
-            value = alpha_match.groups()[0]
-            alpha_scalar = Scalar.parse(value)
+    for token in tokens:
+        if token.value == "auto":
+            is_auto = True
+        elif token.name == "scalar":
+            alpha_scalar = Scalar.parse(token.value)
+            if alpha_scalar.unit != Unit.PERCENT:
+                raise DeclarationError(
+                    property_name, token, "alpha must be given as a percentage."
+                )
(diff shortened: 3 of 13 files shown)


src/textual/color.py
```diff
@@ -145,6 +146,11 @@ def is_transparent(self) -> bool:
         """Check if the color is transparent, i.e. has 0 alpha."""
         return self.a == 0
 
+    @property
+    def is_auto(self) -> bool:
+        """Check if the color is "auto"."""
+        return self.r == AUTO.r and self.g == AUTO.g and self.b == AUTO.b
```
DrBenton, line 152: Would probably be more elegant with `(self.r, self.g, self.b) == (AUTO.r, AUTO.g, AUTO.b)`, but since we're aiming at optimising for performance in most cases I guess creating a tuple takes a bit more time than doing this triple equality condition? :slightly_smiling_face:

sandbox/color_names.py
```diff
@@ -35,8 +34,7 @@ def on_mount(self):
     def compose(self) -> ComposeResult:
         for color_name, color in COLOR_NAME_TO_RGB.items():
             color_placeholder = ColorDisplay(name=color_name)
-            is_dark_color = sum(color) < 400
-            color_placeholder.styles.color = "white" if is_dark_color else "black"
+            color_placeholder.styles.color = "auto 90%"
```
DrBenton, line 37: This "is_dark_color" test is no longer required, since the new `auto` property does exactly this :blush: 

The result:
Screenshot from 2022-06-29 14-33-56

src/textual/color.py
```diff
@@ -63,7 +63,8 @@ class Lab(NamedTuple):
 rgb{OPEN_BRACE}({DECIMAL}{COMMA}{DECIMAL}{COMMA}{DECIMAL}){CLOSE_BRACE}$|
 rgba{OPEN_BRACE}({DECIMAL}{COMMA}{DECIMAL}{COMMA}{DECIMAL}{COMMA}{DECIMAL}){CLOSE_BRACE}$|
 hsl{OPEN_BRACE}({DECIMAL}{COMMA}{PERCENT}{COMMA}{PERCENT}){CLOSE_BRACE}$|
-hsla{OPEN_BRACE}({DECIMAL}{COMMA}{PERCENT}{COMMA}{PERCENT}{COMMA}{DECIMAL}){CLOSE_BRACE}$
+hsla{OPEN_BRACE}({DECIMAL}{COMMA}{PERCENT}{COMMA}{PERCENT}{COMMA}{DECIMAL}){CLOSE_BRACE}$|
+(auto)(?:\s+({DECIMAL})%)?$
```
willmcgugan, line 67: I'm not sure we need the decimal percentage here. We can already apply alpha with a color and a percentage.

We also don't want to enforce an order. i.e. "50% red"  should be the same as "red 50%"

src/textual/color.py
willmcgugan: `self[:3] == AUTO[:3]` is probably fastest. But I'm afraid I'm going to veto this sentinel color object.

tests/test_integration_color.py
```diff
@@ -0,0 +1,77 @@
+import pytest
```
willmcgugan, line 1: These tests run a lot of code that is unrelated to the PR.

I'd like to focus on unit tests, for instance:

- Test a color can be marked as auto
- Test Color.parse("auto") produces an auto color
- Test setting background and color on a Styles object, produces the expected `Styles.rich_style`
- Test parsing the CSS produces the expected rules

None of the above require a running app or async context. They are less likely to break from an unrelated API change.

darrenburns (approved):
Looks good to me :)

willmcgugan (changes requested):
I think we need an alternative way of specifying the "auto" foreground color.

It may not be necessary to modify Color at all. When parsing, we could set a bool on the Styles object such as "auto_color". I think this would capture the requirement that only `color` can be auto.

Please try to minimize the use of integration tests right now. I would like to aim to have as much coverage as possible with unit tests.

tests/test_integration_color.py
DrBenton: With only unit tests how do we check that the "auto" takes into account the colours and opacity levels of all the Node's ancestors?

willmcgugan: That is a good point. I think you can create a hierarchy without running the app. Create the App, call `app.mount`.

willmcgugan: Yeah, keep this integration test, since it's done. But let's add some unit test for what we can easily test. We may have to solve a few such issues a `active_app` to make unit tests more practical in the future. 

Integration tests are more likely to break with API changes. And the breakage is unlikely to tell me anything I didn't know already.

Integration tests will come in to their own when the APIs are more stable. They can catch unintended API changes rather than intentional API changes.

tests/css/test_parse.py
```diff
@@ -907,18 +907,23 @@ def test_background(self):
...
+            ("rgb(1,255,50)", Color(1, 255, 50)),
+            ("rgb( 1, 255,50 )", Color(1, 255, 50)),
+            ("rgba( 1, 255,50,0.3 )", Color(1, 255, 50, 0.3)),
+            ("rgba( 1, 255,50, 1.3 )", Color(1, 255, 50, 1.0)),
+            ("hsl( 180, 50%, 50% )", Color(64, 191, 191)),
+            ("hsl(180,50%,50%)", Color(64, 191, 191)),
+            ("hsla(180,50%,50%,0.25)", Color(64, 191, 191, 0.25)),
+            ("hsla( 180, 50% ,50%,0.25 )", Color(64, 191, 191, 0.25)),
+            ("hsla( 180, 50% , 50% , 1.5 )", Color(64, 191, 191)),
+            ("red 50%", Color(255, 0, 0, 0.5)),
```
DrBenton, line 922: I just added these 2 cases, and Black did the rest :smile:

tests/test_integration_color.py
```diff
@@ -0,0 +1,113 @@
...
+        # "auto" on a white background --> foreground is black
+        ["auto", "white", BLACK],
+        # "auto" on a black background --> foreground is white
+        ["auto", "black", WHITE],
+        # And now with a pinch of alpha:
+        ["auto 50%", "white", WHITE.get_contrast_text(0.5)],
+        ["30% auto", "black", BLACK.get_contrast_text(0.3)],
+    ),
+)
+def test_color_auto_single_widget(
```
DrBenton, line 22: here is a "closer to the metal" test for the "auto" colour, that can only test that a Widget adapts its "auto" colour to its own background :-)

willmcgugan:
More complicated maybe, but I think this a better abstraction. If you consider that an object should have one reason to exist, a Color class should only ever abstract a color (and "auto" is not a color).  "auto" to my mind is more of a way of enabling functionality on the Style object.

src/textual/css/_color.py
```diff
@@ -0,0 +1,57 @@
+"""
```
willmcgugan, line 1: We're doing some custom parsing here, which is different from the parsing done in the CSS. I'd like to avoid solving the same problem twice.

We could use `tokenize_value` in textual.css.tokenize for this -- which is what the CSS parser uses.

Here's an example:

```python
>>> from textual.css.tokenize import tokenize_value
>>> list(tokenize_value("hsl( 180, 50%, 50% ) 30%", "expression"))
[Token(name='color', value='hsl( 180, 50%, 50% )', path='expression', code='hsl( 180, 50%, 50% ) 30%', location=(0, 0)), Token(name='whitespace', value=' ', path='expression', code='hsl( 180, 50%, 50% ) 30%', location=(0, 20)), Token(name='scalar', value='30%', path='expression', code='hsl( 180, 50%, 50% ) 30%', location=(0, 21))]
```

This should give you the same tokens passed to the builder object.

How about the property uses `tokenize` and calls a method to extract a Color from the tokens. The builder could call this same method.

src/textual/css/_style_properties.py
```diff
@@ -794,6 +795,37 @@ def __set__(self, obj: StylesBase, names: str | tuple[str] | None = None):
...
+
+        Args:
+            obj (Styles): The ``Styles`` object
+            value (bool | None): The value to set.
+        """
+        _rich_traceback_omit = True
+        if value is None:
+            obj.clear_rule("color_auto")
+        else:
+            obj.set_rule("color_auto", value)
```
willmcgugan, line 826: I think "auto_color" reads a little better.

src/textual/css/_styles_builder.py
```diff
@@ -570,31 +571,26 @@ def process_color(self, name: str, tokens: list[Token]) -> None:
...
-                    color = Color.parse(token.value)
-                except Exception as error:
-                    self.error(
-                        name,
-                        token,
-                        color_property_help_text(name, context="css", error=error),
-                    )
-            else:
-                self.error(name, token, color_property_help_text(name, context="css"))
+        expression = _join_tokens(tokens)
```
willmcgugan, line 576: We're parsing the color, combining the tokens, and re-parsing the string again.  See my suggestion re tokenize above. I think we can simplify by extracting the color from a list of tokens.

src/textual/css/styles.py
```diff
@@ -182,6 +183,9 @@ class StylesBase(ABC):
     layout = LayoutProperty()
 
     color = ColorProperty(Color(255, 255, 255))
+    # N.B. This one is a pseudo-property, in the sense that we cannot set it via CSS via a dedicated property:
```
willmcgugan, line 186: We can also set it via code can't we? `self.styles.color_auto = True` ?

src/textual/css/_color.py
```diff
@@ -0,0 +1,57 @@
...
+RE_ALPHA = (
+    # Color expressions themselves can contain percentage values (e.g. `hsl( 180, 50%, 50% )`), so we'll only
+    # consider a percentage expression to be an alpha value if it's at the start or the end of the color expression.
+    # (i.e. `color: red 50%` or `color: 50% red`)
+    re.compile(rf"^({DECIMAL})%"),  # starting with a percentage
+    re.compile(rf"({DECIMAL})%$"),  # ending with a percentage
+)
+
+
+def parse_color_expression(color_text: str) -> tuple[Color, bool]:
```
willmcgugan, line 22: As a pure function this is easy to test. Can we have some unit tests?

willmcgugan (changes requested):
I like the update, but I think we need one more iteration on parsing the colors.

src/textual/css/_color.py
DrBenton: My reasoning was that simply doing a `join` over the tokens in the StylesBuilder was much faster than tokenizing the values in the ColorProperty - and the 2 regexes used to find the percentage values likely have a fast lookup since they only search at the beginning and the end of a string.
But ok, I'll do that :slightly_smiling_face:

src/textual/css/_style_properties.py
DrBenton: it might read better, however my choice of `color_auto` was to collocate this property with its attached `color` as they are sister properties, so they are together when we debug for example:
Screenshot from 2022-07-01 09-15-30
It would also group the "auto" properties with their "main" one if we add more "_resolved at runtime 'auto' values_" like this one later one.

But no worries, I'll rename it :slightly_smiling_face:

willmcgugan: No, you're right. We should follow the established convention. Ok, lets stick to `color_auto`

src/textual/css/styles.py
DrBenton: We can yes, what my comment was meaning is that we cannot set it via a CSS string or file - as there is no `process_color_auto` method in the StylesBuilder. 
That's why to me it's a pseudo-property, as it can only be set via another CSS property, "color" :slightly_smiling_face:

willmcgugan: Gotcha

tests/test_color.py
```diff
@@ -115,22 +119,28 @@ def test_color_parse(text, expected):
...
     assert Color.parse(input) == output
 
 
-@pytest.mark.parametrize("input,output", [
-    ("rgb(300, 300, 300)", Color(255, 255, 255)),
-    ("rgba(300, 300, 300, 300)", Color(255, 255, 255, 1.0)),
-    ("hsl(400, 200%, 250%)", Color(255, 255, 255, 1.0)),
-    ("hsla(400, 200%, 250%, 1.9)", Color(255, 255, 255, 1.0)),
-])
+@pytest.mark.parametrize(
```
DrBenton, line 135: same here :black_circle:

tests/test_color.py
```diff
@@ -141,7 +151,8 @@ def test_color_parse_hsl_negative_degrees():
 
 def test_color_parse_hsla_negative_degrees():
     assert Color.parse("hsla(-45, 50%, 50%, 0.2)") == Color.parse(
-        "hsla(315, 50%, 50%, 0.2)")
+        "hsla(315, 50%, 50%, 0.2)"
```
DrBenton, line 154: ditto

DrBenton:
@willmcgugan Feedback (hopefully) addressed! :slightly_smiling_face:  [28edacf](https://github.com/Textualize/textual/pull/596/commits/28edacfa57e9f72cf66399be19ff39a60bc58bc2)

src/textual/css/_color.py
```diff
@@ -0,0 +1,77 @@
...
+        DeclarationError: if we come across an unexpected value
+    """
+    property_name = property_name.replace("-", "_")
+
+    alpha = 1.0
+    color = None
+    is_auto = False
+
+    for token in tokens:
+        if token.value == "auto":
```
willmcgugan, line 36: It's not enough to look at just the value. We also need to change "name" as well, which I think should be "token".  

Consider if we add quoted strings to the parser, we wouldn't want to interpret `"auto"` in the same way as `auto`

src/textual/css/styles.py
```diff
@@ -684,7 +688,11 @@ def append_declaration(name: str, value: str) -> None:
             assert self.layout is not None
             append_declaration("layout", self.layout.name)
 
-        if has_rule("color"):
+        if get_rule("color_auto", False):
+            append_declaration(
+                "color", f"auto{f' {self.color.a * 100}%' if self.color.a < 1 else ''}"
```
willmcgugan, line 693: Can we do this without nested f strings? I find that hard to parse.

tests/test_color.py
```diff
@@ -219,3 +230,24 @@ def test_rgb_lab_rgb_roundtrip():
...
+        ["red", Color(255, 0, 0), False],
+        ["rgba(255, 125, 33, 0.6)", Color(255, 125, 33, 0.6), False],
+        ["40% #405060", Color(64, 80, 96, 0.4), False],
+        ["hsl(45,25%,25%) 70%", Color(80, 72, 48, 0.7), False],
+        ["auto", Color(ANY, ANY, ANY, 1), True],
+        ["auto 25%", Color(ANY, ANY, ANY, 0.25), True],
+        ["32% auto", Color(ANY, ANY, ANY, 0.32), True],
+    ),
+)
+def test_parse_color_expression(
```
willmcgugan, line 247: Can we have one or two invalid color strings tests? i.e. the case where it fails.

src/textual/css/_style_properties.py
```diff
@@ -837,15 +870,18 @@ def __set__(self, obj: StylesBase, color: Color | str | None):
             if obj.set_rule(self.name, color):
                 obj.refresh()
         elif isinstance(color, str):
+            tokens = tokenize_value(color, "expression")
             try:
-                parsed_color = Color.parse(color)
+                parsed_color, is_auto = parse_color_tokens(self.name, tokens)
```
willmcgugan, line 875: This can raise `DeclarationError`s can't it?

This would be entirely unexpected in this context. It should raise `StyleValueError`s

src/textual/css/styles.py
willmcgugan, line 693:
```suggestion
                "color", f"auto {self.color.a * 100}%" if self.color.a < 1 else "auto"
```

willmcgugan (changes requested):
Consider this...

```python
self.styles.color = "red"
```

Fantastic, we have red text. I now do this:

```python
self.styles.color = "auto 90%"
```

Great, the text is now white-ish or black-ish. All good so far.

```python
self.styles.color_auto = False
````

Now what color is the text? I suspect it will be `Color(1,2,3, a=0.9)` which would be a surprise to the developer.

My gut feeling is that it should be red again.

I think we should solve this. Could you have a think about a solution and let's have a chat about it when you have a plan?

**Comment by darrenburns** [comment]
I have to say I'm really not a fan of the `color_auto` - it's so tightly coupled to `color` and having 2 properties interfering with each other like that is just confusing. If we have to guess what the behaviour is for something that is conceptually so simple, I think we're already going to the wrong route.

I liked the idea of a sentinel value but agree the previous approach broke the "color" abstraction.

IMO it'd be nice to have the color property value have a type `Color | Auto` (i.e. a union where `Auto` is some container or sentinel to mark the special case), rather than having a separate flag. The approach as it stands feels like "using a flag to simulate a union type".

**Comment by willmcgugan** [comment]
So `self.styles.color` could return a Color or an Auto?

What kind of interface would the Auto support? Can I write `self.styles.color.r` or would I have to do `is isinstance(self.styes.color)` beforehand?

**Comment by willmcgugan** [comment]
I did not mean to do that!

**Comment by willmcgugan** [comment]
I agree that `styles.color` should probably not be anything other than a Color. It would surprise me if that wasn't the case..

For the Python interface, I think we will need `color_auto` as a boolean, but also `color_auto_alpha` as a float to capture the alpha color. I think that would avoid the edge case I mention above.

@DrBenton I see where you are going with another CSS property to capture the auto color state, but find the declaration of "color: auto 90%" to be quite elegant. And another rule might confuse, especially if there is a "color" *and* a "color-auto".

**Comment by willmcgugan** [comment]
Just for the record. After a 3 way call I think we've agreed on the following:

- The `Styles.color` property *setter* can accept "auto" with an optional alpha, e.g. "auto 90%"
- The `Styles.color` property *getter* returns the calculated contrasting color.
- We add a `Styles.color_auto` boolean (*getter* only) that exposes if "auto" is in effect.

I think this alleviates any edge cases, and is reasonably easy to wrap ones head around.

Thumbs up if you're in favor...

**Comment by DrBenton** [comment]
@willmcgugan here you go! [4439ca1](https://github.com/Textualize/textual/pull/596/commits/4439ca1308926c8c5ac44e60f31ea66c385df4e4)

File: src/textual/css/_style_properties.py [hunk 18]
```diff
@@ -808,14 +810,47 @@ def __get__(
...
+        # We're dealing with a "color" property that has an "auto" value: let's traverse the Node ancestors to
+        # calculate what the resulting value should be.
+        from ..dom import DOMNode
+
+        styles_node = getattr(obj, "node", None)
+        if not isinstance(styles_node, DOMNode):
+            raise ValueError(
+                "Dynamic 'auto' color can only be used with an subclass of StylesBase that has a 'node' attribute"
+            )
+        # We just repeat what `DOMNode.color` does, in a slightly simplified way.
```
willmcgugan, line 836: Have you considered rolling this logic in to the `color` property itself?

src/textual/css/_style_properties.py
```diff
@@ -808,14 +810,47 @@ def __get__(
...
+
         Args:
             obj (Styles): The ``Styles`` object.
             objtype (type[Styles]): The ``Styles`` class.
 
         Returns:
             Color: The Color
         """
-        return cast(Color, obj.get_rule(self.name, self._default_color))
+        if self.name != "color" or not obj.color_auto:
```
willmcgugan, line 823: Rather than overload the functionality of this object based on its name. We could define a property specifically for the foreground color.

This would make it easier to keep the functionality distinct. 

I think this may already be an issue. For instance, if you do `styles.background = "auto 90%"` I'm guessing that would set the foreground to auto? Which would not be desirable.

src/textual/css/_style_properties.py
DrBenton: The thing is that `DOMNode.colors` returns 4 computed colours - foreground and background colours, and its parent's colours. Whereas the Style properties each manage their single property

src/textual/css/_style_properties.py
```diff
@@ -891,13 +860,74 @@ def __set__(self, obj: StylesBase, color: Color | str | None):
...
+        styles_node = getattr(obj, "node", None)
+        if not isinstance(styles_node, DOMNode):
+            raise ValueError(
+                "Dynamic 'auto' color can only be used with an subclass of StylesBase that has a 'node' attribute"
+            )
+        # We just repeat what `DOMNode.color` does, in a slightly simplified way.
+        # N.B. `DOMNode.color` will call this `__get__` method while it is itself traversing the DOMNode ancestors,
+        # so we loop over ancestors in a loop that is itself looping over ancestors.
+        # That's a deliberate implementation choice.
+        background = Color(0, 0, 0, 0)
```
DrBenton, line 913: this dynamic resolution is now only made for the "color" property

DrBenton:
@willmcgugan I added a new ForegroundColorProperty, and moved the dynamic resolution of "_color: auto_" there: [ea1a429](https://github.com/Textualize/textual/pull/596/commits/ea1a429c412dcfa2e014f9653a93b5f7d28eceb4)

Compared to the very first one, this implementation is:
 - 3-4 times more verbose (in terms of lines of code)
 - More complex
 - Slower (the runtime cost of the 1st implementation was just a one-line "if" added in the DomNode's `colors` getter)
 - And all this without any difference in the end result, or the public API that Textual users will see... So, well, I must say that I'm not a big fan of it myself :sweat_smile: 

But if it fits better your idea of how the design for that new feature should look like, I guess that's worth the cost :slightly_smiling_face:

PR #602, opened 2022-07-04
Render cache

willmcgugan:
A new `StylesCache` responsible for applying CSS styles.

The problem I was solving was that updating a small portion of the window (such as the data table) would require re-rendering the entire widget on order to apply any styles.

The `StylesCache` solves this by being able to apply the style a line at a time, without rendering the content again. It also has the welcome benefit of being faster as it doesn't require multiple calls to `console.render_lines`.

The dev should never have to interact with the styles cache directly. With the new API you can either implement `Widget.render` to return a renderable, or you can implement `Widget.render_line` to return a line. The former is easier, but the latter can be used to build more sophisticated widgets.

- Added "ascii" border type for a bit of retro appeal.
- Renamed some properties on Widget, `Widget.size` now returns the size of the content area (not including padding and border) which I suspect 99% of devs will expect.


src/textual/_compositor.py
```diff
@@ -386,33 +386,23 @@ def add_widget(
...
-                    )
-                else:
-                    map[widget] = MapGeometry(
-                        child_region + layout_offset,
-                        order,
-                        clip,
-                        child_region.size,
-                        container_size,
-                    )
+                map[widget] = MapGeometry(
```
willmcgugan, line 399: These changes mean the compositor doesn't need to treat scrollable widgets differently, which is satisfying.

src/textual/_segment_tools.py
```diff
@@ -58,3 +63,64 @@ def line_crop(segments: list[Segment], start: int, end: int, total: int):
...
+        segments = [
+            Segment(" " * pad_left, style),
+            *segments,
+        ]
+    elif pad_right:
+        segments = [
+            *segments,
+            Segment(" " * pad_right, style),
+        ]
+    return list(segments)
```
DrBenton, line 126: [trivial comment] I guess using _early returns_ would avoid having to create a new list object at the end just to make MyPy happy? :slightly_smiling_face: 

```suggestion
    if pad_left and pad_right:
        return [
            Segment(" " * pad_left, style),
            *segments,
            Segment(" " * pad_right, style),
        ]
    if pad_left:
        return [
            Segment(" " * pad_left, style),
            *segments,
        ]

    return [
        *segments,
        Segment(" " * pad_right, style),
    ]

```

DrBenton (approved):
The PR is too big, and my knowledge of the internals of Textual's rendering machinery too small, to be able to fully understand the code diff, but... Looks good to me! :smile:

src/textual/_styles_cache.py
```diff
@@ -0,0 +1,333 @@
...
+    ┃                      ┃ └─ B. border + padding +
+    ┃   Lorem ipsum dolor  ┃◀┐         border
+    ┃   sit amet,          ┃ │
+    ┃   consectetur        ┃ └─ C. border + padding +
+    ┃   adipiscing elit,   ┃     content + padding +
+    ┃   sed do eiusmod     ┃           border
+    ┃   tempor incididunt  ┃
+    ┃                      ┃
+    ┃                      ┃
+    ┗━━━━━━━━━━━━━━━━━━━━━━┛
```
darrenburns, line 46: Not following this diagram at all 😄

src/textual/_styles_cache.py
```diff
@@ -0,0 +1,333 @@
...
+            y (int): The y coordinate of the line (relative to widget screen offset).
+            size (Size): Size of the widget.
+            content_size (Size): Size of the content area.
+            padding (Spacing): Padding.
+            base_background (Color): Background color of widget beneath this line.
+            background (Color): Background color of widget.
+            render_content_line (RenderLineCallback): Callback to render a line of content.
+
+        Returns:
+            list[Segment]: _description_
```
darrenburns, line 197: Missing return description

src/textual/geometry.py
```diff
@@ -306,21 +311,21 @@ def __bool__(self) -> bool:
...
-            tuple[int, int]: Pair of x coordinates (row numbers).
+            tuple[int, int]: Pair of x coordinates (column numbers).
         """
         return (self.x, self.x + self.width)
 
     @property
-    def y_extents(self) -> tuple[int, int]:
-        """Get the starting and ending x coord.
+    def line_span(self) -> tuple[int, int]:
+        """Get the star and end line number (y coord).
```
darrenburns, line 326:
```suggestion
        """Get the start and end line number (y coord).
```

src/textual/_segment_tools.py
willmcgugan: Good point. Although it would still need one case if neither pad is set.

src/textual/_styles_cache.py
darrenburns: Yeah saw those comments and understood them about the 3 code paths. Just didn't really get what the diagram was showing - wasn't getting that the arrows were meant to be point at a line/horizontal slice through the content box. I get it now though.

willmcgugan: I'll add a bit more text to explain, for future explorers of the code!

PR #607, opened 2022-07-12
Register callbacks at message pump level, invoke them after refresh

darrenburns:


src/textual/message_pump.py
```diff
@@ -392,8 +393,16 @@ def post_message_from_child_no_wait(self, message: Message) -> bool:
...
-    async def on_callback(self, event: events.Callback) -> None:
-        await invoke(event.callback)
+    def on_callback(self, event: events.Callback) -> None:
+        self._callbacks.append(event.callback)
+
+    async def invoke_and_clear_callbacks(self) -> None:
+        """Invoke all callbacks that are waiting to be executed, and then clear them.
+        Callbacks will be invoked in the same order they were registered.
+        """
+        for callback in self._callbacks:
```
willmcgugan, line 403: Suspect it is entirely possible for this list to be modified during iteration. Suggest making a copy and then clearing it on self before invoking the callbacks.

willmcgugan (changes requested):
Yeah, that is pretty much it. Couple of requests.

src/textual/widget.py
```diff
@@ -1049,6 +1049,7 @@ def on_idle(self, event: events.Idle) -> None:
             self.screen.post_message_no_wait(messages.Layout(self))
         self._layout_required = False
         self._repaint_required = False
+        await self.invoke_and_clear_callbacks()
```
willmcgugan, line 1052: Sorry, it occurs to me that the screen won't be refreshed at this point. A message has been sent to `screen` but the actual update won't occur until screen idle.

I think the actual redraw occurs in `_on_update`

src/textual/message_pump.py
darrenburns: Good point

src/textual/events.py
```diff
@@ -210,7 +214,7 @@ def is_printable(self) -> bool:
         Returns:
             bool: True if the key is printable. False otherwise.
         """
-        return self.key not in KEY_VALUES
+        return self.key == Keys.Space or self.key not in KEY_VALUES
```
darrenburns, line 217: `Keys` was being used for non-printable characters only. Now that "space" is inside it, it broke this usage and so I've added this special case to resolve it.

src/textual/widgets/text_input.py
```diff
@@ -40,6 +40,8 @@ def on_key(self, event: events.Key) -> None:
         key = event.key
         if key == "escape":
             return
+        elif key == "space":
```
darrenburns, line 43: Otherwise when we press the space bar, it would literally write "space" into the text input.

src/textual/screen.py
```diff
@@ -120,14 +123,36 @@ def on_idle(self, event: events.Idle) -> None:
...
     def _on_update(self) -> None:
         """Called by the _update_timer."""
         # Render widgets together
         if self._dirty_widgets:
             self._compositor.update_widgets(self._dirty_widgets)
             self.app._display(self._compositor.render())
             self._dirty_widgets.clear()
+
         self.update_timer.pause()
+        self.post_message_no_wait(events.PostScreenUpdate(self))
```
darrenburns, line 138: Since we can't easily just make this method async without the function colouring problem arising, we make use of the queue to schedule execution of the async "invoke callbacks" functionality.

willmcgugan (changes requested):
On the right track but I'd like to look in to that freezing issue you saw.

src/textual/screen.py
darrenburns: Done, and renamed it to match this new behaviour.

src/textual/screen.py
```diff
@@ -44,6 +45,8 @@ def __init__(self, name: str | None = None, id: str | None = None) -> None:
         self._compositor = Compositor()
         self._dirty_widgets: set[Widget] = set()
         self._update_timer: Timer | None = None
+        self._callbacks: list[Callable[[], Awaitable[None]]] = []
+        self._callbacks_ready: bool = False
```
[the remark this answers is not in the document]
darrenburns: Removed

darrenburns:
@willmcgugan The freeze was due to using events instead of messages. We were ending up in an infinite loop:

1. `MessagePump` sent an event to itself, picked up by its handler, which then forwards it to `Screen` (implying they both had the same `on_` method names.
2. Since it's an event, the MRO of `Screen` is queried and handlers invoked. The MRO of `Screen` contains `MessagePump`.
3. Since `MessagePump` contains a relevant handler, it was getting invoked.
4. Go to 1...

src/textual/messages.py
```diff
@@ -37,6 +37,18 @@ def can_replace(self, message: Message) -> bool:
         return isinstance(message, Layout)
 
 
+@rich.repr.auto
+class InvokeLater(Message, verbosity=3):
+    def __init__(
+        self, sender: MessagePump, callback: Callable[[], Awaitable[None]]
```
willmcgugan, line 43: I think this may be inherited from the current version, but are all callbacks awaitable? Not sure they are. Perhaps we need to define a callback type that accepts awaitables and plain old functions?

willmcgugan (approved):
LGTM, just one question.

PR #627, opened 2022-07-29
Adds Layers and Docks

willmcgugan:
Replaces the dock layout with a simpler dock rule.

The "dock" rule attribute accepts "top", "right", "bottom", and "left", and will move the widget to that edge, and exclude it from scrolling. Similar to an absolutely positioned div in HTML.

Additionally, this also implements *layers*. A container widget specified a list of layers, such as "base menu modal" to define a gross ordering for its children. Widgets specify a "layer" (singular) attribute to specify which layer a widget will go in to.


src/textual/_layout.py
```diff
@@ -25,6 +27,7 @@ class WidgetPlacement(NamedTuple):
     region: Region
     widget: Widget | None = None  # A widget of None means empty space
     order: int = 0
+    fixed: bool = False
```
willmcgugan, line 30: Sort of equivalent to `position: fixed` in CSS. Basically excludes a widget from scrolling.

src/textual/_styles_cache.py
```diff
@@ -246,7 +246,9 @@ def post(segments: Iterable[Segment]) -> list[Segment]:
         line: Iterable[Segment]
         # Draw top or bottom borders (A)
         if (border_top and y == 0) or (border_bottom and y == height - 1):
-            border_color = border_top_color if y == 0 else border_bottom_color
+            border_color = background + (
```
willmcgugan, line 249: Adding the background allows border colors to respect alpha.

src/textual/color.py
```diff
@@ -161,9 +165,10 @@ def clamped(self) -> Color:
     @property
     def rich_color(self) -> RichColor:
         """This color encoded in Rich's Color class."""
-        # TODO: This isn't cheap as I'd like - cache in a LRUCache ?
         r, g, b, _a = self
-        return RichColor.from_rgb(r, g, b)
+        return RichColor(
```
willmcgugan, line 169: Essentially doing the same, but cutting out a few intermediate steps.

e2e_tests/test_apps/basic.py
```diff
@@ -83,71 +88,112 @@ def render(self) -> Text:
...
-            sidebar=Widget(
-                Widget(classes="title"),
-                Widget(classes="user"),
-                OptionItem(),
-                OptionItem(),
-                OptionItem(),
-                Widget(classes="content"),
+            id="header",
+        )
+        yield from (
```
willmcgugan, line 111: Quite like this way of adding a bunch of widgets.

darrenburns (approved):

PR #629, opened 2022-07-30
Document CSS Styles

willmcgugan:
- Adds docs for most of the Styles. There are a few styles remaining which are not quite final
- Adds a textual CLI command to view borders

https://user-images.githubusercontent.com/554369/182879603-78976cd9-1a6e-48b2-98e6-786cb6cbd976.mov

https://user-images.githubusercontent.com/554369/182879628-0b306141-7439-44e8-bc02-59bd12a8cdfb.mov


docs/styles/box_sizing.md
```diff
@@ -0,0 +1,42 @@
+# Box-sizing
+
+The `box-sizing` rule impacts how `width` and `height` rules are translated in to screen dimensions, when combined with `padding` and `border`.
+
+The default value is `border-box` which means that padding and border are included in in the width and height. This setting means that if you add padding and/or border the widget will not change in size, but you will have less space for content.
```
darrenburns, line 5:
```suggestion
The default value is `border-box` which means that padding and border are included in the width and height. This setting means that if you add padding and/or border the widget will not change in size, but you will have less space for content.
```

docs/styles/box_sizing.md
```diff
@@ -0,0 +1,42 @@
+# Box-sizing
```
darrenburns, line 1: Struggled to follow this intro section. I don't have any suggestions for rewording at the moment, but just thought I'd note it here as something to potentially review. (Although I do think this concept is 100x easier to explain visually anyway)

docs/styles/color.md
```diff
@@ -0,0 +1,41 @@
...
+
+```sass
+/* Blue text */
+color: blue;
+
+/* 20% red text */
+color: red 20%;
+
+/* RGB color */
+color: rgb(100,120,200);
```
darrenburns, line 26: Might want to mention RGBA, HSL, HSLA? Hex syntax? Maybe that stuff belongs elsewhere? We could say the value should be a "Color", then we link off to all the information for how you can specify a "Color" (otherwise we may end up duplicating explanations across each rule that uses Color values)

```suggestion
/* RGB color */
color: rgb(100,120,200);

/* RGBA color (50% opacity) */
color: rgb(100,120,200, 0.5);  
```

docs/styles/color.md
```diff
@@ -0,0 +1,41 @@
...
+```
+
+## Python
+
+You can use the same syntax as CSS, or explicitly set a Color object.
```
darrenburns, line 31:
```suggestion
You can use the same syntax as CSS, or explicitly set a `Color` object.
```

docs/styles/display.md
```diff
@@ -0,0 +1,48 @@
+# Display
+
+The `display` property defines if a Widget is displayed or not. The default value is `"block"` which will display the widget as normal. Setting the property to `"none"` will effectively make it invisible.
```
darrenburns, line 3: Doesn't it also affect the layout of other widgets? The Widget isn't invisible - it's no longer there at all. Might be worth expanding on that now or in the future in order to differentiate this from `visiblility: hidden`.

docs/styles/height.md
```diff
@@ -0,0 +1,37 @@
...
+
+```sass
+/* Explicit cell height */
+height: 10;
+
+/* Percentage height */
+height: 50%;
+
+/* Automatic height */
+width: auto
```
darrenburns, line 28: Without saying what `auto` means this isn't very valuable. In fact, I've totally forgotten what it means and would love if these docs told me 😄

docs/styles/visibility.md
```diff
@@ -0,0 +1,48 @@
+# Visibility
+
+The `visibility` rule may be used to make a widget invisible while still reserving spacing for it. The default value is `"visible"` which will cause the Widget to be displayed as normal. Setting the value to `"hidden"` will cause the Widget to be removed from the screen.
```
darrenburns, line 3: Change to emphasise that the Widget is still there, but it's invisible. "Removed from screen" sounds like the Widget is no longer in the DOM.
```suggestion
The `visibility` rule may be used to make a widget invisible while still reserving spacing for it. The default value is `"visible"` which will cause the Widget to be displayed as normal. Setting the value to `"hidden"` will cause the Widget to become invisible.
```

src/textual/layouts/horizontal.py
```diff
@@ -30,7 +30,7 @@ def arrange(
         total_fraction = sum(
             [int(style.width.value) for style in styles if style.width.is_fraction]
         )
-        fraction_unit = Fraction(size.height, total_fraction or 1)
+        fraction_unit = Fraction(size.width, total_fraction or 1)
```
darrenburns, line 33: :)

darrenburns (approved):
A bunch of suggestions, take or leave any of them. Docs are something we can iterate on easily anyway.

docs/styles/height.md
willmcgugan: Ah yeah. Will add.

docs/examples/styles/padding.py
```diff
@@ -0,0 +1,32 @@
...
+TEXT = """I must not fear.
+Fear is the mind-killer.
+Fear is the little-death that brings total obliteration.
+I will face my fear.
+I will permit it to pass over me and through me.
+And when it has gone past, I will turn the inner eye to see its path.
+Where the fear has gone there will be nothing. Only I will remain."""
+
+
+class PaddingApp(App):
```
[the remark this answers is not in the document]
willmcgugan: Makes sense. Will leave that for a second pass then.

docs/styles/border.md
```diff
@@ -0,0 +1,68 @@
...
+
+Borders may also be set individually for the four edges of a widget with the `border-top`, `border-right`, `border-bottom` and `border-left` rules.
+
+## Border command
+
+The `textual` CLI has a subcommand which will let you explore the various border types:
+
+```
+textual borders
+```
```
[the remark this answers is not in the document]
willmcgugan: Will leave that for the moment. I'll come back to it when there is docs for these commands.

docs/styles/display.md
willmcgugan: It won't affect the layout if display is none. Will revisit this when there are docs that describe layout.

src/textual/layouts/horizontal.py
willmcgugan: My bad.

docs/styles/height.md
willmcgugan: Actually, there is rather a lot to explain regarding dimensions, not just auto. I'll come back to this one.

PR #638, opened 2022-08-05
Focus order

darrenburns:
Updates focus order to work in left to right then top to bottom order.
Anchor point is top left corner of the widget region **including** the margin spacing of the widget.

### DIFF
--- a/src/textual/dom.py
+++ b/src/textual/dom.py
@@ -19,7 +19,9 @@ from .css.errors import StyleValueError
 from .css.parse import parse_declarations
 from .css.styles import Styles, RenderStyles
 from .css.query import NoMatchingNodesError
+from .geometry import Region
 from .message_pump import MessagePump
+from .widget import Widget
 
 if TYPE_CHECKING:
     from .app import App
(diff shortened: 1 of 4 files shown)


src/textual/dom.py
```diff
@@ -596,3 +600,8 @@ def has_pseudo_class(self, *class_names: str) -> bool:
 
     def refresh(self, *, repaint: bool = True, layout: bool = False) -> None:
         pass
+
+
+def _focus_sort_key(widget: Widget) -> tuple[int, int]:
```
willmcgugan, line 605: I have a feeling we can optimize this with a combination of `attrgetter` and a new property.

It's not crazy to imagine a layout with 1000s of widgets.

I'm thinking that widget could grow a `_focus_sort_key` attribute. which return a tuple of (y, x), and we can get that with `attrgetter`

src/textual/widget.py
```diff
@@ -494,6 +494,16 @@ def window_region(self) -> Region:
         window_region = self.region.at_offset(self.scroll_offset)
         return window_region
 
+    @property
+    def region_with_margin(self) -> Region:
```
willmcgugan, line 498: Maybe this should be named `virtual_region_with_margin`. I've used the undecorated term "region" to mean the region relative to the screen.

willmcgugan (changes requested):
Nice simple solution.

Just a request which I think will cut a few cpu cycle.

src/textual/widget.py
darrenburns: Updated

src/textual/dom.py
darrenburns: I've made this change, and also moved `focusable_children` down to `Widget`. It's only used in the context of `Widget` types elsewhere, and in calculating it requires knowing the `virtual_region`, which doesn't exist at the `DOMNode` level.

willmcgugan (approved):

PR #653, opened 2022-08-06
Add remove

willmcgugan:
Implements a `remove()` method to remove a widget from the DOM.

### DIFF
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -884,6 +884,16 @@ class App(Generic[ReturnType], DOMNode):
         for _widget_id, widget in name_widgets:
             widget.post_message_no_wait(events.Mount(sender=parent))
 
+    def unregister(self, widget: Widget) -> None:
+        """Unregister a widget.
+
+        Args:
+            widget (Widget): _description_
+        """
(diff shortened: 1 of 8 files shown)


src/textual/app.py
```diff
@@ -856,7 +856,7 @@ def _register_child(self, parent: DOMNode, child: DOMNode) -> bool:
             return True
         return False
 
-    def register(
+    def _register(
```
willmcgugan, line 859: Devs should never need to call this directly.

There are probably a number of other methods that should be private. Will be an ongoing process to identify them

src/textual/css/query.py
```diff
@@ -36,14 +37,16 @@ def __init__(
         self,
         node: DOMNode | None = None,
         selector: str | None = None,
-        nodes: list[DOMNode] | None = None,
+        nodes: list[Widget] | None = None,
```
willmcgugan, line 40: Queries now return Widgets rather than DOMNodes. This means that it excludes App, which is find since it is a special object that doesn't have any visuals. Although you can still reference App in a selector.

darrenburns (approved):

PR #656, opened 2022-08-07
center layout

willmcgugan:
- Implements a "center" layout.
- Fixes an issues with scrollbars. If the mouse was captured when a scrollbar was hidden due to resizing the window, it was impossible to use the other scrollbar.
- Fixes an issue with refresh(layout=True)

https://user-images.githubusercontent.com/554369/183286651-631fa4e4-e2a2-4a8a-9def-420adc8958b3.mov

### DIFF
--- a/src/textual/layout.py
+++ b/src/textual/layout.py
@@ -1,35 +1,37 @@
 from .widget import Widget
 
 
-class Vertical(Widget):
-    """A container widget to align children vertically."""
+class Container(Widget):
+    """Simple container widget, with vertical layout."""
 
     CSS = """
-    Vertical {
+    Container {
         layout: vertical;       
         overflow: auto;
     }    
     """
 
 
-class Horizontal(Widget):
+class Vertical(Container):
+    """A container widget to align children vertically."""
+
+
+class Horizontal(Container):
     """A container widget to align children horizontally."""
 
     CSS = """
     Horizontal {
-        layout: horizontal;
-        overflow: auto;
+        layout: horizontal;        
     }    
     """
 
 
-class Center(Widget):
+class Center(Container):
     """A container widget to align children in the center."""
 
     CSS = """
     Center {
-        layout: center;
-        overflow: auto;
+        layout: center;        
     }
     
     """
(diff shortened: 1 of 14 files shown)


src/textual/layout.py
```diff
@@ -1,23 +1,37 @@
 from .widget import Widget
 
 
-class Vertical(Widget):
-    """A container widget to align children vertically."""
+class Container(Widget):
```
willmcgugan, line 4: A base class for common styles, and for containers where you might want to dynamically tweak the layout.

src/textual/layouts/factory.py
```diff
@@ -1,11 +1,15 @@
...
+
 from .._layout import Layout
-from ..layouts.vertical import VerticalLayout
+from .horizontal import HorizontalLayout
+from .vertical import VerticalLayout
+from .center import CenterLayout
 
 
-LAYOUT_MAP = {
+LAYOUT_MAP: dict[str, type[Layout]] = {
```
willmcgugan, line 9: Fixes a typing error below. Mypy couldn't figure this one out by itself.

src/textual/scrollbar.py
```diff
@@ -231,10 +231,14 @@ def render(self) -> RenderableType:
             style=scrollbar_style,
         )
 
-    async def on_enter(self, event: events.Enter) -> None:
+    def on_hide(self, event: events.Hide) -> None:
```
willmcgugan, line 234: If the scrollbar is hidden it releases the mouse. I'm wondering if this should be standard functionality, but there are some exotic use cases for a hidden widget capturing the mouse.

src/textual/widget.py
```diff
@@ -1040,7 +1040,9 @@ def refresh(
 
         if layout:
             self._layout_required = True
-            self._clear_arrangement_cache()
+            if isinstance(self._parent, Widget):
```
willmcgugan, line 1043: When a widget's dimensions / visibility change, we want to clear the *parent's* arrangement cache.

darrenburns (approved):

PR #660, opened 2022-08-10
unified events and messages

willmcgugan:
This unifies Events and Messages so that there is a single mechanism (calling `on_*`), and also introduces a namespace to make name collisions less likely.

There still are Events and Messages. Message is the base class, and are reserved for use by the system. Devs are expected to work with Events only.

- Provides a single unified mechanism to handle events and messages
- Adds a namespace to messages
- Removes some deprecated code


sandbox/darren/file_search.py
```diff
@@ -58,7 +58,7 @@ def on_mount(self) -> None:
         self.mount(file_table_wrapper=Widget(self.file_table))
         self.mount(search_bar=self.search_bar)
 
-    def handle_changed(self, event: TextWidgetBase.Changed) -> None:
+    def on_text_widget_base_changed(self, event: TextWidgetBase.Changed) -> None:
```
willmcgugan, line 61: We might want to explicitly set the namespace here. Reads a little clumsily. Guess we could set the namespace to "text_widget" so the method would be "on_text_widget_changed".

src/textual/events.py
```diff
@@ -28,11 +28,6 @@ def __init_subclass__(cls, bubble: bool = True, verbosity: int = 1) -> None:
         super().__init_subclass__(bubble=bubble, verbosity=verbosity)
 
 
-class Null(Event, verbosity=3):
```
willmcgugan, line 31: This used to serve a purpose, but was replaced by Prompt.

src/textual/message_pump.py
```diff
@@ -51,7 +54,27 @@ def __gt__(self, other: object) -> bool:
         return self.priority > other.priority
 
 
-class MessagePump:
+class MessagePumpMeta(type):
```
willmcgugan, line 57: A wild metaclass appears!

src/textual/message_pump.py
```diff
@@ -290,20 +316,32 @@ async def _process_messages(self) -> None:
 
         log("CLOSED", self)
 
-    async def dispatch_message(self, message: Message) -> bool | None:
+    async def dispatch_message(self, message: Message) -> None:
+        """Dispatch a message received form the message queue.
```
darrenburns, line 320:
```suggestion
        """Dispatch a message received from the message queue.
```

darrenburns (approved):

PR #664, opened 2022-08-10
Filling the gap between horizontal and vertical scrollbars

darrenburns:
Introduces new inline and CSS property `scrollbar-corner-color` which sets the color of the cells that fill the gap between the scrollbars. Default color is 1 level darker than the hovered state of the scrollbar background.

image


src/textual/scrollbar.py
```diff
@@ -287,6 +288,16 @@ async def on_mouse_move(self, event: events.MouseMove) -> None:
             await self.emit(ScrollTo(self, x=x, y=y))
 
 
+class ScrollBarCorner(Widget):
```
willmcgugan, line 291: Class docstring please

willmcgugan (approved):
LGTM just a request for a docstring

PR #665, opened 2022-08-11
CSS tie breaker

willmcgugan:
Fixes the Horizontal container returning a `<vertical>` layout.

The problem was that every time you add a widget it adds rules for the base classes. If widgets share the same base classes you get repeat rules, with the same specificity.

This made the rules order dependant. It was undefined which rules would win.

The solution was to add a tie breaker to the specificity that adds an integer for every base class that specifies its order in the inheritance chain. So that base class rules are in the right order.

- Adds a query_one() method to get a single object via a selector.


willmcgugan:
Taking the liberty of merging. @darrenburns feel free to do a retrospective review when you're back in the office.

PR #667, opened 2022-08-11
partial tree control fix

willmcgugan:
Partially fixes the tree control, which was broken in the new API.

This is mostly for demos at the moment. I think we should combine tree functionality in to the data table, for combined tree + tables. It will be more efficient that the old API which renders the entire tree each change

https://user-images.githubusercontent.com/554369/184134408-b7cfd225-6936-4c17-bbcf-fb3887dfb8f5.mov

.

### DIFF
--- a/sandbox/will/basic.css
+++ b/sandbox/will/basic.css
@@ -24,6 +24,23 @@ App > Screen {
   
 }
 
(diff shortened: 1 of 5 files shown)


willmcgugan:

PR #683, opened 2022-08-16
Animation callback

darrenburns:


src/textual/_animator.py
```diff
@@ -241,7 +259,11 @@ def __call__(self) -> None:
             animation_keys = list(self._animations.keys())
             for animation_key in animation_keys:
                 animation = self._animations[animation_key]
-                if animation(animation_time):
+                animation_complete = animation(animation_time)
+                if animation_complete:
+                    completion_callback = animation.on_complete
+                    if completion_callback is not None:
+                        completion_callback()
```
willmcgugan, line 266: I think we will need to use `invoke` here so that you can also pass an awaitable.

src/textual/_animator.py
```diff
@@ -48,6 +60,7 @@ class SimpleAnimation(Animation):
     end_value: float | Animatable
     final_value: object
     easing: EasingFunction
+    on_complete: Callable[[], None] | None = None
```
willmcgugan, line 63: This may need to be typed as a "CallbackType"

willmcgugan (changes requested):
Looks good, just a request re using invoke for callbacks

src/textual/_animator.py
darrenburns: Updated and extracted CallbackType to `textual._types` since it's now used in more than one module.

src/textual/_animator.py
darrenburns: I updated this to use `invoke`, awaited it, and updated the signature to `async __call__`. The only place that I believe that Animator is called also uses `invoke`, so there was no propagation of `async`.

willmcgugan (approved):
LGTM

PR #684, opened 2022-08-16
Delay transition

darrenburns:
Stacked on top of #683 - review that one first.


willmcgugan (changes requested):
This solution will invoke the callback during the "delay" period. The end result will be correct, in that the user will see the delay. The downside is that the animator is working during the delay period and not updating anything.

Ideally the animator would not do any work until the delay has expired.

If this looks particularly complex to implement, suggest putting in a early-out condition. i.e. if the delay hasn't expired return as early as possible.

src/textual/_animator.py
```diff
@@ -141,7 +143,7 @@ def __call__(
 class Animator:
     """An object to manage updates to a given attribute over a period of time."""
 
-    def __init__(self, target: MessageTarget, frames_per_second: int = 60) -> None:
+    def __init__(self, target: App, frames_per_second: int = 60) -> None:
```
willmcgugan, line 146: Is the Animator only associated with the app?

If so, may as well rename `target` to `app` ?

willmcgugan (approved):
Just one thing to consider.

src/textual/_animator.py
darrenburns: Done

PR #695, opened 2022-08-18
Docs scrollbar gutter

darrenburns:


docs/styles/scrollbar_gutter.md
```diff
@@ -0,0 +1,34 @@
+# Scrollbar gutter
```
willmcgugan, line 1: Could you add a Syntax section. See the docs-intro branch for examples...

willmcgugan (changes requested):
Great. I added a Syntax section to the styles in the docs-intro branch. Could you please add something similar to scrollbar gutter.

.pre-commit-config.yaml
```diff
@@ -2,11 +2,12 @@
 # See https://pre-commit.com/hooks.html for more hooks
 repos:
 -   repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v3.2.0
+    rev: v4.3.0
     hooks:
     -   id: trailing-whitespace
     -   id: end-of-file-fixer
     -   id: check-yaml
+        args: ['--unsafe']
```
darrenburns, line 10: This hook was failing due to the `format: !!python/name:textual._doc.format_svg` syntax in `mkdocs.yml`. This arg prevents attempting to load the YAML:

```python
            'Instead of loading the files, simply parse them for syntax.  '
            'A syntax-only check enables extensions and unsafe contstructs '
            'which would otherwise be forbidden.  Using this option removes '
            'all guarantees of portability to other yaml implementations.  '
```

docs/styles/scrollbar_gutter.md
darrenburns: Added `Syntax` section, with a child `Values` section.

The `Values` section is probably where we'd link off to "types" e.g. color, scalar, etc.

image

docs/styles/scrollbar_gutter.md
```diff
@@ -0,0 +1,48 @@
...
+
+=== "scrollbar_gutter.py"
+
+    ```python
+    --8<-- "docs/examples/styles/scrollbar_gutter.py"
+    ```
+
+=== "Output"
+
+    ```{.textual path="docs/examples/styles/scrollbar_gutter.py"}
```
darrenburns, line 34: Absolutely love this plugin

docs/styles/scrollbar_gutter.md
willmcgugan: I like the Values section. Will add a task to review existing styles.

docs/styles/scrollbar_gutter.md
willmcgugan: It is kind of magical!

willmcgugan (approved):
LGTM

PR #700, opened 2022-08-19
Docs content align

darrenburns:
image

image

### DIFF
--- a/docs/examples/styles/content_align.py
+++ b/docs/examples/styles/content_align.py
@@ -3,29 +3,11 @@ from textual.widgets import Static
 
 
 class ContentAlignApp(App):
-    CSS = """
-    #box1 {
-        content-align: left top;
-        background: $panel;
-    }
-    #box2 {
-        content-align: center middle;
-        background: $panel-darken-1;
-    }
-    #box3 {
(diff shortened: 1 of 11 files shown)


docs/examples/styles/content_align.py
```diff
@@ -0,0 +1,31 @@
+from textual.app import App
+from textual.widgets import Static
+
+
+class ContentAlignApp(App):
```
willmcgugan, line 5: Could you separate this in to Python + CSS. 

I want to discourage people from embedding the CSS because the auto-refresh won't work. You also won't need the `ContentAlignApp` prefix to make it work.

Also, could you avoid using variables. I don't want to introduce too many concepts.

willmcgugan (changes requested):
Great, just a request to break the examples in to python + CSS

docs/examples/styles/content_align.py
darrenburns: I've split them up, and done the same for scrollbar-gutter.

willmcgugan (changes requested):
Nice

willmcgugan (approved):

PR #702, opened 2022-08-23
Fix up a bunch of type-hint errors

diceroll123:
Just strolling around the internet and found this lib, thought I'd fix a few type hints since I noticed some incorrectness.

I _highly_ recommend installing Pylance if you use VSCode. 😄

Open for questions, if requested. Thanks for readin'.

### DIFF
--- a/src/textual/_callback.py
+++ b/src/textual/_callback.py
@@ -7,12 +7,12 @@ from typing import Any, Callable
 
 
 @lru_cache(maxsize=2048)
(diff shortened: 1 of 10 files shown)


tusharsadhwani:
Hey, unfortunately most of this code is stale as of now, as most development is going on in the `css` branch. That branch also has a CI job on it that type checks it with mypy, so its types should probably be good.

PR #703, opened 2022-08-23
Docs intro

willmcgugan:
- Adds introduction.md and example code
- Docstrings
- Adds docs for CSS
- Docstrings
- Fixes a few bugs I encountered when working on examples
- Docstrings

Fixes a long standing bug that has likely been causing random-ish glitches for months. The priority queue system which allows timer events to jump the queue wasn't working. It was sometimes returning events out of order.

I've replaced the priority queue with a simple queue. The priority queue may have been a premature optimisation. We might implement it properly at some point if it transpires we actually need it.


docs/events/key.md
```diff
@@ -0,0 +1,11 @@
...
+
+The `Key` event is sent to a widget when the user presses key on the keyboard.
+
+- [x] Bubbles
+
+## Attributes
+
+| attribute | type | purpose                  |
+| --------- | ---- | ------------------------ |
+| `key`     | str  | The key that was pressed |
```
rodrigogiraoserrao, line 11: The “purpose” column in the attributes tables for events looks more like a “description” column, no? Exhibit A.

docs/events/key.md
```diff
@@ -0,0 +1,11 @@
+# Key
+
+The `Key` event is sent to a widget when the user presses key on the keyboard.
```
rodrigogiraoserrao, line 3:
```suggestion
The `Key` event is sent to a widget when the user presses a key on the keyboard.
```

docs/events/load.md
```diff
@@ -0,0 +1,11 @@
+# Mount
```
rodrigogiraoserrao, line 1:
```suggestion
# Load
```

docs/events/mouse_click.md
```diff
@@ -0,0 +1,20 @@
+# Click
+
+The `Click` event is sent to a widget when the user clicks a mouse button.
+
+- [x] Bubbles
```
rodrigogiraoserrao, line 5: All other `mouse_xxx.md` files seem to have a `Mousexxx` event. Is this event purposefully different, or should this be a `MouseClick` event?

docs/events/mouse_click.md
```diff
@@ -0,0 +1,20 @@
...
+| attribute  | type | purpose                                   |
+| ---------- | ---- | ----------------------------------------- |
+| `x`        | int  | Mouse x coordinate, relative to Widget    |
+| `y`        | int  | Mouse y coordinate, relative to Widget    |
+| `delta_x`  | int  | Change in x since last mouse event        |
+| `delta_y`  | int  | Change in x since last mouse event        |
+| `button`   | int  | Index of mouse button                     |
+| `shift`    | bool | Shift key pressed if True                 |
+| `meta`     | bool | Meta key pressed if True                  |
+| `ctrl`     | bool | Shift key pressed if True                 |
```
rodrigogiraoserrao, line 18:
```suggestion
| `ctrl`     | bool | Ctrl key pressed if True                 |
```

rodrigogiraoserrao, line 18: I'd find it clearer if it had been phrased the other way around:

```suggestion
| `shift`    | bool | True if Shift key pressed                 |
| `meta`     | bool | True if Meta key pressed                  |
| `ctrl`     | bool | True if Shift key pressed                 |
```

docs/events/mouse_down.md
```diff
@@ -0,0 +1,20 @@
...
+- [x] Bubbles
+
+## Attributes
+
+| attribute  | type | purpose                                   |
+| ---------- | ---- | ----------------------------------------- |
+| `x`        | int  | Mouse x coordinate, relative to Widget    |
+| `y`        | int  | Mouse y coordinate, relative to Widget    |
+| `delta_x`  | int  | Change in x since last mouse event        |
+| `delta_y`  | int  | Change in x since last mouse event        |
```
rodrigogiraoserrao, line 14:
```suggestion
| `delta_y`  | int  | Change in y since last mouse event        |
```

docs/events/mouse_click.md
```diff
@@ -0,0 +1,20 @@
+# Click
+
+The `Click` event is sent to a widget when the user clicks a mouse button.
+
+- [x] Bubbles
+
+## Attributes
```
rodrigogiraoserrao, line 7: Other mouse events have the exact same table, so whatever changes are accepted here, should be copied and pasted over to `MouseDown`, `MouseMove`, and `MouseUp`.

docs/events/mouse_release.md
```diff
@@ -0,0 +1,11 @@
+# MouseRelease
+
+The `MouseRelease` event is sent to a widget it is no longer receiving mouse events outside of its borders.
```
rodrigogiraoserrao, line 3:
```suggestion
The `MouseRelease` event is sent to a widget when it is no longer receiving mouse events outside of its borders.
```

docs/events/mouse_scroll_down.md
```diff
@@ -0,0 +1,12 @@
+# MouseScrollDown
+
+The `MouseScrollDown` event is sent to a widget when scroll wheel (or trackpad equivalent) is moved _down_.
```
rodrigogiraoserrao, line 3:
```suggestion
The `MouseScrollDown` event is sent to a widget when the scroll wheel (or trackpad equivalent) is moved _down_.
```

docs/events/mouse_scroll_up.md
```diff
@@ -0,0 +1,12 @@
+# MouseScrollUp
+
+The `MouseScrollUp` event is sent to a widget when scroll wheel (or trackpad equivalent) is moved _up_.
```
rodrigogiraoserrao, line 3:
```suggestion
The `MouseScrollUp` event is sent to a widget when the scroll wheel (or trackpad equivalent) is moved _up_.
```

docs/events/resize.md
```diff
@@ -4,20 +4,10 @@ The `Resize` event is sent to a widget when its size changes and when it is firs
...
-: The size of the widget's container.
-
-## Code
-
-::: textual.events.Mount
+| attribute        | type | purpose                                           |
+| ---------------- | ---- | ------------------------------------------------- |
+| `size`           | Size | The new size of the Widget.                       |
+| `virtual_size`   | Size | The virtual size (scrollable area) of the Widget. |
+| `container_size` | Size | The size of the container (parent widget).        |
```
rodrigogiraoserrao, line 13:
```suggestion
| `size`           | Size | The new size of the Widget                       |
| `virtual_size`   | Size | The virtual size (scrollable area) of the Widget |
| `container_size` | Size | The size of the container (parent widget)        |
```

Removed full stops for consistency with the other attribute tables.

docs/events/mouse_click.md
darrenburns: Me too :)

docs/events/mouse_click.md
willmcgugan: I see your point, but the mouse prefix is redundant for click. There is no other kind of click. I think I might have to be pragmatic and accept a little inconsistency.

docs/guide/CSS.md
```diff
@@ -0,0 +1,395 @@
...
+  height: 3;
+  content-align: center middle;
+  background: blue;
+  color: white;
+}
+```
+
+The lines inside the curly braces contains CSS _rules_, which consist of a rule name and rule value separated by a colon and ending in a semi-colon. Such rules are typically written one per line, but you could add additional rules as long as they are separated by semi-colons.
+
+The first rule in the above example reads `"dock: top;"`. The rule name is `dock` which tells Textual to place the widget on a edge of the screen. The text after the colon is `top` which tells Textual to dock to the _top_ of the screen. Other valid values for dock are "right", "bottom", or "left"; but `top` is naturally appropriate for a header.
```
rodrigogiraoserrao, line 57:
```suggestion
The first rule in the above example reads `"dock: top;"`. The rule name is `dock` which tells Textual to place the widget on an edge of the screen. The text after the colon is `top` which tells Textual to dock to the _top_ of the screen. Other valid values for dock are "right", "bottom", or "left"; but `top` is naturally appropriate for a header.
```

docs/guide/CSS.md
```diff
@@ -0,0 +1,395 @@
...
+
+You may be able to guess what some of the the other rules do. We will cover those later.
+
+## The DOM
+
+The DOM, or _Document Object Model_, is a term borrowed from the web world. Textual doesn't use documents but the term has stuck. In Textual CSS, the DOM is a an arrangement of widgets that forms a tree of sorts.
```
rodrigogiraoserrao, line 63:
```suggestion
The DOM, or _Document Object Model_, is a term borrowed from the web world. Textual doesn't use documents but the term has stuck. In Textual CSS, the DOM is an arrangement of widgets that forms a tree of sorts.
```

rodrigogiraoserrao, line 63: Not a (personal) fan of “forms a tree of sorts” unless this is the kind of tone you want in the docs.

 - ... the DOM is an arrangement of widgets you can visualise in a tree-like hierarchical structure.
 - ... the DOM is the (tree-like) hierarchy that establishes the relationships between the widgets.
 - ...?

docs/guide/CSS.md
```diff
@@ -0,0 +1,395 @@
...
+
+Some widgets contain other widgets: for instance a list control widget will likely also have item widgets, or a dialog widget may contain button widgets. These _child_ widgets form the branches of the tree.
```
rodrigogiraoserrao, line 65:
```suggestion
Some widgets contain other widgets: for instance, a list control widget will likely also have item widgets, or a dialog widget may contain button widgets. These _child_ widgets form the branches of the tree.
```

docs/guide/CSS.md
```diff
@@ -0,0 +1,395 @@
...
+    ```python
+    --8<-- "docs/examples/guide/dom2.py"
+    ```
+
+=== "Output"
+
+    ```{.textual path="docs/examples/guide/dom2.py"}
+    ```
+
+With a header and a footer widget the DOM look the this:
```
rodrigogiraoserrao, line 101:
```suggestion
With a header and a footer widget the DOM looks the this:
```

docs/guide/CSS.md
willmcgugan: I'll rephrase. Needs to be a little informal though, for novices devs without a background in computer science.

docs/guide/CSS.md
```diff
@@ -0,0 +1,395 @@
...
+
+## CSS files
+
+To add a stylesheet we need to pass the path to a CSS file via the app classes' `css_path` argument:
+
+```python hl_lines="23"
+--8<-- "docs/examples/guide/dom4.py"
+```
+
+You may have noticed that some of the constructors have additional keywords argument: `id` and `classes`. These are used by the CSS to identify parts of the DOM. We will cover these in the next section.
```
rodrigogiraoserrao, line 150:
```suggestion
You may have noticed that some of the constructors have additional keyword argument: `id` and `classes`. These are used by the CSS to identify parts of the DOM. We will cover these in the next section.
```

docs/guide/CSS.md
```diff
@@ -0,0 +1,395 @@
...
+
+Here's the CSS file we are applying:
+
+```sass
+--8<-- "docs/examples/guide/dom4.css"
+```
+
+The CSS contains a number of rules sets with a selector and a list of rules. You can also add comments with text between `/*` and `*/` which will be ignored by Textual. Add comments to leave yourself reminders or to temporarily disable selectors.
```
rodrigogiraoserrao, line 158:
```suggestion
The CSS contains a number of rule sets with a selector and a list of rules. You can also add comments with text between `/*` and `*/` which will be ignored by Textual. Add comments to leave yourself reminders or to temporarily disable selectors.
```

docs/guide/CSS.md
```diff
@@ -0,0 +1,395 @@
...
+
+```bash
+textual run my_app.py --dev
+```
+
+Being able to iterate on the design without restarting the Python code can make it easier and faster to design beautiful interfaces.
+
+## Selectors
+
+A selector is the text which precedes the curly braces in a set of rules. It tells textual which widgets it should apply the rules to.
```
rodrigogiraoserrao, line 184:
```suggestion
A selector is the text which precedes the curly braces in a set of rules. It tells Textual which widgets it should apply the rules to.
```

docs/guide/CSS.md
```diff
@@ -0,0 +1,395 @@
...
+
+Selectors can target a kind of widget or a very specific widget. For instance you could have a selector that modifies all buttons, or you could target an individual button used in one dialog. This gives you a lot of flexibility in customization your user interface.
```
rodrigogiraoserrao, line 186:
```suggestion
Selectors can target a kind of widget or a very specific widget. For instance you could have a selector that modifies all buttons, or you could target an individual button used in one dialog. This gives you a lot of flexibility in customizing your user interface.
```

docs/guide/CSS.md
```diff
@@ -0,0 +1,395 @@
...
+  background: blue;
+  border: rounded white;
+}
+```
+
+!!! note
+
+    The fact that the type selector matches base classes is a departure from browser CSS which doesn't have the same concept.
+
+You may have noticed that the `border` rule exists in both Static and Button. When this happens, Textual will use the most recently defined sub-class within a list of bases. So Button wins over Static, and Static wins over Widget (the base class of all widgets). Hence if both rules were in a stylesheet, the buttons would be "solid blue" and not "rounded white".
```
rodrigogiraoserrao, line 222: May be useful to draw an explicit parallel to how method resolution works. If `Button` and `Static` both define the method `foo`, calling `Button().foo()` actually runs `Button.foo` and not `Static.foo`.

docs/guide/CSS.md
```diff
@@ -0,0 +1,395 @@
...
+}
+```
+
+A Widget's `id` attribute can not be changed after the Widget has been constructed. 
+
+### Class-name selector
+
+Every widget can have a number of class names applied. The term "class" here is borrowed from web CSS, and has a different meaning to a Python class. You can think of a CSS class as a tag of sorts. Widgets with the same tag will share styles.
+
+CSS classes are set via the widgets `classes` parameter in the constructor. Here's an example:
```
rodrigogiraoserrao, line 248:
```suggestion
CSS classes are set via the widget's `classes` parameter in the constructor. Here's an example:
```

docs/events/mouse_click.md
rodrigogiraoserrao: I suspected this was intentional but decided to raise the issue either way, just to be extra sure.

docs/guide/CSS.md
```diff
@@ -0,0 +1,395 @@
...
+
+Class name selectors may be _chained_ together by appending another full stop and class name. The selector will match a widget that has _all_ of the class names set. For instance, the following sets a red background on widgets that have both `error` _and_ `disabled` class names.
+
+```sass
+.error.disabled {
+  background: darkred;
+}
+```
+
+Unlike the `id` attribute a Widget's classes can be changed after the Widget was created. Adding and removing CSS classes is the recommended way of changing the display while your app is running. There are a few methods you can use to manage CSS classes.
```
rodrigogiraoserrao, line 283: I feel some inconsistencies in the casing of Python class names/nouns of the things we are describing.
E.g., I would capitalised `Widget` differently:

```suggestion
Unlike the `id` attribute, a widget's classes can be changed after the widget was created. Adding and removing CSS classes is the recommended way of changing the display while your app is running. There are a few methods you can use to manage CSS classes.
```

```suggestion
Unlike the `id` attribute, classes can be changed after the Widget instance was created. Adding and removing CSS classes is the recommended way of changing the display while your app is running. There are a few methods you can use to manage CSS classes.
```

docs/guide/CSS.md
```diff
@@ -0,0 +1,395 @@
...
+
+- [add_class()][textual.dom.DOMNode.add_class] Adds one or more classes to a widget.
+- [remove_class()][textual.dom.DOMNode.remove_class] Removes class name(s) from a widget.
+- [toggle_class()][textual.dom.DOMNode.toggle_class] Removes a class name if it is present, or adds the name if its not already present.
```
rodrigogiraoserrao, line 287:
```suggestion
- [toggle_class()][textual.dom.DOMNode.toggle_class] Removes a class name if it is present, or adds the name if it's not already present.
```

docs/guide/CSS.md
```diff
@@ -0,0 +1,395 @@
...
+- [has_class()][textual.dom.DOMNode.has_class] Checks if a class(es) is set on a widget.
+- [classes][textual.dom.DOMNode.classes] Is a frozen set of the class(es) set on a widget.
+
+
+### Universal selector
+
+The _universal_ selectors is denoted by an asterisk and will match _all_ widgets.
```
rodrigogiraoserrao, line 294:
```suggestion
The _universal_ selector is denoted by an asterisk and will match _all_ widgets.
```

docs/guide/CSS.md
```diff
@@ -0,0 +1,395 @@
...
+
+For example, the following will draw a red outline around all widgets:
+
+```sass
+* {
+  outline: solid red;
+}
+```
+
+### Pseudo classes
```
rodrigogiraoserrao, line 304: Pseudo class or pseudo selector? Both are used below, seemingly interchangeably. Might be worth picking one, or clarifying they mean the same thing.

docs/guide/CSS.md
```diff
@@ -0,0 +1,395 @@
...
+
+Let's say we want to make the text of the buttons in the dialog bold, but we _don't_ want to change the Button in the sidebar. We can do this with the following rule:
+
+```sass hl_lines="1"
+#dialog Button {
+  text-style: bold;
+}
+```
+
+The `#dialog Button` selector matches all buttons that are below the widget with an id of "dialog". No other buttons will be matched.
```
rodrigogiraoserrao, line 343:
```suggestion
The `#dialog Button` selector matches all buttons that are below the widget with an ID of "dialog". No other buttons will be matched.
```

ID was uppercase elsewhere.

docs/guide/CSS.md
```diff
@@ -0,0 +1,395 @@
...
+
+As with all selectors you can combine as many as you wish. The following will match a `Button` that is under a `Horizontal` widget _and_ under a widget with an id of `"dialog"`:
```
rodrigogiraoserrao, line 345: Not too familiar with CSS, so this question arises: are we matching a `Button` that is under a `Horizontal` that is in turn under a `#dialog`? Or am I matching all `Button` for which some parent is a `Horizontal` and for which some parent is a `#dialog`? Does this match if a single parent is both `Horizontal` _and_ a `#dialog`?

docs/guide/CSS.md
rodrigogiraoserrao, line 343: This matches `Button` instances directly below `#dialog` or at any level of nesting under `#dialog`?

docs/guide/devtools.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Run
+
+You can run Textual apps with the `run` subcommand. If you supply a path to a Python file it will load and run the application.
+
+```bash
+textual run my_app.py
+```
+
+The `run` sub-command assumes you have a App instance called `app` in the global scope of your Python file. If the application is called something different, you can specify it with a colon following the filename:
```
rodrigogiraoserrao, line 19:
```suggestion
The `run` sub-command assumes you have an App instance called `app` in the global scope of your Python file. If the application is called something different, you can specify it with a colon following the filename:
```

docs/guide/devtools.md
```diff
@@ -0,0 +1,51 @@
...
+textual run my_app.py:alternative_app
+```
+
+!!! note
+
+    If the Python file contains a call to app.run() then you can launch the file as you normally would any other Python program. Running your app via `textual run` will give you access to a few Textual features such as live editing of CSS files.
+
+## Console
+
+When running any terminal application, you can no longer use `print` when debugging (or log to the console). This is because anything you write to standard output would overwrite application content, making it unreadable. Fortunately Textual supplies a debug console of it's own which has some super helpful features.
```
rodrigogiraoserrao, line 31:
```suggestion
When running any terminal application, you can no longer use `print` when debugging (or log to the console). This is because anything you write to standard output would overwrite application content, making it unreadable. Fortunately Textual supplies a debug console of its own which has some super helpful features.
```

docs/introduction.md
```diff
@@ -2,150 +2,449 @@
...
+Here's what the finished app will look like:
+
+
+```{.textual path="docs/examples/introduction/stopwatch.py" press="tab,enter,_,tab,enter,_,tab,_,enter,_,tab,enter,_,_"}
 ```
 
-Enter the following command to run the application:
+### Get the code
+
+If you want to try the finished Stopwatch app and following along with the code first make sure you have [textual installed](getting_started.md) then check out the [Textual](https://github.com/Textualize/textual) GitHub repository:
```
rodrigogiraoserrao, line 28:
```suggestion
If you want to try the finished Stopwatch app and follow along with the code, first make sure you have [Textual installed](getting_started.md) and then check out the [Textual](https://github.com/Textualize/textual) GitHub repository:
```

I am not sure my added “and” should be there.

docs/introduction.md
```diff
@@ -2,150 +2,449 @@
...
-python intro01.py
+cd textual/docs/examples/introduction
+python stopwatch.py
+```
+
+## Type hints (in brief)
+
+!!! tip inline end
+
+    Type hints are entirely optional in Textual. We've included them in the example code but it's up to you wether you add them to your own projects.
```
rodrigogiraoserrao, line 59:
```suggestion
    Type hints are entirely optional in Textual. We've included them in the example code but it's up to you whether you add them to your own projects.
```

docs/introduction.md
```diff
@@ -2,150 +2,449 @@
...
+
+We're a big fan of Python type hints at Textualize. If you haven't encountered type hinting, its a way to express the types of your data, parameters, and return values. Type hinting allows tools like [Mypy](https://mypy.readthedocs.io/en/stable/) to catch potential bugs before your code runs.
```
rodrigogiraoserrao, line 61:
```suggestion
We're a big fan of Python type hints at Textualize. If you haven't encountered type hinting, it's a way to express the types of your data, parameters, and return values. Type hinting allows tools like [Mypy](https://mypy.readthedocs.io/en/stable/) to catch potential bugs before your code runs.
```

docs/introduction.md
```diff
@@ -2,150 +2,449 @@
...
---8<-- "docs/examples/introduction/intro01.py"
+```python title="stopwatch01.py" hl_lines="5-19"
+--8<-- "docs/examples/introduction/stopwatch01.py"
 ```
 
-The `run` method will put your terminal in to "application mode" which disables the prompt and allows Textual to take over input and output. When you press ++ctrl+c++ the application will exit application mode and re-enable the command prompt.
+The App class is where most of the logic of Textual apps is written. It is responsible for loading configuration, setting up widgets, handling keys, and more.
 
-## Handling Events
+There are three methods in our stopwatch app currently.
```
rodrigogiraoserrao, line 115:
```suggestion
Currently, there are three methods in our stopwatch app.
```

docs/introduction.md
```diff
@@ -2,150 +2,449 @@
...
+-  **`compose()`** is where we construct a user interface with widgets. The `compose()` method may return a list of widgets, but it is generally easier to _yield_ them (making this method a generator). In the example code we yield instances of the widget classes we imported, i.e. the header and the footer.
 
-Each event type is represented by an event object, which is an instance of a class containing information you may need to respond the the event. For instance the `Key` event contains the key the user pressed and a `Mouse` event will contain the coordinates of the mouse cursor.
+- **`on_load()`** is an _event handler_ method. Event handlers are called by Textual in response to external events like keys and mouse movements, and internal events needed to manage your application. Event handler methods begin with `on_` followed by the name of the event (in lower case). Hence, `on_load` it is called in response to the Load event which is sent just after the app starts. We're using this event to call `App.bind()` which connects a key to an _action_.
```
rodrigogiraoserrao, line 119:
```suggestion
- **`on_load()`** is an _event handler_ method. Event handlers are called by Textual in response to external events like keys and mouse movements, and internal events needed to manage your application. Event handler methods begin with `on_` followed by the name of the event (in lower case). Hence, `on_load` is called in response to the Load event which is sent just after the app starts. We're using this event to call `App.bind()` which connects a key to an _action_.
```

docs/introduction.md
```diff
@@ -2,150 +2,449 @@
...
+
+- **`action_toggle_dark()`** defines an _action_ method. Actions are methods beginning with `action_` followed by the name of the action. The call to `bind()` in `on_load()` binds this the ++d++ key to this action. The body of this method flips the state of the `dark` boolean to toggle dark mode.
```
rodrigogiraoserrao, line 121:
```suggestion
- **`action_toggle_dark()`** defines an _action_ method. Actions are methods beginning with `action_` followed by the name of the action. The call to `bind()` in `on_load()` binds this the ++d++ key to this action. The body of this method flips the state of the `dark` boolean to toggle dark mode.
```
```suggestion
- **`action_toggle_dark()`** defines an _action_ method. Actions are methods beginning with `action_` followed by the name of the action. The call to `bind()` in `on_load()` binds the ++d++ key to this action. The body of this method flips the state of the `dark` boolean to toggle dark mode.
```

docs/introduction.md
```diff
@@ -2,150 +2,449 @@
...
+### Extending widget classes
 
-!!! note
+We've imported two new widgets in this code: `Button`, which creates a clickable button, and `Static` which is a base class for a simple control. We've also imported `Container` from `textual.layout`. As the name suggests, `Container` is a Widget which contains other widgets. We will use this container to create a scrolling list of stopwatches.
+
+We're extending Static as a foundation for our `TimeDisplay` widget. There are no methods on this class yet. 
+
+The Stopwatch class also extends Static to define a new widget. This class has a `compose()` method which yields its child widgets, consisting of three `Button` objects and a single `TimeDisplay`. These are all we need to build a stopwatch as in the sketch.
+
+The Button constructor takes a label to be displayed in the button ("Start", "Stop", or "Reset"). Additionally some of the buttons set the following parameters:
```
rodrigogiraoserrao, line 167: widget vs Widget vs `Widget`

Are there really three levels? “widget” should be the English noun and `Widget` is the Python class. “Widget” (capitalised, no backticks) can be dropped, right?
Then, it is “just” a matter of being careful and consistent with all other words.
Or not! If not, let me know and I will stop pointing these things out.

docs/introduction.md
```diff
@@ -2,150 +2,449 @@
...
 
-    The "beep" is your terminal's *bell*. Some terminals may be configured to play different noises or a visual indication of a bell rather than a noise.
+- **`id`** is an identifier we can use to tell the buttons apart in code and apply styles. More on that later.
+- **`variant`** is a string which selects a default style. The "success" variant makes the button green, and the "error" variant makes it red. 
 
-There are two event handlers in this app. Event handlers start with the text `on_` followed by the name of the event in lower case. Hence `on_mount` is called for the `Mount` event, and `on_key` is called for the `Key` event.
+### Composing the widgets
 
-The first event handler to run is `on_mount`. The `Mount` is sent to your application immediately after entering application mode.
+To see our widgets with we first need to yield them from the app's `compose()` method:
```
rodrigogiraoserrao, line 174:
```suggestion
To see our widgets we first need to yield them from the app's `compose()` method:
```

docs/introduction.md
```diff
@@ -2,150 +2,449 @@
...
-    You may have noticed there is no function call to repaint the screen in this example. Textual is generally quite smart in detecting when a refresh is required, and updating the screen automatically.
+!!! info inline end
 
-The second event handler will receive `Key` events whenever you press a key on the keyboard:
+    Don't worry if you have never worked with CSS before. The dialect of CSS we use is greatly simplified over web based CSS and easy to learn!
 
-```python hl_lines="22 23 24 25" title="intro02.py"
---8<-- "docs/examples/introduction/intro02.py"
+    
+While its possible to set all styles for an app this way, it is rarely necessary. Textual has support for CSS (Cascading Style Sheets), a technology used by web browsers. CSS files are data files loaded by your app which contain information about styles to apply to your widgets.
```
rodrigogiraoserrao, line 202:
```suggestion
While it's possible to set all styles for an app this way, it is rarely necessary. Textual has support for CSS (Cascading Style Sheets), a technology used by web browsers. CSS files are data files loaded by your app which contain information about styles to apply to your widgets. 
```

docs/introduction.md
```diff
@@ -2,150 +2,449 @@
...
+While its possible to set all styles for an app this way, it is rarely necessary. Textual has support for CSS (Cascading Style Sheets), a technology used by web browsers. CSS files are data files loaded by your app which contain information about styles to apply to your widgets. 
+
+Let's add a CSS file to our application.
+
+```python title="stopwatch03.py" hl_lines="39"
+--8<-- "docs/examples/introduction/stopwatch03.py"
 ```
 
-This method has an `event` positional argument which will receive the event object; in this case the `Key` event. The body of the method sets the background to a corresponding color in the `COLORS` list when you press one of the digit keys. It also calls `bell()` which is a method on App that plays your terminal's bell.
+Adding the `css_path` attribute to the app constructor tells textual to load the following file when it starts the app:
```
rodrigogiraoserrao, line 210:
```suggestion
Adding the `css_path` attribute to the app constructor tells Textual to load the following file when it starts the app:
```

docs/introduction.md
```diff
@@ -2,150 +2,449 @@
...
+
+The `Button` block sets the width (`width`) of buttons to 16 cells (character widths).
 
-```python title="clock01.py" hl_lines="7 8 9 10 11 12 13"
---8<-- "docs/examples/introduction/clock01.py"
+The last 3 blocks have a slightly different format. When the declaration begins with a `#` then the styles will be applied to widgets with a matching "id" attribute. We've set an ID on the Button widgets we yielded in compose. For instance the first button has `id="start"` which matches `#start` in the CSS.
+
+The buttons have a `dock` style which aligns the widget to a given edge. The start and stop buttons are docked to the left edge, while the reset button is docked to the right edge.
+
+You may have noticed that the stop button (`#stop` in the CSS) has `display: none;`. This tells Textual to not show the button. We do this because we don't want to dsplay the stop button when the timer is *not* running. Similarly we don't want to show the start button when the timer is running. We will cover how to manage such dynamic user interfaces in the next section.
```
rodrigogiraoserrao, line 287:
```suggestion
You may have noticed that the stop button (`#stop` in the CSS) has `display: none;`. This tells Textual to not show the button. We do this because we don't want to display the stop button when the timer is *not* running. Similarly we don't want to show the start button when the timer is running. We will cover how to manage such dynamic user interfaces in the next section.
```

docs/introduction.md
```diff
@@ -2,150 +2,449 @@
...
+
+The rule is `"display: none"` which tells Textual to _hide_ the button.
+
+### Manipulating classes
+
+Modifying a widget's CSS classes it a convenient way to modify visuals without introducing a lot of display related code which tends to be hard to maintain.
+
+You can add and remove CSS classes with the `add_class()` and `remove_class()` methods. We will use these methods to connect the started state to the Start / Stop buttons.
+
+The following code adds a event handler for the `Button.Pressed` event.
```
rodrigogiraoserrao, line 323:
```suggestion
The following code adds an event handler for the `Button.Pressed` event.
```

docs/introduction.md
```diff
@@ -2,150 +2,449 @@
...
+- We add the "started" class when the Stopwatch is started, and remove it when it is stopped. This will update the Stopwatch visuals via CSS.
 
-```python title="clock01.py" hl_lines="17 18"
---8<-- "docs/examples/introduction/clock01.py"
+If you run stopwatch06.py you will be able to use the stopwatches independently.
+
+```{.textual path="docs/examples/introduction/stopwatch06.py" title="stopwatch06.py" press="tab,enter,_,_,tab,enter,_,tab"}
+```
+
+The only remaining feature of the Stopwatch app let to implement is the ability to add and remove timers.
```
rodrigogiraoserrao, line 421:
```suggestion
The only remaining feature of the Stopwatch app left to implement is the ability to add and remove timers.
```

docs/introduction.md
```diff
@@ -2,150 +2,449 @@
...
+
+## Dynamic widgets
+
+It's convenient to build a user interface with the `compose` method. We may also want to add or remove widgets while the app is running.
+
+To add a new child widget call `mount()` on the parent. To remove a widget, call it's `remove()` method.
```
rodrigogiraoserrao, line 427:
```suggestion
To add a new child widget call `mount()` on the parent. To remove a widget, call its `remove()` method.
```

docs/reference/message_pump.md
```diff
@@ -0,0 +1,5 @@
+A message pump is a class that processes messages.
+
+It is a base class for the App, Screen, and Widgets.
```
rodrigogiraoserrao, line 3:
```suggestion
It is a base class for the `App`, `Screen`, and `Widget` classes.
```

docs/styles/height.md
```diff
@@ -2,14 +2,28 @@
...
 
+## Syntax
+
+```
+height: <SCALAR>;
+```
+
 ## Example
 
+This examples applies a widget with a height of 50% of the screen.
```
rodrigogiraoserrao, line 13:
```suggestion
This examples creates a widget with a height of 50% of the screen.
```

docs/styles/margin.md
```diff
@@ -10,14 +10,30 @@ The `margin` rule adds space around the entire widget. Margin may be specified w
 
 Margin may also be set individually by setting `margin-top`, `margin-right`, `margin-bottom`, or `margin-left` to an single value.
```
rodrigogiraoserrao, line 11:
```suggestion
Margin may also be set individually by setting `margin-top`, `margin-right`, `margin-bottom`, or `margin-left` to a single value.
```

docs/styles/margin.md
```diff
@@ -10,14 +10,30 @@ The `margin` rule adds space around the entire widget. Margin may be specified w
...
+
+```
+margin: <INTEGER>;
+margin: <INTEGER> <INTEGER>;
+margin: <INTEGER> <INTEGER> <INTEGER> <INTEGER>;
+```
+
 ## Example
 
+In this example we add a large margin to a some static text.
```
rodrigogiraoserrao, line 23:
```suggestion
In this example we add a large margin to some static text.
```

docs/styles/max_height.md
```diff
@@ -2,6 +2,12 @@
 
 The `max-height` rule sets a maximum width for a widget.
```
rodrigogiraoserrao, line 3:
```suggestion
The `max-height` rule sets a maximum height for a widget.
```

docs/styles/overflow.md
```diff
@@ -12,18 +12,32 @@ The default value for overflow is `"auto auto"` which will show scrollbars autom
...
+overflow: [auto|hidden|scroll];
+overflow-x: [auto|hidden|scroll];
+overflow-y: [auto|hidden|scroll];
+```
+
 ## Example
 
 Here we split the screen in to left and right sections, each with three vertically scrolling widgets that do not fit in to the height of the terminal.
 
 The left side has `overflow-y: auto` (the default) and will automatically show a scrollbar. The right side has `overflow-y: hidden` which will prevent a scrollbar from being show.
```
rodrigogiraoserrao, line 27:
```suggestion
The left side has `overflow-y: auto` (the default) and will automatically show a scrollbar. The right side has `overflow-y: hidden` which will prevent a scrollbar from being shown.
```

docs/styles/padding.md
```diff
@@ -10,6 +10,14 @@ The padding rule adds space around the content of a widget. You can specify padd
 
 Padding may also be set individually by setting `padding-top`, `padding-right`, `padding-bottom`, or `padding-left` to an single value.
```
rodrigogiraoserrao, line 11:
```suggestion
Padding may also be set individually by setting `padding-top`, `padding-right`, `padding-bottom`, or `padding-left` to a single value.
```

docs/guide/CSS.md
rodrigogiraoserrao: It is hard to strike a balance between informal and precise, but I'm sure it is attainable.

rodrigogiraoserrao:
I went through the `*.md` files and left a bunch of suggestions and questions.
I only looked at the parts of the files that were shown in the diff;
i.e., I did not necessarily read all files in their entirety if they already existed and were _modified_ in this PR.

Also, I'd like to share something I do on my personal projects, when writing Markdown.
I tend to write one sentence per file line, or sometimes even just a phrase per line.
This does make the file larger in number of lines, but:

 1. it makes it easier to find a specific passage in the source file;
 2. makes it _much_ easier to read diffs; and
 3. simplifies the process of making edit suggestions.

docs/guide/CSS.md
willmcgugan: Depends wether I'm refering to the concept of widgets or the Widget class. It's not clear sometimes!

docs/guide/CSS.md
willmcgugan: They don't quite mean the same thing. Pseudo classes are the attributes on the widget, pseudo selectors are the CSS syntax.

docs/guide/CSS.md
willmcgugan: If its as space then it is "some parent" It won't match a parent with both. THere's diagrams in the doc that might help clarify. I'll do another pass on this doc eventually.

docs/introduction.md
willmcgugan: George will be turning in his grave.

docs/introduction.md
willmcgugan: If its referring to the class, then it probably should be backticked.

src/textual/timer.py
```diff
@@ -25,6 +33,19 @@ class EventTargetGone(Exception):
...
 class Timer:
+    """A class to send timer-based events.
+
+    Args:
+        event_target (MessageTarget): The object which will receive the timer events.
+        interval (float): The time between timer events.
+        sender (MessageTarget): The sender of the event.
+        name (str | None, optional): A name to assign the event (for debugging). Defaults to None.
+        callback (TimerCallback | None, optional): A optional callback to invoke when the event is handled. Defaults to None.
+        repeat (int | None, optional): The number of times to repeat the timer, or None for no repeat. Defaults to None.
```
darrenburns, line 44:
```suggestion
        repeat (int | None, optional): The number of times to repeat the timer, or None to repeat forever. Defaults to None.
```

src/textual/widget.py
```diff
@@ -144,16 +150,54 @@ def __init__(
...
+        May be overridden if you want different logic regarding allowing scrolling.
+
+        Returns:
+            bool: True if the widget may scroll _horizontally_.
+        """
         return self.is_scrollable and self.show_horizontal_scrollbar
 
+    @property
+    def _allow_scroll(self) -> bool:
+        """Check if both axis may be scrolled.
```
darrenburns, line 192:
```suggestion
        """Check if both axes may be scrolled.
```

src/textual/widget.py
```diff
@@ -551,20 +655,31 @@ def virtual_region_with_margin(self) -> Region:
...
+
+        """
         focusable = [
             child for child in self.children if child.display and child.visible
         ]
         return sorted(focusable, key=attrgetter("_focus_sort_key"))
 
     @property
     def _focus_sort_key(self) -> tuple[int, int]:
+        """Key function to sort widgets in to tfocus order."""
```
darrenburns, line 671:
```suggestion
        """Key function to sort widgets in to focus order."""
```

darrenburns (changes requested):

docs/events/mouse_scroll_down.md
```diff
@@ -0,0 +1,12 @@
+# MouseScrollDown
+
+The `MouseScrollDown` event is sent to a widget when the scroll wheel (or trackpad equivalent) is moved _down_.
```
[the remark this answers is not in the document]
darrenburns: Don't think the second paragraph is relevant since your change this afternoon.

docs/guide/CSS.md
rodrigogiraoserrao: I'm sure there are nuances, I'm just suggesting to make an analogy here because the two things look similar.

docs/events/mouse_scroll_down.md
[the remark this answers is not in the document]
willmcgugan: Good point are platform. Will look at that in another update.

docs/events/screen_resume.md
```diff
@@ -0,0 +1,9 @@
+# ScreenResume
+
+The `ScreenResume` event is sent to a **Screen** when it becomes current.
```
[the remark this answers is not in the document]
willmcgugan: Yeah

docs/examples/demo.py
```diff
@@ -0,0 +1,235 @@
...
+            Tweet(TweetBody()),
+            Widget(
+                Static(
+                    Syntax(CODE, "python", line_numbers=True, indent_guides=True),
+                    classes="code",
+                ),
+                classes="scrollable",
+            ),
+            table,
+            Widget(DirectoryTree("~/projects/textual"), id="tree-container"),
```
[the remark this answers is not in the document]
willmcgugan: Aha. Probably simplest to remove that. Will have some better demos for docs at some point.

docs/guide/CSS.md
```diff
@@ -0,0 +1,395 @@
+# Textual CSS
+
+Textual uses CSS to apply style to widgets. If you have any exposure to web development you will have encountered CSS, but don't worry if you haven't: this section will get you up to speed.
```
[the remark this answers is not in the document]
willmcgugan: Don't think this is the right place to do that. Muddys the water a bit. The reader will be expecting a description of what the difference is.

docs/guide/CSS.md
```diff
@@ -0,0 +1,395 @@
...
+Header {
+  dock: top;
+  height: 3;
+  content-align: center middle;
+  background: blue;
+  color: white;
+}
+```
+
+The first line is a _selector_ which tells Textual which Widget(s) to modify. In the above example, the styles will be applied to a widget defined in the Python class `Header`.
```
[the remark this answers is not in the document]
willmcgugan: "widgets named Header` isn't clear to me. Where is the name defined, is it an attribute? I want to say it is literally a Python class name. i.e:

```python
class Header:
    pass
```

docs/guide/CSS.md
```diff
@@ -0,0 +1,395 @@
...
+
+```css hl_lines="2 3 4 5 6"
```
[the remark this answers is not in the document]
willmcgugan: You can. I discovered that some time after I wrote this.

docs/guide/CSS.md
```diff
@@ -0,0 +1,395 @@
...
+
+The first rule in the above example reads `"dock: top;"`. The rule name is `dock` which tells Textual to place the widget on an edge of the screen. The text after the colon is `top` which tells Textual to dock to the _top_ of the screen. Other valid values for dock are "right", "bottom", or "left"; but `top` is naturally appropriate for a header.
+
+You may be able to guess what some of the the other rules do. We will cover those later.
+
+## The DOM
+
+The DOM, or _Document Object Model_, is a term borrowed from the web world. Textual doesn't use documents but the term has stuck. In Textual CSS, the DOM is a an arrangement of widgets you can visualize as a tree-like structure.
+
+Some widgets contain other widgets: for instance, a list control widget will likely also have item widgets, or a dialog widget may contain button widgets. These _child_ widgets form the branches of the tree.
```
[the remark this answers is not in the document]
willmcgugan: I'm trying to explain why its a tree for those without a background in computer science. Trees have branches not nodes...

docs/guide/CSS.md
```diff
@@ -0,0 +1,395 @@
...
+
+When you run this code you will have an instance of an `ExampleApp` in memory. This app class will also create a `Screen` object. In DOM terms, the Screen is a _child_ of the app.
+
+With the above example, the DOM will look like the following:
+
+<div class="excalidraw">
+--8<-- "docs/images/dom1.excalidraw.svg"
+</div>
+
+This doesn't look much like a tree yet. Let's add a header and a footer to this application, which will create more _branches_ of the tree:
```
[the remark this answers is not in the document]
willmcgugan: Nope. "branches"

docs/guide/CSS.md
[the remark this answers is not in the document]
darrenburns: Fair enough, but "a widget defined in the Python class `Header`" is just confusing. What does "in" mean here?

docs/guide/CSS.md
```diff
@@ -0,0 +1,395 @@
...
+Button:hover {
+  background: green;
+}
+```
+
+The `background: green` is only applied to the Button underneath the mouse cursor. When you move the cursor away from the button it will return to its previous background color.
+
+Here are some other such pseudo classes:
+
+- `:focus` Matches widgets which have input focus.
```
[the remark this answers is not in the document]
willmcgugan: Probably "input focus". Might make it clearer what focus is to the uninitiated.

docs/guide/CSS.md
```diff
@@ -0,0 +1,395 @@
...
+- `:focus-within` Matches widgets with a focused a child widget.
+
+## Combinators
+
+More sophisticated selectors can be created by combining simple selectors. The rule that combines selectors is know as a _combinator_.
```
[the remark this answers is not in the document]
willmcgugan: Probably not "operator" though. Too computer sciencey.

docs/guide/CSS.md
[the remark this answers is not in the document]
willmcgugan: Defined "by" the Python class Header?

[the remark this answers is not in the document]
darrenburns: That sounds better

docs/index.md
```diff
@@ -1,11 +1,76 @@
...
-Welcome to the [Textual](https://github.com/Textualize/textual) framework documentation, built with ❤️ by [Textualize.io](https://www.textualize.io)
+Welcome to the [Textual](https://github.com/Textualize/textual) framework documentation. Built with ❤️ by [Textualize.io](https://www.textualize.io)
 
-## Getting started
+<hr>
 
-Textual is a Python framework which you can install via Pypi.
+Textual is a framework for building applications that run within your terminal. Such Text User Interfaces (TUIs) have a number of advantages over traditional web and desktop apps.
+
+<div class="grid cards" markdown>
```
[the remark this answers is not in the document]
willmcgugan: That's one of the additions in the Insider addition of mkdocs-material. Here's what it looks like on my machine:

Screenshot 2022-08-26 at 10 42 52

Will get this installed on your machine when your back in the office.

docs/introduction.md
```diff
@@ -2,150 +2,449 @@
...
+Let's see what happens when we run "stopwatch02.py".
+
+```{.textual path="docs/examples/introduction/stopwatch02.py" title="stopwatch02.py"}
 ```
 
-This `on_mount` method sets the `background` attribute of `self.styles` to `"darkblue"` which makes the background blue when the application starts. There are a lot of other properties on the Styles object, which define how your app looks. We will explore what you can do with this object later.
+The elements of the stopwatch application are there. The buttons are clickable and you can scroll the container but it doesn't look like the sketch. This is because we have yet to apply any _styles_ to our new widgets.
 
-!!! note
+## Writing Textual CSS
```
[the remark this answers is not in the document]
willmcgugan: Don't follow. Why is it jarring?

docs/introduction.md
```diff
@@ -2,150 +2,449 @@
...
+
+Every widget has a `styles` object with a number of attributes that impact how the widget will appear. Here's how you might set white text and a blue background for a widget:
```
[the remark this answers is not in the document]
willmcgugan: Doesn't read right to me.

docs/introduction.md
```diff
@@ -2,150 +2,449 @@
...
+
+!!! info
+
+    The `time` attribute is created with `Reactive.init` which calls watch methods when the widget is mounted. See below for an explanation of watch methods.
+
+In the `on_mount` method the call to `set_interval` creates a timer object which runs `self.update_time` sixty times a second. This `update_time` method calculates the time elapsed since the widget started and assigns it to `self.time`. Which brings us to one of Reactive's super-powers.
+
+If you implement a method that begins with `watch_` followed by the name of a reactive attribute (making it a _watch method_), that method will be called when the attribute is modified.
+
+Because `watch_time` watches the `time` attribute, when we update `self.time` 60 times a second we also implicitly call `watch_time` which converts the elapsed time in to a string and updates the widget with a call to `self.update`.
```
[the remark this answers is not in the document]
willmcgugan: I concur, but your version doesn't read much better for me.

src/textual/app.py
```diff
@@ -824,6 +923,19 @@ def set_focus(self, widget: Widget | None) -> None:
                 widget.post_message_no_wait(events.Focus(self))
                 widget.emit_no_wait(events.DescendantFocus(self))
 
+    def _reset_focus(self, widget: Widget) -> None:
+        """Reset the focus when a widget is removed
+
+        Args:
+            widget (Widget): A widget that is removed.
+        """
+        for sibling in widget.siblings:
```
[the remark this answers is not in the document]
willmcgugan: I don't think that is the right thing to do, although its not clear what the right thing to do here is. It produce the right behaviour for the stopwatch example.

src/textual/css/stylesheet.py
```diff
@@ -320,13 +320,16 @@ def apply(self, node: DOMNode, animate: bool = False) -> None:
...
+        # I think we can pre-calculate which rules may be applicable to a given node
+
         # Dictionary of rule attribute names e.g. "text_background" to list of tuples.
         # The tuples contain the rule specificity, and the value for that rule.
         # We can use this to determine, for a given rule, whether we should apply it
         # or not by examining the specificity. If we have two rules for the
         # same attribute, then we can choose the most specific rule and use that.
         rule_attributes: dict[str, list[tuple[Specificity6, object]]]
-        rule_attributes = defaultdict(list)
+        rule_attributes = {}
```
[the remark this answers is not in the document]
willmcgugan: Yes! Turns out setdefault is slower. I think because every call will create a new blank list. defaultdict avoids that.

src/textual/reactive.py
```diff
@@ -40,11 +39,55 @@ def __init__(
         *,
         layout: bool = False,
         repaint: bool = True,
+        init: bool = False,
```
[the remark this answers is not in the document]
willmcgugan: I think "call_watchers_on_initialization" would be most apt. But too verbose.

I also wanted to keep it compact from the devs POV:

```python
Reactive.init("foo")
```

docs/events/mouse_release.md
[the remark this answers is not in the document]
willmcgugan: It's related to mouse capture. You "capture" the mouse to get all mouse events even outside of the widgets borders, then "release" the mouse when you're done. There is precedence in other UI libs.

docs/guide/CSS.md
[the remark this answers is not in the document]
darrenburns: Fair enough - I didn't realise this guide was aimed at people with no experience of data structures like trees. Just not a fan of trying to define our own terminology for things like this as, especially since "branch" already has a totally different meaning in the context of the data structure.

docs/index.md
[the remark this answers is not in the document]
darrenburns: Ah cool, looks nice

docs/introduction.md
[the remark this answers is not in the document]
darrenburns: Yeah, you're right

docs/introduction.md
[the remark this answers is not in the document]
darrenburns: Because the title is "Writing Textual CSS", and the first paragraph has nothing to do with CSS

src/textual/css/stylesheet.py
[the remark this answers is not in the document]
darrenburns: This PR moved away from `defaultdict` and to `setdefault` though 😄

PR #704, opened 2022-08-25
`text-align` CSS support

darrenburns:
image


src/textual/widget.py
```diff
@@ -982,11 +982,13 @@ def post_render(self, renderable: RenderableType) -> RenderableType:
         """
 
         if isinstance(renderable, str):
-            renderable = Text.from_markup(renderable)
+            renderable = Text.from_markup(renderable, justify=self.styles.text_justify)
 
         rich_style = self.rich_style
         if isinstance(renderable, Text):
             renderable.stylize(rich_style)
+            if not renderable.justify:
```
darrenburns, line 990: Wasn't sure whether to do this `if` or not...

src/textual/widget.py
```diff
@@ -1580,3 +1577,22 @@ def key_pageup(self) -> bool:
...
+
+    Returns:
+        JustifyMethod: The Rich JustifyMethod that corresponds to the text-align
+            value
+    """
+    assert css_align in VALID_TEXT_ALIGN
+    return {
+        "start": "left",
+        "end": "right",
+        "justify": "full",
```
darrenburns, line 1597: Since Textual doesn't support RTL, we can just directly map start -> left and so on.

tests/test_auto_refresh.py
```diff
@@ -25,4 +25,4 @@ def test_auto_refresh():
     elapsed = app.run(quit_after=1, headless=True)
     assert elapsed is not None
     # CI can run slower, so we need to give this a bit of margin
-    assert elapsed >= 0.3 and elapsed < 0.6
+    assert 0.3 <= elapsed < 0.8
```
darrenburns, line 28: Got some failures where elapsed was a bit greater than 0.6

src/textual/widget.py
willmcgugan: Seems reasonable...

willmcgugan (approved):

PR #705, opened 2022-08-25
Optimize applying CSS styles

willmcgugan:
Optimizes applying CSS styles.

- Only applies CSS rules to visible widgets on mouse hover change
- Limits rules applied to Node 
- Adds an early-out for rules


willmcgugan:
@darrenburns Knock yourself out

darrenburns (approved):

PR #715, opened 2022-08-31
fix: 'old example' -> old_example

Abdur-rahmaanJ:


willmcgugan:
That entire directory is going to be deleted.

Abdur-rahmaanJ:
@willmcgugan ok but in the meantime the space is tripping for contributors who want to view examples

willmcgugan:
Tripping? In what way?

Abdur-rahmaanJ:
@willmcgugan cd into old examples

willmcgugan:
```
cd old\ examples/
```

or 

```
cd "old examples"
```

But regardless, the old examples don't work and will be removed very soon...

willmcgugan:
Thanks anyway

Abdur-rahmaanJ:
@willmcgugan Yes, they don't work. Yes you can but like people have to type twice as it is not expected.

PR #720, opened 2022-08-31
Changes to opacity

darrenburns:
* Rename `opacity` to `text-opacity`
* Add reference docs for `text-opacity`
* Adds new `opacity` property for blending whole widget into background
* Docs for new opacity

Example of a green background widget fading into a red background:

image

I thought it looked a little weird, but confirmed it matches what Photoshop does (main picture is simulating the scenario in PS using same colours, and snippet at bottom right is from Textual itself).

opacity-test

Chose slightly less jarring colours for the docs example output though:

image


src/textual/_opacity.py
```diff
@@ -0,0 +1,46 @@
...
+    _Segment = Segment
+    for segment in segments:
+        text, style, _ = segment
+        if not style:
+            yield segment
+            continue
+
+        blended_style = style
+        if style.color:
+            color = Color.from_rich_color(style.color)
```
willmcgugan, line 33: Can we hoist `from_rich_color` and `from_color` out of the loop? As a tight loop, it can make a real difference!

darrenburns: Grabbed an early reference to 3 methods

PR #724, opened 2022-09-01
CSS Path Relative

darrenburns:
`css_path` is taken to be relative from the app location.

### DIFF
--- a/src/textual/file_monitor.py
+++ b/src/textual/file_monitor.py
@@ -1,4 +1,3 @@
-from __future__ import annotations
 import os
 from pathlib import PurePath
 from typing import Callable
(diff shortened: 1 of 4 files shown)


src/textual/file_monitor.py
```diff
@@ -12,7 +11,7 @@
 class FileMonitor:
     """Monitors a file for changes and invokes a callback when it does."""
 
-    def __init__(self, path: str | PurePath, callback: Callable) -> None:
+    def __init__(self, path: PurePath, callback: Callable) -> None:
```
darrenburns, line 14: Is this class going to remain part of the public API? If so, I can change this back to give users flexibility.

If it's just going to be internal though I think we should convert all str | PurePath at the API boundary so we have less propagation of unions like this.

darrenburns:
@willmcgugan hold fire on reviewing this, it's not working with `textual run`...

darrenburns:
Seems like it's not an issue with this PR, it's happening for me on `css` branch too. `python just_a_box.py` loads the CSS, `textual run just_a_box.py` doesn't.

src/textual/file_monitor.py
willmcgugan: This probably should be private. I agree about reducing unions.

willmcgugan (approved):

PR #730, opened 2022-09-02
validate identifiers

willmcgugan:
Perform validation on node id and class names, so that the Python data can match CSS syntax.

### DIFF
--- a/src/textual/dom.py
+++ b/src/textual/dom.py
@@ -111,7 +111,6 @@ class DOMNode(MessagePump):
 
         _classes = classes.split() if classes else []
         check_identifiers("class name", *_classes)
-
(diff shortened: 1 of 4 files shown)


src/textual/dom.py
```diff
@@ -75,9 +103,17 @@ def __init__(
         id: str | None = None,
         classes: str | None = None,
     ) -> None:
+        self._classes = set()
         self._name = name
-        self._id = id
-        self._classes: set[str] = set() if classes is None else set(classes.split())
+        self._id = None
```
willmcgugan, line 108: Ignoring `name` for now. We don't have support in CSS for querying names,

darrenburns (approved):

PR #732, opened 2022-09-04
added char attribute to Text event

willmcgugan:
Adds a `char` attribute to the Key event, which will contain a character for simple keys. Note that `char` may not always be printable.


willmcgugan:
Taking the liberty of merging. @darrenburns please do a retroactive review.

PR #744, opened 2022-09-07
tree fix

willmcgugan:
- Updates the tree control to the new CSS API.
- Theme colors
- Scrolling for page up / down home / end

https://user-images.githubusercontent.com/554369/188905647-fca8010a-95db-481c-b164-e123b4046b2f.mov

### DIFF
--- a/src/textual/message_pump.py
+++ b/src/textual/message_pump.py
@@ -522,10 +522,12 @@ class MessagePump(metaclass=MessagePumpMeta):
         Args:
             event (events.Key): A key event.
         """
-        key_method = getattr(self, f"key_{event.key_name}", None)
+        key_method = getattr(self, f"key_{event.key_name}", None) or getattr(
+            self, f"_key_{event.key_name}", None
+        )
         if key_method is not None:
-            if await invoke(key_method, event):
-                event.prevent_default()
+            await invoke(key_method, event)
+            event.prevent_default()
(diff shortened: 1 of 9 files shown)


src/textual/message_pump.py
```diff
@@ -522,10 +522,12 @@ async def dispatch_key(self, event: events.Key) -> None:
...
             event (events.Key): A key event.
         """
-        key_method = getattr(self, f"key_{event.key_name}", None)
+        key_method = getattr(self, f"key_{event.key_name}", None) or getattr(
+            self, f"_key_{event.key_name}", None
+        )
         if key_method is not None:
-            if await invoke(key_method, event):
-                event.prevent_default()
+            await invoke(key_method, event)
```
willmcgugan, line 529: Sight change in semantics here. If there is a `key_` method defined for a given key then it won't attempt to try the base classes.

src/textual/widget.py
```diff
@@ -1080,7 +1093,7 @@ def scroll_to_region(
             self.scroll_relative(
                 delta.x or None,
                 delta.y or None,
-                animate=animate,
+                animate=animate if abs(delta_y) > 1 else False,
```
willmcgugan: Don't animate movement of a single line.

darrenburns (approved):

PR #746, opened 2022-09-08
Code browser example

willmcgugan:
Adds code_browser example and related fixes

https://user-images.githubusercontent.com/554369/189101351-256b79e4-9c9d-4807-899a-b694108e5cc8.mov

### DIFF
--- a/examples/code_browser.py
+++ b/examples/code_browser.py
@@ -31,9 +31,15 @@ class CodeBrowser(App):
     def on_directory_tree_file_click(self, event: DirectoryTree.FileClick) -> None:
         code_view = self.query_one("#code", Static)
         try:
-            syntax = Syntax.from_path(event.path, line_numbers=True, word_wrap=True)
+            syntax = Syntax.from_path(
+                event.path,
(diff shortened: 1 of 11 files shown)


examples/code_browser.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def watch_show_tree(self, show_tree: bool) -> None:
+        self.set_class(show_tree, "-show-tree")
+
+    def on_load(self) -> None:
+        self.bind("t", "toggle_tree", description="Toggle Tree")
+        self.bind("q", "quit", description="Quit")
+
+    def compose(self) -> ComposeResult:
+        path = "./" if len(sys.argv) < 2 else sys.argv[1]
```
darrenburns, line 23: If you run this with `textual run examples/code_browser.py`, it fails with `FileNotFoundError: [Errno 2] No such file or directory: 'run'`

willmcgugan: Works now, but you will need to wrap the args in quotes. i.e. `textual run "code_browser ../"`

PR #747, opened 2022-09-08
Scroll issue

willmcgugan:
- Makes scroll to visible the default
- Prevents an issue where clicking a window would scroll it then send a click somewhere other than where you clicked


willmcgugan:
Taking the liberty of merging. It's a "take my word for it" kind of thing anyway.

PR #748, opened 2022-09-08
Initial layout guide stuff

darrenburns:


docs/guide/layout.md
```diff
@@ -1,68 +1,140 @@
...
-There probably isn't a practical use for such overlapping widgets. In practice this layout is probably only useful where you have a single child widget.
+=== "Output"
+
+    ```{.textual path="docs/examples/guide/horizontal_layout.py"}
+    ```
+
+We've changed the `layout` to `horizontal` inside our CSS file, and set the child `.box` widgets to 100% height to ensure they take up the entire height of the screen.
+As a result, the widgets are now arranged from left to right instead of top to bottom.
+
+TODO: ??? Do we draw attention to the fact that we had to set height 100%, for the widgets to fill the height of their parent? This wasn't required in the vertical example, so might be worth explaining why it's required here.
```
willmcgugan, line 78: Yes, I think so.

docs/guide/layout.md
```diff
@@ -72,7 +144,11 @@ Widgets have a relative offset which is added to the widget's location, after it
 --8<-- "docs/images/layout/offset.excalidraw.svg"
 </div>
 
+TODO: What happens when you check the position of a widget that has offset? Do you get the position inclusive of offset back, or the original position (excluding offset)?
 
-TODO: Diagram
-TODO: Offsets
+## Combining Layouts (????)
 
+TODO: The sections above show layouts in isolation. Maybe we should mention how we can combine multiple layouts in a single app?
```
willmcgugan, line 151: Yes, good idea. A simple example combining layouts would be great.

darrenburns:
@willmcgugan Feel free to review down to "Adjusting cell spans"

willmcgugan (changes requested):
This is great! We'll probably tighten up the prose a little. It's a little wordy in places, particularly in grid (which I know is harder to describe). 

Only one technical thing to mention. The grid-size declaration can take two integer. With two values it sets a fixed columns and rows, and you don't get that auto new row behaviour.

Could you add that detail, and we should be good to merge.

We'll do another pass on all the docs to pick up typos / grammar later.

PR #751, opened 2022-09-08
focus level bindings

willmcgugan:
- Changed to how bindings are specified. Bindings should be defined by a classvar called BINDINGS, which may be a tuple of three values for simple bindings, or a `textual.binding.Binding` where more parameters are required.

```python
class MyApp(App):
    BINDINGS = [
        ("f1", "help", "Launch help"),
        Binding("s", "save", "Save",)
    ]
```

Bindings may also be set on widgets. If that widget is focused, its bindings will take precedence.

https://user-images.githubusercontent.com/554369/189309255-e2bd92dd-f651-42e9-860a-70aec48a46d7.mov

- Tweaked easing example to use more theme colors
- Some micro-optimizations
- Added get_component_rich_style
- Added aliases for Reactive descriptor; `reactive` and `var`
- Updates tutorial and stopwatch example accordingly


src/textual/widget.py
```diff
@@ -1464,6 +1477,9 @@ def release_mouse(self) -> None:
     async def broker_event(self, event_name: str, event: events.Event) -> bool:
         return await self.app._broker_event(event_name, event, default_namespace=self)
 
+    def _on_syles_updated(self) -> None:
```
darrenburns, line 1480: Typo here

darrenburns (changes requested):
One comment regarding a typo

src/textual/widget.py
willmcgugan: Good spot

darrenburns (approved):

PR #755, opened 2022-09-09
Layout order

willmcgugan:
- Fixes the issue with widgets being draw in the reverse order
- Optimizes compositor by caching sorted widgets


willmcgugan:

PR #763, opened 2022-09-12
auto color

willmcgugan:


src/textual/design.py
```diff
@@ -126,6 +129,8 @@ def generate(self) -> dict[str, str]:
         else:
             panel = self.panel
 
+        boost = self.boost or background.get_contrast_text(1.0).with_alpha(0.07)
```
darrenburns, line 132: What's this `$boost` intended to be used for? The name "boost" isn't suggesting anything to me. Is 7% the correct alpha here?

darrenburns (approved):

PR #774, opened 2022-09-14
Minimise rendering when text-opacity/opacity is zero

darrenburns:
* For text-opacity we no longer render text as the same colour as the background. We instead render empty cells of the same cell length. This will also improve the "minimum contrast" situation.
* For opacity, the compositor will ignore widgets with 0 opacity (same as when visibility is "hidden" - by still reserving space for the widget).

### DIFF
--- a/src/textual/_compositor.py
+++ b/src/textual/_compositor.py
@@ -604,19 +604,28 @@ class Compositor:
         # up to this point.
         _rich_traceback_guard = True
 
+        def is_visible(widget: Widget) -> bool:
+            """Return True if the widget is (literally) visible by examining various
+            properties which affect whether it can be seen or not."""
+            return (
+                widget.visible
+                and not widget.is_transparent
+                and widget.styles.opacity > 0
+            )
+
         if self.map:
             if crop:
                 overlaps = crop.overlaps
                 mapped_regions = [
                     (widget, region, order, clip)
                     for widget, (region, order, clip, *_) in self.map.items()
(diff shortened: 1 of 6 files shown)


src/textual/renderables/text_opacity.py
```diff
@@ -67,11 +68,16 @@ def process_segments(
 
             color = style.color
             bgcolor = style.bgcolor
-            if color and color.triplet and bgcolor and bgcolor.triplet:
-                color_style = _get_blended_style_cached(bgcolor, color, opacity)
-                yield _Segment(text, style + color_style)
+
+            if opacity > 0:
```
willmcgugan, line 72: This condition will be constant in the loop. If we move it outside of the loop then we can write an optimized version for the case of opacity == 0

src/textual/_compositor.py
```diff
@@ -604,19 +604,28 @@ def _get_renders(
         # up to this point.
         _rich_traceback_guard = True
 
+        def is_visible(widget: Widget) -> bool:
```
willmcgugan, line 607: This is a fairly tight loop.

Humor me and unroll the code in is_visible. Essentially replace the properties with immediate code, so it makes no implicity function calls.

willmcgugan (changes requested):
Looks good, but I'd like to go the extra mile to squeeze a few more cycles out of this.

darrenburns:
@willmcgugan Made those changes, also fixed a link in the docs.

willmcgugan (approved):

PR #775, opened 2022-09-15
Docs styles

willmcgugan:
- Documents Styles
- Tweaks to app docs
- Changed HLS to HSL in Color class to match browser CSS
- Added "rich" docs extension, used to build the color table in color.py


docs/guide/styles.md
```diff
@@ -1,15 +1,320 @@
...
+
+Widgets will wrap text by default. If you were to replace `"Textual"` with a long paragraph of text, the widget will expand downwards to fit.
+
+## Colors
+
+There are a number of style attribute which accept colors. The most commonly used are [color](../styles/color.md) which sets the default color of text on a widget, and [background](..styles/background/md) which sets the background color (beneath the text).
+
+You can set a color value to one of a number of pre-defined color constants, such as "crimson", "lime", and "palegreen". You can find a full list in the [Color reference](../reference/color.md#textual.color--named-colors).
+
+Here's how you would set the screen background to lime:
```
darrenburns, line 54: The text says "screen", and the example shows an arbitrary widget, suggesting:

```suggestion
Here's how you would set the background of a widget to lime:
```

docs/guide/styles.md
```diff
@@ -1,15 +1,320 @@
...
+```
+
+In addition to color names, you can also use any of the following ways of expressing a color:
+
+- RGB hex colors starts with a `#` followed by three pairs of one or two hex digits; one for the red, green, and blue color components. For example, `#f00` is an intense red color, and `#9932CC` is *dark orchid*.
+- RGB decimal color start with `rgb` followed by a tuple of three numbers in the range 0 to 255. For example `rgb(255,0,0)` is intense red, and `rgb(153,50,204)` is *dark orchid*.
+- HSL colors start with `hsl` followed by a angle between 0 and 360 and two percentage values, representing Hue, Saturation and Lightness. For example `hsl(0,100%,50%)` is intense red and `hsl(280,60%,49%)` is *dark orchid*
+
+
+The background and color styles will also accept a [color][textual.color.Color] object which can be used to create colors dynamically.
```
darrenburns, line 67: A bit nitpicky:
```suggestion
The background and color styles also accept a [Color][textual.color.Color] object which can be used to create colors dynamically.
```

docs/guide/styles.md
```diff
@@ -1,15 +1,320 @@
...
+Here is the output:
+
+```{.textual path="docs/examples/guide/styles/colors01.py"}
+```
+
+### Alpha
+
+Textual represents color internally as a tuple of three values for the red, green, and blue components.
+
+Textual support a common fourth value called *alpha* which can make a color translucent. If you set alpha on a background color, Textual will blend the background with the color beneath it. If you set alpha on the text color, then Textual will blend the text with the background color.
```
darrenburns, line 84:
```suggestion
Textual supports a common fourth value called *alpha* which can make a color translucent. If you set alpha on a background color, Textual will blend the background with the color beneath it. If you set alpha on the text color, then Textual will blend the text with the background color.
```

docs/guide/styles.md
```diff
@@ -1,15 +1,320 @@
...
+```
+
+Notice that an alpha of 0.1 the background almost matches the screen, but at 1.0 it is a solid color. 
+
+```{.textual path="docs/examples/guide/styles/colors02.py"}
+```
+
+## Dimensions
+
+Widgets occupy a rectangular region of the screen, which may be as small as a single character or as large as the screen. Potentially *larger* if [scrolling](#) is enabled.
```
darrenburns, line 105: Take it or leave it: "Potentially larger if scrolling is enabled." doesn't sound right to me as a standalone sentence.

```suggestion
Widgets occupy a rectangular region of the screen, which may be as small as a single character or as large as the screen. If [scrolling](#) is enabled, widgets can be even larger than the screen.
```

docs/guide/styles.md
```diff
@@ -1,15 +1,320 @@
...
+
+```{.textual path="docs/examples/guide/styles/dimensions02.py"}
+```
+
+#### Units
+
+Textual offers a few different *units* which allow you to specify dimensions relative to the screen or container. Relative units can better make use of available space if the user resizes the terminal.
+
+- Percentage units are given as a string containing a number followed by a percentage symbol, e.g. `"50%"`. Setting a dimension to a percentage unit will cause it to fit in that percentage of the available space. For instance, setting width to `"50%"` will cause the width of the widget to be half the available space.
+- View units are similar to percentage units, but explicitly reference a dimension. The `vw` unit sets a dimension to a percentage of the terminal *width*, and `vh` sets a dimension to a percentage of the terminal *height*.
```
darrenburns, line 158: Nicely put

docs/guide/styles.md
```diff
@@ -1,15 +1,320 @@
...
+- The `w` unit sets a dimension to a percentage of the available width (which may be smaller than the terminal size if the widget is within another widget).
```
darrenburns, line 159: Appreciate the text in the brackets here. I was like "but how does that differ from the bullet above" and my question was immediately answered!

docs/guide/styles.md
```diff
@@ -1,15 +1,320 @@
...
+
+    ```{.textual path="docs/examples/guide/styles/dimensions03.py" columns="80" lines="30"}
+    ```
+
+=== "120 x 40"
+
+    ```{.textual path="docs/examples/guide/styles/dimensions03.py" columns="120" lines="40"}
+    ```
+
+Percentage units are useful for widgets that occupy a relative portion of the screen, but it can be problematic for some proportions. For instance if we want to divide the screen in to thirds, we would have to set a dimension to `33.3333333333%` which is awkward. Textual supports `fr` units which is often better than percentage based units for these situations.
```
darrenburns, line 187: Felt off so Grammarly'd it 😄 take it or leave it:
```suggestion
Percentage units are useful for widgets that occupy a relative portion of the screen, but they can be problematic for some proportions. For instance, if we want to divide the screen into thirds, we would have to set a dimension to `33.3333333333%` which is awkward. Textual supports `fr` units which are often better than percentage-based units for these situations.
```

docs/guide/styles.md
```diff
@@ -1,15 +1,320 @@
...
+
+When specifying `fr` units for a given dimensions, Textual will divide the available space by the total `fr` units on a dimension. That space will then be divided amongst the widgets as a proportion of their individual `fr` value.
```
darrenburns, line 189:
```suggestion
When specifying `fr` units for a given dimension, Textual will divide the available space by the sum of the `fr` units on that dimension. That space will then be divided amongst the widgets as a proportion of their individual `fr` values.
```

docs/guide/styles.md
```diff
@@ -1,15 +1,320 @@
...
+
+Let's look at an example. We will create two widgets, one with a height of `"2fr"` and one with a height of `"1fr"`.
+
+```python title="dimensions04.py" hl_lines="24-25"
+--8<-- "docs/examples/guide/styles/dimensions02.py"
```
darrenburns, line 194: The wrong code snippet was used here.

```suggestion
--8<-- "docs/examples/guide/styles/dimensions04.py"
```

docs/guide/styles.md
```diff
@@ -1,15 +1,320 @@
...
+```python title="padding02.py" hl_lines="22"
+--8<-- "docs/examples/guide/styles/padding02.py"
+```
+
+Compare the output of this example to the previous example:
+
+```{.textual path="docs/examples/guide/styles/padding02.py"}
+```
+
+You can also set padding to a tuple of four values which applies padding to each edge individually. The first value it the padding for the top of the widget, followed by the right of the widget, then bottom, then left.
```
darrenburns, line 235:
```suggestion
You can also set padding to a tuple of four values which applies padding to each edge individually. The first value is the padding for the top of the widget, followed by the right of the widget, then bottom, then left.
```

docs/guide/styles.md
```diff
@@ -1,15 +1,320 @@
...
+There are many other border types. Run the following from the command prompt to preview them.
+
+
+```bash
+textual borders
+```
+
+### Outline
+
+[Outline](../styles/outline.md) is similar to border and is set in the same way. The difference is that outline will not change the size of the widget, and may overlap the content area. The following examples sets an outline on a widget:
```
darrenburns, line 261:
```suggestion
[Outline](../styles/outline.md) is similar to border and is set in the same way. The difference is that outline will not change the size of the widget, and may overlap the content area. The following example sets an outline on a widget:
```

docs/guide/styles.md
```diff
@@ -1,15 +1,320 @@
...
+```
+
+Notice how the outline overlaps the text in the widget.
+
+```{.textual path="docs/examples/guide/styles/outline01.py"}
+```
+
+Outline can be useful to emphasize a widget, but be mindful that it may obscure your content.
+
+### Box sizing
```
darrenburns, line 274: I remember having difficulty grokking the box-sizing description in the styles reference, but this section is great. Very well explained.

docs/guide/styles.md
```diff
@@ -1,15 +1,320 @@
...
+
+When you set padding or border it reduces the size of the widget's content area. In other words, setting padding or border won't change the width or height of the widget.
+
+This is generally desirable when you arrange things on screen as you can add border or padding without breaking your layout. Occasionally though you may want to keep the size of the content area constant and grow the size of the widget to fit padding and border. The [box-sizing](../styles/box_sizing.md) style allows you to switch between these two modes.
+
+If you set `box_sizing` to `"content-box"` then space required for padding and border will be added to the widget dimensions. The default value of `box_sizing` is `"border-box"`. Compare the box model diagram for content box to the previous box model diagram.
```
darrenburns, line 280: Only thing I'd say here is "the previous box model diagram" is so far back up the page I'd almost forgotten about it.
When I read this my instinct is to want to compare them. I'd probably like to see both of them side by side, or at least in tabs that I can click between instead of having to scroll away back up to spot the difference.

docs/guide/styles.md
```diff
@@ -1,15 +1,320 @@
...
+
+The following example creates two widgets which have a width of 30, a height of 6, and a border and padding of 1. The second widget has `box_sizing` set to `"content-box"`.
+
+```python title="box_sizing01.py" hl_lines="33"
+--8<-- "docs/examples/guide/styles/box_sizing01.py"
+```
+
+The padding and border of the first widget is subtracted from the height leaving only 2 lines in the content area. The second widget also has a height of 6, but the padding and border adds additional height so that the content area remains 6 lines. 
+
+```{.textual path="docs/examples/guide/styles/box_sizing01.py"}
```
darrenburns, line 294: Aside/suggestion: The code snippet here:
```python
        self.widget1.styles.background = "purple"
        self.widget2.styles.background = "darkgreen"
        self.widget1.styles.width = 30
        self.widget2.styles.width = 30
        self.widget2.styles.height = 6
        self.widget1.styles.height = 6
        self.widget2.styles.height = 6
        self.widget1.styles.border = ("heavy", "white")
        self.widget2.styles.border = ("heavy", "white")
        self.widget1.styles.padding = 1
        self.widget2.styles.padding = 1
        self.widget2.styles.box_sizing = "content-box"
```

Makes me wish for an API like:

```python
self.widget1.styles.update(
    background="purple",
    width=30,
    height=6,
    padding=1,
)
```

Inline styles are definitely more verbose than they need to be atm, which is good for selling CSS, but I think both stories can be great 😄

docs/guide/styles.md
```diff
@@ -1,15 +1,320 @@
...
+```
+
+Notice how each widget has an additional two rows and columns around the border.
+
+```{.textual path="docs/examples/guide/styles/margin01.py"}
+```
+
+!!! note
+
+    In the above example both widgets have a margin of 2, but there are only 2 lines of space between the widgets. This is because margins of consecutive widgets *overlap*. In other words when there are two widgets next to each other Textual picks the greater of the two margins.
```
darrenburns, line 314: Slightly embarrassed to admit I didn't know this 😄

darrenburns (changes requested):
Looks good, comments are pretty minor. I did make some typo/grammar suggestions, but have used suggestion blocks so should be quick to accept/decline.

docs/guide/styles.md
willmcgugan: Updated the example. Wanted to stick to screen for now.

docs/guide/styles.md
willmcgugan: I'll add a link

darrenburns (approved):

PR #793, opened 2022-09-20
Snapshot testing

darrenburns:
Adds snapshot testing features to the test suite, and a set of layout-related example tests. We can craft better example apps to do snapshots on in the future.

### Basic view:

image

### After toggling the "show difference" switch:

image

### Link to the report will be written to your terminal in the event of a snapshot mismatch:

image

### If you're running in CI, it'll be saved as a build artifact which you can download and view:

image


.github/workflows/pythonpackage.yml
```diff
@@ -43,3 +43,9 @@ jobs:
         run: |
           source $VENV
           python e2e_tests/sandbox_basic_test.py basic 2.0
+      - name: Upload snapshot report
+        if: always()
+        uses: actions/upload-artifact@v3
+        with:
+          name: snapshot-report-textual
+          path: tests/snapshot_tests/output/snapshot_report.html
```
darrenburns, line 51: This runs on both failure and success. On success, there's no snapshot report, but the step still succeeds.

Makefile
```diff
@@ -2,6 +2,8 @@ test:
 	pytest --cov-report term-missing --cov=textual tests/ -vv
 unit-test:
 	pytest --cov-report term-missing --cov=textual tests/ -vv -m "not integration_test"
+test-snapshot-update:
+	pytest --cov-report term-missing --cov=textual tests/ -vv --snapshot-update
```
darrenburns, line 6: Since this runs all tests, it'll update all snapshots.

notes/snapshot_testing.md
```diff
@@ -0,0 +1,43 @@
+# Snapshot Testing
```
darrenburns, line 1: Internal docs

pyproject.toml
```diff
@@ -48,13 +48,15 @@ pre-commit = "^2.13.0"
 pytest-aiohttp = "^1.0.4"
 time-machine = "^2.6.0"
 Jinja2 = "<3.1.0"
+syrupy = "^3.0.0"
```
darrenburns, line 51: This is the snapshot testing library we're wrapping which handles associating pytest test functions with snapshots on disk.

pyproject.toml
```diff
@@ -48,13 +48,15 @@ pre-commit = "^2.13.0"
...
 
 [tool.black]
 includes = "src"
 
 [tool.pytest.ini_options]
 asyncio_mode = "auto"
 testpaths = ["tests"]
+addopts = "--strict-markers"
```
darrenburns, line 59: IMO this should be the default - if you use an undefined marker it'll fail fast - good for catching typos.

tests/snapshot_tests/test_snapshots.py
```diff
@@ -0,0 +1,30 @@
+def test_grid_layout_basic(snap_compare):
```
darrenburns, line 1: A selection of layout related tests to get us started.

willmcgugan (approved):

PR #797, opened 2022-09-23
Fix comma separated string in "Bindings"

darrenburns:


willmcgugan (changes requested):
I'm thinking that `Binding` should be immutable. So nobody is tempted to change it and have it do something sane.

Making the Binding a frozen dataclass is probably all we need.

willmcgugan (approved):

PR #832, opened 2022-10-05
Note timers ID added to Container object

kimvanwyk:
Noting that the "timers" ID has been added to the Container object in the final example.


willmcgugan:
Thanks, Kim

PR #833, opened 2022-10-05
 Highlight changed set_interval line in stopwatch06 example

kimvanwyk:
Highlight set_interval line in stopwatch06 example as it grew a paused argument


willmcgugan:
Thanks

PR #841, opened 2022-10-06
fix error on exit from textual console

willmcgugan:
- Fixes traceback on exit from `textual console`
- Makes the textual app runable via `python textual.cli`


darrenburns (approved):
Fixes it for me 👍

PR #844, opened 2022-10-06
Fix minor readability issue in modifying CSS class

driscollis:


docs/tutorial.md
```diff
@@ -320,7 +320,7 @@ The rule is `"display: none"` which tells Textual to hide the button.
 
 ### Manipulating classes
 
-Modifying a widget's CSS classes it a convenient way to modify visuals without introducing a lot of messy display related code.
+Modifying a widget's CSS class is a convenient way to modify visuals without introducing a lot of messy display related code.
```
willmcgugan, line 323: There are multiple classes, so it should be plural. But good spot the grammar is off!

```suggestion
Modifying a widget's CSS classes is a convenient way to update visuals without introducing a lot of messy display related code.
```

willmcgugan (changes requested):

PR #854, opened 2022-10-07
Fix issues with `Styles.css`, update "Grid" styles doc page

darrenburns:
There were a few lookups using `-` instead of `_`, a few cases where `elif` was used when it should be `if`, and a bunch of properties missing from `Styles.css`.

Also updates grid styles documentation and adds an example for it.


willmcgugan (approved):

willmcgugan:
LGTM

PR #856, opened 2022-10-08
monochrome tests

willmcgugan:
Adds a simple color filter system, and implements NO_COLOR with it.


willmcgugan:
#executivedecision

PR #860, opened 2022-10-10
Another concept for Tab widget

yoann9344:
It is working but probably not the best conception. But can give ideas

Other concepts:  
https://github.com/Textualize/textual/pull/54

### DIFF
--- a/examples/tab_bar.py
+++ b/examples/tab_bar.py
@@ -1,0 +1,50 @@
+from rich.markdown import Markdown
+
+from textual.app import App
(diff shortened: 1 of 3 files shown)


willmcgugan:
Can I see a screenshot? :)

yoann9344:
Yep c'est mieux ^^
textual_mr_860

willmcgugan:
This is cool. We like it, but we have some other ideas for an official tab control. Thanks for your input!

PR #863, opened 2022-10-11
Cli colors

willmcgugan:
Added a preview for the color system.

At some point we might extend this in to a color scheme editor. But that should probably wait until we have a config system.

```
textual colors
```

Screenshot 2022-10-11 at 10 19 59

### DIFF
--- a/src/textual/_arrange.py
+++ b/src/textual/_arrange.py
@@ -9,7 +9,6 @@ from .geometry import Region, Size, Spacing
 from ._layout import DockArrangeResult, WidgetPlacement
 from ._partition import partition
 
-
 if TYPE_CHECKING:
     from .widget import Widget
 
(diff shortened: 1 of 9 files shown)


src/textual/_arrange.py
```diff
@@ -115,7 +114,7 @@ def arrange(
                         for placement in layout_placements
                     ]
                 ).size
-                placement_offset += styles._align_size(placement_size, size)
+                placement_offset += styles._align_size(placement_size, size).clamped
```
willmcgugan, line 117: Stop centered things from being clipped.

darrenburns (approved):
NP: The padding looks a bit too much at the top of each container

Otherwise LGTM 👍

willmcgugan:
Better?

Screenshot 2022-10-11 at 10 52 47

davep (approved):

PR #888, opened 2022-10-13
Modify the screen examples to import Screen from textual.screen

davep:
The surrounding text in the documentation states that Screen is imported from textual.screen, but the code was importing the class from textual.app. Further, I was getting warnings from pyright that I was using a class that wasn't exported from the module I was importing from.

### DIFF
--- a/docs/examples/guide/screens/screen01.py
+++ b/docs/examples/guide/screens/screen01.py
@@ -1,4 +1,5 @@
-from textual.app import App, Screen, ComposeResult
+from textual.app import App, ComposeResult
+from textual.screen import Screen
 from textual.widgets import Static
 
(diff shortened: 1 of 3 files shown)


willmcgugan (dismissed):

willmcgugan:
@davep Just noticed this will break the code highlghting feature in the docs.

If you search for those files in the docs (.md files), you will find those examples there. You will need to update the "hl_lines" attribute accordingly. 

You can run `make docs-serve` to confirm.

davep:
Thanks for pointing that out @willmcgugan -- now hopefully corrected.

willmcgugan (approved):
LGTM

PR #898, opened 2022-10-13
Reviewing CSS docs

davep:
Collection of tweaks and typo-fixes while reading through the CSS docs.


davep:
I think this is going to get a wee bit too large, so from now on I think I'll break this up into one PR per page.

docs/guide/app.md
```diff
@@ -141,7 +141,7 @@ You may have noticed that we subclassed `App[str]` rather than the usual `App`.
 --8<-- "docs/examples/app/question01.py"
 ```
 
-The addition of `[str]` tells Mypy that `run()` is expected to return a string. It may also return `None` if [App.exit()][textual.app.App.exit] is called without a return value, so the return type of `run` will be `str | None`. Replace the `str` in `[str]` with the type of the value you intend to call the exit method with.
+The addition of `[str]` tells mypy that `run()` is expected to return a string. It may also return `None` if [App.exit()][textual.app.App.exit] is called without a return value, so the return type of `run` will be `str | None`. Replace the `str` in `[str]` with the type of the value you intend to call the exit method with.
```
willmcgugan, line 144: TIL that mypy shouldn't be capitalized.

davep: Actually, I'm starting to question that; even in their own docs and on their own website I've seen mypy, Mypy, MyPy and of course My[Py].

docs/guide/devtools.md
```diff
@@ -23,7 +23,7 @@ You can run Textual apps with the `run` subcommand. If you supply a path to a Py
 textual run my_app.py
 ```
 
-The `run` sub-command assumes you have an App instance called `app` in the global scope of your Python file. If the application is called something different, you can specify it with a colon following the filename:
+The `run` sub-command will look to see if you have an `App` instance called `app` in the global scope of your Python file and will use that if it does. If not it will then look to see if there is a class that inherits from `App` and it will create and `run` an instance of that. Finally, you can specify the application instance to run with a colon following the filename:
```
willmcgugan, line 26:
```suggestion
The `run` sub-command will first look for a `App` instance called `app` in the global scope of your Python file. If there is no `app`, it will create an instance of the first `App` class it finds and run that.

Alternatively, you can add the name of an `App` instance or class after a colon to run a specific app in the Python file. Here's an example: 
```

willmcgugan (changes requested):
Great stuff. Love the attention to detail. Added an alternative wording re running apps.

willmcgugan (approved):
👍

PR #904, opened 2022-10-13
CSS docs review - Styles

davep:
Reviewing the Styles page of the docs.

### DIFF
--- a/docs/guide/styles.md
+++ b/docs/guide/styles.md
@@ -96,7 +96,7 @@ The following example shows what happens when you set alpha on background colors
 --8<-- "docs/examples/guide/styles/colors02.py"
 ```
 
-Notice that an alpha of 0.1 the background almost matches the screen, but at 1.0 it is a solid color.
+Notice that at an alpha of 0.1 the background almost matches the screen, but at 1.0 it is a solid color.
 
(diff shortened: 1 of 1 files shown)


darrenburns (approved):

docs/guide/styles.md
```diff
@@ -88,15 +88,15 @@ There are a few ways you can set alpha on a color in Textual.
...
 
-The following examples shows what happens when you set alpha on background colors:
+The following example shows what happens when you set alpha on background colors:
 
 ```python title="colors01.py" hl_lines="12-15"
 --8<-- "docs/examples/guide/styles/colors02.py"
 ```
 
-Notice that an alpha of 0.1 the background almost matches the screen, but at 1.0 it is a solid color. 
+Notice that at an alpha of 0.1 the background almost matches the screen, but at 1.0 it is a solid color.
```
willmcgugan, line 99: Superfluous `at`?

davep: Perhaps, but then I think that bit of text might need rewording: `Notice that an alpha of 0.1 the background almost` -- or perhaps `with` rather than `at`. Without something like that it seems a wee bit off.

PR #912, opened 2022-10-14
wait for mount

willmcgugan:
This PR should hopefully resolve the general class of issue where you can't query a child from a parent's `on_mount`

- Wait for composed children to be mounted, before mounting parent.
- Added a mount awaitable to allow awaiting self.mount.
- Allowed focus() to work from copose()


willmcgugan:
@darrenburns @davep Complex one. Deserves a deep dive.

davep:
> Complex one. Deserves a deep dive.

Coffee and whiteboard basics to problem to solution on Monday possible?

src/textual/app.py
```diff
@@ -1164,9 +1179,16 @@ def _start_widget(self, parent: Widget, widget: Widget) -> None:
         """
         widget._attach(parent)
         widget._start_messages()
-        widget.post_message_no_wait(events.Mount(sender=parent))
 
     def is_mounted(self, widget: Widget) -> bool:
```
darrenburns, line 1183: The difference between register and mount is confusing me - is registering simply "adding the widget to the registry", and mounting is "adding the widget to the DOM"?

Should this be `is_registered`?

src/textual/app.py
```diff
@@ -1110,15 +1129,15 @@ def _register_child(self, parent: DOMNode, child: Widget) -> bool:
 
     def _register(
         self, parent: DOMNode, *anon_widgets: Widget, **widgets: Widget
-    ) -> None:
+    ) -> list[Widget]:
         """Mount widget(s) so they may receive events.
```
darrenburns, line 1133:
```suggestion
        """Register widget(s) so they may receive events.
```

src/textual/app.py
willmcgugan: Pretty much. "register" is the internal name. "Mounting" is the full procedure which includes registering and sending events.

Should still be `is_mounted` since if it is in the registry the mount procedure will be complete.

src/textual/app.py
```diff
@@ -1008,23 +1011,36 @@ async def _process_messages(
...
+            self._running = True
+
+            try:
+                await self._process_messages_loop()
+            except asyncio.CancelledError:
+                pass
+            finally:
+                self._running = False
+                for timer in list(self._timers):
+                    await timer.stop()
```
darrenburns, line 1042: The timers get stopped inside MessagePump `_close_messages`, which gets called below. Do we need this loop here?

src/textual/app.py
```diff
@@ -1110,15 +1129,15 @@ def _register_child(self, parent: DOMNode, child: Widget) -> bool:
 
     def _register(
         self, parent: DOMNode, *anon_widgets: Widget, **widgets: Widget
-    ) -> None:
-        """Mount widget(s) so they may receive events.
+    ) -> list[Widget]:
```
darrenburns, line 1132: Could you update the docstring to include the new return type?

src/textual/app.py
```diff
@@ -696,21 +696,24 @@ def update_styles(self, node: DOMNode | None = None) -> None:
         self._require_stylesheet_update.add(self.screen if node is None else node)
         self.check_idle()
 
-    def mount(self, *anon_widgets: Widget, **widgets: Widget) -> None:
+    def mount(self, *anon_widgets: Widget, **widgets: Widget) -> AwaitMount:
```
darrenburns, line 699: Docstring needs update (especially since it's part of the public API)

darrenburns (commented):
A couple of docstrings need updating, and one question about code that looks redundant.

PR #929, opened 2022-10-16
Fixes a typo, changes a 'then' to 'when'

tsvenson:
I'm fairly sure it should be a "when" and not a "then" here.


willmcgugan (approved):

willmcgugan:
Thanks

PR #941, opened 2022-10-18
key bindings refactor

willmcgugan:
- Fixed footer click to run binding
- Takes in to account a widget's namespace when invoking bindings
- Adds "universal" keys which are processed first, but never sent to the focused widgert

### DIFF
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -196,7 +196,7 @@ class App(Generic[ReturnType], DOMNode):
 
         self._logger = Logger(self._log)
 
-        self._bindings.bind("ctrl+c", "quit", show=False, allow_forward=False)
+        self._bindings.bind("ctrl+c", "quit", show=False, universal=True)
         self._refresh_required = False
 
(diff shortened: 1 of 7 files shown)


src/textual/app.py
```diff
@@ -1275,22 +1263,42 @@ def bell(self) -> None:
         if not self.is_headless:
             self.console.bell()
 
-    async def check_bindings(self, key: str) -> bool:
+    @property
+    def _binding_chain(self) -> list[tuple[DOMNode, Bindings]]:
+        """Get a chain of nodes and bindings to consider. Goes from focused widget to app, or
+        screen to app.
```
darrenburns, line 1269: To me "goes from focused widget to app" didn't make it clear that it goes via all the ancestors.

```suggestion
        """Get a chain of nodes and bindings to consider. If no widget is focused, returns the bindings from both the screen and the app level bindings. Otherwise, combines all the bindings from the currently focused node up the DOM to the root App.
```

darrenburns (approved):

PR #943, opened 2022-10-18
Try and better settle focus after a focused widget is removed

### LINKED ISSUE #939: stopwatch key presses stop working after remove with focus
From the `css` branch on version `0.2.0b7`, it looks like there is a bug in the `docs/example/tutorial/stopwatch.py` example or maybe something to do with focus. 

To reproduce, run `stopwatch.py`, use `tab` to focus the "Reset" button on the **_last_** stopwatch, and then press `r` to remove it. After this, the app no longer responds to any key presses, including `ctrl+c`. Using the mouse to click a button seems to get things working again.

davep:
Seeks to resolve #939.

### DIFF
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -1419,12 +1419,12 @@ class App(Generic[ReturnType], DOMNode):
         widget = event.widget
         parent = widget.parent
 
-        widget.reset_focus()
-
         remove_widgets = widget.walk_children(
             Widget, with_self=True, method="depth", reverse=True
         )
         for child in remove_widgets:
+            if self.focused is child:
+                child.screen.focus_previous()
             await child._close_messages()
             self._unregister(child)
         if parent is not None:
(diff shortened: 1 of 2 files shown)


src/textual/app.py
```diff
@@ -1419,14 +1419,36 @@ async def _on_remove(self, event: events.Remove) -> None:
         widget = event.widget
         parent = widget.parent
 
-        widget.reset_focus()
-
-        remove_widgets = widget.walk_children(
-            Widget, with_self=True, method="depth", reverse=True
+        # Get the list of all the widgets that are about to be removed.
+        remove_widgets = list(
```
davep, line 1423: Note I've note made the suggested change to `walk_children` here (have it return a `list`) -- I didn't want to head off into a side-quest while working on this. If we're happy with this particular PR I'll then go tweak `walk_children` and come back and tidy this up.

davep:
Closed in preference to https://github.com/Textualize/textual/pull/954

PR #954, opened 2022-10-19
Try and better settle focus after a focused widget is removed (redux)

davep:
A new approach to solving #939, written to avoid the issues that https://github.com/Textualize/textual/pull/943 would have caused.

### DIFF
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -1419,11 +1419,16 @@ class App(Generic[ReturnType], DOMNode):
         widget = event.widget
         parent = widget.parent
 
-        widget.reset_focus()
-
-        remove_widgets = widget.walk_children(
-            Widget, with_self=True, method="depth", reverse=True
+        remove_widgets = list(
+            widget.walk_children(Widget, with_self=True, method="depth", reverse=True)
         )
+
+        if self.screen.focused in remove_widgets:
+            self.screen._reset_focus(
+                self.screen.focused,
+                [to_remove for to_remove in remove_widgets if to_remove.can_focus],
+            )
+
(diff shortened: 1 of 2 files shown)


src/textual/screen.py
```diff
@@ -224,19 +225,51 @@ def focus_previous(self) -> Widget | None:
...
+            self.set_focus(None)
+            return
+
+        # Now go looking for something before it, that isn't about to be
+        # removed, and which can receive focus, and go focus that.
+        chosen: DOMNode | None = None
+        for candidate in reversed(
+            focusable_widgets[widget_index + 1 :] + focusable_widgets[:widget_index]
+        ):
+            if candidate not in avoiding and candidate.can_focus:
```
willmcgugan, line 267: I think that we can be sure that `candidate.can_focus` is True, can't we?

willmcgugan (approved):
Yep. Looks good! Just comments.

src/textual/screen.py
davep: Need a 🤦 as the real response here.

src/textual/app.py
```diff
@@ -1419,11 +1419,16 @@ async def _on_remove(self, event: events.Remove) -> None:
         widget = event.widget
         parent = widget.parent
 
-        widget.reset_focus()
-
-        remove_widgets = widget.walk_children(
-            Widget, with_self=True, method="depth", reverse=True
+        remove_widgets = list(
```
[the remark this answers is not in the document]
davep: Aye, already noted in 11ddcdd771f54576bd14f4b4ba96ec50e2d26014 on commit.

PR #959, opened 2022-10-19
adds demo

willmcgugan:
- Added demo app. Run with `python -m textual`
- Added `top` parameter to `scroll_*` methods
- Fixed issues where TextLog wouldn't change color in dark mode.
- Added demo to docs.
- Added section re examples.


davep (approved):
More from a "testing the result" point of view than a "looking at the code" point of view: Looks good! Nice intro to things. I think there's three main things stand out:

- Keyboard navigation can be a little confusing depending on how you start navigation and what you click on when using a combination of mouse and keyboard (see conversation we had about that).
- The screenshot facility is really neat but I suspect it needs some sort of feedback to let the user know that a) the screenshot worked and b) where it was written.
- The screenshot facility might need a little bit of `try`...`except` to capture any filesystem or similar exceptions. For example, I can easily make the demo crash if I try and screenshot while the CWD is a directory I don't have write permission to write to.

src/textual/dom.py
```diff
@@ -278,7 +278,7 @@ def screen(self) -> "Screen":
         while node and not isinstance(node, Screen):
             node = node._parent
         if not isinstance(node, Screen):
-            raise NoScreen(f"{self} has no screen")
+            raise NoScreen(f"node has no screen")
```
davep, line 281: Perhaps wanted `node` inside a pair of `{}` here?

davep (approved):
Couple of small observations made inline. Fun to see the focus settling thing keeps expanding.

src/textual/dom.py
willmcgugan: No, the problem was that the repr in the exception message could refer to `self.screen` which would create a infinite loop!

src/textual/widget.py
```diff
@@ -261,6 +261,18 @@ def siblings(self) -> list[Widget]:
...
+    def visible_siblings(self) -> list[Widget]:
+        """A list of siblings which will be shown.
+
+        Returns:
+            list[Widget]: List of siblings.
+        """
+        siblings = [
+            widget for widget in self.siblings if widget.visible and widget.display
+        ]
+        return siblings
```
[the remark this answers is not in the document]
willmcgugan: A little faster perhaps. But I like to keep return statements simple.

src/textual/dom.py
davep: Sorry, perhaps I wasn't clear. In the replacement line you're using an f-string with no variable content; so perhaps the intention was to have `f"{node} has no screen"` rather than `f"node has no screen"`.

davep: Ahh, I see; was being thrown by the intention of it being an f-string with nothing to f.

davep (approved):
Good call! I forgot that had got merged in now.

src/textual/demo.py
```diff
@@ -141,7 +141,7 @@
 - **DataTable** A spreadsheet-like widget for navigating data. Cells may contain text or Rich renderables.
 - **TreeControl** An generic tree with expandable nodes.
 - **DirectoryTree** A tree of file and folders.
-- *... [many more planned](https://textual.textualize.io/roadmap/) ...* 
+- *... many more planned ...*
```
davep, line 144: Awww! I liked the links. What was the issue?

willmcgugan: For some bizarre reason iTerm gets slow when there are links on the page. Just one link makes it stuttery.

src/textual/demo.py
```diff
@@ -0,0 +1,419 @@
...
+        for n in range(20):
+            table.add_row(*[f"Cell ([b]{n}[/b], {col})" for col in range(6)])
+        self.query_one("Welcome Button", Button).focus()
+
+    def action_screenshot(self, filename: str | None = None, path: str = "./") -> None:
+        """Save an SVG "screenshot". This action will save an SVG file containing the current contents of the screen.
+
+        Args:
+            filename (str | None, optional): Filename of screenshot, or None to auto-generate. Defaults to None.
+            path (str, optional): Path to directory. Defaults to "~/".
```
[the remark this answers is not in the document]
willmcgugan: Good point.

It could... Might leave that for another update though.

src/textual/demo.py
davep: Ugh! That's a shame! But, yeah, doesn't make sense to slow the demo for this. :-/

src/textual/demo.py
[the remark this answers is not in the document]
davep: (scratch link suggestion here given the iTerm issue)

davep (approved):

PR #962, opened 2022-10-19
Fix bindings as tuples

darrenburns:
The logic for splitting keys was only being applied to `Binding` when supplied as the dataclass, not when the tuple was used.


davep (approved):
👍

willmcgugan (approved):

PR #963, opened 2022-10-19
Initial pull of 5x5 into the Textual examples directory

davep:
I think this is in a good enough state to start to pull it in now. It has keyboard navigation, mouse navigation, a simple help screen, a simple win state display, etc. It demonstrates *not* using a default screen, etc. Colour needs some work, but I think the core implementation is pretty sound now.

Tweaks to make it more Textually idiomatic are welcome -- this is obviously written by someone who is still getting to know the framework.


examples/five_by_five.py
```diff
@@ -0,0 +1,273 @@
...
+    @property
+    def on_cells(self) -> DOMQuery[GameCell]:
+        """The collection of cells that are currently turned on.
+
+        :type: DOMQuery[GameCell]
+        """
+        return cast(DOMQuery[GameCell], self.query("GameCell.on"))
+
+    @property
+    def on_count(self) -> int:
```
willmcgugan, line 141: The naming of this conflicts with our event handlers. Can we rename to prevent any confusion?

examples/five_by_five.py
```diff
@@ -0,0 +1,273 @@
...
+
+    def __init__(self) -> None:
+        """Constructor."""
+        super().__init__(title="5x5 -- A little annoying puzzle")
+
+    def on_mount(self) -> None:
+        """Set up the application on startup."""
+        self.push_screen(Game())
+
+
```
willmcgugan, line 271: Can we construct the app before the `if __main__`. Which is the convention in other apps.

examples/five_by_five.py
```diff
@@ -0,0 +1,273 @@
...
+    def on_mount(self) -> None:
+        """Get the game started when we first mount."""
+        self.action_new_game()
+
+
+class FiveByFive(App[None]):
+    """Main 5x5 application class."""
+
+    #: The name of the stylesheet for the app.
+    CSS_PATH = Path(__file__).with_suffix(".css")
```
willmcgugan, line 255: I think this can be `CSS_PATH="FiveByFive.css"` now.

examples/five_by_five.py
```diff
@@ -0,0 +1,273 @@
...
+
+        :type: DOMQuery[GameCell]
+        """
+        return cast(DOMQuery[GameCell], self.query("GameCell.on"))
+
+    @property
+    def on_count(self) -> int:
+        """The number of cells that are turned on.
+
+        :type: int
```
willmcgugan, line 144: Could we change the docstrings to use the Google convention.

willmcgugan (changes requested):
Great! Just a few requests.

Once that's in, I might have a go at the stylesheet.

darrenburns (changes requested):
Could you change your import of `Final` to something like this?

```python
import sys
if sys.version_info >= (3, 8):
    from typing import Final
else:
    from typing_extensions import Final
```

examples/five_by_five.py
davep: Good call, kept meaning to do that.

examples/five_by_five.py
davep: As noted in local conversation: this isn't the convention for Textual apps ever since `textual run` was updated to find the `App` class, etc.

examples/five_by_five.py
davep: Changed.

willmcgugan (approved):

darrenburns (approved):

PR #969, opened 2022-10-20
More testing

darrenburns:
- We automatically snapshot the canonical examples for CSS properties.
- There are snapshot tests for each widget canonical example.
- Expose the `App` instance to snapshot tests. This lets us query for DOM nodes in the apps and check internal state is as we expect.


willmcgugan (approved):
Nice

PR #971, opened 2022-10-21
Update package metadata and dev experience

ofek:
## Background

Hello there! As discussed in Discord, this is a PoC showing what a switch to [Hatch](https://hatch.pypa.io/latest/) would look like. Some new users include Black, Pydantic, MkDocs, and FastAPI.

The only thing lost is the lock file, since the format is [not yet standardized](https://discuss.python.org/t/how-should-a-lockfile-pep-665-successor-look-like/17690). This [can be thought of](https://github.com/renovatebot/renovate/issues/10187#issuecomment-1264225788) as an improvement though since Textual is a library for actual apps to use. (examples: https://github.com/ics-py/ics-py/pull/354, https://github.com/P403n1x87/austin-python/pull/9, https://github.com/delb-xml/delb-py/commit/010f5f056572712ce965c33a0a325358347d395a)

## Build backend

The Python packaging ecosystem has standardized on the format for metadata declaration ([PEP 621](https://peps.python.org/pep-0621/)/[PEP 631](https://peps.python.org/pep-0631/)). The build backend `hatchling` (of which I am a maintainer in the [PyPA](https://github.com/pypa)) is now used as that is the default in the official [Python packaging tutorial](https://packaging.python.org/en/latest/tutorials/packaging-projects/).

Hatchling is available on all the major distribution channels such as [Debian](https://salsa.debian.org/python-team/packages/hatchling), [Fedora](https://src.fedoraproject.org/rpms/python-hatchling), [Arch Linux](https://archlinux.org/packages/community/any/python-hatchling/), [conda-forge](https://anaconda.org/conda-forge/hatchling), [Nixpkgs](https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/python-modules/hatchling/default.nix), [Alpine Linux](https://github.com/alpinelinux/aports/blob/master/community/py3-hatchling/APKBUILD), [FreeBSD](https://github.com/freebsd/freebsd-ports/blob/main/devel/py-hatchling/Makefile)/[OpenBSD](https://github.com/openbsd/ports/blob/master/devel/py-hatchling/Makefile), [Gentoo Linux](https://github.com/gentoo/gentoo/tree/master/dev-python/hatchling), [MacPorts](https://github.com/macports/macports-ports/blob/master/python/py-hatchling/Portfile), [OpenEmbedded](https://github.com/openembedded/openembedded-core/commit/846e806181f1349be29cbce78c5041735dfd7e6f), [Spack](https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/py-hatchling/package.py), [MSYS2](https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-python-hatchling/PKGBUILD), etc.

## PyPI landing page

As you can see on https://pypi.org/project/textual/0.2.0b8/, the links to code samples and images are broken. This PR fixes that by using @hynek's awesome [hatch-fancy-pypi-readme](https://github.com/hynek/hatch-fancy-pypi-readme) plugin.

## Dev experience

The dependencies for testing, linting, and documentation are now separated into 3 different environments. This speeds up install/syncing operations and reduces the chance of conflicts or runtime side effects.

- `default` - the dependencies for testing are installed here along with the project and the `dev` optional dependency group. the `default` environment is special in that Hatch will use it always in lieu of a [selection](https://hatch.pypa.io/latest/environment/#selection) e.g. `hatch run test` vs `hatch run docs:serve`
- `docs` - the dependencies for documentation are installed here. the project and the `dev` optional dependency group are also required for `mkdocstrings` auto-generation of docs
- `lint` - only the dependencies for linting are installed here

This now works on Windows! (Makefiles do not without great effort)

## Python version cap

I removed the Python upper bound constraint as upper bounds for libraries are [heavily discouraged](https://iscinumpy.dev/post/bound-version-constraints/) by the community and that was just a thing that Poetry recommended.

## Future

- Move config for Mypy out of `mypy.ini` and into `pyproject.toml`
- Hatch has a large ecosystem of plugins, so for example you could use Git for [versioning](https://github.com/ofek/hatch-vcs) or ship optional wheels compiled with [Mypyc](https://github.com/ofek/hatch-mypyc), as Black [recently did](https://ichard26.github.io/blog/2022/05/31/compiling-black-with-mypyc-part-1/)
- If you want I can do this for Rich too


willmcgugan:
Thanks! Will look at this after the pending release. Need to understand a bit more about Hatch.

ofek:
Happy to help! I'm interested in hearing your thoughts.

hatch.toml
```diff
@@ -0,0 +1,38 @@
+[envs.default]
+features = ["dev"]
+dependencies = [
```
ofek, line 3:
```suggestion
dependencies = [
    "jinja2 < 3.1.0",
```

hatch.toml
```diff
@@ -0,0 +1,38 @@
...
+    "pre-commit >= 2.12.1, <3",
+]
+[envs.lint.scripts]
+format = "black src"
+check = "black --check src"
+typing = "mypy src/textual"
+
+[envs.docs]
+dependencies = [
+    "jinja2 < 3.1.0",
```
ofek, line 29:
```suggestion
```

ofek:
I rebased, would you mind triggering the CI with that button below?

willmcgugan:
Thanks @ofek . We're just about to release a new version. But I reckon we should go all in with Hatch after that.

ofek:
Cool, rebased again

willmcgugan:
Thanks. Next release will be "hatched"!

PR #977, opened 2022-10-22
added version number to demo, bumped beta

willmcgugan:
- Added the version number to the demo
- Textual did not work without [dev]. Hardly surprising since we never test it without. Wish we had spotted that earlier.

Screenshot 2022-10-22 at 16 02 50

### DIFF
--- a/src/textual/__init__.py
+++ b/src/textual/__init__.py
@@ -52,7 +52,7 @@ class Logger:
             app = active_app.get()
         except LookupError:
             raise LoggerError("Unable to log without an active app.") from None
-        if not app.devtools.is_connected:
+        if not app.devtools_enabled:
             return
 
         previous_frame = inspect.currentframe().f_back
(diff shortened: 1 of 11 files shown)


src/textual/__init__.py
```diff
@@ -52,7 +52,7 @@ def __call__(self, *args: object, **kwargs) -> None:
             app = active_app.get()
         except LookupError:
             raise LoggerError("Unable to log without an active app.") from None
-        if not app.devtools.is_connected:
+        if not app.devtools_enabled:
```
willmcgugan, line 55: Checks both availability and if it is connected

willmcgugan:
#executivedecision

@darrenburns @davep feel free to sanity check.

davep:
> Textual did not work without [dev]. Hardly surprising since we never test it without. Wish we had spotted that earlier.

I've made a note of this in my notes for standup on Monday, but as a reminder here too: this feels like a perfect subject for a "HOWTO" in the docs: how to set up for building a Textual app and how to package an app for deployment to others and the how and when you'd use `[dev]`, etc -- perhaps relating it to the common packaging tools.

PR #985, opened 2022-10-23
Add sensible error to query

willmcgugan:
`app.query` would raise an invalid CSS error because it reuses code from CSS parsing. This PR converts those to a more sensibly named error.

### DIFF
--- a/examples/calculator.py
+++ b/examples/calculator.py
@@ -28,7 +28,8 @@ class CalculatorApp(App):
         "plus_minus_sign": "plus-minus",
         "percent_sign": "percent",
         "equals_sign": "equals",
-        "enter": "equals",
+        "minus": "minus",
+        "plus": "plus",
     }
 
     def watch_numbers(self, value: str) -> None:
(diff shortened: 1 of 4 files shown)


examples/calculator.py
```diff
@@ -89,7 +89,9 @@ def press(button_id: str) -> None:
             press("c")
             press("ac")
         else:
-            press(self.NAME_MAP.get(key, key))
```
willmcgugan, line 92: This would blow up in `press` if you pressed "shift+tab"

davep: Good catch!

examples/calculator.py
```diff
@@ -28,7 +28,8 @@ class CalculatorApp(App):
         "plus_minus_sign": "plus-minus",
         "percent_sign": "percent",
         "equals_sign": "equals",
-        "enter": "equals",
```
davep, line 31: Presumably the intention here is to allow full keyboard-only button navigation, which then requires that `Enter` presses the current button rather than act as an alternative for `=`? That's fair.

davep (approved):

PR #987, opened 2022-10-23
Fix broken links in README.md

UriyaHarpeness:
The links for the `code_browser.py` and `calculator.py` examples in the README (and therefore in PyPi as well) are broken.
This is a small fix for it, pointing to the main branch,


willmcgugan:
Thanks

PR #988, opened 2022-10-23
Add highlighting for id="timers" line in tutorial

teschty:
It appears the tutorial is missing a line of highlighting when adding the `timers` id to the timer container. This PR adds highlighting for that line.

Rendered:
highlighting


willmcgugan:
Yup. THanks.

PR #990, opened 2022-10-23
DirectoryTree: remove example code as examples/code_browser.py already covers that

1ace:
We're not all lucky enough to be @willmcgugan :stuck_out_tongue_winking_eye:

### DIFF
--- a/src/textual/widgets/_directory_tree.py
+++ b/src/textual/widgets/_directory_tree.py
@@ -10,7 +10,6 @@ import rich.repr
 
 from rich.text import Text
 
-from .. import events
 from ..message import Message
 from .._types import MessageTarget
 from ._tree_control import TreeControl, TreeNode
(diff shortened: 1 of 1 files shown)


1ace:
I had that fix lying around for ages and just posted it, but actually even with that it's still broken now.

I'll try fixing it properly; closing for now

willmcgugan:
That was test code, which should probably just be removed. Were you trying to get `python -m textual.widgets._directory_tree` to work?

1ace:
Yeah, `python -m textual.widgets._directory_tree` is how I was using that.

It was actually easy to fix, so re-opening with that, but perhaps it should be moved into `examples/` instead?

1ace:
Added a commit to move the example code to `examples/tree.py`, and changed the `$HOME` to `.` as it should be a smaller folder to list (likely will be the textual folder itself)

1ace:
Actually `examples/code_browser.py` already covers the use of `DirectoryTree`, so I've replaced everything in this PR with just deleting that code :upside_down_face:

UriyaHarpeness (approved):

davep (approved):

PR #991, opened 2022-10-23
Fix the hover text for the docs dark mode switch

davep:
While looking at how the docs are done... stumbled on the cause of that odd hover text I reported the other week!


willmcgugan:
How in the blazes did I manage that?

davep:
> How in the blazes did I manage that?

Cosmic rays. 🤷‍♀️

willmcgugan (approved):

Issue #993, opened 2022-10-23
Invalid default screenshot file name in Windows.


kmonson:
Default screenshot filename format in App.save_screenshot (specifically time_format) results in colons in the default file name. This is invalid in Windows. 

`OSError: [Errno 22] Invalid argument: './textual demo 2022-10-23 16:00:23 341765.svg'`

An additional replace call here is probably needed: https://github.com/Textualize/textual/blob/main/src/textual/app.py#L538

davep:
Good catch, thanks.

PR #995, opened 2022-10-24
Extend set of handled reserved characters for screenshots

davep:
This extends the set of reserved characters that are guarded against when auto-generating a screenshot filename. The extension covers the most likely candidates taken from:

   https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file#naming-conventions

This could be spun out into a more general filename cleaning function, which takes OS and filesystem into account and covers every single reserved character, but for the purposes of this bit of code this should be more than enough.

Addresses #993.

### DIFF
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -535,7 +535,8 @@ class App(Generic[ReturnType], DOMNode):
             svg_filename = (
                 f"{self.title.lower()} {datetime.now().strftime(time_format)}.svg"
             )
-            svg_filename = svg_filename.replace("/", "_").replace("\\", "_")
+            for reserved in '<>:"/\\|?*':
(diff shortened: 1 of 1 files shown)


willmcgugan:
@davep  I think maybe we should change the data format to  `YYYYMMDD HHMMSS`.

Could you also add `encoding='utf-8'` to `open`.

davep:
> I think maybe we should change the data format to `YYYYMMDD HHMMSS`.
> 
> Could you also add `encoding='utf-8'` to `open`.

Both done. I've keep the code that strips reserved characters as it's possible someone could pass in a problematic time format, and it seems sensible to guard against this.

willmcgugan (approved):

PR #1016, opened 2022-10-25
Add Python 3.11 to CI, update install-poetry action

darrenburns:


src/textual/app.py
```diff
@@ -641,6 +641,7 @@ async def press_keys() -> None:
                             await asyncio.sleep(0.02)
 
                     await app._animator.wait_for_idle()
+                    await asyncio.sleep(0.05)
```
darrenburns, line 644: On Windows, screenshots seemed to be slightly too early and the final key wasn't being processed. Not ideal of course, but seems to fix the issue.

willmcgugan: THis might be redundant with a pending PR.

willmcgugan (approved):

PR #1017, opened 2022-10-25
Fix some inaccuracies in the tutorial.

UriyaHarpeness:
There's an undocumented difference between `docs/examples/tutorial/stopwatch03.css` and `docs/examples/tutorial/stopwatch04.css`, so I've equalized that part, and sorted the attributes by ABC.
Also, `Header` does not the current time by default and in the stopwatch example, so I updated the docs there.

### DIFF
--- a/docs/examples/tutorial/stopwatch03.css
+++ b/docs/examples/tutorial/stopwatch03.css
@@ -2,8 +2,9 @@ Stopwatch {
     layout: horizontal;
     background: $boost;
     height: 5;
-    padding: 1;
     margin: 1;
+    min-width: 50;
+    padding: 1;
 }
 
 TimeDisplay {
--- a/docs/examples/tutorial/stopwatch04.css
+++ b/docs/examples/tutorial/stopwatch04.css
@@ -2,8 +2,8 @@ Stopwatch {
     layout: horizontal;
     background: $boost;
     height: 5;
-    min-width: 50;
     margin: 1;
+    min-width: 50;
     padding: 1;
 }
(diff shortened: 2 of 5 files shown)


UriyaHarpeness:
I see typos in many files I open on my PyCharm (it warns about them like in the picture shown), maybe one should take a little time to do a quick scan through the docs to fix all of them.
image

UriyaHarpeness:
@davep, can someone review and merge?

UriyaHarpeness:
Of course @willmcgugan, just that the previous PR was merged after 2 minutes~, just wondered if this PR was skipped or missed. No problem.

davep (approved):
Thanks for that.

Could I make a suggestion though? In future, it might be a good idea to split something like this into more than one PR. In here you had very obvious typos -- there's no question about them and fixing them -- plus some changes that were more about choice and writing style, which I feel should be in their own PR so a conversation can be had with the author of the text.

I don't disagree with any of the suggestions here so we'll pull them in.

> maybe one should take a little time to do a quick scan through the docs to fix all of them.

Just so you're aware: many of us have been over the docs quite a few times now, and as I'm sure you'll have seen in the commit history, lots of typos and other small issues have already been squished. Lots of folk have taken more than "a little time" to go over them; but of course things can still get missed. Thanks for adding more eyes to that number.

As for time taken to get to the PR: don't worry, we get to them all eventually; this one, because of the mixture of different things that it touched on, needed a bit more free time to look over.

Thanks again. :-)

PR #1019, opened 2022-10-25
Fixed data_table example app - actually run it

typon:
Minor fix for actually running the data_table app in the example


UriyaHarpeness (approved):

davep (approved):
Good catch, thanks.

PR #1028, opened 2022-10-26
Add highlight line to tutorial preview.

lukexyz:
New timer ID adds `hl_lines="86"` for consistency with other changes.

Line 86 in `stopwatch.py`:
`        yield Container(Stopwatch(), Stopwatch(), Stopwatch(), id="timers")`

Had me briefly stuck when working through the tutorial.

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -430,7 +430,7 @@ The Stopwatch app creates widgets when it starts via the `compose` method. We wi
 
 Let's use these methods to implement adding and removing stopwatches to our app.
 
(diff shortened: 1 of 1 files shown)


davep:
Thanks @lukexyz! However, someone already got there (see https://github.com/Textualize/textual/commit/2d003603bf89c7c2dfb3a857968fa9e53f0bbd70) -- that's kind of on us in that the `css` branch is still kicking about (we're about to remove it) and the website is currently deployed from a version of `main` before https://github.com/Textualize/textual/commit/2d003603bf89c7c2dfb3a857968fa9e53f0bbd70 was done.

Really appreciate you picking up on it though. :-)

PR #1033, opened 2022-10-27
Adds the debounce decorator

lllama:
We add the debounce decorator. It can be used to ensure an event handler does not get overwhelmed with events, or that expensive operations (e.g. network requests) do not happen too frequently.

### DIFF
--- a/src/textual/debounce.py
+++ b/src/textual/debounce.py
@@ -1,6 +1,11 @@
 import asyncio
 from functools import wraps
-from typing import Any, Callable, Optional, ParamSpec
+from typing import Any, Callable, Optional
+
+try:
+    from typing import ParamSpec
+except ImportError:
+    from typing_extensions import ParamSpec
 
 from .app import App
 from .timer import Timer as TimerClass
--- a/src/textual/debounce.py
+++ b/src/textual/debounce.py
@@ -32,9 +32,9 @@ def debounce(timeout: float) -> Callable[[Callable[P, Any]], Callable[P, None]]:
         async def wrapper(self: App, *args: P.args, **kwargs: P.kwargs) -> None:
             nonlocal timer
             if timer:
-                timer.reset()
-            else:
-                timer = self.set_timer(timeout, lambda: f(self, *args, **kwargs))
+                timer.stop_no_wait()
+
+            timer = self.set_timer(timeout, lambda: f(self, *args, **kwargs))
 
         return wrapper
 
--- a/src/textual/debounce.py
+++ b/src/textual/debounce.py
@@ -1,12 +1,14 @@
 import asyncio
 from functools import wraps
-from typing import Callable, Optional
+from typing import Any, Callable, Optional, ParamSpec
 
 from .app import App
 from .timer import Timer as TimerClass
 
+P = ParamSpec("P")
 
-def debounce(timeout: float) -> Callable:
+
+def debounce(timeout: float) -> Callable[[Callable[P, Any]], Callable[P, None]]:
     """Debounce repeated events.
 
     If event handlers perform expensive operations, then it is sometimes desirable to
@@ -18,11 +20,11 @@ def debounce(timeout: float) -> Callable:
         timeout (float): the number of seconds to wait for repeated events.
     """
 
-    def decorator(f: Callable) -> Callable:
+    def decorator(f: Callable[P, Any]) -> Callable[P, None]:
         timer: Optional[TimerClass] = None
 
         @wraps(f)
-        async def wrapper(self: App, *args, **kwargs) -> None:
+        async def wrapper(self: App, *args: P.args, **kwargs: P.kwargs) -> None:
             nonlocal timer
             if timer:
                 timer.reset()
--- a/src/textual/debounce.py
+++ b/src/textual/debounce.py
@@ -34,9 +34,7 @@ def debounce(timeout: float) -> Callable[[Callable[P, Any]], Callable[P, None]]:
             if timer:
                 timer.reset()
             else:
-                timer = self.set_timer(
-                    timeout, lambda: asyncio.create_task(f(self, *args, *kwargs))
-                )
+                timer = self.set_timer(timeout, lambda: f(self, *args, **kwargs))
 
         return wrapper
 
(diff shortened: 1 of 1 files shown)


lllama:
PR against `main` as requested. Let me know what your thoughts are.

src/textual/debounce.py
```diff
@@ -0,0 +1,24 @@
...
+
+def debounce(timeout: float) -> Callable:
+    def decorator(f: Callable) -> Callable:
+        timer: Optional[TimerClass] = None
+
+        @wraps(f)
+        async def wrapper(self: App, *args, **kwargs) -> None:
+            nonlocal timer
+            if timer:
+                timer.stop_no_wait()
```
willmcgugan, line 17: Rather than stopping the timer, we could pause and resume it. This would be more efficient, since we wouldn't need to create tasks every time.

src/textual/debounce.py
```diff
@@ -0,0 +1,24 @@
+import asyncio
+from functools import wraps
+from typing import Callable, Optional
+
+from .app import App
+from .timer import Timer as TimerClass
+
+
+def debounce(timeout: float) -> Callable:
```
willmcgugan, line 9: Are you up for a challenge? Can we use the new `ParamSpec` to type the decorator? Without it the typing of a decorated function is lost.

Here's a blog post on the subject:

https://sobolevn.me/2021/12/paramspec-guide

willmcgugan, line 9: Can we have a docstring for this? In the google format. Should be plenty of examples in the repo.

willmcgugan (changes requested):
A nice addition. Think it is in the right place. Some requests to consider...

src/textual/debounce.py
lllama: I think we want to `reset()` it instead, as otherwise I think we'd need something like a timer to restart the original timer or some such.

I've also just done some testing and it looks like if we do the reset method, then it's the original `args` and `kwargs` that get passed to the handler when the timer fires. e.g. for the dictionary example, it'll be the first letter typed rather than the word.

src/textual/debounce.py
tusharsadhwani: I think it's about time you add mypy to your CI. pre-commit.ci might be the best option!

src/textual/debounce.py
```diff
@@ -0,0 +1,36 @@
...
+
+    Args:
+        timeout (float): the number of seconds to wait for repeated events.
+    """
+
+    def decorator(f: Callable) -> Callable:
+        timer: Optional[TimerClass] = None
+
+        @wraps(f)
+        async def wrapper(self: App, *args, **kwargs) -> None:
```
tusharsadhwani, line 25:
```suggestion
from typing import Any, Callable, Optional, ParamSpec

from .app import App
from .timer import Timer as TimerClass

P = ParamSpec("P")


def debounce(timeout: float) -> Callable[[Callable[P, Any]], Callable[P, None]]:
    """Debounce repeated events.

    If event handlers perform expensive operations, then it is sometimes desirable to
    limit the number of calls to the handler. The debounce decorartor will (re)start
    a timer for every event received and only call the handler if no events have occured
    in the time period provided.

    Args:
        timeout (float): the number of seconds to wait for repeated events.
    """

    def decorator(f: Callable[P, Any]) -> Callable[P, None]:
        timer: Optional[TimerClass] = None

        @wraps(f)
        async def wrapper(self: App, *args: P.args, **kwargs: P.kwargs) -> None:
```

src/textual/debounce.py
```diff
@@ -0,0 +1,38 @@
...
+        timer: Optional[TimerClass] = None
+
+        @wraps(f)
+        async def wrapper(self: App, *args: P.args, **kwargs: P.kwargs) -> None:
+            nonlocal timer
+            if timer:
+                timer.reset()
+            else:
+                timer = self.set_timer(
+                    timeout, lambda: asyncio.create_task(f(self, *args, *kwargs))
```
willmcgugan, line 33: Why create a task here? Could we not just pass in `f` and parameters as the callback.

Launching tasks in asyncio is supper easy, but stopping them is not. Best avoided if possible.

willmcgugan: I think `set_timer` will accept both sync and async callables. So it might just work!

lllama: It does. `invoke` in `_callback.py` checks for awaitable and does what's required.

src/textual/debounce.py
lllama: I've added an `ImportError` check as `ParamSpec` is >3.10 only.

src/textual/debounce.py
lllama: @willmcgugan not sure if you still have a view on this. A basic `reset` doesn't work. We'd need to replace the closure callback in the `Timer` but then we're not using a public api.

tusharsadhwani:
@willmcgugan is there any blocker on this?

lllama:
@willmcgugan if the new `work` decorator grew a `delay` parameter then `delay` + `exclusive` would equal `debounce`...

tusharsadhwani:
@lllama or debounce can be implemented with work, using exclusive and adding delay code on top

lllama:
Yeah. @willmcgugan is a crafty one. Just stick an `await asyncio.sleep` in your function and set `exclusive`.

lllama:
Some potential ideas for other options to go with `exclusive`: http://ember-concurrency.com/docs/task-concurrency

Maybe also some docs inspiration.

ethael:
@lllama @willmcgugan I was following this topic, because I would happily use this feature, but I am not sure what is the result now. Is this going to land in near future?

lllama:
So if you have a look here: https://textual.textualize.io/guide/workers/#work-decorator

So previously with the `debounce` decorator, your task would be created in a `Timer`. If a new event came in, then the task was cancelled and a new one created. A similar thing now happens with the `@work` decorator when you set the `exclusive=True` parameter.

So to achieve a debounce, you'll create a task using `@work(exclusive=True)` and then the first thing in your code will be an `asyncio.sleep(...)` for whatever your debounce value is. If a new task comes in before the sleep starts, then the task will be cancelled and you'll get the debounce behaviour.

Issue #1064, opened 2022-10-31
Arguments are not working with 'textual run'


jspv:
Per `textual run --help`:
```
  If you are running a file and want to pass command line arguments, wrap the
  filename and arguments in quotes:

      textual run "foo.py arg --option"
```

This is not working.   Using the following sample.py:

```
from textual.app import App, ComposeResult
from textual.widgets import Static
from sys import argv

class MyApp(App):
    def __init__(self, greeting: str="default") -> None:
        self.greeting = greeting
        super().__init__()

    def compose(self) -> ComposeResult:
        yield Static(f"{self.greeting}")

if __name__ == "__main__":
    if len(argv) > 1:
        app = MyApp(argv[1])
    else:
        app = MyApp()
    app.run()
```

When run via:
`python ./sample.py foo`
it returns a textual window with `foo`

When run via:
`textual run "./sample.py foo"`
it returns a textual window with `default`

davep:
The issue here is the bit of code towards the end:

```python
if __name__ == "__main__":
    if len(argv) > 1:
        app = MyApp(argv[1])
    else:
        app = MyApp()
    app.run()
```

The `textual run` command isn't running that code. When presented with your code, `textual run`, not seeing a variable called `app` at the top level (which is fine), goes looking for a class in your module that inherits from `App`; it then creates an instance of that and `.run()`s it.

However, even if you were to do that, I suspect it won't have the effect you're hoping for. As you may imagine, if you're going to `textual run` your application `argv` will end up looking quite different; at the moment anyway. I'm going to dig into this a little more and see if there might be a better way of doing things.

But there is a different approach you can take. Rather than using `textual run` to run your app, you can run it just as normal, with `python`, but have `TEXTUAL=devtools` in your environment. That gives all the benefits of a `textual run` without needing to use `textual run`.

So, for example, in one terminal you can do:

```sh
$ textual console
```

to get the developer console up and running, and then in another do something like (depending on shell):

```sh
$ export TEXTUAL=devtools python ./sample.py. foo
```

and everything should work as you were hoping.

Let me know how you get on and if this solves the issue for you.

willmcgugan:
There's a fix for this in master.

davep:
@jspv Yup, as Will mentions, about 90 minutes before you mentioned it, we did https://github.com/Textualize/textual/pull/1070 which should address this. :-)

jspv:
[...]

PR #1070, opened 2022-10-31
Change the way that `textual run` does the argv monkey-patch

davep:
A `textual run` already made an attempt to monkey-patch `argv` so that the application being run would not see an `argv` any differently to how it would if it had been run "vanilla". However, it could be caught out by code like:

    from sys import argv

    ...

    do_something_with( argv )

beforehand it was okay with:

    import sys

    ...

    do_something_with( sys.argv )

With this change it should work both ways.

Somewhat related to #1064.

### DIFF
--- a/src/textual/_import_app.py
+++ b/src/textual/_import_app.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import os
+import sys
 import runpy
 import shlex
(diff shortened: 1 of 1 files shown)


willmcgugan (approved):
Can we have a changelog entry please, and link to the issue

PR #1076, opened 2022-10-31
Ensure that queries against IDs are always fully case-insensitive

### LINKED ISSUE #1047: Selectors should be case insensitive
It looks like selectors insist on being lower-case, which means that if you use upper case in an ID it won't match. Possibly true of classes as well. Need to investigate. 

### Discussed in https://github.com/Textualize/textual/discussions/1041

<div type='discussions-op-text'>

<sup>Originally posted by **mindset-team** October 28, 2022</sup>
There's seems to be some case sensitivity issue with query/query_one. Here's an example app where widget id *must* be lower-cased in order to return a valid query result (vs. a `NoMatches` exception). The query selector seems to be case-insensitive.  

```python
from textual.app import App

class GridLayoutExample(App):
    def compose(self):
        yield Static("some text here..", id="One")

    async def on_key(self, event):
        print(f'{self.query_one("#One")=}')

app = GridLayoutExample()

if __name__ == "__main__":
    app.run()
```
Above, `on_key` is simply used to trigger the query event and `textual console` opened in another terminal to monitor print output.
The code above fails with `NoMatches: No nodes match <DOMQuery GridLayoutExample(title='GridLayoutExample') filter='#One'`

However, changing `id=One` to `id=one` (in compose method yielding the static widget) works, no exception raised and console printout is valid:

```
[20:58:32] PRINT                                                                                                           latout.py:13
self.query_one("#One")=Static(id='one')
```
Is this behavior expected? It seems counter-intuitive that a selector would be case insensitive, but id would be coerced to be lower-cased for a query to work. I couldn't find any reference to case-[in]sensitivity in the textualize.io guide, or on github (after a _quick_ review of the textual css parsing code and def query). 

Not sure this elevates to a bug (and anyway maintaining all ID's in lower-case would avoid this issue) but since many of the examples have `#One, #Two` identifiers throughout the docs, the issue seems to merit a discussion here, a clear instruction in the docs, or designating as a bug to fix. Thoughts?</div>

davep:
Fixes #1047.


davep:
Retracting this one. After some discussion in the office we've decided that we should follow the web here and be fully case-sensitive, rather than fully case-insensitive. New PR incoming soon.

PR #1082, opened 2022-10-31
optimization

willmcgugan:
A micro-optimization. Tested with timeit. Shaves of a few nanoseconds.


davep (approved):
Insert Knuth quote here.

PR #1086, opened 2022-11-01
Drop support for anonymous vs named widget mounting

davep:
In the past App.mount and Widget.mount allowed you to do something like this:

```python
self.mount( Static("A"), Static("B"), c=Static("C"), d=Static("D") )
```

which would result in two widgets being mounted with no ID, and two widgets being mounted with assigned IDs, taken from the name of the keyword argument. It's been decided that this is unnecessary, especially given that the above can still be done (after this commit) like this:

```python
self.mount( Static("A"), Static("B"), Static("C", id="c"), Static("D", id="d") )
```

### DIFF
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -822,27 +822,27 @@ class App(Generic[ReturnType], DOMNode):
         self._require_stylesheet_update.add(self.screen if node is None else node)
         self.check_idle()
 
-    def mount(self, *anon_widgets: Widget, **widgets: Widget) -> AwaitMount:
-        """Mount widgets. Widgets specified as positional args, or keywords args. If supplied
-        as keyword args they will be assigned an id of the key.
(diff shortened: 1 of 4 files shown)


src/textual/app.py
```diff
@@ -822,27 +822,27 @@ def update_styles(self, node: DOMNode | None = None) -> None:
...
+        Args:
+            *widgets (Widget): The widget(s) to mount.
 
         Returns:
             AwaitMount: An awaitable object that waits for widgets to be mounted.
-
         """
-        mounted_widgets = self._register(self.screen, *anon_widgets, **widgets)
-        return AwaitMount(mounted_widgets)
+        return AwaitMount(self._register(self.screen, *widgets))
```
willmcgugan, line 834: I'd opt to assign a variable here. It makes it easier to insert print statements and makes it self documenting. Not a blocker though!

willmcgugan (approved):

PR #1088, opened 2022-11-01
Introduce _typing module for version-dependent typing imports

darrenburns:


src/textual/_typing.py
```diff
@@ -0,0 +1,11 @@
+import sys
+
+if sys.version_info >= (3, 10):
+    from typing import TypeAlias
+else:  # pragma: no cover
+    from typing_extensions import TypeAlias
+
+if sys.version_info >= (3, 8):
+    from typing import Final, Iterable, Literal, Protocol, TypedDict
```
willmcgugan, line 9: Should `Iterable` be in here? Thats been around for a long time.

willmcgugan (approved):
Great. One possible oversight.

src/textual/_typing.py
darrenburns: Nope, it shouldn't. Removed

Issue #778, opened 2022-09-15
mount should have additional options to insert new widgets


willmcgugan:
Currently, mount() inserts new widgets at the end of the list.

We should provide some way of inserting widgets at the beginning and arbitrary point.

I'm thinking an `before` and `after` parameter which would accept an integer or a CSS query. And insert the new widget at a matching index or query. Indexes can be negative, so `after=-1` would insert at the end of the list.

The default for before and after should be None, which would insert at the end of the list.

davep:
First thing to do, before tacking this, is to remove support for adding "anonymous" widgets vs "named" widgets. This will mean dropping support in the `App` and `Widget` `mount` methods, and also looking through the code, examples and documentation to remove any use and mention of this.

davep:
Having tidied up the parameters for the two `mount` methods (removing support for "named" widgets), the next thing to do is to look at [`NodeList`](https://github.com/Textualize/textual/blob/93e75f3f0e84b86883c253a30e4a4ec1b890ff1f/src/textual/_node_list.py`) with a view to adding other methods of getting a widget into the list, that isn't just a simple append.

davep:
From what I can see there are no unit tests for `NoteList` -- this might be the perfect moment to start to add some.

davep:
Reviewing the higher-level requirement, and thinking about how it could be used, I'd like to just recheck the approach suggested. As I understand it, right now, we have `mount`, which (if we can forgive relating this to web development for a moment) would seem to be equivalent to [`appendChild`](https://developer.mozilla.org/en-US/docs/Web/API/Node/appendChild). If that's the case, I can't help but feel that an analogous interface (which I suspect folk would be looking for), would be something akin to [`insertBefore`](https://developer.mozilla.org/en-US/docs/Web/API/Node/insertBefore) (although of course we may want both before and after versions). What I'm not sure I'm getting is when someone would want to work with numeric indexes.

Taking our stopwatch tutorial, with this in mind:

```python
def action_add_stopwatch(self) -> None:
    """An action to add a timer."""
    new_stopwatch = Stopwatch()
    self.query_one("#timers").mount(new_stopwatch)
    new_stopwatch.scroll_visible()
```

I can imagine wanting to turn this into "add after the focused stopwatch or at the end if one isn't focused", with some query to find the relevant `Stopwatch` within the `#timers` container resulting in something akin to:

```python
self.query_one("#timers").mount_after(focused_stopwatch, new_stopwatch)
```

[...]

PR #1095, opened 2022-11-02
Add support for mounting widgets relative to other widgets

davep:
Addresses the requirements in #778.

In short, `App.mount`, `App.mount_all` and `Widget.mount` now accept either a `before` or `after` keyword argument, which can be an integer, a string selector or a widget, and the new widget(s) will be mounted relative to them.

A quick sandbox app for testing some of this would be:

```python
from textual.widgets import Button, Header, Footer
from textual.reactive import reactive

class Foundation(Button):
    DEFAULT_CSS = """
    Foundation {
        width: 100%;
        border: solid red;
        color: red;
        background: #303030;
        text-align: center;
        margin: 1;
        margin-left: 10;
        margin-right: 10;
    }
    """

class Addition(Foundation):
    DEFAULT_CSS = """
    Addition {
        border: dashed yellow;
        color: yellow;
    }
    """

class MountTest( App[ None ] ):

    before = reactive(True)

    counter = reactive(0)

    BINDINGS = [
        ("t", "tree", "Dump Tree"),
        ("w", "where", "Toggle before/after"),
        ("m", "mount_many", "Mount Many"),
        ("i", "mount_here", "Mount Here"),
        ("s", "mount_pos(0)", "Mount At Start"),
        ("e", "mount_pos(-1)", "Mount At End"),
        *[Binding(str(n), f"mount({n})", f"Before {n}", False) for n in range(10)],
    ]
        yield Header()
        for n in range(10):
            yield Foundation(f"Foundation Widget {n}", id=f"foundation-{n}")
        yield Footer()

    def action_tree(self) -> None:
        self.log( self.tree )

    def action_where(self):
        self.before = not self.before

    def action_mount( self, n: int ) -> None:
        if self.before:
            self.screen.mount( Addition(f"Added {self.counter} BEFORE {n}"), before=f"#foundation-{n}")
        else:
            self.screen.mount( Addition(f"Added {self.counter} AFTER {n}"), after=f"#foundation-{n}")
        self.counter += 1

    def action_mount_many( self ) -> None:
        if self.before:
            self.mount_all(
                [ Addition( f"Many add {n} BEFORE 5") for n in range(3)],
                before=f"#foundation-5"
            )
        else:
            self.mount_all(
                [ Addition( f"Many add {n} AFTER 5") for n in range(3)],
                after=f"#foundation-5"
            )

    def action_mount_pos( self, pos: int ) -> None:
        if self.before:
            self.screen.mount( Addition(f"Added {self.counter} BEFORE INDEX {pos}"), before=pos)
        else:
            self.screen.mount( Addition(f"Added {self.counter} AFTER INDEX {pos}"), after=pos)
        self.counter += 1

    def action_mount_here( self ) -> None:
        if self.focused is not None:
            if self.before:
                self.screen.mount( Addition(f"Added {self.counter} BEFORE INDEX {self.focused!r}"), before=self.focused)
            else:
                self.screen.mount( Addition(f"Added {self.counter} AFTER INDEX {self.focused!r}"), after=self.focused)
            self.counter += 1

if __name__ == "__main__":
    MountTest().run()
```

**PLEASE NOTE**: This PR also contains a couple of small fixes to the `Pilot` as the unit tests for this PR found a couple of small issues.


src/textual/_node_list.py
```diff
@@ -39,6 +39,20 @@ def __len__(self) -> int:
     def __contains__(self, widget: Widget) -> bool:
         return widget in self._nodes
 
+    def _index(self, widget: Widget) -> int:
```
willmcgugan, line 42: I'd be tempted to make this public. Since this object is already list-like.

src/textual/dom.py
```diff
@@ -13,6 +13,7 @@
     TypeVar,
     cast,
     overload,
+    Union,
```
willmcgugan, line 16: Is this used?

willmcgugan (commented):
Looks great. Just comments ATM. Will do a review tomorrow.

src/textual/dom.py
davep: Good catch, nope. Did a lot of changes in there but backed them out in the end so that'll be a hangover.

src/textual/_node_list.py
davep: Will do. For anyone reading: there was a followup chat where we decided that we *do* want the read-only list-a-like methods exposed so they provide some utility when someone does things with `.children`; it's just the methods that modify the `NodeList` that should be treated as private.

davep: Done.

src/textual/app.py
```diff
@@ -1290,23 +1322,64 @@ def _on_idle(self) -> None:
             self._require_stylesheet_update.clear()
             self.stylesheet.update_nodes(nodes, animate=True)
 
-    def _register_child(self, parent: DOMNode, child: Widget) -> bool:
+    def _register_child(
+        self, parent: DOMNode, child: Widget, before: int | None, after: int | None
+    ) -> bool:
```
willmcgugan, line 1327: Docstring please (still like to document 'private' methods. I'm aware that the original version didn't have a docstring...

src/textual/app.py
```diff
@@ -1290,23 +1322,64 @@ def _on_idle(self) -> None:
...
+
+        # Let's be 100% sure that we've not been asked to do a before and an
+        # after at the same time. It's possible that we can remove this
+        # check later on, but for the purposes of development right now,
+        # it's likely a good idea to keep it here to check assumptions in
+        # the rest of the code.
+        if before is not None and after is not None:
+            raise AppError(
+                "A child can only be registered before or after, not before and after"
```
willmcgugan, line 1336:
```suggestion
                "Only one of 'before' and 'after' may be specified."
```

src/textual/app.py
```diff
@@ -1290,23 +1322,64 @@ def _on_idle(self) -> None:
...
+
+        # If we don't already know about this widget...
         if child not in self._registry:
-            parent.children._append(child)
+
+            # Now to figure out where to place it. If we've got a `before`...
+            if before is not None:
+                # ...it's safe to NodeList._insert before that location.
+                parent.children._insert(before, child)
+            elif after is not None and after != -1:
```
willmcgugan, line 1346: Have you checked if this logic works with `after = -2`? or another negative index.

src/textual/widget.py
```diff
@@ -375,16 +375,103 @@ def _get_virtual_dom(self) -> Iterable[Widget]:
...
+        if before is not None and after is not None:
+            raise self.MountError(
+                "Only one of `before` or `after` can be handled -- not both"
+            )
+
+        # Decide the final resting place depending on what we've been asked
+        # to do.
+        if before is not None:
+            parent, before = self._find_mount_point(before)
+            self.log.debug(f"MOUNT under {parent!r} before {before!r} ")
```
willmcgugan, line 464: We probably don't need these debug logs in now.

src/textual/app.py
davep: Added. Also drop the return value as it is never used by the single instance of a call.

src/textual/app.py
davep: Two more unit tests added that test `before=-2` and `after=-2` so we can be 100% sure.

PR #1101, opened 2022-11-03
Event queue

darrenburns:
This PR was to add a queue for input events, in order to make feeding events more predictable.

After some consideration, this *may* not be the right approach. So we are parking it for now.


willmcgugan:
Closing because the code is too stale. Not entirely given up on the idea.

Issue #1096, opened 2022-11-02
query_one should raise an exception


willmcgugan:
The `query_one` method should raise an exception if there isn't exactly one result from the query.

davep:
Once done, a TODO item in `test_widget_mounting.py` can be uncommented and put into play (if/when #1095 is merged).

PR #1102, opened 2022-11-03
Change query_one so that it raises an error on multiple hits

davep:
As part of this, add a only_one method to DOMQuery.

Addresses #1096.


willmcgugan (approved):
LGTM

PR #1104, opened 2022-11-03
writer thread

willmcgugan:
An experiment that uses a thread to write to stdout.

I discovered that writing to stdout was surprisingly slow. Some writes were in the order of 20ms. I suspect because it includes the time the terminal took to update the display.

Doing the writes in a thread makes them non-blocking, and allows us to work on the next frame at the same time.

### DIFF
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -143,7 +143,7 @@ class _WriterThread(threading.Thread):
 
     def __init__(self) -> None:
         super().__init__(daemon=True)
-        self._queue: Queue[str | None | threading.Event] = Queue(16)
+        self._queue: Queue[str | None] = Queue(32)
         self._file = sys.__stdout__
 
     def write(self, text: str) -> None:
@@ -156,21 +156,18 @@ class _WriterThread(threading.Thread):
         return self._file.fileno()
 
     def flush(self) -> None:
-        event = threading.Event()
-        self._queue.put(event)
-        event.wait()
-        self._file.flush()
+        return
 
     def run(self) -> None:
         write = self._file.write
+        flush = self._file.flush
+        get = self._queue.get
         while True:
-            text: str | threading.Event | None = self._queue.get()
-            if isinstance(text, threading.Event):
-                text.set()
-                continue
+            text: str | None = get()
             if text is None:
                 break
             write(text)
+            flush()
 
     def stop(self) -> None:
         self._queue.put(None)
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -162,12 +162,13 @@ class _WriterThread(threading.Thread):
         get = self._queue.get
         qsize = self._queue.qsize
         while True:
-            if not qsize():
-                flush()
             text: str | None = get()
+            empty = qsize() == 0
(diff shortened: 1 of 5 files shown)


src/textual/screen.py
```diff
@@ -19,7 +19,7 @@
 
 
 # Screen updates will be batched so that they don't happen more often than 60 times per second:
```
UriyaHarpeness, line 21:
```suggestion
# Screen updates will be batched so that they don't happen more often than 120 times per second:
```

src/textual/app.py
```diff
@@ -128,13 +129,54 @@ class CssPathError(Exception):
...
     def flush(self) -> None:
         pass
 
 
+class _WriterThread(threading.Thread):
+    """A thread / file-like to do writes to stdout in the background."""
+
+    def __init__(self) -> None:
+        super().__init__(daemon=True)
+        self._queue: Queue[str | None] = Queue(32)
```
davep, line 146: What's the significance of `32` here? Might it be a good idea to make it a constant with a docstring to explain it, or at least a comment?

src/textual/app.py
```diff
@@ -128,13 +129,54 @@ class CssPathError(Exception):
...
+    """A file-like where writes go nowhere."""
+
     def write(self, text: str) -> None:
         pass
 
     def flush(self) -> None:
         pass
 
 
+class _WriterThread(threading.Thread):
```
davep, line 141: Does the docstring rule hold for methods of a "private" class, in the same way that it holds for private methods?

src/textual/screen.py
```diff
@@ -18,8 +18,8 @@
 from .widget import Widget
 
 
-# Screen updates will be batched so that they don't happen more often than 60 times per second:
-UPDATE_PERIOD: Final = 1 / 60
+# Screen updates will be batched so that they don't happen more often than 120 times per second:
```
davep, line 21: Worth turning the comment into a docstring?

src/textual/app.py
willmcgugan: Could use a constant.

src/textual/app.py
willmcgugan: In general, yes! Although this class is a file-like, which already as a well defined and documented interface. Seems superfluous to add docstrings.

willmcgugan: But you have shamed me in to adding docstrings!

src/textual/screen.py
willmcgugan: Don't think so. No reason for this to be in docs.

src/textual/app.py
willmcgugan: Its the number of enqueued writes before it blocks. Changed to a constant..

src/textual/screen.py
davep: That makes sense, although don't gorget that docstrings serve anyone working on the code as any good editor setup will show the documentation when focus is around that item (personally I try and write docstrings with that purpose in mind -- which is presumably why we also do them for `_private` methods and the like too?).

PR #1109, opened 2022-11-04
docs fixes

aaronst:
Couple small things.

### DIFF
--- a/docs/styles/border.md
+++ b/docs/styles/border.md
@@ -7,11 +7,11 @@ Borders may also be set individually for the four edges of a widget with the `bo
 ## Syntax
 
 ```
-border: [<COLOR>] [<BORDER VALUE>];
-border-top: [<COLOR>] [<BORDER VALUE>];
-border-right: [<COLOR>] [<BORDER VALUE>];
-border-bottom: [<COLOR>] [<BORDER VALUE>];
-border-left: [<COLOR>] [<BORDER VALUE>];
+border: [<BORDER VALUE>] [<COLOR>];
+border-top: [<BORDER VALUE>] [<COLOR>];
+border-right: [<BORDER VALUE>] [<COLOR>];
+border-bottom: [<BORDER VALUE>] [<COLOR>];
+border-left: [<BORDER VALUE>] [<COLOR>];
 ```
 
 ### Values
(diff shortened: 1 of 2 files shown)


docs/styles/border.md
```diff
@@ -7,11 +7,11 @@ Borders may also be set individually for the four edges of a widget with the `bo
 ## Syntax
 
 ```
-border: [<COLOR>] [<BORDER VALUE>];
-border-top: [<COLOR>] [<BORDER VALUE>];
-border-right: [<COLOR>] [<BORDER VALUE>];
-border-bottom: [<COLOR>] [<BORDER VALUE>];
-border-left: [<COLOR>] [<BORDER VALUE>];
+border: [<BORDER VALUE>] [<COLOR>];
```
willmcgugan, line 10: I think there are actually order-independant.

Which we don't have a nice way of expressing in this format currently.

aaronst: Ah okay, good to know! I was going off of the line above and examples.

> A border is set with a border value (see below) followed by a color.

willmcgugan: Yeah, that should probably be updated. Something like "A border is set with a border value *and* a color.

Of course, we could just enforce one way or the other, but it seems kind of arbitrary.

aaronst: Taking a closer look at all the different examples throughout the documentation, I'm noticing that `border` also accepts a single value or color? Should we update this page with examples of that, similar to `margin`?

aaronst: For example,

https://github.com/Textualize/textual/blob/66d3e571f35ad8cd841afdb8b6b019dac5f4d20f/docs/examples/guide/layout/combining_layouts.css#L18

willmcgugan: Yeah, probably.

willmcgugan: I might have to think about the pros and cons of being strict or relaxed. My gut feeling is that browser CSS is fairly relaxed when there is no ambiguity.

aaronst: Okay, would it be best if I left `border` alone then and save that for a larger effort? This PR can just be to update the `Footer` docstring.

aaronst: Looks like @davep got this one in #1136. I'll close this PR, happy to open another later.

davep: Ahh damn, sorry about that - didn't see the matching fix within this (someone reported the same late last evening so I fixed it quickly).

aaronst: Oh no worries at all, was happy to see the change get in! Just trying to be mindful of clutter.

PR #1112, opened 2022-11-04
Replace f-string without interpolated values with string

overflowy:


overflowy:
Closed for committing with the wrong email address.

Issue #1106, opened 2022-11-03
Enable `textual run` to run a command that doesn't have a `.py` extension


davep:
Currently, if you have a Textual app that is in a file that doesn't have a `.py` extension -- something you may do if you're creating a command that uses `#!/usr/bin/env python` to do the expected thing -- `textual run` can't run the command.

PR #1116, opened 2022-11-04
Let `textual run` run files that don't end in .py

davep:
It's not uncommon to create "standalone" commands in Python that don't end in a .py and which:

  `#!/usr/bin/env python`

or a variation on that theme. This commit makes it so that `textual run` will treat a file that's named and set to run like that as if it were a file that ends in .py.

See #1106.

### DIFF
--- a/src/textual/_import_app.py
+++ b/src/textual/_import_app.py
@@ -28,9 +28,9 @@ def shebang_python(candidate: Path) -> bool:
     try:
         with candidate.open("rb") as source:
             first_line = source.readline()
-        return first_line.startswith(b"#!") and b"python" in first_line
     except IOError:
         return False
+    return first_line.startswith(b"#!") and b"python" in first_line
 
 
 def import_app(import_name: str) -> App:
--- a/src/textual/_import_app.py
+++ b/src/textual/_import_app.py
@@ -16,7 +16,7 @@ class AppFail(Exception):
     pass
 
 
-def shebang_python(candidate: Path):
+def shebang_python(candidate: Path) -> bool:
     """Does the given file look like it's run with Python?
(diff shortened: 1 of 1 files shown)


src/textual/_import_app.py
```diff
@@ -16,6 +16,23 @@ class AppFail(Exception):
     pass
 
 
+def shebang_python(candidate: Path):
```
willmcgugan, line 19: Return type por favor.

src/textual/_import_app.py
```diff
@@ -16,6 +16,23 @@ class AppFail(Exception):
...
+    Args:
+        candidate (Path): The candidate file to check.
+
+    Returns:
+        bool: ``True`` if it looks to #! python, ``False`` if not.
+    """
+    try:
+        with candidate.open("rb") as source:
+            first_line = source.readline()
+        return first_line.startswith(b"#!") and b"python" in first_line
```
willmcgugan, line 31: Could you put the return after the except? Just to keep the exception block as focused as possible.

willmcgugan (changes requested):
small requests.

src/textual/_import_app.py
davep: 🤦

src/textual/_import_app.py
davep: Aye, good call, will do.

src/textual/_import_app.py
davep: Done.

src/textual/_import_app.py
davep: Done.

Issue #1090, opened 2022-11-01
Docs: Specify dependencies for example apps


bryophyta:
First I just wanted to say that I really like the way that the new docs are taking shape! And I really like the inclusion of some example apps with instructions on how to run them.

The issue I wanted to raise is about how the dependencies for the examples are documented. The [Examples section](https://github.com/Textualize/textual/blob/main/docs/getting_started.md#examples) of `Getting Started` seems to imply that if you follow the steps in that section then you'll be able to run the examples. However, there's a dependency on `textual` and for some (e.g. `dictionary.py`) on `httpx`, too. So the code samples will only run if the user either has `textual` installed globally, or else is in an activated virtual environment which has it installed.

To my mind, there are two likely user journeys for this section, both of which might run into some trouble:

## 1. Relative newcomers to python
The current assumption seems to be that users will have a global install of `textual` because they will have run `pip install "textual[dev]"` or `pip install textual` after reading the 'Installation' section earlier on the page.

I can see an argument for not getting into the weeds of virtual environments, in order to keep things as approachable as possible. But to my mind the way that the section is written doesn't make it clear that Examples depends on Installation. (Contrast the 'Demo' section, which begins, `Once you have Textual installed...`.)

## 2. More experienced python developers
More experienced devs are (I think) unlikely to run the global install command when reading the Installation section. Because the Examples section started with cloning the `textual` repo, my personal assumption was that we'd create a new virtual environment for the project, which would have the dependencies needed to run the examples.

The natural options in my mind are either:

1. The examples can be run from the project's own dependencies, as specified in `poetry.lock`, or
2. Create a new virtual environment for the examples themselves, most likely in the `/examples/` directory.

Because `httpx` isn't in the project dependencies, (1) doesn't seem like the way to go. If (2) is the intended route for readers who aren't running a global install, I wonder whether it would be possible to add a `requirements.txt` file to the `/examples/` dir? (I know that there are arguments against using `pip` but the `requirements.txt` pattern is a widely understood convention, and pip is already being recommended in the docs and [in the example files themselves](https://github.com/Textualize/textual/blob/cc376a7cd5547286557c5068f211e4e8119953d0/examples/dictionary.py#L8).)

## tl:dr

[...]

PR #1126, opened 2022-11-05
Document requirements for example apps

bryophyta:
Adds a `requirements.txt` file in `./examples/`, and some documentation of dependencies in the README.

Specifically, the direct dependencies at the time of writing are Textual (for all of the example apps) and httpx (for `dictionary.py`) the file includes both of these, plus their dependencies, according to `pip freeze`. The `requirements.txt` file in this PR is the output of:

```sh
python -m venv venv
venv bin activate
pip install textual
pip install httpx
pip freeze -r requirements.txt
```

See #1090

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -2,7 +2,7 @@
 
 This directory contains example Textual applications.
 
-To run them, navigate to the examples directory and enter `python` followed buy the name of the Python file.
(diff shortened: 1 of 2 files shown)


willmcgugan:
Don't think we need this. It will make things harder to keep up to date.

Issue #1107, opened 2022-11-03
Can Textual widgets be made pylint-friendly?


davep:
Like many folk I'm a fan of using `pylint` to keep me honest when writing my Python code. However, given code like this:

```python
from textual.widgets import Static, Header, Footer
```

I get complaints from `pylint` like this:

```
pipenv run pylint mandelplot
************* Module mandelplot
mandelplot:24:0: E0611: No name 'Static' in module 'textual.widgets' (no-name-in-module)
mandelplot:24:0: E0611: No name 'Header' in module 'textual.widgets' (no-name-in-module)
mandelplot:24:0: E0611: No name 'Footer' in module 'textual.widgets' (no-name-in-module)

------------------------------------------------------------------
Your code has been rated at 8.51/10 (previous run: 8.51/10, +0.00)
```

Presumably because of the lazy-loading mechanism in use.

Is there a good way of keeping pylint happy here?

aaronst:
Widgets could be imported under `if typing.TYPE_CHECKING`? Probably a hacky solution but it seems to make `pylint` happy.

```python
if typing.TYPE_CHECKING:
    from ..widget import Widget
    from ._button import Button
    from ._checkbox import Checkbox
    from ._data_table import DataTable
    from ._directory_tree import DirectoryTree
    from ._footer import Footer
    from ._header import Header
    from ._placeholder import Placeholder
    from ._pretty import Pretty
    from ._static import Static
    from ._input import Input
    from ._text_log import TextLog
    from ._tree_control import TreeControl
    from ._welcome import Welcome
```

willmcgugan:
Good idea!

PR #1127, opened 2022-11-05
fix: pylint no-name-in-module for widgets

aaronst:
For #1107, we can import Widgets under `if typing.TYPE_CHECKING:` to avoid `no-name-in-module` errors from `pylint` when importing from `textual.widgets`. I also updated and moved the comment to reflect the additional step needed for new built-in Widgets.


davep (approved):
Looks good.

PR #1130, opened 2022-11-07
Support focusable containers

darrenburns:
Seems like if something was a container we never allowed it to focus at all, even if can_focus was True.


willmcgugan (approved):

willmcgugan (approved):
LGTM

PR #1139, opened 2022-11-08
Minimal refactoring

overflowy:


overflowy:
This is so weird, I have absolutely no idea why this is failing.

src/textual/css/scalar.py
```diff
@@ -259,13 +259,12 @@ def parse(cls, token: str, percent_unit: Unit = Unit.WIDTH) -> Scalar:
         """
         if token.lower() == "auto":
             scalar = cls(1.0, Unit.AUTO, Unit.AUTO)
-        else:
-            match = _MATCH_SCALAR(token)
-            if match is None:
-                raise ScalarParseError(f"{token!r} is not a valid scalar")
-            value, unit_name = match.groups()
-            scalar = cls(float(value), SYMBOL_UNIT[unit_name or ""], percent_unit)
-        return scalar
```
willmcgugan, line 268: This return statement is missing from the RHS

overflowy: Thank you, I was looking at the wrong commit

willmcgugan (approved):

willmcgugan:
Thanks

PR #1143, opened 2022-11-09
List view

darrenburns:
image


src/textual/dom.py
```diff
@@ -259,7 +259,7 @@ def screen(self) -> "Screen":
         from .screen import Screen
 
         node = self
-        while node and not isinstance(node, Screen):
+        while node is not None and not isinstance(node, Screen):
```
darrenburns, line 262: This is the line that ultimately resulted in those weird `NoScreen` errors. Because ListView has a __len__, the condition was evaluating and terminating the search for the Screen.

src/textual/widgets/_list_view.py
```diff
@@ -0,0 +1,130 @@
...
+from textual.geometry import clamp
+from textual.message import Message
+from textual.reactive import reactive
+from textual.widgets._list_item import ListItem
+
+
+class ListView(Vertical, can_focus=True, can_focus_children=False):
+    DEFAULT_CSS = """
+    ListView {
+        scrollbar-size-vertical: 1;
```
darrenburns, line 18: Subjective, but I prefer width 1 scrollbars for scrollbars that aren't Screen-level.

src/textual/dom.py
willmcgugan: Ah. Good catch!

src/textual/widgets/_list_item.py
```diff
@@ -0,0 +1,39 @@
...
+    ListItem {
+        color: $text;
+        height: auto;
+        background: $panel-lighten-1;
+        overflow: hidden hidden;
+    }
+    ListItem > Widget :hover {
+        background: $boost;
+    }
+    ListView ListItem.--highlight {
```
willmcgugan, line 18: Should we add a `>` here? For the (hopefully rare) scenario of nested ListViews

src/textual/widgets/_list_view.py
```diff
@@ -0,0 +1,145 @@
...
+    }
+    """
+
+    BINDINGS = [
+        Binding("down", "down", "Down"),
```
willmcgugan, line 23: In the tree view I've called these `cursor_up` and `cursor_down`. I think we should adopt that as a convention. So we can standardise bindings.

src/textual/widgets/_list_view.py
```diff
@@ -0,0 +1,145 @@
...
+        name: str | None = None,
+        id: str | None = None,
+        classes: str | None = None,
+    ) -> None:
+        super().__init__(*children, name=name, id=id, classes=classes)
+        self.index = initial_index
+
+    @property
+    def highlighted_child(self) -> ListItem | None:
+        """Get the currently highlighted ListItem
```
willmcgugan, line 43: I think we're going to adopt property style docstrings in this format:

"""ListItem | None: description"""

src/textual/widgets/_list_view.py
```diff
@@ -0,0 +1,145 @@
...
+        if self._is_valid_index(new_index):
+            new_child = self.children[new_index]
+            new_child.highlighted = True
+        else:
+            new_child = None
+
+        self._scroll_highlighted_region()
+        self.emit_no_wait(self.Highlighted(self, new_child))
+
+    async def append(self, item: ListItem) -> None:
```
willmcgugan, line 82: Could we make this non-async?

Use `emit_no_wait` and return the awaitable from the `mount` method.

Hopefully that won't break anything.

Ditto for `clear`

willmcgugan (changes requested):
Nice. Some requests to consider.

We're kind of finding our feet here. As we build these widgets, I think we will formulate some best practices / standards.

src/textual/widgets/_list_view.py
```diff
@@ -0,0 +1,145 @@
...
+    def action_select(self) -> None:
+        selected_child = self.highlighted_child
+        self.emit_no_wait(self.Selected(self, selected_child))
+
+    def on_list_item__child_clicked(self, event: ListItem._ChildClicked) -> None:
+        self.focus()
+        self.index = self.children.index(event.sender)
+        self.emit_no_wait(self.Selected(self, event.sender))
+
+    def key_up(self, event: events.Key) -> None:
```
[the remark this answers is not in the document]
darrenburns: Yeah, this is where the handling is actually done. The bindings for up/down are dummies just to populate the footer. As far as I know when you use bindings you have no way to stop the event and so it was doing Textuals default scrolling behaviour when you pressed up and down.

I've hit a similar issue several times now - maybe I'm just missing something though. In my case I actually very rarely want the scrolling to occur when up/down is pressed.

src/textual/widgets/_list_view.py
darrenburns: I've updated them all to match the tree view exactly.

src/textual/widgets/_list_view.py
darrenburns: Technically yes, but I think this will cause a lot confusion/subtle bugs. It needs to be async because we need to `await` the call to `remove` inside it. Otherwise we'll send a `ChildrenUpdated` with undefined `children` (could contain children before or after update, who knows).

If I receive a `ChildrenUpdated` message, I won't know inside my `on_children_updated` handler if the children have actually been updated yet or not, and I'll have no easy way of `await`ing it (I won't have access to the `AwaitRemove` in there). It would mean we can't actually trust that message any more since the state of the children is undefined. Doing anything to my ListView on receipt of this message would be dangerous.

willmcgugan: Do we need a children update message? Generally it is the parent that is doing the updating, so it will know there has been an updated.

Do you have a use-case in mind?

src/textual/widgets/_list_view.py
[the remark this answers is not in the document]
willmcgugan: Ah, I think that is because the Widget class was using key events. I changed that in the tree pr to use bindings. You might find it will work with bindings now.

src/textual/widgets/_list_view.py
```diff
@@ -0,0 +1,140 @@
...
+    def _clamp_index(self, index: int) -> int:
+        """Clamp the index to a valid value given the current list of children"""
+        last_index = max(len(self.children) - 1, 0)
+        return clamp(index, 0, last_index)
+
+    def _is_valid_index(self, index: int | None) -> bool:
+        """Return True if the current index is valid given the current list of children"""
+        if index is None:
+            return False
+        return 0 <= index < len(self.children)
```
rodrigogiraoserrao, line 62: I find something like ↓ nicer. Is it, or should I see a doctor about my preferences?

```suggestion
        return index is not None and 0 <= self.index < len(self.children)
```

src/textual/widgets/_list_view.py
```diff
@@ -0,0 +1,140 @@
...
+    def highlighted_child(self) -> ListItem | None:
+        """ListItem | None: The currently highlighted ListItem,
+        or None if nothing is highlighted.
+        """
+        if self.index is None:
+            return None
+        elif 0 <= self.index < len(self.children):
+            return self.children[self.index]
+
+    def validate_index(self, index: int | None) -> int | None:
```
rodrigogiraoserrao, line 48: Missing docstring.

src/textual/widgets/_list_view.py
```diff
@@ -0,0 +1,140 @@
...
+
+    BINDINGS = [
+        Binding("down", "cursor_down", "Down", show=False),
+        Binding("up", "cursor_up", "Up", show=False),
+        Binding("enter", "select_cursor", "Select", show=False),
+    ]
+
+    index = reactive(0, always_update=True)
+
+    def __init__(
```
rodrigogiraoserrao, line 27: Missing docstring.

rodrigogiraoserrao (changes requested):

docs/widgets/list_view.md
```diff
@@ -0,0 +1,78 @@
...
+You can select a list item by pressing ++enter++ while it is highlighted,
+or by clicking on it.
+
+- [x] Bubbles
+
+#### Attributes
+
+| attribute | type       | purpose                     |
+|-----------|------------|-----------------------------|
+| `item`    | `ListItem` | The item that was selected. |
```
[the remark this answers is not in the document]
darrenburns: Yeah, this is mostly just an ergonomics thing, the one from the ListItem gets sent to the ListView to say "hey, I'm the child object that was selected". The ListView captures that then emits its own message (keeps both messages on ListView).

src/textual/widgets/_list_view.py
darrenburns: I prefer to keep the two cases as I find it more readable :)

src/textual/widgets/_list_view.py
darrenburns: Added

src/textual/widgets/_list_view.py
darrenburns: Added

tests/snapshot_tests/test_snapshots.py
```diff
@@ -92,6 +92,10 @@ def test_header_render(snap_compare):
     assert snap_compare(WIDGET_EXAMPLES_DIR / "header.py")
 
 
+def test_list_view(snap_compare):
```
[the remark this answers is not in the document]
darrenburns: I don't think that would add much since we can't assert that events are fired. I think we spoke in the past about how it would be nice to assert that, for example, when I press Enter, a ListView.Selected event is fired.

src/textual/widgets/_list_view.py
darrenburns: I've removed it, don't think it's necessary.

It was to give the ancestors a single hook (the children update handler) that they could use when the children changed. e.g. a single place they can update a label to reflect the number of items in the list rather.

rodrigogiraoserrao (approved):

willmcgugan (approved):

PR #1144, opened 2022-11-09
Add easing parameter to Widget.scroll_* methods

darrenburns:


willmcgugan (changes requested):
Just had a thought. Slight change from the issue...

I think what we should do is allow easing to be `None` by default. And in `scroll_to` we should have a `DEFAULT_SCROLL_EASING` constant.

That way we don't need to repeat "out_cubic" in these methods. It will also allow us to change the scroll easing, or move it in to config at a later date.

willmcgugan (approved):

PR #1146, opened 2022-11-09
Add get_child_by_id and get_widget_by_id

darrenburns:
- Maintains dict of ID -> Widget inside the NodeList for quick access children by ID
- Removes DOMNode.get_child
- Removes App.get_child
- Adds Widget.get_child_by_id
- Adds App.get_child_by_id (delegates to self.screen.get_child_by_id)
- Adds Widget.get_widget_by_id
- Adds App.get_widget_by_id (delegates to self.screen.get_widget_by_id)


src/textual/_node_list.py
```diff
@@ -62,6 +71,9 @@ def _append(self, widget: Widget) -> None:
         if widget not in self._nodes_set:
             self._nodes.append(widget)
             self._nodes_set.add(widget)
+            widget_id = widget.id
+            if widget_id is not None and widget_id not in self._nodes_by_id:
```
willmcgugan, line 75: I think I see an issue.

Widgets can have duplicate IDs currently. If we add widget A and widget B with an id of "foo". Then widget A gets removed, I think that widget B will not be in the `_nodes_by_id` cache.

I think this means that `_nodes_by_id` should be a dict of lists.

willmcgugan (changes requested):
Noted an potential issue with duplicate IDs.

src/textual/_node_list.py
darrenburns: Nice one, very good catch

src/textual/_node_list.py
```diff
@@ -73,8 +90,19 @@ def _insert(self, index: int, widget: Widget) -> None:
...
+            if widget_id is not None:
+                self._ensure_unique_id(widget_id)
+                self._nodes_by_id[widget_id] = widget
             self._updates += 1
 
+    def _ensure_unique_id(self, widget_id: str) -> None:
+        if widget_id in self._nodes_by_id:
+            raise DuplicateIds(
+                f"Tried to insert a widget with ID {widget_id!r}, but a widget {self._nodes_by_id[widget_id]!r} "
+                f"already exists with that ID. Widget IDs must be unique."
```
willmcgugan, line 103: Probably worth specifying that widgets should be unique within a given children list. Could you rephrase to cover that?

src/textual/widget.py
```diff
@@ -333,6 +335,44 @@ def offset(self) -> Offset:
...
+        Args:
+            id (str): The ID to search for in the subtree
+
+        Returns:
+            DOMNode: The first descendant encountered with this ID.
+
+        Raises:
+            NoMatches: if no children could be found for this ID
+        """
+        for child in self.walk_children(method="depth"):
```
willmcgugan, line 369: This will walk the entire DOM since `walk_children` returns a list.

so performance will unfortunately be much the same. I think we might need a more specialised way of walking children for this method.

I'm also think that the walk should be breadth first. If there happens to be multiple IDs, I think you would expect the most immediate matching node to be returned.

src/textual/widget.py
```diff
@@ -460,6 +500,20 @@ def mount(
             provided a ``MountError`` will be raised.
         """
 
+        # Check for duplicate IDs in the incoming widgets
+        ids_to_mount = [widget.id for widget in widgets if widget.id is not None]
+        unique_ids = set(ids_to_mount)
+        num_unique_ids = len(unique_ids)
+        num_widgets_with_ids = len(ids_to_mount)
+        if num_unique_ids != num_widgets_with_ids:
```
willmcgugan, line 508: Bravo for detailed error messages. I have a feeling this may be overkill though. Would it be enough to say "Can't mount widgets with duplicate IDs" ?

willmcgugan: I withdraw my objection.

src/textual/widget.py
willmcgugan: Scratch that. I'm thinking you were right the first time in `"depth"`. Other wise `query_one` and `get_widget_by_id` could return different values.

willmcgugan:
@darrenburns Could you do a quick pass on this one next change you get.

src/textual/_node_list.py
darrenburns: I've reworded it

src/textual/widget.py
darrenburns: Have updated to use `walk_depth_first`

willmcgugan (approved):

PR #1147, opened 2022-11-09
Improve the speed of `DOMQuery.remove`

davep:
See https://github.com/Textualize/textual/discussions/1110 for some context. In short, this changes the `Remove` event so that it can take a collection of widgets to remove, rather than a single widget. The way that removal is done is then changed to dismantle the relevant parts of the DOM first, perform the closedown of the widgets involved, and then finally do a single refresh operation.

Note that, along the way, what was `DOMNode.ancestors` has become `DOMNode.ancestors_with_self` and `DOMNode.ancestors` now no longer contains `self`.

### DIFF
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -1859,10 +1859,11 @@ class App(Generic[ReturnType], DOMNode):
         # In other words: find the smallest set of ancestors in the DOM that
         # will remove the widgets requested for removal, and also ensure
         # that all knock-on effects happen too.
+        request_remove = set(event.widgets)
         pruned_remove = [
             widget
             for widget in event.widgets
-            if dedupe_to_remove.isdisjoint(widget.ancestors)
+            if request_remove.isdisjoint(widget.ancestors)
         ]
 
         # Now that we know that minimal set of widgets, we go through them
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -1863,13 +1863,17 @@ class App(Generic[ReturnType], DOMNode):
             )
         ]
 
+        # Now that we know that minimal set of widgets, we go through them
+        # and get their parents to forget about them. This has the effect of
+        # snipping each affected branch from the DOM.
         for widget in pruned_remove:
             if widget.parent is not None:
                 widget.parent.children._remove(widget)
 
-        # Now that we have the minimal set of widgets that need to be
-        # removed from the DOM, to get the effect of removing everything
-        # affected, let's go prune them.
+        # Havnig done that, it's now safe for us to start the process of
+        # winding down all of the affected widgets. We do that by pruning
+        # just the roots of each affected branch, and letting the normal
+        # prune process take care of all the offspring.
         for widget in pruned_remove:
             await self._prune_node(widget)
 
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -1652,7 +1652,9 @@ class App(Generic[ReturnType], DOMNode):
                 (self, self._bindings),
             ]
         else:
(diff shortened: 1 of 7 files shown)


src/textual/dom.py
```diff
@@ -561,6 +564,12 @@ def ancestors(self) -> list[DOMNode]:
             node = node._parent
         return cast("list[DOMNode]", nodes)
 
+    @property
+    def ancestors(self) -> list[DOMNode]:
+        """list[DOMNode]: A list of ancestor nodes Nodes by tracing ancestors all the way back to App."""
+        nodes = self.ancestors_with_self
+        return nodes[1:] if nodes else nodes
```
willmcgugan, line 571: I think we can guarantee that nodes has at least one node.

src/textual/app.py
```diff
@@ -1822,23 +1824,66 @@ async def _on_resize(self, event: events.Resize) -> None:
...
+        # and reduce it to the minimal set of things that need removing. In
+        # other words, for any given branch that is going to have things
+        # removed, we need to just keep the topmost node in the DOM that is
+        # a candidate for removal. Note that we retain the ordering of the
+        # original walk at the top as it's depth-first and so should be the
+        # order we want to go in from here on.
+        pruned_remove: list[Widget] = [
+            widget
+            for widget in everything_to_remove
+            if not any(
```
willmcgugan, line 1866: I think we can replace the if condition with a set operation:

```python
isdisjoint = everything_to_remove
pruned_remove = [
    widget for widget in everything_to_remove
    if isdisjoint(widget.ancestors)
]
```

src/textual/app.py
```diff
@@ -1822,23 +1824,66 @@ async def _on_resize(self, event: events.Resize) -> None:
...
+        ]
+
+        # Now that we know that minimal set of widgets, we go through them
+        # and get their parents to forget about them. This has the effect of
+        # snipping each affected branch from the DOM.
+        for widget in pruned_remove:
+            if widget.parent is not None:
+                widget.parent.children._remove(widget)
+
+        # Havnig done that, it's now safe for us to start the process of
```
willmcgugan, line 1878:
```suggestion
        # Having done that, it's now safe for us to start the process of
```

willmcgugan (changes requested):
Great. Suggested an optimization.

src/textual/dom.py
davep: Good point. :-D

src/textual/app.py
davep: Like this? https://github.com/Textualize/textual/pull/1147/commits/3cb00f93d5e2b000f2fa19951b5338ac0ee59180

src/textual/app.py
```diff
@@ -1822,23 +1824,64 @@ async def _on_resize(self, event: events.Resize) -> None:
...
+        # Next, we want to go through the full set of everything to remove
+        # and reduce it to the minimal set of things that need removing. In
+        # other words, for any given branch that is going to have things
+        # removed, we need to just keep the topmost node in the DOM that is
+        # a candidate for removal. Note that we retain the ordering of the
+        # original walk at the top as it's depth-first and so should be the
+        # order we want to go in from here on.
+        pruned_remove = [
+            widget
+            for widget in everything_to_remove
```
willmcgugan, line 1865: I have a feeling we are doing more work than required here. I think we could get the same result by iterating over `event.widgets`.

davep: I think you're right, good catch.

davep: Done: a888e52c30191b17e1d885b64f3a0e2b61faacdf

willmcgugan (approved):

PR #1149, opened 2022-11-09
call later

willmcgugan:
`call_later` has been renamed `call_after_refresh` which guarantees a callback will be done after a refresh of the display.
A new `call_later` invokes a callback after all messages have been processed.


darrenburns (approved):
Looks good

Issue #1148, opened 2022-11-09
Ability to stop and start/resume application/driver


costrouc:
To be added by @willmcgugan

I'm looking at creating an application similar to [k9s](https://k9scli.io/) which has the feature where within the curses application you can edit text via whatever your $EDITOR environment variable is set to. It does the following:
 - suspends k9s
 - opens given editor (emacs, vim, vi, etc.) via exec
 - resume k9s for terminal keyboard events etc.

 I have started to experiment with exposing this in textual but can't seem to "resume" textual after stopping the driver. I'm likely doing this the wrong way. It doesn't seem to respond to keyboard events after calling `self._driver.start_application_mode()`.

If possible it would be nice to have a way to easily suspend/resume the given textual application.

```python
import tempfile
import subprocess

from rich.markdown import Markdown

from textual.app import App, ComposeResult
from textual import events
from textual.containers import Container
from textual.widgets import Footer, Header, Static


class GithubApp(App):
    """A working 'desktop' github interface for managing issues/discussions."""

    BINDINGS = [
        ("q", "quit", "Quit"),
        ("e", "edit", "Edit"),
    ]

    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)

    def compose(self) -> ComposeResult:
        """Compose our UI."""
        yield Header()
        yield Container(
            Static(id="text", markup=True),
            Static(id="render", markup=True),
        )
        yield Footer()

    async def action_edit(self) -> None:
        self._driver.stop_application_mode()
        with tempfile.NamedTemporaryFile() as tempf:
            subprocess.run(f"$EDITOR {tempf.name}", shell=True)
            with open(tempf.name) as f:
                contents = f.read()
                self.query_one("#text", Static).update(contents)
                self.query_one("#render", Static).update(Markdown(contents))
        self._driver.start_application_mode()



if __name__ == "__main__":
    GithubApp().run()
```

JoshKarpel:
> It doesn't seem to respond to keyboard events after calling `self._driver.start_application_mode()`.

I should have checked the issues - I just posted a PR related to this! https://github.com/Textualize/textual/pull/1150

TomJGooding:
I think this is now resolved by #4064?

Issue #1093, opened 2022-11-02
Launching subprocesses like Vim from Textual apps


darrenburns:
Right now if you launch a subprocess of say Vim or Nano from a Textual app, there's a race condition between Textual and the subprocess - some of the input makes it into the subprocess, but some gets eaten by Textual's driver thread.

Also, if the subprocess changes the terminal state, then after it exits the terminal state may not be what Textual expects. For example, the cursor may be visible.

We should explore how to make this easier, e.g. via a context manager/utility method for launching subprocesses which pause the driver and resumes after the subprocess exits.

davep:
See also https://github.com/Textualize/textual/discussions/165

JoshKarpel:
Just put up a related PR (I should have checked the issues first!) #1150

rustybrooks:
I had a similar problem but took a different approach.  In my case, I wanted to be able to run an interactive program within a widget in a textual app.  The attached code is pretty rough and dirty but it mostly works.  Some missing things:
1. it doesn't do color
2. there's no cursor - I'd have to track the cursor location and display it, which is doable
what it does is
1. opens a pseudo tty
2. opens whatever you want to run, giving have the ptty pair to the process as stdin/out/err
3. uses the "pyte" library to create a virtual screen sized the same size as the component
4. runs a scheduled job to check to see if there's any output from the ptty.  If there is, stream it to the pyte screen and replace the widget contents with the rendered screen
5. capture keystrokes and send them to the ptty input

I want to stress that this needs refinement and is just the first POC that actually worked for me.

You need to install "pyte" and of course textual.  You can run a test like
python ./lib/components/interactive_shell.py
right now it'll just try to run vi

```
import asyncio
import os
import pty
import select
import subprocess

import pyte
from textual import events
from textual.app import App
from textual.widgets import Static


class InteractiveShell(Static, can_focus=True):
    _screen = None
    _stream = None
    _pty = _tty = None

    DEFAULT_CSS = """
    InteractiveShell {
        width: 100%;
        height: 100%;
    }"""

    def __init__(
        self,
        command,
        exit_command,
        *,
        name: str | None = None,
        id: str | None = None,
        classes: str | None = None,
    ) -> None:
        super().__init__(
            "",
            name=name,
            id=id,
            classes=classes,
        )
        self.command = command
        self.exit_command = exit_command

    async def on_mount(self):
        self.focus()
        self.call_later(self._run)

    def on_key(self, event: events.Key) -> None:
[...]

PR #1150, opened 2022-11-10
Allow `Driver`s to handle key events after being restarted

JoshKarpel:
Change: clear driver exit events after key-handling threads are joined in the two concrete `Driver` implementations, which allows `Driver.start_application_mode()` to be called again successfully after `Driver.stop_application_mode()`.

This lets something like this work:

```python
import sys
import time
from contextlib import contextmanager, redirect_stdout, redirect_stderr
from typing import Iterator

from textual.widgets import Footer, Header


class SuspendApp(App):
    BINDINGS = [
        ("d", "toggle_dark", "Toggle dark mode"),
        ("s", "suspend", "Suspend and print a message"),
    ]

        """Create child widgets for the app."""
        yield Header()
        yield Footer()

    def action_toggle_dark(self) -> None:
        """An action to toggle dark mode."""
        self.dark = not self.dark

    def action_suspend(self) -> None:
        with self.suspend():
            print("Hi!")
            print("Resuming soon...")
            time.sleep(2)

    @contextmanager
    def suspend(self) -> Iterator[None]:
        driver = self._driver

        if driver is not None:
            driver.stop_application_mode()
            with redirect_stdout(sys.__stdout__), redirect_stderr(sys.__stderr__):
                yield
            driver.start_application_mode()


if __name__ == "__main__":
    SuspendApp().run()
```

Without this change, when application modes resumes at the end of `SuspendApp.suspend`, the `Driver`'s key thread will be started but will immediately stop because the exit event is still set, so the bindings won't work anymore. With this change, bindings work after the suspend finishes.

I put the event reset immediately after the key thread is joined to follow the general feeling of `Driver.stop_application_mode` trying to put the system back in its initial state.

---

I made a few attempts to write tests for this but was stymied by the dependence of the `Driver` on the event loop and a message pump to send message to - I didn't see an existing pattern in the test suite to copy for tests that rely on the event loop or message pumps, but maybe I just missed it? 

I've tested manually on Linux locally (using the above example script), but not on Windows or Mac. Happy to continue hacking on tests for this behavior since it'd be easy to inadvertently regress it, but I think I need some guidance on how to do so.

(For context, I'm using this trick to play around with dropping out of the Textual app and into another CLI app, namely IPython https://github.com/JoshKarpel/spiel/blob/9da45e1f22b4a284ac9691fa9125329598ed28ad/spiel/app.py#L151-L176)

### DIFF
--- a/src/textual/drivers/linux_driver.py
+++ b/src/textual/drivers/linux_driver.py
@@ -176,6 +176,7 @@ class LinuxDriver(Driver):
                 self.exit_event.set()
                 if self._key_thread is not None:
                     self._key_thread.join()
+                self.exit_event.clear()
                 termios.tcflush(self.fileno, termios.TCIFLUSH)
         except Exception as error:
             # TODO: log this
--- a/src/textual/drivers/windows_driver.py
+++ b/src/textual/drivers/windows_driver.py
@@ -84,6 +84,7 @@ class WindowsDriver(Driver):
                 if self._event_thread is not None:
                     self._event_thread.join()
                     self._event_thread = None
+                self.exit_event.clear()
         except Exception as error:
             # TODO: log this
             pass


src/textual/drivers/linux_driver.py
```diff
@@ -249,4 +250,4 @@ class MyApp(App):
         async def on_mount(self, event: events.Mount) -> None:
             self.set_timer(5, callback=self._close_messages)
 
-    MyApp.run()
+    MyApp().run()
```
JoshKarpel, line 253: Unrelated, just noticed this bug while poking around.

costrouc:
@JoshKarpel can confirm this works when I tested it on my code to drop the user into an editor and resuming textual. This will resolve issues #1148,  #1093 and discussion https://github.com/Textualize/textual/discussions/165.

darrenburns:
Thanks for this @JoshKarpel, it looks really useful - we'll try and review it soon!

darrenburns:
Can confirm this fixed the issue inside my file manager app :)

willmcgugan (approved):

JoshKarpel:
Thanks all!

PR #1157, opened 2022-11-10
Awaitable remove

davep:
This adds the ability to `await` a `Widget.remove` or a `DOMQuery.remove`. Doing this adds a new type -- `AwaitRemove` -- and also changes the return type of `DOMQuery.remove` (which used to return `self` but now returns an awaitable object).


willmcgugan (approved):
Great!

PR #1160, opened 2022-11-11
factored out walk

willmcgugan:
Factored out DOM walk functionality in to its own file.

The methods in `textual.walk` are the *low level* functionality. They are implemented as generators, which means we can use them to bail out early when searching. They are used by `Widget.walk_children`.

Debated wether `walk.py` should be private or note. Settled on a big warning in the docstrings. It could be useful to app developers, but they should be aware of the issues of walking while iterating over the DOM.

### DIFF
--- a/src/textual/dom.py
+++ b/src/textual/dom.py
@@ -659,7 +659,7 @@ class DOMNode(MessagePump):
         nodes = list(node_generator)
         if reverse:
             nodes.reverse()
-        return cast("list[WalkType]", nodes)
+        return cast("list[DOMNode]", nodes)
 
     def get_child(self, id: str) -> DOMNode:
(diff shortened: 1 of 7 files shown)


src/textual/dom.py
```diff
@@ -645,51 +645,20 @@ def walk_children(
...
+            else walk_breadth_first(self, check_type, with_root=with_self)
         )
 
         # We want a snapshot of the DOM at this point So that it doesn't
         # change mid-walk
         nodes = list(node_generator)
         if reverse:
             nodes.reverse()
-        return nodes
+        return cast("list[WalkType]", nodes)
```
willmcgugan, line 661: Couldn't avoid the cast. Any idea why?

PR #1165, opened 2022-11-11
feat: initial datatable reactive values

aaronst:
For #1164, adds the [six reactive attributes listed in the docs](https://textual.textualize.io/widgets/data_table/#reactive-attributes) for `DataTable` to `__init__`. I noticed the class has three more reactive attributes (`cursor_type`, `cursor_cell`, and `hover_cell`) that aren't listed, but I left those out. Let me know if any of them should also be included.

PR #1168, opened 2022-11-13
fix deadlock on tests

willmcgugan:
Fixed an issue where a failed snapshot test caused a deadlock. Highlighted by https://github.com/Textualize/textual/pull/1165

Essentially `animator.wait_for_idle()` was awaiting an event which never occurred.

### DIFF
--- a/src/textual/_animator.py
+++ b/src/textual/_animator.py
@@ -189,9 +189,12 @@ class Animator:
     async def stop(self) -> None:
         """Stop the animator task."""
         try:
-            await self._timer.stop()
-        except asyncio.CancelledError:
-            pass
+            try:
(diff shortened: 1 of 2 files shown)


src/textual/_animator.py
```diff
@@ -189,9 +189,12 @@ async def start(self) -> None:
     async def stop(self) -> None:
         """Stop the animator task."""
         try:
-            await self._timer.stop()
-        except asyncio.CancelledError:
-            pass
+            try:
```
darrenburns, line 192: Why the double nested try?

willmcgugan: That's a mistake, will fix.

PR #1170, opened 2022-11-13
WIP new tree control

willmcgugan:
New tree control

### DIFF
--- a/src/textual/widgets/_tree.py
+++ b/src/textual/widgets/_tree.py
@@ -1,0 +1,245 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import ClassVar, Generic, TypeVar
+
+import rich.repr
+from rich.segment import Segment
+from rich.style import Style
+from rich.text import Text, TextType
+
+
+from ..binding import Binding
+from ..geometry import clamp, Region
+from .._loop import loop_last
+from .._cache import LRUCache
+from ..reactive import reactive
+from .._segment_tools import line_crop
+from ..scroll_view import ScrollView
+
+from .. import events
+
+TreeDataType = TypeVar("TreeDataType")
+
+
+@dataclass
+class _TreeLine:
+    parents: list[TreeNode]
+    node: TreeNode
+    last: bool
+
+
+@rich.repr.auto
+class TreeNode(Generic[TreeDataType]):
+    def __init__(
+        self,
+        tree: Tree,
+        parent: TreeNode[TreeDataType] | None,
+        label: Text,
+        data: TreeDataType,
+        *,
+        expanded: bool = True,
+    ) -> None:
+        self._tree = tree
+        self._parent = parent
+        self.label = label
+        self.data: TreeDataType = data
+        self.expanded = expanded
+        self.children: list[TreeNode] = []
+
+    def __rich_repr__(self) -> rich.repr.Result:
+        yield self.label.plain
+        yield self.data
+
+    @property
+    def last(self) -> bool:
+        if self._parent is None:
+            return True
+        return bool(
+            self._parent.children and self._parent.children[-1] == self,
+        )
+
+    def add(
+        self, label: TextType, data: TreeDataType, expanded: bool = True
+    ) -> TreeNode[TreeDataType]:
+        if isinstance(label, str):
+            text_label = Text.from_markup(label)
+        else:
+            text_label = label
+        node = TreeNode(
+            self._tree,
+            self,
+            text_label,
+            data,
+            expanded=expanded,
+        )
+        self.children.append(node)
+        self._tree.invalidate()
+        return node
+
+
+class Tree(Generic[TreeDataType], ScrollView, can_focus=True):
+
+    BINDINGS = [
+        Binding("up", "cursor_up", "Cursor Up"),
+        Binding("down", "cursor_down", "Cursor Down"),
+    ]
+
+    DEFAULT_CSS = """
+    Tree {
+        background: $surface;
+        color: $text;
+    }
+    Tree > .tree--guides {
+        color: $success;
+    }
+    Tree > .tree--cursor {
+        background: $secondary;
+        color: $text;
+    }
+    Tree > .tree--highlight {
+        
+        text-style: underline;
+    }
+    
+    """
+
+    show_root = reactive(True)
+
+    COMPONENT_CLASSES: ClassVar[set[str]] = {
+        "tree--guides",
+        "tree--cursor",
+        "tree--highlight",
+    }
+
+    hover_line = reactive(-1, repaint=False)
+    cursor_line = reactive(0, repaint=False)
+
+    def __init__(
+        self,
+        label: TextType,
+        data: TreeDataType,
+        *,
+        name: str | None = None,
+        id: str | None = None,
+        classes: str | None = None,
+    ) -> None:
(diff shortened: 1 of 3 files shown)


src/textual/css/styles.py
```diff
@@ -551,6 +551,22 @@ def _align_size(self, child: tuple[int, int], parent: tuple[int, int]) -> Offset
             self._align_height(height, parent_height),
         )
 
+    @property
+    def partial_rich_style(self) -> Style:
+        """Get the style properties associated with this node only (not including parents in the DOM).
+
+        Returns:
+            Style: Rich Style object
```
UriyaHarpeness, line 559:
```suggestion
            Style: Rich Style object.
```

src/textual/widget.py
```diff
@@ -374,7 +381,7 @@ def get_widget_by_id(self, id: str) -> Widget:
                 pass
         raise NoMatches(f"No descendant found with id={id!r}")
 
-    def get_component_rich_style(self, name: str) -> Style:
+    def get_component_rich_style(self, name: str, *, partial: bool = False) -> Style:
```
UriyaHarpeness, line 384: `partial` not documented

UriyaHarpeness (commented):
Saw it's a draft only after..

src/textual/css/styles.py
```diff
@@ -551,6 +551,22 @@ def _align_size(self, child: tuple[int, int], parent: tuple[int, int]) -> Offset
             self._align_height(height, parent_height),
         )
 
+    @property
+    def partial_rich_style(self) -> Style:
+        """Get the style properties associated with this node only (not including parents in the DOM).
```
davep, line 556: Perhaps use a more property-style docstring here rather than a function style?

src/textual/widget.py
```diff
@@ -966,6 +983,18 @@ def content_region(self) -> Region:
         content_region = self.region.shrink(self.styles.gutter)
         return content_region
 
+    @property
+    def scrollable_content_region(self) -> Region:
+        """Gets an absolute region containing the scrollable content (minus padding, border, and scrollbars).
```
davep, line 988: Would likely be better with a property-style docstring.

src/textual/widgets/_tree.py
```diff
@@ -0,0 +1,838 @@
...
+
+
+@dataclass
+class _TreeLine:
+    path: list[TreeNode]
+    last: bool
+
+    @property
+    def node(self) -> TreeNode:
+        """The node associated with this line."""
```
davep, line 41:
```suggestion
        """TreeNode: The node associated with this line."""
```

src/textual/widgets/_tree.py
```diff
@@ -0,0 +1,838 @@
...
+        data: TreeDataType | None = None,
+        *,
+        expand: bool = False,
+        allow_expand: bool = True,
+    ) -> TreeNode[TreeDataType]:
+        """Add a node to the sub-tree.
+
+        Args:
+            label (TextType): The new node's label.
+            data (TreeDataType): Data associated with the new node.
```
davep, line 189:
```suggestion
            data (TreeDataType | None, optional): Data associated with the new node.
```

src/textual/widgets/_tree.py
```diff
@@ -0,0 +1,838 @@
...
+        self._tree_lines_cached: list[_TreeLine] | None = None
+        self._cursor_node: TreeNode[TreeDataType] | None = None
+
+    @property
+    def cursor_node(self) -> TreeNode[TreeDataType] | None:
+        """TreeNode | Node: The currently selected node, or ``None`` if no selection."""
+        return self._cursor_node
+
+    @property
+    def last_line(self) -> int:
```
davep, line 360: Missing docstring

src/textual/widgets/_tree.py
```diff
@@ -0,0 +1,838 @@
...
+        self._updates += 1
+        return node
+
+    def render_label(
+        self, node: TreeNode[TreeDataType], base_style: Style, style: Style
+    ) -> Text:
+        """Render a label for the given node. Override this to modify how labels are rendered.
+
+        Args:
+            node (TreeNode[TreeDataType]): A tree node.
```
davep, line 397: `base_style` and `style` are missing from the docstring.

src/textual/widgets/_tree.py
```diff
@@ -0,0 +1,838 @@
...
+            TreeNode[TreeDataType] | None: A tree node, or ``None`` if there is no node at that line.
+        """
+        try:
+            line = self._tree_lines[line_no]
+        except IndexError:
+            return None
+        else:
+            return line.node
+
+    def validate_cursor_line(self, value: int) -> int:
```
davep, line 472: Docstring?

src/textual/widgets/_tree.py
```diff
@@ -0,0 +1,838 @@
...
+
+        self.virtual_size = Size(width, len(lines))
+        if self.cursor_line != -1:
+            if self.cursor_node is not None:
+                self.cursor_line = self.cursor_node._line
+            if self.cursor_line >= len(lines):
+                self.cursor_line = -1
+        self.refresh()
+
+    def render_line(self, y: int) -> list[Segment]:
```
davep, line 640: Docstring?

davep (approved):
Handful of "perhaps needs a docstring" and "perhaps needs to be a property-style docstring rather than function-style", that's all I could see.

src/textual/css/styles.py
willmcgugan: Do you mean the "int: description"?

I've found that mkdocstrings doesn't render them correctly. It just omits the type :-(

src/textual/widgets/_tree.py
```diff
@@ -0,0 +1,838 @@
...
+        "tree--label",
+        "tree--guides",
+        "tree--guides-hover",
+        "tree--guides-selected",
+        "tree--cursor",
+        "tree--highlight",
+        "tree--highlight-line",
+    }
+
+    show_root = reactive(True)
```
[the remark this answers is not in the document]
willmcgugan: Probably should, yeah.

src/textual/css/styles.py
davep: Yes and no. While that'd be ideal (and it does render really nicely in Emacs and VSCode hints, from what I've seen), I'm meaning more the wording. `"Get..."` sort of describing it as a function that does something, rather than as a property that has a value that has a particular meaning. So in this case likely just `"The style properties associated with this node only (not including parents in the DOM)`".

It's unfortunate that mkdocstrings doesn't handle it given it's, as far as I can tell, a documented aspect of the Google docstring convention.

PS: Just omits or just emits?

src/textual/widgets/_tree.py
davep: Good call; I've been on and off wondering about this too. I guess in the end too much documentation is better than none.

src/textual/css/styles.py
willmcgugan: The type in properties docstrings is just omitted. You see the description but no type.

With a function style, you do see the type.

Will assume that's a bug in mkdocstrings, and go for property style docs then.

davep: I may see about raising that with mkdocstrings then; it's unfortunate it does that.

PR #1185, opened 2022-11-15
Support callables in App.SCREENS

darrenburns:
Allows you to pass callables rather than Screen instances.


davep (approved):

src/textual/app.py
```diff
@@ -1532,14 +1536,14 @@ async def _close_all(self) -> None:
 
         # Close all screens on the stack
         for screen in self._screen_stack:
-            if screen._running:
+            if isinstance(screen, Screen) and screen._running:
```
willmcgugan, line 1539: Are we going to have something on the Screen stack that is not a screen?

darrenburns: No, good catch. Have removed this isinstance check

willmcgugan (approved):

Issue #1190, opened 2022-11-16
Add a `Label` widget type


davep:
The need to display text on a screen is obviously common, and often `Static` will be used. The problem here is that if you want to select all your "labels" in a query, or style them all in a particular way, you can't just style `Static` as it is an ancestor of some other widgets too.

So... perhaps let's add a `Label` widget that simply inherits from `Static` to help separate out text from other forms of widgets?

PR #1193, opened 2022-11-16
Add a Label widget

davep:
For the moment this does nothing more than inherit from a Static; but what it does do is make it easier for someone to add text to their application and to style it by styling all the Labels. Before now it would be common to use a Static but if you try and style (or query) all Statics, you'd also get things like Buttons, which inherit from Static.

See #1190

### DIFF
--- a/docs/widgets/label.md
+++ b/docs/widgets/label.md
@@ -3,7 +3,7 @@
 A widget which displays static text, but which can also contain more complex Rich renderables.
 
 - [ ] Focusable
-- [x] Container
+- [ ] Container
 
 ## Example
 
--- a/docs/widgets/label.md
+++ b/docs/widgets/label.md
@@ -1,0 +1,33 @@
+# Label
+
+A widget which displays static text, but which can also contain more complex Rich renderables.
+
+- [ ] Focusable
+- [x] Container
+
+## Example
+
+The example below shows how you can use a `Label` widget to display some text.
+
+=== "Output"
+
+    ```{.textual path="docs/examples/widgets/label.py"}
+    ```
+
+=== "label.py"
+
+    ```python
+    --8<-- "docs/examples/widgets/label.py"
+    ```
+
+## Reactive Attributes
+
+This widget has no reactive attributes.
+
+## Messages
+
+This widget sends no messages.
+
+## See Also
+
+* [Label](../api/label.md) code reference
--- a/src/textual/widgets/_label.py
+++ b/src/textual/widgets/_label.py
@@ -4,4 +4,4 @@ from ._static import Static
 
 
 class Label(Static):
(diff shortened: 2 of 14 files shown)


src/textual/widgets/_label.py
```diff
@@ -0,0 +1,7 @@
+"""Provides a simple Label widget."""
+
+from ._static import Static
+
+
+class Label(Static):
+    """A simple label widget for displaying text-oriented rendenrables."""
```
rodrigogiraoserrao, line 7:
```suggestion
    """A simple label widget for displaying text-oriented renderables"""
```

UriyaHarpeness:
This feels weird...
I get the idea but if a user would want to change the style of only part of them he'd still need to do work.
And if going with this approach, I think a warning message should be added if constructing a Static directly and not through inheriting it.
What do you think @davep?

davep:
> I get the idea but if a user would want to change the style of only part of them he'd still need to do work.

Obviously, hence classes and styles and the like in CSS. This early split between a more "baseline" renderable container and a more "has a specific job" renderable container isn't about reducing or eliminating that sort of work (which, as we both know, would be very little anyway), it just starts to set up what seems like a fairly natural split between a basic renderable container that's about showing text, and more involved widgets.

It's possible that at some point in the near future `Label` may acquire some extra functionality that wouldn't make much sense in `Static` but which anyone using `Label` would optionally want. This wee addition acknowledges that.

> And if going with this approach, I think a warning message should be added if constructing a Static directly and not through inheriting it.

I'm not really seeing much value in emitting warnings for this. As it stands, using `Static` rather than `Label` makes zero difference and has zero cost. What we're likely to do though is update the documentation and example code to naturally show that labels in an application will use `Label`, buttons will use `Button`, etc... and `Static` will sort of transitional into being a more foundational type.

docs/widgets/label.md
```diff
@@ -0,0 +1,33 @@
+# Label
+
+A widget which displays static text, but which can also contain more complex Rich renderables.
+
+- [ ] Focusable
+- [x] Container
```
willmcgugan, line 6: A Label can't contain children, which means it isn't a container. I think this is inherited from Static, which is also incorrect.

willmcgugan (approved):
Just one request for a doc tweak.

docs/widgets/label.md
davep: Good catch, I'll fix `Static` and then sync this.

davep: > A Label can't contain children, which means it isn't a container. I think this is inherited from Static, which is also incorrect.

Actually... this isn't this isn't quite true is it? Especially not for static. A Static can comprise of other widgets. Consider [the stopwatch tutorial](https://textual.textualize.io/tutorial/#custom-widgets) for example.

PR #1213, opened 2022-11-17
Sensible default key displays + allow users to override key displays at the `App` level

darrenburns:

- [x] Updated documentation

Example output from snapshot test in which user-land app overrides some stuff, and question mark comes from Textual defaults.

image


davep (approved):

aaronst:
Nice!

willmcgugan:
Looks good. Is `^q` a fairly standard convention then?

willmcgugan (approved):

davep:
> Looks good. Is `^q` a fairly standard convention then?

(Chiming in from the older days...) I'd absolutely read ^q as "control Q", it's been a convention as long as I've known.

darrenburns:
Note that this PR will not make the `^` appear in the footer - that's just what I done in my test of the user overrides. Textual will still show CTRL+Q.

PR #1219, opened 2022-11-18
fix remove freeze

willmcgugan:
Fixes a deadlock when removing widgets from the app.

Pruning the nodes is now down in a task. There's is a lock around the DOM to prevent it from rendering in the middle of a prune operation.


CHANGELOG.md
```diff
@@ -41,6 +41,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Fixed containers with transparent background not showing borders https://github.com/Textualize/textual/issues/1175
 - Fixed auto-width in horizontal containers https://github.com/Textualize/textual/pull/1155
 - Fixed Input cursor invisible when placeholder empty https://github.com/Textualize/textual/pull/1202
+- Fixed deadlock when remove widgets from the App https://github.com/Textualize/textual/pull/1219
```
davep, line 44: Perhaps "removing" rather than "remove" here?

davep (approved):

davep (approved):

Issue #1200, opened 2022-11-17
Build a "Placeholder" widget


willmcgugan:
We had a Placeholder widget which could be used to design a layout before you built a full widget. This use an older API.

We should refresh the placeholder idea, make it look nicer that the previous iteration, and add some features

- Pick harmonious alternating colors, by mixing brackground with ~30% from [this palette](https://iamkate.com/data/12-bit-rainbow/)
- Argument to select various modes: plain color, size and location, CSS, lorem ipsum text, etc
- Add click handler to cycle between modes

rodrigogiraoserrao:
With the commit https://github.com/Textualize/textual/commit/83c8a9b55f4ad9a141f23ebe9b2348ba12b06d52 we have the scaffolding for the widget `Placeholder` which:

 - has 5 different variants that display different things; and
 - cycles through the different variants when you click the placeholder.

 1. The `"default"` variant shows a solid colour (from the palette linked) that was associated to the placeholder upon instantiation.
 2. The `"state"` variant shows whether the placeholder has focus (or not) and whether the mouse is over the placeholder or not.
 3. The `"size"` variant shows the size of the placeholder.
 4. The `"css"` variant shows the CSS of the placeholder.
 5. The `"text"` variant shows some Lorem Ipsum text.

rodrigogiraoserrao:
Questions left unanswered:

 - Exactly _how_ should each variant be styled by default?
 - Should the background colour (from the palette linked) be customisable?
 - Should the placeholders scroll if the content is too large? (E.g., the css variant tends to have lots of content that won't fit small placeholders.)

rodrigogiraoserrao:
After some discussion, the variants for state and CSS have been removed from consideration.
Thus, placeholders have 3 variants.

PR #1229, opened 2022-11-18
Add Placeholder widget

rodrigogiraoserrao:

- [x] Updated documentation

Add the widget `Placeholder` as per #1200


davep (approved):

CHANGELOG.md
```diff
@@ -30,6 +30,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Support lazy-instantiated Screens (callables in App.SCREENS) https://github.com/Textualize/textual/pull/1185
 - Display of keys in footer has more sensible defaults https://github.com/Textualize/textual/pull/1213
 - Add App.get_key_display, allowing custom key_display App-wide https://github.com/Textualize/textual/pull/1213
+- Added widget `Placeholder`
```
davep, line 33: This may need moving up the CHANGELOG to be under 0.6.0.

src/textual/widgets/_placeholder.py
```diff
@@ -1,63 +1,166 @@
...
+        """Reset the placeholder background color cycle."""
+        cls.COLORS = cycle(_PLACEHOLDER_BACKGROUND_COLORS)
 
     def __init__(
-        # parent class constructor signature:
         self,
-        *children: Widget,
+        variant: PlaceholderVariant = "default",
+        *,
+        label: str | None = None,
```
willmcgugan, line 86: I think the label should be a positional arg. A placeholder without a label is of limited use, so we should require it.

willmcgugan (changes requested):
Looks good! Just one request.

src/textual/widgets/_placeholder.py
rodrigogiraoserrao: I don't understand. "We should require it" means it shouldn't have a default value of `None`, right?
It could be either positional or keyword.

When a placeholder doesn't get a label, it uses its ID (or the string "Placeholder" if there is no ID) so I think a placeholder without an explicit label _is_ useful. Especially when you have a bunch of placeholders and are playing around with them with live CSS editing.

So, do I make it positional or leave it as a keyword parameter?
And do I remove the default value of `None` so users have to pass in `None` explicitly, or do I remove the ability to receive `None` altogether?

willmcgugan: I think positional with no default. You wouldn’t need an id for a placeholder typically, because when you come to style it you would use a custom widget.

rodrigogiraoserrao: Your push in this direction makes me think I misunderstood how placeholders will be used.

My premise is that (1) placeholders are used _before_ the custom widgets exist, so to create and style a layout that uses placeholders can only work with IDs.

The only other alternative (2) that comes to me is to have empty classes for future placeholders inherit from `Placeholder`, something like:

```py
class MyFutureWidget(Placeholder):
    ...
```

If (2) is the intended usage pattern, then I 100% agree with your push, I'll make the change, and also edit the docs to explain this is the intended usage pattern.

If (2) is not how we want this to be used/if (1) is what we think will happen, then I think a placeholder should require the least amount of things possible, and thus I would push for `label` having a default value of `None`.

Either way, probably makes sense to move `label` to be the first parameter.

darrenburns: > A placeholder without a label is of limited use

I don't understand this because the old placeholders didn't require labels and they were very useful - almost every time I've used a placeholder in the past has been without a label. Almost every example of the old placeholder in the Textual repo was also without a label.

willmcgugan: Works you not want some kind of label? By that I mean some text that identifies what the placeholder will become?

My thinking is the first arg should be something like “options side bar” or “user message”

Maybe then the label should be positional but with a default of None, defaulting to id as it is now.

rodrigogiraoserrao: Commit [9cd592b](https://github.com/Textualize/textual/pull/1229/commits/9cd592bd044b2175bfb54358f3bdcebff2dc6d14) reorders the parameters of the `Placeholder` so that the first parameter is `label=None`.

willmcgugan (approved):
LGTM

PR #1238, opened 2022-11-21
expect_type for get_child_by_id and get_widget_by_id

aaronst:
For #1237, adds an `expect_type` parameter to `get_child_by_id` and `get_widget_by_id` methods. I tried to mimic how this was done for `query_one`.

I also noticed Pylance was not happy with `NodeList._get_by_id` being protected so I made it public, apologies if this is an unwanted change.

- [ ] Updated documentation

### DIFF
--- a/src/textual/_node_list.py
+++ b/src/textual/_node_list.py
@@ -64,8 +64,7 @@ class NodeList(Sequence):
         return self._nodes.index(widget)
 
     def get_by_id(self, widget_id: str) -> Widget | None:
-        """Get the widget for the given widget_id, or None if there's no matches in this list
-        """
+        """Get the widget for the given widget_id, or None if there's no matches in this list"""
         return self._nodes_by_id.get(widget_id)
 
     def _append(self, widget: Widget) -> None:
@@ -101,9 +100,9 @@ class NodeList(Sequence):
     def _ensure_unique_id(self, widget_id: str) -> None:
         if widget_id in self._nodes_by_id:
             raise DuplicateIds(
-                f"Tried to insert a widget with ID {widget_id!r}, but a widget"
-                f" {self._nodes_by_id[widget_id]!r} already exists with that ID in this"
-                " list of children. The children of a widget must have unique IDs."
+                f"Tried to insert a widget with ID {widget_id!r}, but a widget {self._nodes_by_id[widget_id]!r} "
+                "already exists with that ID in this list of children. "
+                "The children of a widget must have unique IDs."
             )
 
     def _remove(self, widget: Widget) -> None:
--- a/src/textual/_node_list.py
+++ b/src/textual/_node_list.py
@@ -63,7 +63,7 @@ class NodeList(Sequence):
         """
         return self._nodes.index(widget)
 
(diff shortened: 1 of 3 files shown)


src/textual/app.py
```diff
@@ -892,23 +893,52 @@ async def _on_css_change(self) -> None:
...
 
         Returns:
-            DOMNode: The first child of this node with the specified ID.
+            ExpectType | Widget: The first child of this node with the specified ID.
 
         Raises:
             NoMatches: if no children could be found for this ID
+            WrongType: if the wrong type was found.
         """
-        return self.screen.get_child_by_id(id)
```
willmcgugan, line 909: Do we need this condition? The second parameter can be None can’t it?

src/textual/_node_list.py
```diff
@@ -63,7 +63,7 @@ def index(self, widget: Widget) -> int:
         """
         return self._nodes.index(widget)
 
-    def _get_by_id(self, widget_id: str) -> Widget | None:
```
willmcgugan, line 66: What was PyLance complaining about?

willmcgugan (commented):
Looks great! Just some questions for now.

src/textual/_node_list.py
aaronst: When used in `Widget.get_child_by_id`, it shows:
> "_get_by_id" is protected and used outside of the class in which it is declared

willmcgugan: Ah.

PyLance is using an overly strict interpretation of leading underscores there. Which doesn’t follow from pep8. Actually are you sure that’s PyLance and not a linter? I didn’t recall seeing any red underlines in vscode.

We usd leading underscores to mean “not part of the public API”. I don’t think it makes sense for this method to be public.

src/textual/app.py
aaronst: I was trying to copy what is done here for `query_one`:
https://github.com/Textualize/textual/blob/f7867178a310a0e577c760a0ef3be0b0cab54316/src/textual/dom.py#L758

And noticed that when I tried:
```python
return self.screen.get_child_by_id(id, expect_type=expect_type)
```

Pylance was yelling at me with:
> Return type is unknown

So I decided it was best to follow the `if/else` pattern. Would really love to learn more about the decision either way though, a lot of this is still new to me.

src/textual/_node_list.py
aaronst: Yeah it gives me this link to `pyright`: https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportPrivateUsage

Understood on the different interpretations, I'll change it back!

aaronst: I do have __Type Checking Mode__ set to `strict`, maybe that's what did it.

src/textual/app.py
willmcgugan: I’m not sure why either. But not fussed either way.

willmcgugan (approved):

willmcgugan:
All good. Thanks!

PR #1259, opened 2022-11-23
Set default container height to 1fr.

rodrigogiraoserrao:

- [x] Updated documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 
 - Rebuilt `DirectoryTree` with new `Tree` control.
+- Container widgets now have default height of `1fr`.
 
 ## [0.5.0] - 2022-11-20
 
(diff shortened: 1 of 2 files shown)


darrenburns:
@rodrigogiraoserrao note that a lot of the examples are covered by snapshot tests, so you won't need to check them. Would just be things that aren't covered by the snapshots.

rodrigogiraoserrao:
This breaks `docs/examples/guide/dom4.py`:

Screenshot_1

The app screen has this CSS:

```CSS
#dialog {
    margin: 4 8;
    background: $panel;
    color: $text;
    border: tall $background;
    padding: 1 2;
}
```

Maybe when the inner container sees its default `height: 1fr`, it computes its height with the entire screen height instead of taking into account the space reduction from the margin?

willmcgugan:
I’ll look at that broken example. Please don’t merge before that.

rodrigogiraoserrao:
FYI no other examples seem to break.

PR #1264, opened 2022-11-23
Docs show empty app screenshot

rodrigogiraoserrao:

- [x] Updated documentation

The docs page [Guide/App](https://textual.textualize.io/guide/app/) has an empty screenshot immediately above the section [Async Events](https://textual.textualize.io/guide/app/#async-events):

Screenshot_1

Adding a pause before the screenshot seems to fix this issue (the screenshot below is from a local docs build with the fix):

Screenshot_2

### DIFF
--- a/docs/guide/CSS.md
+++ b/docs/guide/CSS.md
@@ -68,7 +68,7 @@ Let's look at a trivial Textual app.
 
 === "Output"
 
-    ```{.textual path="docs/examples/guide/dom1.py"}
+    ```{.textual path="docs/examples/guide/dom1.py" press="_"}
     ```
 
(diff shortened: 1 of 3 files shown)


rodrigogiraoserrao:
Turns out I was over-optimistic in my initial commit because there were some other screenshots that were empty but this PR fixes those.

willmcgugan (approved):

willmcgugan:
I’m guessing this only occurs on your machine? There is a trace condition there. We haven’t found a good solution for waiting until all events have been processed. In the meantime, this will have to do.

rodrigogiraoserrao:
> I’m guessing this only occurs on your machine? 

I don't think so. I had seen some of these in the live docs website before I ever deployed the docs from my machine, so I think these also happened on other machines.

PR #1268, opened 2022-11-24
Optimisations and fixes 

willmcgugan:
This started life as a refactor to optimize scrolling. Alas, it didn't work and I backed out. I did encounter a few optimisations which were worth doing.


src/textual/_styles_cache.py
```diff
@@ -52,6 +54,20 @@ def style_links(
     return segments
 
 
+@lru_cache(1024 * 8)
```
willmcgugan, line 57: When we paint widgets we tend to store a lot of blank strings for backgrounds. This caches those blank segments, more for memory than cpu.

src/textual/css/_style_properties.py
```diff
@@ -305,7 +305,7 @@ def __set__(self, obj: Styles, border: tuple[EdgeType, str | Color] | None):
             current_value: tuple[str, Color] = cast(
                 "tuple[str, Color]", obj.get_rule(self.name)
             )
-            has_edge = current_value and current_value[0]
+            has_edge = bool(current_value and current_value[0])
```
willmcgugan, line 308: This bug was causing a layout when anything with borders was updated.

src/textual/widget.py
```diff
@@ -755,14 +754,12 @@ def watch_hover_style(
     def watch_scroll_x(self, new_value: float) -> None:
         if self.show_horizontal_scrollbar:
             self.horizontal_scrollbar.position = int(new_value)
-            self.horizontal_scrollbar.refresh()
-            self.refresh(layout=True)
+            self.refresh(layout=True, repaint=False)
```
willmcgugan, line 757: We don't need to repaint things when scrolling.

src/textual/css/styles.py
```diff
@@ -594,8 +594,10 @@ def clear_rule(self, rule: str) -> bool:
         Returns:
             bool: ``True`` if a rule was cleared, or ``False`` if it was already not set.
         """
-        self._updates += 1
-        return self._rules.pop(rule, None) is not None
+        changed = self._rules.pop(rule, None) is not None
+        if changed:
+            self._updates += 1
```
willmcgugan, line 599: No need to invalidate cache if nothing changed.

davep (approved):

PR #1292, opened 2022-11-26
2022-11-26 DevLog from davep

davep:


willmcgugan (approved):
In Airport. No time to review.

willmcgugan (commented):
Those PNGs are very large. Could you run pngcrush over them?

davep:
Don't think I know that but I'll hunt it down, sounds like a handy thing.

PR #1294, opened 2022-11-27
Squish some PNGs

davep:
Ran `pngcrush` over them, as suggested by @willmcgugan.


willmcgugan (approved):
Thanks. Reckon we will move the docs to a new repos at some point. Won’t take long before we have an excessive amount of media.

davep:
> Reckon we will move the docs to a new repos at some point. Won’t take long before we have an excessive amount of media.

I was thinking the exact same thing yesterday, that the docs and blog would be better off in their own repos.

PR #1311, opened 2022-12-03
Fix 1309

rodrigogiraoserrao:

### DIFF
--- a/src/textual/_layout.py
+++ b/src/textual/_layout.py
@@ -67,7 +67,7 @@ class Layout(ABC):
                 )
                 width = child_width if width is None else max(width, child_width)
         if width is None:
-            width = container.width
+            width = 0
 
         return width
 
@@ -86,7 +86,7 @@ class Layout(ABC):
             int: Content height (in lines).
         """
         if not widget.displayed_children:
-            height = container.height
+            height = 0
         else:
             placements, *_ = widget._arrange(Size(width, container.height))
             height = max(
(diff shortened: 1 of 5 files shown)


rodrigogiraoserrao:
Will had mentioned that making empty containers collapse would likely come back to bite us.
It already did, because my change broke the widget `Input`.
In this case, and in the future, to make an empty widget visible, you have to set its `min-height` or `min-width`, depending on the dimension that you don't want to collapse.

In the case of an empty `input`, we don't want it to collapse when it has no text in it, so we set `min-height: 1`.

willmcgugan (approved):
Seems reasonable! 👍

willmcgugan:
@rodrigogiraoserrao Could you give this one a better name?

rodrigogiraoserrao:
> @rodrigogiraoserrao Could you give this one a better name?

What is “this” referring to?

willmcgugan:
The PR

willmcgugan:
You can put `fixes <ISSUE URL>` in the description, and GitHub will close the issue when the PR is merged.

rodrigogiraoserrao:
> You can put `fixes <ISSUE URL>` in the description, and GitHub will close the issue when the PR is merged.

I don't get what you want. Are you asking for two things now?

 1. Give a better name to the PR; and
 2. Link the issue to this PR?

Because I already linked the issue:

Screenshot_1

willmcgugan:
I was just asking for a better title. Which I see you have changed. Thanks

Issue #1316, opened 2022-12-04
`Label` should have a `DEFAULT_CSS` that sets both `width` and `height` to `auto`


davep:
Prompted by https://github.com/Textualize/textual/discussions/1315.

`Label`, at the moment, is nothing more than a `pass` inherit from `Static`. The `DEFAULT_CSS` of `Static` is `height: auto;` and nothing else. This results in, apparently, a `Label` having a default width being whatever the width of the container is. I think this would be surprising to most people -- the reasonable expectation being that the width should be the width of the content, by default.

davep:
This code helps illustrate the issue:

```python
"""https://github.com/Textualize/textual/discussions/1315"""

from textual.app import App, ComposeResult
from textual.containers import Horizontal
from textual.widgets import Label

class LabelAuto( App[ None ] ):

    CSS = """
    Horizontal {
        border: solid blue;
        height: auto;
    }

    Label {
        border: solid red;
    }

    .width {
        width: auto;
    }

    .height {
        height: auto;
    }

    .one-fr {
        width: 1fr;
    }
    """

    def compose(self) -> ComposeResult:
        yield Horizontal(
            Label( "Raw Label" ),
            Label( "Raw Label" ),
            Label( "Raw Label" ),
            Label( "Raw Label" ),
            Label( "Raw Label" )
        )
        yield Horizontal(
            Label( "width: auto;", classes="width" ),
            Label( "height: auto;", classes="height" ),
            Label( "width: auto; height: auto;", classes="width height" )
        )
        yield Horizontal( Label( "width: auto;", classes="width" ) )
        yield Horizontal( Label( "height: auto;", classes="height" ) )
        yield Horizontal( Label( "width: auto; height: auto;", classes="width height" ) )
        yield Horizontal(
            Label( "width: auto;", classes="width" ),
            Label( "width: auto;", classes="width" ),
            Label( "width: auto;", classes="width" )
        )
        yield Horizontal(
            Label( "width: 1fr;", classes="one-fr" ),
            Label( "width: 1fr;", classes="one-fr" )
        )
        yield Horizontal(
            Label( "width: 1fr;", classes="one-fr" ),
            Label( "width: 1fr;", classes="one-fr" ),
            Label( "width: 1fr;", classes="one-fr" ),
            Label( "width: 1fr;", classes="one-fr" ),
            Label( "width: 1fr;", classes="one-fr" ),
            Label( "width: 1fr;", classes="one-fr" )
        )

[...]

PR #1318, opened 2022-12-05
Change the default width of a label to auto

davep:
See #1315 for the discussion that prompted this, #1316 for the issue where this particular aspect of it was isolated as something to address.

I'm not 100% convinced that `Static` does the right thing here either, but I don't want to go changing too make things at once. However, I think a good argument can be made that `Label` should be `width: auto` by default, this seems to make perfect sense to me. I think everyone would expect, by default, that a label's width is the width of the content (plus any decoration, of course).

### DIFF
--- a/src/textual/widgets/_label.py
+++ b/src/textual/widgets/_label.py
@@ -5,3 +5,11 @@ from ._static import Static
 
 class Label(Static):
     """A simple label widget for displaying text-oriented renderables."""
+
+    DEFAULT_CSS = """
+    Static {
+        width: auto;
+        height: auto;
+    }
+    """
+    """str: The default styling of a `Label`."""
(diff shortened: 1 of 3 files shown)


src/textual/widgets/_label.py
```diff
@@ -5,3 +5,11 @@
 
 class Label(Static):
     """A simple label widget for displaying text-oriented renderables."""
+
+    DEFAULT_CSS = """
+    Static {
```
rodrigogiraoserrao, line 10: Why didn't you write the rule as `Label { ... }`?

davep: Too early in the morning and not enough coffee. 😛

davep: Fixed. Good catch. :-)

rodrigogiraoserrao (approved):

Issue #1313, opened 2022-12-04
Visibility rule doesn't impact children


willmcgugan:
Applying visibility: hidden to a container, makes the container hidden, but not the children.

Pretty sure browser CSS would also make the children hidden, so we should replicate that functionality.

The follow example demonstrates the issue:

```python
from textual.app import App, ComposeResult

from textual.widget import Widget
from textual.widgets import Static


class VisApp(App):

    CSS = """
    #container {
        border: solid red;
        visibility: hidden;
    }
    """

    def compose(self) -> ComposeResult:
        yield Widget(
            Static("foo"),
            Static("bar"),
            id="container",
        )


if __name__ == "__main__":
    app = VisApp()
    app.run()
```

With the above, both foo and bar are visible, while the container is not.

PR #1327, opened 2022-12-07
fix for visiblity

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/1313

Essentially "visibility: hidden" only changed the one widget were it was set. This PR makes the visibility apply to children. It also allows children to change their visibility.

Consider implementing a dialog. A layer can be a container with a center aligned child widget. The container is hidden, but the child is unhidden. So the dialog floats over the content below.


rodrigogiraoserrao (commented):
The changes _seem_ sensible but I am not comfortable enough with the compositor to say the changes are OK.

davep (approved):

davep:
> Consider implementing a dialog. A layer can be a container with a center aligned child widget. The container is hidden, but the child is unhidden. So the dialog floats over the content below.

I know it's out of the scope of this change, but I suspect this isn't how you'd want to implement a dialog most of the time, right? All sorts of binding issues to be had here as I'm thinking about it. (which brings us back to screens that can be transparent and float over other screens).

willmcgugan:
> > Consider implementing a dialog. A layer can be a container with a center aligned child widget. The container is hidden, but the child is unhidden. So the dialog floats over the content below.
> 
> I know it's out of the scope of this change, but I suspect this isn't how you'd want to implement a dialog most of the time, right? All sorts of binding issues to be had here as I'm thinking about it. (which brings us back to screens that can be transparent and float over other screens).

Don't see why not. Wouldn't solve the focus issues, but seems like a good way of rendering a floating thing.

PR #1329, opened 2022-12-08
Fix for auto dimensions on containers, with relative children.

willmcgugan:
A change in how auto dimensions are calculated for containers.

Essentially children with relative dimensions are ignored in the auto calculation, and only known dimensions are considered.


src/textual/demo.py
```diff
@@ -322,7 +322,7 @@ def compose(self) -> ComposeResult:
         example_css = "\n".join(Path(self.css_path[0]).read_text().splitlines()[:50])
         yield Container(
             Sidebar(classes="-hidden"),
-            Header(show_clock=True),
+            Header(show_clock=False),
```
davep, line 325: Out of PR scope cosmetic choice in passing, or accidental change?

davep (approved):

darrenburns (approved):
Looks good

src/textual/demo.py
willmcgugan: I added this to the snapshot tests, because there was a recent breakage we didn't notice. Alas the clock meant that the snapshot would only pass if I run it at exactly the same time of day!

PR #1332, opened 2022-12-08
Fix container offsets

willmcgugan:
Fixes offsets when applied to containers.


willmcgugan:
This PR brought to you from "changing stuff randomly until it works".

willmcgugan:
#executivedecision

PR #1333, opened 2022-12-09
textlog documentation

willmcgugan:
Adds documentation for TextLog.

Also makes a few significant changes:

`events.Show` is sent *after* resize. So that by the time you handle that, the widget has a size.

I've added an `events.Ready` event which is sent to the App only, when the screen has been mounted and resized. This is essentially a DOM ready event that @davep requested. It's a bit experimental (I'm not certain this is the right way of doing it), so I'm not going to document it quite yet.


willmcgugan:
#executivedecision

Issue #1335, opened 2022-12-09
Default CSS overrides more specific rules when custom widgets are instantiated in another screen


rodrigogiraoserrao:
Consider this scenario:
 - app with two screens `A` and `B`;
 - each screen has some custom widgets that are only displayed on _that_ screen, respectively `WA` and `WB`;
 - one custom widget has `DEFAULT_CSS` set (`WA`) while the other doesn't (`WB`);
 - both widgets share an ancestor **with** `DEFAULT_CSS` set.

If you create instances of `WB` while there are no instances of `WA` mounted, the `DEFAULT_CSS` of the common ancestor of both widgets will override `WA`'s `DEFAULT_CSS` and change its style.

Here is a video demo:

https://user-images.githubusercontent.com/5621605/206695369-26a0b418-1097-40f4-b4cd-7f57e67888bf.mp4

<details>
<summary><b>CODE for the repro</b></summary>

```py
from textual.app import App, ComposeResult
from textual.screen import Screen
from textual.widgets import Footer, Label, Static


class _Static(Static):
    DEFAULT_CSS = """
    _Static {
        background: red;
        height: 10;
        width: 5;
    }
    """


class CustomWidget1(_Static):
    DEFAULT_CSS = """
    CustomWidget1 {
        background: blue;
        height: 3;
        width: 20;
    }
    """
    def compose(self) -> ComposeResult:
        yield Label("Blue, wide")


class CustomWidget2(_Static):
    def compose(self) -> ComposeResult:
        yield Label("Red\ntall")


class Screen1(Screen):
    BINDINGS = [("a", "add", "add"), ("c", "clear", "clear"), ("p", "pop", "pop")]
    def compose(self) -> ComposeResult:
        yield Footer()

    def action_pop(self) -> None:
        self.app.pop_screen()

    def action_add(self) -> None:
        self.mount(CustomWidget1())

    def action_clear(self) -> None:
        for w in self.query(CustomWidget1):
            w.remove()


class Screen2(Screen):
    BINDINGS = [("a", "add", "add"), ("c", "clear", "clear"), ("p", "push", "push")]
    def compose(self) -> ComposeResult:
        yield Footer()

    def action_push(self) -> None:
        self.app.push_screen("screen1")

    def action_add(self) -> None:
        self.mount(CustomWidget2())

    def action_clear(self) -> None:
        for w in self.query(CustomWidget2):
            w.remove()


class MyApp(App):
    SCREENS = {
        "screen1": Screen1,
[...]

PR #1336, opened 2022-12-09
Fix default CSS retrieval from widgets with no `DEFAULT_CSS` that inherit from widgets that do have `DEFAULT_CSS`

rodrigogiraoserrao:

- [x] ~~Docstrings on all new or modified functions / classes~~ I _think_ no docstring updates are needed because I think the docstring for `get_default_css` remains consistent with its behaviour. We _could_ add a note saying that we only get CSS out of classes that introduce new CSS.
- [x] ~~Updated documentation~~ No docs updates required.

This PR implements the suggested fix in #1335.
I struggled to write the fix in an elegant way, so I am open to suggestions.


src/textual/dom.py
```diff
@@ -265,9 +265,12 @@ def get_path(base: Type[DOMNode]) -> str:
             except TypeError:
                 return f"{base.__name__}"
 
-        for tie_breaker, base in enumerate(self._node_bases):
-            css = base.DEFAULT_CSS.strip()
-            if css:
+        default_css = [base.DEFAULT_CSS.strip() for base in self._node_bases]
```
willmcgugan, line 268: I think I see the problem you are trying to solve.

I gather your loop is checking if the `DEFAULT_CSS` is the same as the base class, then skipping it?

Rather than go through the class attribute mechanism, what if we were to look at `__dict__` directly?

```python
base.__dict__get("DEFAULT_CSS")
```

Instead of 

```python
base.DEFAULT_CSS
```

I think the first will return None if the attribute is not explicitly defined on the class, where the second will return the value defined on a base class.

willmcgugan (commented):
Kudos for figuring this out. I've proposed an modification that I think will be more efficient.

src/textual/dom.py
rodrigogiraoserrao: That's well thought-of. I tried to come up with a way to check if `DEFAULT_CSS` had been set explicitly in a given class or in its parent classes but I didn't think of `__dict__`. Rewrote the PR so that it uses `__dict__.get(..., "")` to check for the default CSS.

willmcgugan (approved):

willmcgugan:
LGTM Ready to merge?

Issue #1343, opened 2022-12-12
As of 0.6.0, bindings on (at least) `left`, `right`, `up` and `down` no longer appear to work


davep:
With Textual 0.5.0 I had bindings on the arrow keys that worked fine. Having upgraded to 0.6.0 those bindings no longer seem to work. Isolating the issue I can recreate with this code (some other keys thrown in as control tests)

```python
from textual.app import App, ComposeResult
from textual.widgets import TextLog, Header, Footer
from textual.binding import Binding
from textual.events import Key

class Binder( App[ None ] ):

    CSS = """
    TextLog {
        background: #222200;
        color: #BBBB00;
        text-style: bold;
    }
    """

    BINDINGS = [
        Binding( "up",    "log( 'up' )", "Up" ),
        Binding( "down",  "log( 'down' )", "Down" ),
        Binding( "left",  "log( 'left' )", "Left" ),
        Binding( "right", "log( 'right' )", "Right" ),
        Binding( "a", "log( 'a' )", "a" ),
        Binding( "s", "log( 's' )", "s" ),
        Binding( "d", "log( 'd' )", "d" ),
        Binding( "f", "log( 'f' )", "f" ),
        Binding( "left_square_bracket", "log( '[' )", "[" ),
        Binding( "right_square_bracket", "log( ']' )", "]" ),
    ]

    def compose( self ) -> ComposeResult:
        yield Header()
        yield TextLog()
        yield Footer()

    def on_mount( self ) -> None:
        self.query_one( TextLog ).write( "Ready..." )
        self.query_one( TextLog ).write( "Key bindings being looked for:" )
        self.query_one( TextLog ).write( ", ".join(
            [ binding.key for binding in self.BINDINGS ]
        ) )

    def action_log( self, name: str ) -> None:
        self.query_one( TextLog ).write( f"Via binding: {name}" )

    def on_key( self, event: Key ) -> None:
        self.query_one( TextLog ).write( f"Via event: {event!r}" )

if __name__ == "__main__":
    Binder().run()
```

Running the above, pressing the arrow keys vs some of the other keys...

![Screenshot 2022-12-12 at 11 25 12](https://user-images.githubusercontent.com/28237/207033752-5dba92cc-a860-4112-b0b3-d80fe810f813.png)

davep:
This seems to be related to the `universal` keyword parameter on `Binding`. Again, the above code as shown above:

[...]

PR #1346, opened 2022-12-12
Add support for a `PRIORITY_BINDINGS` classvar to work alongside `BINDINGS`

davep:
~~This PR aims to explore, add tests for, and hopefully resolve the issue raised in #1343. For now this is just a draft PR as I work to build up a set of (hopefully comprehensive) unit tests for the issue. As this grows and ideally ends up implementing a solution for the problem, I'll edit this description to better describe what's gone into it.~~

~~As I started to pull on the thread first found as #1343, it became obvious that a fairly comprehensive set of unit tests were needed to document the issue, test what does and doesn't work right now, document expectations and also cover things that work now and should carry on working should any changes be made.~~

~~I feel it makes sense to merge these tests into `main` first and *then* move on to considering options for how to resolve the issue. Hence repurposing this PR for that and that alone.~~

This PR has now turned back into a fix for #1343, taking the approach of allowing the library user to say if the bindings on a particular widget should be considered to have priority or not, by use of a `PRIORITY_BINDINGS` classvar. Additionally, `App` and `Screen` are set so that all of their bindings are priority bindings by default.

Side quest: The binding system now handles situations where a library user may bind `"a, b, c, d"` rather than `"a,b,c,d"`.

### DIFF
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -298,7 +298,7 @@ class App(Generic[ReturnType], DOMNode):
 
         self._logger = Logger(self._log)
 
-        self._bindings.bind("ctrl+c", "quit", show=False, universal=True)
+        self._bindings.bind("ctrl+c", "quit", show=False, priority=True)
         self._refresh_required = False
 
         self.design = DEFAULT_COLORS
@@ -1729,12 +1729,12 @@ class App(Generic[ReturnType], DOMNode):
             ]
         return namespace_bindings
 
-    async def check_bindings(self, key: str, universal: bool = False) -> bool:
+    async def check_bindings(self, key: str, priority: bool = False) -> bool:
         """Handle a key press.
 
         Args:
             key (str): A key
-            universal (bool): Check universal keys if True, otherwise non-universal keys.
+            priority (bool): If `True` check from `App` down, otherwise from focused up.
 
         Returns:
             bool: True if the key was handled by a binding, otherwise False
@@ -1742,7 +1742,7 @@ class App(Generic[ReturnType], DOMNode):
 
         for namespace, bindings in self._binding_chain:
             binding = bindings.keys.get(key)
-            if binding is not None and binding.universal == universal:
+            if binding is not None and binding.priority == priority:
                 await self.action(binding.action, default_namespace=namespace)
                 return True
         return False
@@ -1762,7 +1762,7 @@ class App(Generic[ReturnType], DOMNode):
                 self.mouse_position = Offset(event.x, event.y)
                 await self.screen._forward_event(event)
             elif isinstance(event, events.Key):
-                if not await self.check_bindings(event.key, universal=True):
+                if not await self.check_bindings(event.key, priority=True):
                     forward_target = self.focused or self.screen
                     await forward_target._forward_event(event)
             else:
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -1743,7 +1743,7 @@ class App(Generic[ReturnType], DOMNode):
         """
 
         for namespace, bindings in (
-            reversed(self._binding_chain) if priority else self._binding_chain
+            list(reversed(self._binding_chain)) if priority else self._binding_chain
         ):
             binding = bindings.keys.get(key)
             if binding is not None and binding.priority == priority:
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -1743,7 +1743,7 @@ class App(Generic[ReturnType], DOMNode):
         """
 
         for namespace, bindings in (
-            list(reversed(self._binding_chain)) if priority else self._binding_chain
+            reversed(self._binding_chain) if priority else self._binding_chain
         ):
             binding = bindings.keys.get(key)
             if binding is not None and binding.priority == priority:
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -231,6 +231,8 @@ class App(Generic[ReturnType], DOMNode):
     }
     """
 
+    PRIORITY_BINDINGS = True
+
     SCREENS: dict[str, Screen | Callable[[], Screen]] = {}
     _BASE_PATH: str | None = None
     CSS_PATH: CSSPathType = None
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -1742,9 +1742,10 @@ class App(Generic[ReturnType], DOMNode):
             bool: True if the key was handled by a binding, otherwise False
         """
 
-        for namespace, bindings in (
-            list(reversed(self._binding_chain)) if priority else self._binding_chain
-        ):
+        # for namespace, bindings in (
+        #     list(reversed(self._binding_chain)) if priority else self._binding_chain
+        # ):
+        for namespace, bindings in self._binding_chain:
             binding = bindings.keys.get(key)
             if binding is not None and binding.priority == priority:
                 await self.action(binding.action, default_namespace=namespace)
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -1742,7 +1742,9 @@ class App(Generic[ReturnType], DOMNode):
             bool: True if the key was handled by a binding, otherwise False
         """
 
-        for namespace, bindings in self._binding_chain:
+        for namespace, bindings in (
+            reversed(self._binding_chain) if priority else self._binding_chain
+        ):
             binding = bindings.keys.get(key)
             if binding is not None and binding.priority == priority:
                 await self.action(binding.action, default_namespace=namespace)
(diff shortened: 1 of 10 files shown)


tests/test_binding_inheritance.py
```diff
@@ -0,0 +1,484 @@
...
+    def on_mount(self) -> None:
+        self.query_one(FocusableWidgetWithBindings).focus()
+
+
+async def test_focused_child_widget_with_movement_bindings() -> None:
+    """A focused child widget with movement bindings should handle its own actions."""
+    async with AppWithWidgetWithBindings().run_test() as pilot:
+        await pilot.press(*AppKeyRecorder.ALL_KEYS)
+        await pilot.pause(2/100)
+        pilot.app.all_recorded("locally_")
```
rodrigogiraoserrao, line 266: Shouldn't the prefix `"locally_"` here match the prefix `"local_"` in line 244?

tests/test_binding_inheritance.py
```diff
@@ -0,0 +1,484 @@
...
+    def compose(self) -> ComposeResult:
+        yield Container(FocusableWidgetWithNoBindings())
+
+    def on_mount(self) -> None:
+        self.query_one(FocusableWidgetWithNoBindings).focus()
+
+class AppWithScreenWithBindingsWrappedWidgetNoBindings(AppKeyRecorder):
+    """An app with a non-default screen that handles movement key bindings."""
+
+    SCREENS = {"main":ScreenWithMovementBindings}
```
rodrigogiraoserrao, line 343:
```suggestion
    SCREENS = {"main": ScreenWithMovementBindings}
```

rodrigogiraoserrao (changes requested):
I found a couple of minor things that I really think you should look at again, but overall I ended up leaving more question-comments than you-should-change-this-comments.

In particular, I don't think I really grokked how your `AppKeyRecoder` worked, given that you had several tests where you used `AppKeyRecoder.mk_bindings` with a specific prefix but then ended up testing against a different prefix with `"ly"` in the end.

tests/test_binding_inheritance.py
```diff
@@ -0,0 +1,484 @@
...
+        self.push_screen("main")
+
+
+@pytest.mark.xfail(
+    reason="Screen is incorrectly starting with bindings for movement keys [issue#1343]"
+)
+async def test_app_screen_has_no_movement_bindings() -> None:
+    """A screen with no bindings should not have movement key bindings."""
+    async with AppWithScreenNoBindings().run_test() as pilot:
+        assert sorted(list(pilot.app.screen._bindings.keys.keys())) != sorted(MOVEMENT_KEYS)
```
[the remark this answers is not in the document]
davep: The point this is documenting, for the purposes of this test, is that the movement keys haven't leaked into the screen's bindings. It's explicitly telling that story.

davep:
> In particular, I don't think I really grokked how your AppKeyRecoder worked, given that you had several tests where you used AppKeyRecoder.mk_bindings with a specific prefix but then ended up testing against a different prefix with "ly" in the end.

`mk_bindings` takes a prefix for the actions. The actions themselves emit the markers that say that the action was run.

tests/test_binding_inheritance.py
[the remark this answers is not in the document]
davep: Keep in mind: these tests are likely to get further updates and reworking once we've decided what we're going to do about the binding problem. The tests aim to tell the story of what the problem is right now and how end developers may encounter them.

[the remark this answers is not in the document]
rodrigogiraoserrao: Right. I get that it is the movement keys that are leaking, it is just that it is "easy" to make this test pass without actually fixing the problem: just make it so that the bindings get an extra random binding that shouldn't be there and all of a sudden the test passes. That's the only reason why I decided to comment on this test.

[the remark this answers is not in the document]
davep: Aye, longer-term that'd make sense; this was about telling the story right now. But you're right, it works either way so I may as well go with that now. Good catch.

davep:
Making this a draft again. #1343 is going in a slightly different direction that initially anticipated, and hopefully in a less-drastic direction too, so there's less need to split the initial testing out from whatever work follows.

davep:
@willmcgugan I believe that this PR is around the point of implementing what we spoke about this morning, with tests to check too. Before I think about what's needed for any documentation, can we get a chat to go over it first? Want to be sure I've gone about this the correct way before investing any time on all the tidying up.

tests/test_binding_inheritance.py
```diff
@@ -0,0 +1,610 @@
...
+
+
+async def test_app_screen_with_bindings() -> None:
+    """Test a screen with a single key binding defined."""
+    async with AppWithScreenThatHasABinding().run_test() as pilot:
+        # The screen will contain all of the movement keys, because it
+        # inherits from Widget. That's fine. Let's check they're there, but
+        # also let's check that they all have a non-priority binding.
+        assert all(
+            pilot.app.screen._bindings.get_key(key).priority == False
```
willmcgugan, line 122: The `is` operator may be more appropriate for a test.

tests/test_binding_inheritance.py
```diff
@@ -0,0 +1,610 @@
...
+    """Base application class that can be used to record keystrokes."""
+
+    ALPHAS = "abcxyz"
+    """str: The alpha keys to test against."""
+
+    ALL_KEYS = [*ALPHAS, *MOVEMENT_KEYS]
+    """list[str]: All the test keys."""
+
+    @staticmethod
+    def mk_bindings(action_prefix: str = "") -> list[Binding]:
```
willmcgugan, line 182: No abbreviates please.

tests/test_binding_inheritance.py
```diff
@@ -0,0 +1,610 @@
...
+        Binding("0", "record('app_0')", "0", priority=False),
+        Binding("a", "record('app_a')", "a", priority=True),
+        Binding("b", "record('app_b')", "b", priority=False),
+        Binding("c", "record('app_c')", "c", priority=False),
+        Binding("d", "record('app_d')", "c", priority=True),
+        Binding("e", "record('app_e')", "e", priority=True),
+        Binding("f", "record('app_f')", "f", priority=False),
+    ]
+
+    SCREENS = {"main": PriorityOverlapScreen()}
```
willmcgugan, line 595: This may be the source of your test fail.

src/textual/app.py
```diff
@@ -1729,20 +1731,22 @@ def _binding_chain(self) -> list[tuple[DOMNode, Bindings]]:
...
-            universal (bool): Check universal keys if True, otherwise non-universal keys.
+            priority (bool): If `True` check from `App` down, otherwise from focused up.
 
         Returns:
             bool: True if the key was handled by a binding, otherwise False
         """
 
-        for namespace, bindings in self._binding_chain:
+        for namespace, bindings in (
+            list(reversed(self._binding_chain)) if priority else self._binding_chain
```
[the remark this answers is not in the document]
davep: You appear to have caught this mid-experiment. That was added to eliminate a potential issue.

tests/test_binding_inheritance.py
davep: Interesting, I'll give that a try. I *think* I turned it into an instance at some point as a process of elimination but forgot to swap back.

Any idea as to why this would cause a problem, especially on different Python versions? (the issues I'm seeing, so far, seem to be down to Python version).

willmcgugan: Not entirely sure. I think it is something to do with creating async primitives in different loops, as created by pytest.

davep: FWIW, that did fix it.

willmcgugan (approved):

tests/test_binding.py
```diff
@@ -6,19 +6,26 @@
 
 BINDING1 = Binding("a,b", action="action1", description="description1")
 BINDING2 = Binding("c", action="action2", description="description2")
+BINDING3 = Binding(" d   , e ", action="action3", description="description3")
```
jayaddison, line 9: Is there a motivating use-case for binding keys that have surrounding whitespace?

Although I get that it provides more leeway for developers ([Postel's Law](https://en.wikipedia.org/wiki/Robustness_principle), perhaps), there could be a lot said in support of keeping the binding format concise (would unique-representation-per-key be achievable?).

davep: Being more robust was my motivation. Space is bound as the name of the key, not the printable character, and comma obviously has to be bound by name too. It's likely some folk might write those binding lists similar to how they'd write a list anyway ("a, b, c, d") so it seemed to make sense to accept that, given there was zero cost in doing so.

Issue #1347, opened 2022-12-12
Removing (and potentially adding?) child widgets to a scrolling container widget can throw off the position and scrollbar


davep:
See #1166 for the original context, but it was decided that this was a different issue altogether. Simply put, while this code...

```python
from textual.app import App, ComposeResult
from textual.containers import Vertical
from textual.widget import Widget
from textual.widgets import Header, Footer, Label

class ScrollApp( App[ None ] ):

    BINDINGS = [
        ( "d", "decimate", "Decimate" )
    ]

    def widgets( self, count: int ) -> list[ Widget ]:
        return [
            Label( f"This is label {n}", id=f"line{n}" ) for n in range( count )
        ]

    def compose(self) -> ComposeResult:
        yield Header()
        yield Vertical( *self.widgets( 500 ) )
        yield Footer()

    def action_decimate( self ) -> None:
        widgets = self.query( "Vertical > *" )
        new_count = int( len( widgets ) * 0.9 )
        widgets.remove()
        self.query_one( Vertical ).mount( *self.widgets( new_count ) )

if __name__ == "__main__":
    ScrollApp().run()
```

appears to behave as expected, adding some `async`/`await` to `action_decimate`...

```python
from textual.app import App, ComposeResult
from textual.containers import Vertical
from textual.widget import Widget
from textual.widgets import Header, Footer, Label

class ScrollApp( App[ None ] ):

    BINDINGS = [
        ( "d", "decimate", "Decimate" )
    ]

    def widgets( self, count: int ) -> list[ Widget ]:
        return [
            Label( f"This is label {n}", id=f"line{n}" ) for n in range( count )
        ]

    def compose(self) -> ComposeResult:
        yield Header()
        yield Vertical( *self.widgets( 500 ) )
        yield Footer()

    async def action_decimate( self ) -> None:
        widgets = self.query( "Vertical > *" )
        new_count = int( len( widgets ) * 0.9 )
        await widgets.remove()
        self.query_one( Vertical ).mount( *self.widgets( new_count ) )

if __name__ == "__main__":
    ScrollApp().run()
```

...this causes problems with the display. It doesn't always happen on the first hit of `d`, but it's rare that it doesn't start to show after the second or third press.

Note that for best effect you should scroll the `Vertical` to the very bottom before hitting `d`.

_Originally posted by @davep in https://github.com/Textualize/textual/issues/1166#issuecomment-1332758017_

PR #1350, opened 2022-12-13
move refresh

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/1347

I think what was happening is that refresh was being done prematurely.


rodrigogiraoserrao (approved):
To me, it makes sense that we only want to refresh after the task to prune widgets has done its best to prune the widgets.

PR #1352, opened 2022-12-13
Fix binding merging when binding inheritance is set to `False`.

### LINKED ISSUE #1351: A `Widget`-derived widget with `inherit_bindings=False`, but no `BINDINGS`, appears to still inherit bindings
This is *very* much related to #1343, but I think merits an issue of its own to consider first. The reasons *why* this happens seem clear enough, but I feel the setup may cause confusion. With Textual as it is of 0.6.0 (with the bindings for movement keys being on `Widget`), consider this widget:

```python
class Odradek( Widget, inherit_bindings=False ):
    pass
````

vs this:

```python
class Odradek( Widget, inherit_bindings=False ):
    BINDINGS = []
````

In the former case the bindings of `Widget` are still inherited. In the latter they're not.

rodrigogiraoserrao:
This fixes #1351 with a strategy _very_ similar to that of #1336.


willmcgugan (approved):
Lgtm

PR #1359, opened 2022-12-14
Call validator on first set

### LINKED ISSUE #1220: Validators and watchers should run hand-in-hand
When `init=True` in a `reactive` attribute, the `watch_` method for that attribute run, but the `validate_` method does not.
This seems rather confusing and we think that validators and watchers should always run hand-in-hand.
If the watcher is called, you should be confident that the incoming `new_value` has already been validated by the `validate_` method.

darrenburns:
Fixes #1220 

- [x] Updated documentation

### DIFF
--- a/src/textual/reactive.py
+++ b/src/textual/reactive.py
@@ -177,8 +177,8 @@ class Reactive(Generic[ReactiveType]):
         validate_function = getattr(obj, f"validate_{name}", None)
         # Check if this is the first time setting the value
         first_set = getattr(obj, f"__first_set_{self.internal_name}", True)
-        # Call validate, but not on first set.
-        if callable(validate_function) and not first_set:
+        # Call validate
+        if callable(validate_function):
             value = validate_function(value)
         # If the value has changed, or this is the first time setting the value
         if current_value != value or first_set or self._always_update:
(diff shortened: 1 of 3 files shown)


src/textual/reactive.py
```diff
@@ -177,8 +177,8 @@ def __set__(self, obj: Reactable, value: ReactiveType) -> None:
         validate_function = getattr(obj, f"validate_{name}", None)
         # Check if this is the first time setting the value
         first_set = getattr(obj, f"__first_set_{self.internal_name}", True)
-        # Call validate, but not on first set.
-        if callable(validate_function) and not first_set:
```
darrenburns, line 181: Do you remember the reasoning for this? The comment and the code suggest this behaviour was intentional, but I'm struggling to think of why that might be.

willmcgugan: I think it was a workaround to avoid an issue with the default value invoking validators before the DOM was set up. Maybe prior to `init`. Could very well be deprecated.

darrenburns:
I've left the first_set variable because it's used further down. Not 100% sure it's needed, but I think when I tried to remove it before it broke something, so I'll probably follow up investigating that while looking at some of the other reactive issues.

willmcgugan (approved):

PR #1362, opened 2022-12-14
Export the types that _typing.py imports and makes available

davep:
Mostly adding this so that it turns this:

$ poetry run mypy --strict binding.py
binding.py:8: error: Module "textual._typing" does not explicitly export attribute "TypeAlias"  [attr-defined] Found 1 error in 1 file (checked 1 source file)

into this:

$ poetry run mypy --strict binding.py
Success: no issues found in 1 source file

### DIFF
--- a/src/textual/_typing.py
+++ b/src/textual/_typing.py
@@ -9,3 +9,5 @@ if sys.version_info >= (3, 8):
     from typing import Final, Literal, Protocol, TypedDict
 else:
     from typing_extensions import Final, Literal, Protocol, TypedDict
+
+__all__ = ["TypeAlias", "Final", "Literal", "Protocol", "TypedDict"]


willmcgugan (approved):

willmcgugan:
I wonder if this will fix PyLance complaining about these imports.

davep:
It was pyright complaints about these[^1] (which is related to pylance IIRC) that prompted me to check this; didn't make the pyright complaints go away (a different change I made did, but I backed out of that as it was a side quest I didn't need right then), but did keep mypy happy in strict mode so that felt like a start.

[^1]: Actually about the binding types, which lean on `TypeAlias`, and `pyright` is *really* unhappy with how Textual uses `TypeAlias`, it seems.

PR #1363, opened 2022-12-14
Remove unused import of MounrError in app.py

davep:
Noticed while passing by.

### DIFF
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -63,7 +63,7 @@ from .messages import CallbackType
 from .reactive import Reactive
 from .renderables.blank import Blank
 from .screen import Screen
-from .widget import AwaitMount, MountError, Widget
+from .widget import AwaitMount, Widget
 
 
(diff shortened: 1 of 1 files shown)


rodrigogiraoserrao:
Does this break the docstrings? A couple of them mention `MountError` and automatic linking may or may not depend on this `MountError` having been imported or not.

davep:
I've never seen a Python doc system that's resolves linking that way; [and looking at our currently-published docs the one we use doesn't](https://textual.textualize.io/api/app/#textual.app.App.mount_all).

rodrigogiraoserrao:
Great, then! I was just afraid something might break, but looks like it won't :)

rodrigogiraoserrao (approved):

davep:
It does raise the question though: why do our docs link `Widget` but not `MountError` -- my wild guess is it's about what's in the docs? I might add an issue to think about checking on that at some point.

PR #1379, opened 2022-12-16
fix scroll to top

willmcgugan:
- Fixed `scroll_to_widget` when `top=True`, which didn't scroll when the widget was visible.
- Fixed broken bindings action (Couldn't click binding in footer).
- Fixed clean exit from Load event (previously calling exit in Load would go in to app mode when it shouldn't).
- Fixed Tree refresh issue


willmcgugan:
#executivedecition

@darrenburns @davep @rodrigogiraoserrao Please do a retroactive review!

PR #1324, opened 2022-12-06
Validate the content of binding keys during construction of bindings

jayaddison:
Although some (todo: be more precise?) binding keys may display as-expected when rendered within the application, they aren't matched by the corresponding keypress event keys (for example, pressing `.` on the keyboard parses as `full_stop`, and won't match a binding for key `.`).

This changeset adds validation for binding keys, and displys a suggested alternative value for invalid entries when possible.

May resolve #101.

**Please review the following checklist.**

- [x] Docstrings on all new or modified functions / classes 
- [ ] Updated documentation
- [ ] Updated CHANGELOG.md (where appropriate)

PR #1380, opened 2022-12-16
Add `keys` as a command to the `textual` tool

davep:
With this someone who has `textual[dev]` installed can run:

```sh
$ textual keys
```

and explore which keys make it through their terminal to Textual, and also explore what those keys are called for the purposes of binding (although https://github.com/Textualize/textual/pull/1324 if it makes it in will also be helpful here too).


willmcgugan:
Closing in favor of https://github.com/Textualize/textual/pull/1409

PR #1395, opened 2022-12-18
optional actions

willmcgugan:
- Adds a `SkipAction` exception which, when raised from an action, will give the parents an opportunity to handle the binding.
- Fixes an issue with `focus-within`, which wasn't updating on blur


tests/test_binding_inheritance.py
```diff
@@ -607,3 +608,40 @@ async def test_overlapping_priority_bindings() -> None:
             "app_e",
             "screen_f",
         ]
+
+
+async def test_skip_action() -> None:
+    """Test that a binding may be skipped by an action returning False"""
+
```
darrenburns, line 615:
```suggestion
    """Test that a binding may be skipped by an action raising SkipAction"""

```

darrenburns (approved):

PR #1408, opened 2022-12-20
Review styles reference

rodrigogiraoserrao:

- [x] Updated documentation – well, that _is_ the point of this PR :D


docs/css_types/_template.md
```diff
@@ -0,0 +1,40 @@
...
+    rule: type-value-3;
+}
+```
+
+### Python
+
+<!-- Same examples as above. -->
+
+```py
+type_name = type_value_1
```
willmcgugan, line 37: I think we need to need to show that these are properties on Widget somehow. 

```python
my_widget.rule = value
```

docs/css_types/border.md
```diff
@@ -0,0 +1,59 @@
...
+textual borders
+```
+
+## Examples
+
+### CSS
+
+```sass
+* {
+    rule: ascii   /* A border with plus, hyphen, and vertical bar characters. */
```
willmcgugan, line 23: I guess the `rule` is meant as a placeholder?

This doesn't really serve as an example, since it is invalid CSS.

I think what we need from the examples section is just a few examples of use so the reader can orient themselves. Doesn't have to be an exhaustive list, which is covered nicely in the Syntax section.

Suggest small fragments. For the border something like this:

```sass
#container {
    border: heavy red;
}

#heading {
    border-bottom: solid blue;
}
```

Something similar for the Python section.

docs/snippets/type_syntax/color.md
```diff
@@ -0,0 +1,9 @@
+The legal values for a [`<color>`](/css_types/color) depend on the [class `Color`][textual.color.Color] and include:
```
willmcgugan, line 1: I don't think we need the reference to the Color class here.

```suggestion
A [`<color>`](/css_types/color) should be in one of the following formats:
```

docs/snippets/type_syntax/color.md
```diff
@@ -0,0 +1,9 @@
...
+
+ - a recognised [named color](../../api/color#textual.color--named-colors) (e.g., `red`);
+ - a hexadecimal number representing the RGB values of the color (e.g., `#F35573`);
+ - a color description in the RGB system (e.g., `rgb(23,78,200)`);
+ - a color description in the HSL system (e.g., `hsl(290,70%,80%)`); and
+
+For more details about the exact formats accepted, see [the class method `Color.parse`][textual.color.Color.parse].
```
willmcgugan, line 8: I think this page should be exhaustive regarding the color formats. Since we're talking about CSS, I don't think we should direct people to the code reference.

willmcgugan (commented):
This is looking great! 

I see that the CSS type information is linked to from Styles, but also repeated on the Styles page.

I'm thinking that it may be redundant when devs become more familiar with the CSS. And for beginners it may appear an intimidating amount of information.

Maybe we should remove Values from the styles reference and leave in those links.

rodrigogiraoserrao:
> This is looking great!
> 
> I see that the CSS type information is linked to from Styles, but also repeated on the Styles page.
> 
> I'm thinking that it may be redundant when devs become more familiar with the CSS. And for beginners it may appear an intimidating amount of information.
> 
> Maybe we should remove Values from the styles reference and leave in those links.

Addressed in 42baa0e and 69a1434.

docs/css_types/_template.md
rodrigogiraoserrao: Fixed in 81ddae7

docs/css_types/border.md
rodrigogiraoserrao: Fixed in 81ddae7

docs/snippets/type_syntax/color.md
rodrigogiraoserrao: Fixed in 1ef20eb.

Is it worth adding a formal syntax description of the values accepted, much like we have for the styles?
Something like

```
color: named-color
     | hex-rgb
     | functional-rgb
     | functional-hsl

hex-rgb: # hex hex hex
       | # hex hex hex hex
       | # hex hex hex hex hex hex
       | # hex hex hex hex hex hex hex hex
...
```

docs/snippets/type_syntax/color.md
rodrigogiraoserrao: Fixed in 1ef20eb.

rodrigogiraoserrao:
Please explicitly review 40cbbc3 before accepting this PR.

docs/css_types/border.md
```diff
@@ -0,0 +1,54 @@
...
+| `outer`   | Solid border with additional space around content.       |
+| `round`   | Rounded corners.                                         |
+| `solid`   | Solid border.                                            |
+| `tall`    | Solid border with additional space top and bottom.       |
+| `vkey`    | Vertical key-line border.                                |
+| `wide`    | Solid border with additional space left and right.       |
+
+## Border command
+
+The `textual` CLI has a subcommand which will let you explore the various border types interactively, when applied to the CSS rule [`border`](../styles/border.md):
```
davep, line 29: Perhaps link this to `/guide/devtools/`?

davep (approved):
This looks excellent to me. Shockingly I've not read *every* line (I know, I KNOW!, but scanning through it and taking a sample of various docs and changes it looks spot on. From my perspective I feel any nits or issues that may exist will probably show up during documentation consumption.

willmcgugan:
Amazing work @rodrigogiraoserrao !

I think we will want to make some very minor updates. But rather than make updates to this PR, I think it might be better to merge then make smaller PRs for any edits.

docs/css_types/border.md
rodrigogiraoserrao: As per [Will's comment](https://github.com/Textualize/textual/pull/1408#issuecomment-1377347910) I'll merge this and then we can work on these edits.

PR #1417, opened 2022-12-20
keys refactor

willmcgugan:
- Deprecated PRIORITY_BINDINGS classvar
- Renamed attributes on Key Events
- Documented Key Event object in input.md


src/textual/binding.py
```diff
@@ -71,17 +70,16 @@ def make_bindings(bindings: Iterable[BindingType]) -> Iterable[Binding]:
                 # be a list of keys, so now we unroll that single Binding
                 # into a (potential) collection of Binding instances.
                 for key in binding.key.split(","):
+                    key = key.strip()
+                    if len(key) == 1:
+                        key = _character_to_key(key)
                     yield Binding(
                         key=key.strip(),
```
davep, line 77: This strip can go now.

src/textual/events.py
```diff
@@ -183,55 +182,62 @@ def __rich_repr__(self) -> rich.repr.Result:
...
-    __slots__ = ["key", "char", "key_aliases"]
+    __slots__ = ["key", "character", "aliases"]
 
     def __init__(self, sender: MessageTarget, key: str, char: str | None) -> None:
         super().__init__(sender)
         self.key = key
-        self.char = (key if len(key) == 1 else None) if char is None else char
-        self.key_aliases = [_normalize_key(alias) for alias in _get_key_aliases(key)]
+        self.character = (key if len(key) == 1 else None) if char is None else char
+        self.aliases = _get_key_aliases(key)
```
davep, line 207: `character` and `aliases` appear to be public but not documented.

src/textual/events.py
```diff
@@ -183,55 +182,62 @@ def __rich_repr__(self) -> rich.repr.Result:
...
     Args:
-        sender (MessageTarget): The sender of the event (the App).
-        key (str): A key name (textual.keys.Keys).
-        char (str | None, optional): A printable character or None if it is not printable.
+        sender (MessageTarget): The sender of the event (always the App).
+        key (str): The key that was pressed.
+        char (str | None, optional): A printable character or ``None`` if it is not printable.
 
     Attributes:
         key_aliases (list[str]): The aliases for the key, including the key itself
```
davep, line 198: This property has been removed so this bit of documentation can go.

src/textual/events.py
```diff
@@ -183,55 +182,62 @@ def __rich_repr__(self) -> rich.repr.Result:
...
         yield "is_printable", self.is_printable
-        yield "key_aliases", self.key_aliases, [self.key_name]
+        yield "aliases", self.aliases, [self.key]
 
     @property
-    def key_name(self) -> str | None:
+    def name(self) -> str | None:
         """Name of a key suitable for use as a Python identifier."""
-        return _normalize_key(self.key)
+        return _key_to_identifier(self.key).lower()
```
davep, line 219: The return type says `str | None` but the actual return is just `str`.

tests/test_keys.py
```diff
@@ -0,0 +1,47 @@
...
+        ("?", "question_mark"),
+        ("£", "pound_sign"),
+        (",", "comma"),
+    ],
+)
+def test_character_to_key(character: str, key: str) -> None:
+    assert _character_to_key(character) == key
+
+
+async def test_character_bindings():
```
darrenburns, line 26: Should we include the `ctrl+char` case in one of these tests or is that covered elsewhere?

tests/test_keys.py
```diff
@@ -0,0 +1,47 @@
...
+
+    class BindApp(App):
+        BINDINGS = [(".,~,space", "increment", "foo")]
+
+        def action_increment(self) -> None:
+            nonlocal counter
+            counter += 1
+
+    app = BindApp()
+    print(app._bindings)
```
darrenburns, line 38:
```suggestion
```

src/textual/binding.py
```diff
@@ -32,7 +33,7 @@ class Binding:
     """bool: Show the action in Footer, or False to hide."""
     key_display: str | None = None
     """str | None: How the key should be shown in footer."""
-    priority: bool | None = None
+    priority: bool = False
     """bool | None: Is this a priority binding, checked form app down to focused widget?"""
```
davep, line 41: Type in docs can be changed to just `bool` now.

src/textual/events.py
```diff
@@ -183,55 +182,62 @@ def __rich_repr__(self) -> rich.repr.Result:
...
 
     Attributes:
-        key_aliases (list[str]): The aliases for the key, including the key itself
+        aliases (list[str]): The aliases for the key, including the key itself.
     """
 
-    __slots__ = ["key", "char", "key_aliases"]
+    __slots__ = ["key", "character", "aliases"]
 
     def __init__(self, sender: MessageTarget, key: str, char: str | None) -> None:
```
darrenburns, line 203: Could we rename this to `character`, since that's what is in the docstring?

davep (approved):
Few comments about missing/needing-update docs, mostly, but looks good!

CHANGELOG.md
```diff
@@ -17,9 +18,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Added `textual.actions.SkipAction` exception which can be raised from an action to allow parents to process bindings.
 - Added `textual keys` preview.
+- Added ability to bind to a character in addition to key name. i.e. you can bind to "." or "full_stop".
+- Added TextLog.shrink attribute to allow renderable to reduce in size to fit width.
 
 ### Changed
 
+- Deprecated `PRIORITY_BINDINGS` class variable.
```
rodrigogiraoserrao, line 26: Isn't there a DEPRECATED section just for this?

docs/guide/input.md
```diff
@@ -23,25 +23,52 @@ The most fundamental way to receive input is via [Key](./events/key) events. Let
...
+!!! tip
 
-There are two main attributes on a key event. The `key` attribute is the _name_ of the key which may be a single character, or a longer identifier. Textual ensures that the `key` attribute could always be used in a method name.
+    For a more feature feature rich version of this example, run `textual keys` from the command line.
 
-Key events also contain a `char` attribute which contains a single character if it is printable, or ``None`` if it is not printable (like a function key which has no corresponding character).
+### Key Event
 
-To illustrate the difference between `key` and `char`, try `key01.py` with the space key. You should see something like the following:
+The key event contains following attributes which your app can use to know how to respond.
```
rodrigogiraoserrao, line 34:
```suggestion
The key event contains the following attributes which your app can use to know how to respond.
```

docs/guide/input.md
```diff
@@ -127,24 +154,28 @@ Note how the footer displays bindings and makes them clickable.
...
-!!! note
-
-    Ordinarily a binding on a focused widget has precedence over the same key binding at a higher level. However, bindings at the `App` or `Screen` level always have priority.
+The tuple of three strings may be enough for simple bindings, but you can also replace the tuple with a [Binding][textual.binding.Binding] instance which exposes a few more options.
 
-    The priority of a single binding can be controlled with the `priority` parameter of a `Binding` instance. Set it to `True` to give it priority, or `False` to not.
+### Priority bindings
 
-    The default priority of all bindings on a class can be controlled with the `PRIORITY_BINDINGS` class variable. Set it to `True` or `False` to set the default priroty for all `BINDINGS`.
+Textual checks bindings for the focused widget first. If there is no matching binding, then the parent widget(s) will be checked. Occasionally, you may want bindings to be checked *before* the focused widget. Typically to create hot-keys.
```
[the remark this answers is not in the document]
willmcgugan: reworded.

docs/guide/input.md
```diff
@@ -23,25 +23,52 @@ The most fundamental way to receive input is via [Key](./events/key) events. Let
...
-```
+The `key` attribute is a string which identifies the key that was pressed. The value of `key` will be a single character for letters and numbers, or a longer identifier for other keys.
+
+Some keys may be combined with the ++shift++ key. In the case of letters, this will result in a capital letter as you might expect. For non-printable keys, the `key` attribute will be prefixed with `shift+`. For example, ++shift+home++ will produce an event with `key="shift+home"`.
+
+Many keys can also be combined with ++ctrl++ which will prefix the key with `ctrl+`. For instance, ++ctrl+p++ will produce an event with `key="ctrl+p"`.
+
+!!! warning
+
+    Not all keys combinations are supported in terminals, and some keys may be intercepted by your OS. If in doubt, run `textual keys` from the command line.
```
rodrigogiraoserrao, line 46:
```suggestion
    Not all keys combinations are supported in terminals and some keys may be intercepted by your OS. If in doubt, run `textual keys` from the command line.
```

docs/guide/input.md
```diff
@@ -23,25 +23,52 @@ The most fundamental way to receive input is via [Key](./events/key) events. Let
...
+
+#### character
+
+If the key has an associated printable character then `character` will contain a string with a single unicode character. If there is no printable character for the key (such as for function keys) then `character` will be `None`.
```
rodrigogiraoserrao, line 50:
```suggestion
If the key has an associated printable character, then `character` will contain a string with a single Unicode character. If there is no printable character for the key (such as for function keys) then `character` will be `None`.
```

docs/guide/input.md
```diff
@@ -23,25 +23,52 @@ The most fundamental way to receive input is via [Key](./events/key) events. Let
...
+
+For example the ++p++ key will produce `character="p"` but ++f2++ will produce `character=None`.
+
+#### name
+
+The `name` attribute is similar to `key` but, unlike `key`, is guaranteed to be valid within a Python function name. Textual derives `name` from the `key`  attribute by lower casing it and replacing `+` with `_`. Upper case letters are prefixed with `upper_` to distinguish them from lower case names.
```
rodrigogiraoserrao, line 56:
```suggestion
The `name` attribute is similar to `key` but, unlike `key`, is guaranteed to be valid within a Python function name. Textual derives `name` from the `key` attribute by lower casing it and replacing `+` with `_`. Upper case letters are prefixed with `upper_` to distinguish them from lower case names.
```

tests/test_keys.py
willmcgugan: THere's no equivalent singe character for that though.

docs/guide/input.md
```diff
@@ -23,25 +23,52 @@ The most fundamental way to receive input is via [Key](./events/key) events. Let
...
+#### aliases
+
+Some keys or combinations of keys can produce the same event. For instance, the ++tab++ key is indistinguishable from ++ctrl+i++ in the terminal. For such keys, Textual events will contain a list of the possible keys that may have produced this event. In the case of ++tab++, the `aliases` attribute will contain `["tab", "ctrl+i"]`
 
-Note that the `key` attribute contains the word "space" while the `char` attribute contains a literal space.
 
 ### Key methods
 
-Textual offers a convenient way of handling specific keys. If you create a method beginning with `key_` followed by the name of a key, then that method will be called in response to the key.
+Textual offers a convenient way of handling specific keys. If you create a method beginning with `key_` followed by the key name (the event's `name` attribute), then that method will be called in response to the key.
```
rodrigogiraoserrao, line 71:
```suggestion
Textual offers a convenient way of handling specific keys. If you create a method beginning with `key_` followed by the key name (the event's `name` attribute), then that method will be called in response to the key press.
```

CHANGELOG.md
willmcgugan: Not in this changelog format AFAIK

src/textual/events.py
```diff
@@ -183,55 +182,62 @@ def __rich_repr__(self) -> rich.repr.Result:
...
     def is_printable(self) -> bool:
-        """Return True if the key is printable. Currently, we assume any key event that
-        isn't defined in key bindings is printable.
+        """Check if the key is printable (produces a unicode character).
 
         Returns:
             bool: True if the key is printable.
         """
-        return False if self.char is None else self.char.isprintable()
+        return False if self.character is None else self.character.isprintable()
```
rodrigogiraoserrao, line 235:
```suggestion
        return self.character is not None and self.character.isprintable()
```

Is this too much..? I think it reads more linearly than with the `val if cond else default`.

src/textual/keys.py
```diff
@@ -256,3 +256,14 @@ def _get_key_display(key: str) -> str:
     if unicode_character.isprintable():
         return unicode_character
     return upper_original
+
+
+def _character_to_key(character: str) -> str:
+    """Convert a single character to a key value."""
+    assert len(character) == 1
+    if not character.isalnum():
+        key = unicodedata.name(character).lower().replace("-", "_").replace(" ", "_")
```
rodrigogiraoserrao, line 265: Using `str.translate` with a pre-built table might be faster 🤞

rodrigogiraoserrao (approved):
LGTM

src/textual/events.py
willmcgugan: The ternary operator has generally replaced these short circuiting expressions.

rodrigogiraoserrao: "Generally" in Textual or "generally" in Python? But I'll refrain from suggesting them, then.

willmcgugan: Generally in Python. The short circuiting used to be the only way to do that in a one-liner, but Guido thought it too ugly.

src/textual/keys.py
willmcgugan: True! But doubt it would make a material difference in this case, with such small strings.

Issue #1369, opened 2022-12-15
Setting self.dark in on_load doesn't work: No Screens on stack


willmcgugan:
Suspect we could make this work.

davep:
Minimal code to demonstrate the issue:

```python
from textual.app        import App, ComposeResult
from textual.containers import Container
from textual.widgets    import Header, Footer

class EarlyDark( App[ None ] ):

    CSS = """
    Container {
        background: $panel;
        border: solid $panel-lighten-2;
    }
    """

    BINDINGS = [
        ( "d", "toggle", "Toggle Light/Dark" )
    ]

    def compose( self ) -> ComposeResult:
        yield Header()
        yield Container()
        yield Footer()

    def action_toggle( self ) -> None:
        self.dark = not self.dark

    def on_load( self ) -> None:
        self.dark = False

if __name__ == "__main__":
    EarlyDark().run()
```

The result of running this is the following error on startup:

```
╭──────────────────────────────────────────────────────────── Traceback (most recent call last) ─────────────────────────────────────────────────────────────╮
│ /Users/davep/develop/python/textual-sandbox/early_dark.py:27 in on_load                                                                                    │
│                                                                                                                                                            │
│   24 │   │   self.dark = not self.dark                                                                                                                     │
│   25 │                                                                                                                                                     │
│   26 │   def on_load( self ) -> None:                                                                                                                      │
│ ❱ 27 │   │   self.dark = False                                                                                                                             │
│   28                                                                                                                                                       │
│   29 if __name__ == "__main__":                                                                                                                            │
│   30 │   EarlyDark().run()                                                                                                                                 │
│                                                                                                                                                            │
│ ╭─────────────────────────── locals ───────────────────────────╮                                                                                           │
│ │ self = EarlyDark(title='EarlyDark', classes={'-light-mode'}) │                                                                                           │
│ ╰──────────────────────────────────────────────────────────────╯                                                                                           │
│                                                                                                                                                            │
│ /Users/davep/develop/python/textual-sandbox/.venv/lib/python3.10/site-packages/textual/app.py:463 in watch_dark                                            │
│                                                                                                                                                            │
│    460 │   │   """Watches the dark bool."""                                                                                                                │
│    461 │   │   self.set_class(dark, "-dark-mode")                                                                                                          │
│    462 │   │   self.set_class(not dark, "-light-mode")                                                                                                     │
│ ❱  463 │   │   self.refresh_css()                                                                                                                          │
[...]

PR #1421, opened 2022-12-21
Allow `App.dark` to be set as early as `App.on_load`

davep:
This PR addresses #1369 by swallowing any `ScreenStackError` when attempting to update the styles within `App.watch_dark`; the idea being that if there isn't a screen, one will be coming soon, and the styles will get taken up then anyway.


willmcgugan:
Can we have a regression test that sets `dark` on load please.

davep:
> Can we have a regression test that sets `dark` on load please.

Added.

willmcgugan (approved):

PR #1444, opened 2022-12-26
adds Strip primitive

willmcgugan:
- Introduces a new `Strip` primitive which represents a line worth of segments.
- Added `hits` and `misses` attributes to LRUCache
- Fixed issue where scroll views were refreshed unnecessarily
- Added `textual._cache.FIFOCache`
- Removed thread locks from caches

Because Strip is immutable this allows for caching that wouldn't otherwise have been possible. Not all these optimizations have been fully realized yet


src/textual/_cache.py
```diff
@@ -70,10 +86,9 @@ def grow(self, maxsize: int) -> None:
 
     def clear(self) -> None:
         """Clear the cache."""
-        with self._lock:
-            self._cache.clear()
-            self._full = False
-            self._head = []
+        self._cache.clear()
+        self._full = False
+        self._head = []
```
TimelessP, line 91: Should it also clear hits and misses counts?

willmcgugan: The hits and misses are a debugging aid to see how well the cache is performing. You generally want to know about past performance even when it is cleared.

willmcgugan:
@darrenburns @davep @rodrigogiraoserrao Merging this. You might want to do a retroactive review at some point.

#executivedecision

PR #1466, opened 2023-01-02
don't sleep small intervals

willmcgugan:
A little polish for win sleep. Handle the situation where ctypes is not available (a core dev tells me this is possible on some builds).

### DIFF
--- a/src/textual/_win_sleep.py
+++ b/src/textual/_win_sleep.py
@@ -12,7 +12,6 @@ WAIT_FAILED = 0xFFFFFFFF
 CREATE_WAITABLE_TIMER_HIGH_RESOLUTION = 0x00000002
 
 
-
 def sleep(sleep_for: float) -> None:
     """A replacement sleep for Windows.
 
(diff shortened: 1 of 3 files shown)


src/textual/_win_sleep.py
```diff
@@ -1,55 +1,69 @@
...
+        sleep_for = max(0, secs - 0.001)
+        if sleep_for < 0.0005:
+            # Less than 0.5ms and its not worth doing the sleep
+            return
+
+        handle = kernel32.CreateWaitableTimerExW(
+            None,
+            None,
+            CREATE_WAITABLE_TIMER_HIGH_RESOLUTION,
+            0x1F0003,
```
davep, line 47: For the sake of future readers, perhaps define a constant for `EVENT_ALL_ACCESS` so it's obvious what this is for (or can that be imported from somewhere -- I'm not familiar with Python on Windows).

davep (approved):

PR #1480, opened 2023-01-04
Prevent scrollbar rules from being ignored

rodrigogiraoserrao:

- ~[x] Updated documentation~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## Unreleased
+
+### Fixed
+
+- The styles `scrollbar-background-active` and `scrollbar-color-hover` are no longer ignored https://github.com/Textualize/textual/pull/1480
+
 ## [0.9.1] - 2022-12-30
 
 ### Added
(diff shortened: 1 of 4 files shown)


willmcgugan:
Looks like the scrollbars have a rather sickly yellow color. I'm guessing because colors that were previously ignored are now updated?

You might need to change the defaults, to restore the original look.

Screenshot 2023-01-05 at 11 48 34

rodrigogiraoserrao:
@willmcgugan I only get that “sicklish” tone when scrollbars are being hovered and I thought that was the default styling for `scrollbar-color-hover` (which was being ignored).

Here is a screenshot: on the left, I have my mouse over the scrollbar (which also means that I **do not** have my mouse over the scrollbar on the right).

Screenshot at Jan 05 12-04-06

What color would you prefer, instead, for `scrollbar-color-hover`?

willmcgugan:
Same color as before this update, which I believe didn't change the scrollbar thumb at all...

rodrigogiraoserrao:
Yes, before this update the scrollbar thumb wasn't updated when the scrollbar was hovered.

Upon hovering a scrollbar, do we want the default `scrollbar-color-hover` to _match_ `scrollbar-color` or do we want it to be explicitly that specific shade of blue?

willmcgugan:
Set it to the same shade should be fine. It still allows custom CSS to override it if desired.

willmcgugan:
Looks good!

willmcgugan (approved):

Issue #1396, opened 2022-12-19
[Tree] Add `label` property that calls `set_label`


davep:
It would be really helpful to have a `TreeNode.label` property so that code that's potentially unrelated to what build a `Tree` can access a node's label for display elsewhere.

PR #1487, opened 2023-01-05
Add public access (both get and set) to the `label` of a `TreeNode`

davep:
As per #1396 this adds support for public access to the label of a `TreeNode`, via a `label` property. Both get and set access is provided. In addition to that main change, the constructor of `TreeNode` has been slightly modified so that the incoming label is processed via `Tree.process_label` to match the way that `TreeNode.set_label` works. Without this there was an asymmetry to how a `str` label was set.


rodrigogiraoserrao (approved):
LGTM

src/textual/widgets/_tree.py
```diff
@@ -163,6 +163,17 @@ def toggle(self) -> None:
...
+    @property
+    def label(self) -> TextType:
+        """TextType: The label for the node."""
+        return self._label
+
+    @label.setter
+    def label(self, new_label: TextType) -> TextType:
+        """TextType: The label for the node."""
+        self.set_label(new_label)
+        return self.label
```
rodrigogiraoserrao, line 175: By the way, what does it mean for the setter to return the new value?

davep: Actually, come to think of it, in Python, nothing useful at all. Hangover from elsewhere (languages where assignments are also expressions).

davep: Ditto Clipper with the likes of Class(y), for those of a particular vintage (if I'm recalling correctly).

Issue #1397, opened 2022-12-19
[Tree] Add a read-only parent property to TreeNode


davep:
It would be really helpful if `TreeNode` exposed `parent` as a public property. In combination with something like #1396 it would make for some easy path-building code that wouldn't always need to have knowledge of the underlying data for the `Tree`.

PR #1488, opened 2023-01-05
Add a public read-only parent property to TreeNode

davep:
See #1397.

### DIFF
--- a/tests/tree/test_tree_node_parent.py
+++ b/tests/tree/test_tree_node_parent.py
@@ -1,0 +1,10 @@
+from textual.widgets import TreeNode, Tree
+
+def test_tree_node_parent() -> None:
+    """It should be possible to access a TreeNode's parent."""
+    tree = Tree[None]("Anakin")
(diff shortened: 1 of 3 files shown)


willmcgugan (approved):

tests/tree/test_tree_node_parent.py
```diff
@@ -0,0 +1,10 @@
+from textual.widgets import TreeNode, Tree
+
+def test_tree_node_parent() -> None:
+    """It should be possible to access a TreeNode's parent."""
+    tree = Tree[None]("Anakin")
+    child = tree.root.add("Leia")
+    grandchild = child.add("Ben")
+    assert tree.root.parent is None
+    assert grandchild.parent == child
```
nitzan-shaked, line 9: (apologies if this in nit'ting, I don't mean it as such) I think it might be better to change this `==` and the one below to `is`. Basically I submit that this is the contract -- you get back a node that IS the parent, not just "looks the same". Or at least that's what I think the contract should be.

Issue #1398, opened 2022-12-19
[Tree] Make TreeNode an immutable Sequence to allow for walking the tree


davep:
Especially to allow some initial "just so" setup of a `Tree`, but also to make it easy to write code to walk a `Tree`, it would be useful to have public access to the `children` of a `TreeNode`.

nitzan-shaked:
👍 I just needed this for writing code that "focuses" the tree on a specific node, collapsing all nodes other than one specific node. In my use-case I get only the *names* of the nodes (labels), not the nodes themselves, along them path; so have to traverse children.

willmcgugan:
Making the list public invites a few problems. A dev might expect to be able to extend, remove etc and the tree to update accordingly.

I think we should provide for the use-case of iterating children, and walking a sub-tree without exposing the list.

willmcgugan:
I'm thinking perhaps we should make TreeNode a Sequence, to allow for iterating. This would allow for walking a tree while not exposing internals.

PR #1495, opened 2023-01-05
Add read-only access to the children of a `TreeNode`

davep:
Adds a generic `ImmutableSequence` wrapper class (name and location totally up for debate, although the name of the class got the nod in a Slack chat so I suspect that's okay) and then goes on to use that to wrap the internal child list of a `TreeNode`, making it available in a non-destructive way.

See #1398.

### DIFF
--- a/src/textual/_collections.py
+++ b/src/textual/_collections.py
@@ -15,7 +15,7 @@ class ImmutableSequence(Generic[T]):
         Args:
             wrap (Iterable[T]): The iterable value being wrapped.
         """
-        self._list = list(wrap)
+        self._wrap = tuple(wrap)
 
     @overload
     def __getitem__(self, index: int) -> T:
@@ -27,25 +27,25 @@ class ImmutableSequence(Generic[T]):
 
     def __getitem__(self, index: int | slice) -> T | ImmutableSequence[T]:
         return (
-            self._list[index]
+            self._wrap[index]
             if isinstance(index, int)
-            else ImmutableSequence[T](self._list[index])
+            else ImmutableSequence[T](self._wrap[index])
         )
 
     def __iter__(self) -> Iterator[T]:
-        return iter(self._list)
+        return iter(self._wrap)
 
     def __len__(self) -> int:
-        return len(self._list)
+        return len(self._wrap)
 
     def __length_hint__(self) -> int:
         return len(self)
 
     def __bool__(self) -> bool:
-        return bool(self._list)
+        return bool(self._wrap)
 
     def __contains__(self, item: T) -> bool:
-        return item in self._list
+        return item in self._wrap
 
     def index(self, item: T) -> int:
         """Return the index of the given item.
@@ -59,7 +59,7 @@ class ImmutableSequence(Generic[T]):
         Raises:
             ValueError: If the item is not in the sequence.
         """
-        return self._list.index(item)
+        return self._wrap.index(item)
 
     def __reversed__(self) -> Iterator[T]:
-        return reversed(self._list)
+        return reversed(self._wrap)
--- a/src/textual/_collections.py
+++ b/src/textual/_collections.py
@@ -1,0 +1,65 @@
+"""Provides collection-based utility code."""
+
+from __future__ import annotations
+from typing import Generic, TypeVar, Iterator, overload, Iterable
+
+T = TypeVar("T")
+
+
+class ImmutableSequence(Generic[T]):
+    """Class to wrap a sequence of some sort, but not allow modification."""
+
+    def __init__(self, wrap: Iterable[T]) -> None:
+        """Initialise the immutable sequence.
+
+        Args:
+            wrap (Iterable[T]): The iterable value being wrapped.
+        """
+        self._list = list(wrap)
+
+    @overload
+    def __getitem__(self, index: int) -> T:
+        ...
+
+    @overload
+    def __getitem__(self, index: slice) -> ImmutableSequence[T]:
+        ...
+
+    def __getitem__(self, index: int | slice) -> T | ImmutableSequence[T]:
+        return (
+            self._list[index]
+            if isinstance(index, int)
+            else ImmutableSequence[T](self._list[index])
+        )
+
+    def __iter__(self) -> Iterator[T]:
+        return iter(self._list)
+
+    def __len__(self) -> int:
+        return len(self._list)
+
+    def __length_hint__(self) -> int:
+        return len(self)
+
+    def __bool__(self) -> bool:
+        return bool(len(self))
+
+    def __contains__(self, item: T) -> bool:
+        return item in self._list
+
+    def index(self, item: T) -> int:
+        """Return the index of the given item.
+
+        Args:
+            item (T): The item to find in the sequence.
+
+        Returns:
+            int: The index of the item in the sequence.
+
+        Raises:
+            ValueError: If the item is not in the sequence.
+        """
+        return self._list.index(item)
+
+    def __reversed__(self) -> Iterator[T]:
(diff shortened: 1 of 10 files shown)


src/textual/_collections.py
```diff
@@ -0,0 +1,65 @@
...
+class ImmutableSequence(Generic[T]):
+    """Class to wrap a sequence of some sort, but not allow modification."""
+
+    def __init__(self, wrap: Iterable[T]) -> None:
+        """Initialise the immutable sequence.
+
+        Args:
+            wrap (Iterable[T]): The iterable value being wrapped.
+        """
+        self._list = list(wrap)
```
nitzan-shaked, line 18: wouldn't it be better to just hold a reference to the original sequence? Why copy it? We're not allowing modification anyway.

nitzan-shaked: (and, not to retract from the above: if copying, why not make it tuple() while we're at it, to REALLY disallow modification?)

src/textual/_collections.py
```diff
@@ -0,0 +1,65 @@
...
+        return iter(self._list)
+
+    def __len__(self) -> int:
+        return len(self._list)
+
+    def __length_hint__(self) -> int:
+        return len(self)
+
+    def __bool__(self) -> bool:
+        return bool(len(self))
```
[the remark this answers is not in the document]
davep: Yeah, good call, probably makes sense to whittle it down to that (mostly the above is a handover from a slightly earlier version).

src/textual/_collections.py
davep: > wouldn't it be better to just hold a reference to the original sequence? Why copy it? We're not allowing modification anyway.

The request for this was, in part, that it be able to take any reasonable iterable while allowing it to be used list-like.

nitzan-shaked: I'll give it one more shot, then:

1. I was thinking more something like `KeysView` and `ItemsView` from Python's stdlib. They are "aware of perf tradeoffs", so to speak. So here we would have `SequenceView`, and use it where appropriate. There can be also `ImmutableSequence`, but it doesn't mean it has to be used in the particular case of TreeNode's children.
2. How about adding a check in ImmutableSequence's `__init__()`: if `wrap` is an instance of `Sequence` then just store a reference; other convert to tuple.

Apologies for "insisting", I promise this is the last of them, and hope you find the above useful.

davep: Nah, appreciate it. Of course unlike those views I want this to be a subscriptable object. But yes, you're right, makes more sense to only convert when absolutely needed.

src/textual/_collections.py
```diff
@@ -0,0 +1,65 @@
...
+    def __length_hint__(self) -> int:
+        return len(self)
+
+    def __bool__(self) -> bool:
+        return bool(self._wrap)
+
+    def __contains__(self, item: T) -> bool:
+        return item in self._wrap
+
+    def index(self, item: T) -> int:
```
willmcgugan, line 50: There are a couple more parameters to `index`. Check a recently merged typing PR for the typing.

willmcgugan (changes requested):
re naming, I think this is essentially a view. Much like the dict views.

Suggest `ImmutableSequenceView` as a name. I don't know if it qualifies as a collection, so I think maybe just name the file `"_immutable_sequence_view.py"` for the moment.

Since the use case is to provide a immutable view on to a collection. It shouldn't copy the data internally. Let's store a reference to the `Sequence` parameter supplied in the constructor.

davep:
> re naming, I think this is essentially a view. Much like the dict views.
> 
> Suggest `ImmutableSequenceView` as a name. I don't know if it qualifies as a collection, so I think maybe just name the file `"_immutable_sequence_view.py"` for the moment.

Makes sense, I'll make those changes.

> Since the use case is to provide a immutable view on to a collection. It shouldn't copy the data internally. Let's store a reference to the Sequence parameter supplied in the constructor.

I'm not sure I quite follow this though. As the code stands doesn't it store a reference if it's an indexable sequence type, and only retains a copy if the incoming sequence is one that can't be indexed, allowing it ti be indexed. Perhaps I've misunderstood what was asked for here?

src/textual/_collections.py
davep: Do you mean https://github.com/Textualize/textual/pull/1456?

willmcgugan:
> I'm not sure I quite follow this though. As the code stands doesn't it store a reference if it's an indexable sequence type, and only retains a copy if the incoming sequence is one that can't be indexed, allowing it ti be indexed. Perhaps I've misunderstood what was asked for here?

It's only ever going to be used as a shim on top of a sequence. Can't see any use for creating these for iterators.

src/textual/_collections.py
willmcgugan: yup

davep:
> It's only ever going to be used as a shim on top of a sequence. Can't see any use for creating these for iterators.

Ahh, okay, that makes life easier. I only took that into account because you asked that it do so (accept an iterator too) when we spoke about this yesterday. Happy to drop that part.

davep:
Right, that should be those couple or so issues addressed. Thanks.

willmcgugan (approved):

tests/tree/test_tree_node_children.py
```diff
@@ -0,0 +1,27 @@
...
+from textual.widgets import Tree, TreeNode
+
+def label_of(node: TreeNode[None]):
+    """Get the label of a node as a string"""
+    return str(node.label)
+
+
+def test_tree_node_children() -> None:
+    """A node's children property should act like an immutable list."""
+    CHILDREN=23
```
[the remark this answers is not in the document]
davep: Not by hand on the tests, no. `black` is run by pre-commit of course so if there's a non-black-happy-ness in the above, that would suggest that both pre-commit and CI black checks avoid the unit tests?

Checking:

```sh
$ poetry run black --check .
would reformat devtools/__init__.py
would reformat snapshot_tests/snapshot_apps/horizontal_auto_width.py
would reformat snapshot_tests/snapshot_apps/multiple_css/multiple_css.py
would reformat renderables/test_sparkline.py
would reformat snapshot_tests/snapshot_apps/key_display.py
would reformat renderables/test_text_opacity.py
would reformat test_path.py
would reformat test_focus.py
would reformat test_immutable_sequence_view.py
would reformat test_node_list.py
would reformat renderables/test_underline_bar.py
would reformat test_arrange.py
would reformat test_unmount.py
would reformat test_table.py
would reformat test_dom.py
would reformat test_screens.py
would reformat tree/test_tree_node_label.py
would reformat tree/test_tree_node_children.py
would reformat test_text_backend.py
would reformat test_widget_child_moving.py
would reformat test_reactive.py
would reformat test_widget_mounting.py
would reformat test_widget_removing.py
would reformat test_geometry.py

Oh no! 💥 💔 💥
24 files would be reformatted, 66 files would be left unchanged.
```

Guess none of us are?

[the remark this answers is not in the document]
willmcgugan: I think we may be excluding tests.

Can you check you are also running the latest black specified in dev dependencies.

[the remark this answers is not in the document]
davep:
```sh
$ poetry run black --version
black, 22.10.0 (compiled: yes)
Python (CPython) 3.11.1
```

Which seems to be a later version that specified in poetry.lock?

```
[[package]]
name = "black"
version = "22.8.0"
```

Although weirdly:

```
black = "^22.3.0,<22.10.0"  # macos wheel issue on 22.10
```

from pyproject.toml

Issue #1472, opened 2023-01-03
Meta: faqtory workflow can be made to error when using double quotes in issue subject


davep:
See [this issue](https://github.com/Textualize/textual/issues/1471) and [this result](https://github.com/Textualize/textual/actions/runs/3827865074/jobs/6512835190); the cause of it [being this line](https://github.com/Textualize/textual/blob/a33d87d3d40c6a6e88f92975da7e5aadf76f3f8f/.github/workflows/new_issue.yml#L17).

davep:
At first glance [this looks like it may be relevant](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-an-intermediate-environment-variable).

PR #1500, opened 2023-01-06
Sanitise issue titles before running suggest on them

davep:
Applying
https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-an-intermediate-environment-variable to #1472.

Likely best to double-check this as I'm not actually sure how you can check a workflow before actually deploying it, but if I'm reading things correctly this should make the user of issue titles in the FAQtory workflow safer and less likely to run into errors.


willmcgugan:
Looks correct. Let's give it a shot.

PR #1503, opened 2023-01-06
Converting cell position to index

darrenburns:

- [x] Updated documentation


src/textual/_segment_tools.py
```diff
@@ -14,6 +14,47 @@
...
+    try:
+        while segment_end_index < index:
+            segment = next(iter_segments)
+            text = segment.text
+            segment_length = len(text)
+            segment_cell_length = cell_len(text)
+            cell_position_end += segment_cell_length
+            segment_end_index += segment_length
+    except StopIteration:
+        return 0
```
willmcgugan, line 48: Is `0` the expected return if index is not in segments?

My gut feeling is this should either be the total cell position or an error.

darrenburns: I think an error - I could raise `IndexError`?

willmcgugan (approved):

PR #1518, opened 2023-01-08
Make default filename for screenshots use ISO8601 datetime format

edrogers:
Currently, the `save_screenshot` function uses a datetime format by default that (a) introduces spaces to the resulting filename and (b) does not match what is specified in the doc-string. This PR will replace those spaces and bring the datetime format closer to the ISO8601 format. (Since colons are a reserved character, the format will not exactly match ISO8601.)

Before: (the filename for a screenshot may look like)  
`demo 20230107 205720 203616.svg`  

After: (the filename for a screenshot may look like)  
`demo_2023-01-07T22_52_31.512393.svg`

---

- [X] Updated documentation

---

Feedback welcome


src/textual/app.py
```diff
@@ -644,24 +644,27 @@ def save_screenshot(
...
         Returns:
             str: Filename of screenshot.
         """
         if filename is None:
-            svg_filename = (
-                f"{self.title.lower()} {datetime.now().strftime(time_format)}.svg"
-            )
-            for reserved in '<>:"/\\|?*':
+            if time_format is None:
+                dt = datetime.now().isoformat()
```
davep, line 663: We're not opposed to the change, although it doesn't make a whole lot of difference to how screenshots work. However, I see one small wrinkle that should be handled. This approach will result in more than one `.` in the resulting name. While permissible on all operating systems Textual runs on, it could be confusing for some people in some situations.  I think it would be better to result in an ISO8601-related format that doesn't run into that issue.

edrogers: Yes, I agree, it's a small change that doesn't make a whole lot of difference. Also, I have to admit that ISO8601 is not ideal for a file name -- the microsecond spec includes `:` and `.`. My misgivings with the current default are (a) I'm always nervous about spaces in a file name, and (b) the three groups of numbers -- date, time, and microseconds -- without any other separators look a little cryptic at first.

Do you think simply substituting the `.` used for microseconds with `_`, (eg., `demo_2023-01-07T22_52_31_512393.svg`), would be an improvement?

davep: Aye, that would make sense. Mostly I think we'd be expect people to pass in their own name anyway - the current form is only to help discourage a clash if people forget or choose not to. Swapping out the decimal point seems like the best approach here.

davep (approved):
Looks good, thanks!

MrEnder0 (commented):
Good

Issue #1490, opened 2023-01-05
Watch Highlander


willmcgugan:
- [ ] Watch the 1986 Blockbuster movie Highlander, starring Christopher Lambert and Sean Connery
- [ ] Understand that while the lead was a French actor playing the Scottish hero, it totally made sense to have Sean Connery play a Spanish guy
- [ ] Be able to use the catchline "There can be only one" in conversation
- [ ] Know enough of the lyrics to Queen's ["Its a kind of magic"](https://www.youtube.com/watch?v=ASw32qKppMw&ab_channel=TitanFind) to pretend to sing along
- [ ] Accept the truth that 80s movies are the best movies

davep:
- [ ] For bonus points: point out the Neil Young connection.

davep:
Meta: I think we should pin this issue.

liiight:
Might I suggesting adding "Do NOT see the sequels"

tonybaloney:
Bonus points 

[ ] have the catchphrase “ There can be only one” as a meaningful comment or string in the codebase of textual, e.g, when using a singleton pattern in a test fixture

davep:
My greatest regret so far is not trying to sneak in [the obvious exception name when I added this](https://github.com/Textualize/textual/blob/6cfad7c4b33eb4e4a9dac87ac34563635989c973/src/textual/css/query.py#L47).

davidfokkema:
I don't get it. This issue is still open?!

ofek:
At work we have some members that have not seen The Lord of the Rings movies and I'm trying to incorporate that into our OKRs. This is serious stuff.

jpjenkins:
I was recently in a quaint little Portuguese village and got talking to an elderly portuguese lady in a shop, her English wasn't great and she asks me where I'm from, I tell her Scotland, and she replies "AH, SO YOU'RE AN IMMORTAL!". It was so out of place it took me about half a day to realise she was referring to me being a MacCleud of the clan MacCleud descendant :) made my year.

Zimzozaur:
@ofek you can add this task for the Cyber Sec team: every denied access attempt must be logged with "You Shall Not Pass!" 😄

stonebig:
"There can be only one" is a frequent strategy of IT, unfortunately replacing all the tactic constraints, failsafe considerations and business needs.

regisb:
This issue has been open for more than a year and a half. Is this project dead?

comalice:
[...]

PR #1526, opened 2023-01-08
Annotate only_one comment with additional context

tonybaloney:
See #1490

- [x] Updated documentation

### DIFF
--- a/src/textual/css/query.py
+++ b/src/textual/css/query.py
@@ -246,7 +246,7 @@ class DOMQuery(Generic[QueryType]):
             # should *not* be anything there, so we *should* get an
             # IndexError. We *could* have just checked the length of the
             # query, but the idea here is to do the check as cheaply as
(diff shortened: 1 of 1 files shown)


src/textual/css/query.py
```diff
@@ -246,7 +246,7 @@ def only_one(
             # should *not* be anything there, so we *should* get an
             # IndexError. We *could* have just checked the length of the
             # query, but the idea here is to do the check as cheaply as
-            # possible.
+            # possible. "There can be only one!" -- Duncan McLeod
```
davep, line 249: I feel I need to speak up for the much-maligned Kurgan here, given, IIRC, he's the first one we hear utter these immortal words. ¯\_(ツ)_/¯

tonybaloney: Updated.

davep: Perfect! ⚔️

PR #1535, opened 2023-01-10
Add `Tree.get_node_by_id`

davep:
The thinking here is that a user of a Tree may want to relate notes to other parts of their UI, or other data in their application. While this could be done by keeping a reference to the node itself, it could also be beneficial to just track the ID. Given that ID is a public property of a TreeNode, but given it doesn't currently have any other (public) purpose, this seems to add some useful symmetry.


src/textual/widgets/_tree.py
```diff
@@ -513,6 +513,17 @@ def get_node_at_line(self, line_no: int) -> TreeNode[TreeDataType] | None:
         else:
             return line.node
 
+    def get_node_by_id(self, node_id: NodeID) -> TreeNode[TreeDataType]:
```
willmcgugan, line 516: I think this might deserve a custom exception. A `KeyError` would too easily be confused with another logic error. We would also want the opportunity to add a custom message like "No tree node with id 'foo'".

Let's also add a `Raises:` section in the docstring.

willmcgugan (changes requested):

src/textual/widgets/_tree.py
davep: Sold!

davep: Done: 728acc27c0d90a9e00184871a78df840d1bf69d3

src/textual/widgets/_tree.py
```diff
@@ -515,6 +515,26 @@ def get_node_at_line(self, line_no: int) -> TreeNode[TreeDataType] | None:
         else:
             return line.node
 
+    class UnknownID(Exception):
```
willmcgugan, line 518: Contrary to what I said before, I'm thinking that `UnknownNodeID` would be better. We have other IDs in the code that may confuse.

willmcgugan (changes requested):

src/textual/widgets/_tree.py
davep: Sounds good to me.

src/textual/widgets/_tree.py
```diff
@@ -515,6 +515,26 @@ def get_node_at_line(self, line_no: int) -> TreeNode[TreeDataType] | None:
...
+        Returns:
+            TreeNode[TreeDataType]: The node associated with that ID.
+
+        Raises:
+            Tree.UnknownID: Raised if the `TreeNode` ID is unknown.
+        """
+        try:
+            return self._nodes[node_id]
+        except KeyError:
+            raise self.UnknownID(f"Unknown TreeNode ID: {node_id}") from None
```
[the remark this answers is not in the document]
davep: Aye, makes sense, will do.

src/textual/widgets/_tree.py
davep: Done.

src/textual/widgets/_tree.py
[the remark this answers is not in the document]
davep: Modified.

willmcgugan (approved):

Issue #1218, opened 2022-11-18
Synchronous compute methods should be immediately invoked


darrenburns:
The test below fails because the compute method is not called immediately.
This will catch out users and lead to subtle issues because there's no telling when a computed value will be available.
Synchronous compute methods should be invoked immediately rather than queued, so that the user can safely make use of a computed value after updating a reactive, without worrying about whether it has been recomputed or not.

```python
async def test_superpower_method_call_order():
    class CallOrder(App):
        count = reactive(OLD_VALUE, init=False)
        count_times_ten = reactive(OLD_VALUE * 10)
        calls = []

        def validate_count(self, value: int) -> int:
            self.calls.append(f"validate {value}")
            return value + 1

        def watch_count(self, value: int) -> None:
            self.calls.append(f"watch {value}")

        def compute_count_times_ten(self) -> int:
            self.calls.append(f"compute {self.count}")
            return self.count * 10

    app = CallOrder()
    async with app.run_test():
        app.count = NEW_VALUE
        assert app.calls == [
            # The validator receives NEW_VALUE, since that's what the user
            # set the reactive attribute to...
            f"validate {NEW_VALUE}",
            # The validator adds 1 to the new value, and this is what should
            # be passed into the watcher...
            f"watch {NEW_VALUE + 1}",
            # The compute method accesses the reactive value directly, which
            # should have been updated by the validator to NEW_VALUE + 1.
            f"compute {NEW_VALUE + 1}",
        ]
```

willmcgugan:
I think this is the case now. We need to confirm. Close if it is already working, or fix if not.

darrenburns:
There's an xfail test, so if it started working then [this](https://github.com/Textualize/textual/blob/e968ca0e50faf3b9255fd279a3b19836c49f9255/tests/test_reactive.py#L207) test should have started failing (unless the test is wrong).

rodrigogiraoserrao:
Closed by #1558

Issue #1217, opened 2022-11-18
Improve error message for textual run command with incorrect path to package.


rodrigogiraoserrao:
In some situations, the command `textual run package.path.to.app` returns a cryptic error when the `package.path.to.app` is not an actual package path.

 > Unable to find '' in <module 'package.path.to.app' from
'C:\\Users\\rodri\\Documents\\Textualize\\Textual\\package\\to\\app.py'>

To reproduce this issue, go to the root of the repro:

```shell
λ textual run docs.examples.widgets.button
Unable to find '' in <module 'docs.examples.widgets.button' from
'C:\\Users\\rodri\\Documents\\Textualize\\Textual\\docs\\examples\\widgets\\button.py'>

λ textual run docs.examples.widgets.button.ButtonApp
No module named 'docs.examples.widgets.button.ButtonApp'; 'docs.examples.widgets.button' is not a package
```

However, both commands will work if the file defines an `app`:

If you remove the typical guard `if __name__ == "__main__"` and change the end of the file to the code shown below, then both commands work.

```py
# if __name__ == "__main__":
app = ButtonsApp()
print(app.run())
```

However, the user probably meant to use the command `textual run` with a path file, e.g. with `textual run docs/examples/widgets/button.py` in this case.

willmcgugan:
Now it says:

```
Unable to find 'app' in <module 'docs.examples.widgets.button' from '/Users/willmcgugan/projects/textual/docs/examples/widgets/button.py'>
```

Which is fairly clear, but at some point we will had some more detailed help in the errors.

PR #1539, opened 2023-01-10
Fix compute issues

rodrigogiraoserrao:

- [ ] Updated documentation

### DIFF
--- a/src/textual/reactive.py
+++ b/src/textual/reactive.py
@@ -199,7 +199,10 @@ class Reactive(Generic[ReactiveType]):
         """
         _rich_traceback_guard = True
 
-        async def _set_later(obj: Reactable, compute: str, compute_method: Callable) -> None:
+        async def _set_later(
+            obj: Reactable, compute: str, compute_method: Callable
+        ) -> None:
+            """Coroutine to schedule compute method invocation and attribute update."""
             value = await compute_method()
             setattr(obj, compute, value)
 
@@ -213,7 +216,10 @@ class Reactive(Generic[ReactiveType]):
             if isawaitable(compute_method):
                 # Post message to call compute method when possible.
                 obj.post_message_no_wait(
-                    events.Callback(sender=obj, callback=partial(_set_later, obj, compute, compute_method))
+                    events.Callback(
+                        sender=obj,
+                        callback=partial(_set_later, obj, compute, compute_method),
+                    )
                 )
             else:
                 # Compute method is synchronous so we invoke immediately.
--- a/src/textual/reactive.py
+++ b/src/textual/reactive.py
@@ -182,15 +182,47 @@ class Reactive(Generic[ReactiveType]):
         if current_value != value or self._always_update:
             # Store the internal value
             setattr(obj, self.internal_name, value)
-            # Check all watchers
+            # Check all relevant watchers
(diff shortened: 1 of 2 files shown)


rodrigogiraoserrao:
As it stands, the only commit here is an attempt at fixing #1218 which it does – partially. The `xfail` test is now passing but now I broke a test

```
pytest tests/snapshot_tests/test_snapshots.py::test_order_independence -vv
```

Locally, running the app produces the correct output and we can "fix" the test if we start by pausing 50ms before pressing `"t"`.

However, there _must_ be an issue because now we can't even run the following example:

```
textual run --dev docs/examples/guide/reactivity/computed01.py
```

What is seems is that now I am being too eager in calling compute methods.

As of writing this, my changes make it so that **whenever** a reactive attribute changes, we go over **all** compute methods and try to call them. If they are synchronous, they are called immediately. If they are async, we post a Callback event to call them later. We do this after checking the watchers associated with the attribute.

We do not check all compute methods again after a `watch` method is done running because if the `watch` method changed any reactive attribute, the compute methods will react to that change in time.

This sounds like the strategy that should be followed but it looks like it doesn't work very well during app startup.

rodrigogiraoserrao:
1. A potential fix would entail having the user tag compute methods with the reactive attributes it depends on. Then:
    - we would know exactly what compute methods need to be called when a specific reactive attribute changes; and
    - we could ensure all attribute dependencies have been initialised before running a compute method.
 2. Another potential fix would be to try and initialise all reactive attributes of an object before running compute methods for the first time.

rodrigogiraoserrao:
Plot twist, I was trying to fix #1218 but ended up fixing #1217...

rodrigogiraoserrao:
Plot twist twist: synchronous compute methods really are being called immediately, which is the object of issue 1218. However, the expected output of the `xfail` test for issue 1218 doesn't match the output we are getting because we call compute functions more often that what seems to have been assumed when the original test was written.

In particular, as of this comment, every time a reactive attribute changes, we trigger two compute calls:

 1. the first one, that is expected; but then
 2. the compute method finishes, we set the value of the computed attribute, and that triggers a new wave of `compute` calls

src/textual/reactive.py
```diff
@@ -182,15 +182,47 @@ def __set__(self, obj: Reactable, value: ReactiveType) -> None:
...
+    @classmethod
+    def _check_computes(cls, obj: Reactable) -> None:
+        """Invoke all computes.
+
+        Args:
+            obj (Reactable): Reactable object whose computes will be called.
+        """
+        _rich_traceback_guard = True
+
+        async def _set_later(obj: Reactable, compute: str, compute_method: Callable) -> None:
```
willmcgugan, line 202: Docstrings please. Even on private methods and closures.

willmcgugan (approved):

willmcgugan:
@rodrigogiraoserrao I think there are a number of interactions going on here.

The PR is good, but I think I need to review the reactive stuff. I have a feeling that its like a rubic cube, where each PR is solving one side only.

Let's put this on hold until I've done that.

willmcgugan:
On the right track, but there were larger issues that needed to be addressed at the same time! Closed in favor of https://github.com/Textualize/textual/pull/1558

PR #1541, opened 2023-01-10
Add function to temporarily suspend application mode

JoshKarpel:
This PR adds the `App.suspend()` method that I've been using in my Textual projects. This originally came up in #1150 (and related issues), and a few people have since asked how to do this kind of thing on Discord, so it seems that enough people want this behavior that supporting it in core Textual might make sense. 

I added a section for it in the docs, but I wasn't really sure where to do it, so I just put it under the Actions documentation, assuming that most people would be interested in this in the context of something like "open a text editor when the user presses a key". One thing that was a little awkward is that "application mode" isn't really a term used in the docs, so I had to kind of dance around the language.

- [x] Updated documentation


mkdocs.yml
```diff
@@ -258,7 +258,8 @@ plugins:
               - "!^_"
               - "^__init__$"
               - "!^can_replace$"
-
+          import:
+            - https://docs.python.org/3/objects.inv
```
JoshKarpel, line 262: This allows cross-linking to the standard library docs; used here https://github.com/Textualize/textual/pull/1541/files#diff-681b255c8a47359bc04668c27ab9770e5326a66af7931e4e267e8e205cb43aeeR148

docs/examples/guide/actions/suspend.py
```diff
@@ -0,0 +1,27 @@
...
+    def compose(self) -> ComposeResult:
+        yield Static(
+            "Press r to open the Python built-in REPL. Run quit() to return to the app."
+        )
+        yield Footer()
+
+    def action_open_repl(self):
+        with self.suspend():
+            repl = InteractiveConsole()
+            repl.interact()
```
JoshKarpel, line 22: In #1150 I used
```
    def action_suspend(self) -> None:
        with self.suspend():
            print("Hi!")
            print("Resuming soon...")
            time.sleep(2)
```
as a reproduction case, which would simplify this example if desired, but I figured doing something interactive would be more interesting and inspiring.

willmcgugan (commented):
Nice.

Have you checked this against Windows? I never intended start and stop application mode to be repeatable, so I'm not certain it can start after being stopped.

JoshKarpel:
> Have you checked this against Windows?

Good call; it does not seem to work the same way on Windows, at least with the REPL example. Investigating...

JoshKarpel:
The main problem I was hitting was that, after getting out of the terminal (interesting issue with that; I'll get back to it in a moment), I get:
```
│ C:\Users\joshk\projects\textual\src\textual\drivers\win32.py:164 in enable_application_mode                            │
│                                                                                                                        │
│   161 │   terminal_out = sys.stdout                                                                                    │
│   162 │                                                                                                                │
│   163 │   current_console_mode_in = _get_console_mode(terminal_in)                                                     │
│ ❱ 164 │   current_console_mode_out = _get_console_mode(terminal_out)                                                   │
│   165 │                                                                                                                │
│   166 │   def restore() -> None:                                                                                       │
│   167 │   │   """Restore console mode to previous settings"""                                                          │
│                                                                                                                        │
│ ╭──────────────────────────────────────── locals ────────────────────────────────────────╮                             │
│ │ current_console_mode_in = 487                                                          │                             │
│ │             terminal_in = <_io.TextIOWrapper name='<stdin>' mode='r' encoding='utf-8'> │                             │
│ │            terminal_out = <textual.app._NullFile object at 0x000001BEE87E1D90>         │                             │
│ ╰────────────────────────────────────────────────────────────────────────────────────────╯                             │
│                                                                                                                        │
│ C:\Users\joshk\projects\textual\src\textual\drivers\win32.py:147 in _get_console_mode                                  │
│                                                                                                                        │
│   144 │   Returns:                                                                                                     │
│   145 │   │   int: The current console mode.                                                                           │
│   146 │   """                                                                                                          │
│ ❱ 147 │   windows_filehandle = msvcrt.get_osfhandle(file.fileno())                                                     │
│   148 │   mode = wintypes.DWORD()                                                                                      │
│   149 │   KERNEL32.GetConsoleMode(windows_filehandle, ctypes.byref(mode))                                              │
│   150 │   return mode.value                                                                                            │
│                                                                                                                        │
│ ╭────────────────────────── locals ───────────────────────────╮                                                        │
│ │ file = <textual.app._NullFile object at 0x000001BEE87E1D90> │                                                        │
│ ╰─────────────────────────────────────────────────────────────╯                                                        │
╰────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
AttributeError: '_NullFile' object has no attribute 'fileno'
```

Which is happening because we're still inside this block when `suspend()` is running: https://github.com/JoshKarpel/textual/blob/ee088d2d921890b382094e117ae9a2ed116bca47/src/textual/app.py#L1527-L1530 

I believe this can be resolved by using
```python
    terminal_in = sys.__stdin__
    terminal_out = sys.__stdout__
```
in `win32.enable_application_mode()` - those always point to the original stdin and stdout (only stdout is needed here, but I did stdin too for symmetry, though maybe there are other places that are relying on stdin redirection?).

Second issue: no `readline` on non-Unix.

Tertiary issue: running `quit()` in the embedded terminal on Windows kills the Textual app entirely (no error, it just exits), while on Unix it puts you back in the app. I suspect this is because it turns out) that `quit()` in the REPL closes stdin. Though I like this example it might be too thorny to have consistent and unsurprising behavior; I will think a bit on a good replacement that is hopefully still interactive.

JoshKarpel:
Alright, take two, now with better cross-platform support! I realized I could do `subprocess.run(sys.executable)` to get consistent behavior across platforms by decoupling how the REPL closes from the Textual app that's running it. I've also got a more meaningful test there to keep things working.

docs/guide/actions.md
```diff
@@ -132,3 +132,23 @@ Textual supports the following builtin actions which are defined on the app.
 - [action_switch_screen][textual.app.App.action_switch_screen]
 - [action_screenshot][textual.app.App.action_screenshot]
 - [action_toggle_dark][textual.app.App.action_toggle_dark]
+
+## Temporarily suspending the application
```
willmcgugan, line 136: I think "App Basics" might be a better place for this. It's only tangentially related to actions.

src/textual/app.py
```diff
@@ -947,6 +954,32 @@ async def run_app() -> None:
...
+            driver.stop_application_mode()
+            with redirect_stdout(sys.__stdout__), redirect_stderr(sys.__stderr__):
+                # This sleep seems to help avoid timing problems.
+                # Without it, if you do something like
+                #
+                #   with app.suspend():
+                #       print("hi")
+                #
+                # The print will sometimes not go to sys.__stdout__ as expected.
+                time.sleep(0.025)
```
willmcgugan, line 979: This may be masking a race condition. I wonder if we need to do a flush after stopping application mode so no escape sequences arrive after the yield.

willmcgugan (changes requested):

src/textual/app.py
JoshKarpel: 🤦🏻‍♂️  of course! I will experiment.

JoshKarpel: Initial results: `self.console.file.flush()` before `yield` *does not* fix the problem. Given that it's random whether the prints appear or not, the sleep is definitely masking a race condition, but I'm not sure where it is.

But here's where it gets extra weird: I tried adding some debug code to `_NullFile.write` to see if the write were still going there, but it seems like they aren't, which makes it seem like they're just completely vanishing, which doesn't make much sense. The investigation continues...

willmcgugan: Curious! Do you think `"hi!"` is not being written at all?

We have a thread that writes to stdout, so that we can prepare the next frame while the previous frame is being printed. I wonder if that thread is still writing in the background for a fraction of a second.

JoshKarpel: Aha! That's promising - I didn't realize that `self.console.file` was that thread (should have `print`ed it!) https://github.com/Textualize/textual/blob/0be1d96107adffcec748dcd89e04be0422ff61b0/src/textual/app.py#L284-L299

And `_WriterThread.flush` is a no-op! https://github.com/Textualize/textual/blob/0be1d96107adffcec748dcd89e04be0422ff61b0/src/textual/app.py#L190-L192

I bet the problem lies in there - I'll try to find some time to investigate today.

davep:
Currently looking to see if this will play well with #1582 -- right now playing in https://github.com/davep/textual/tree/sleep-resume; nothing concrete yet, more mulling over how it might work.

davep:
Now part of https://github.com/Textualize/textual/pull/1655 and happily helping #1582 happen.

davep:
> Nice! Feel free to close this one and do changes/review in your branch - whatever workflow works for you.

Will do, thanks. I'll keep this one open for the moment, just until I'm happy that the other is heading in the right direction. I would have added to this one but I utterly failed to work out which magic incantation would let me grab a PR from upstream into the origin of my fork and let me work on that. Got to love `git`. 😄 

> I keep getting merge conflicts on the changelog trying to keep up from a different timezone 🤣

Pfft; we're all in the same timezone and that still happens. ¯\\_(ツ)_/¯

davep:
Wee bit later on but... making it over the line as part of #4064.

Issue #1542, opened 2023-01-11
Add 'textual diagnose' CLI command


willmcgugan:
We need a textual diagnose command which writes out some basic diagnostic information. We can ask for this in issue templates.

- Textual version
- Python version
- Path to Python executable
- OS
- Detect terminal, env vars TERM, COLORTERM, other terminal vars
- Output of rich.console.Console.options

The output may be colorized, but should be in an easy cut n pastable format.

willmcgugan:
We need a `textual diagnose` command which writes out some basic diagnostic information. We can ask for this in issue templates.

- Textual version
- Python version
- Path to Python executable
- OS
- Detect terminal, env vars TERM, COLORTERM, other terminal vars
- Output of rich.console.Console.options

The output may be colorized, but should be in an easy cut n pastable format.

PR #1545, opened 2023-01-11
Add a textual diagnose CLI command

davep:
This, in this early version anyway, emits a simple dump of GitHub-issue-friendly markdown text that lists all of the key information we may want to know when someone using having a problem with Textual.

See #1542.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Added read-only public access to the children of a `TreeNode` via `TreeNode.children` https://github.com/Textualize/textual/issues/1398
 - Added `Tree.get_node_by_id` to allow getting a node by its ID https://github.com/Textualize/textual/pull/1535
 - Added a `Tree.NodeHighlighted` message, giving a `on_tree_node_highlighted` event handler https://github.com/Textualize/textual/issues/1400
+- Added `diagnose` as a `texttual` command https://github.com/Textualize/textual/issues/1542
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Added read-only public access to the children of a `TreeNode` via `TreeNode.children` https://github.com/Textualize/textual/issues/1398
 - Added `Tree.get_node_by_id` to allow getting a node by its ID https://github.com/Textualize/textual/pull/1535
 - Added a `Tree.NodeHighlighted` message, giving a `on_tree_node_highlighted` event handler https://github.com/Textualize/textual/issues/1400
+- Added `diagnose` as a `texttual` command https://github.com/Textualize/textual/issues/1542
```
rodrigogiraoserrao, line 17:
```suggestion
- Added `diagnose` as a `textual` command https://github.com/Textualize/textual/issues/1542
```

src/textual/cli/tools/diagnose.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+def _section(title: str, values: dict[str, str]) -> None:
+    """Print a collection of named values within a titled section.
+
+    Args:
+        title (str): The title for the section.
+        values (dict[str, str]): The values to print out.
+    """
+    max_name = len(max(list(values.keys()), key=len))
```
rodrigogiraoserrao, line 17: It's a nitpick, but something like

```suggestion
    max_name = max(map(len, values.keys()))
```

Looks a bit less redundant, no?
(Ditto for max_values.)

rodrigogiraoserrao (approved):

CHANGELOG.md
davep: Good catch!

willmcgugan (approved):

src/textual/cli/tools/diagnose.py
davep: Actually, yeah, I think that's my preference too -- too much in the habit of staying away from `map` and friends as so much of the Python world seems to dislike "proper" functional programming. ;-)

rodrigogiraoserrao: `max(len(k) for k in values.keys())` would also come to mind :D

Issue #1532, opened 2023-01-09
Document compound widgets


willmcgugan:
There is a placeholder section in the Widgets guide for "Compound widgets", i.e. widgets which contain widgets.

This section needs to be written,  with a decent example, and an explanation of setting data on child widgets, and responding to event.

We need to emphasize that the parent widget can set attributes, but child widgets should `emit` events.

rodrigogiraoserrao:
This had been tentatively closed by #1818 (now https://mathspp.com/blog/textual-tutorial-build-a-todo-app-in-python) and has been closed by #1952.

PR #1546, opened 2023-01-11
Make `COMPONENT_CLASSES` inheritable and add classvar to prevent inheritance.

### LINKED ISSUE #1399: Make `COMPONENT_CLASSES` inheritable
Much like `DEFAULT_CSS` and similar things, look at the idea of making `COMPONENT_CLASSES` inheritable. At the moment, if you inherit from a widget that has a `COMPONENT_CLASSES`, and you set your own in your derived widget, you'll erase those of the parent. Right now the best option is to combine the sets. For example:

```python
class ThreeDimensionalTree(Tree[Anaglyph]):
    """Tree, but more so."""

    COMPONENT_CLASSES: ClassVar[set[str]] = Tree.COMPONENT_CLASSES | {
        "anaglyph--behind-screen",
        "anaglyph--in-front-of-screen",
    }
```

This will likely also need the addition of a `inherit_component_classes` to control if the inheriting should happen or not.

rodrigogiraoserrao:

- ~[ ] Updated documentation~ Currently there's no documentation on `COMPONENT_CLASSES` and my suggestion would be to add it with #1532 

This fixes #1399

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Added read-only public access to the children of a `TreeNode` via `TreeNode.children` https://github.com/Textualize/textual/issues/1398
 - Added `Tree.get_node_by_id` to allow getting a node by its ID https://github.com/Textualize/textual/pull/1535
 - Added a `Tree.NodeHighlighted` message, giving a `on_tree_node_highlighted` event handler https://github.com/Textualize/textual/issues/1400
+- Added a `_inherit_component_classes` class variable to control whether or not component classes are inherited from base classes https://github.com/Textualize/textual/issues/1399
 
 ### Changed
 
@@ -21,6 +22,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Fail-fast and print pretty tracebacks for Widget compose errors https://github.com/Textualize/textual/pull/1505
 - Added Widget._refresh_scroll to avoid expensive layout when scrolling https://github.com/Textualize/textual/pull/1524
 - `events.Paste` now bubbles https://github.com/Textualize/textual/issues/1434
+- `COMPONENT_CLASSES` are now inherited from base classes https://github.com/Textualize/textual/issues/1399
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Added read-only public access to the children of a `TreeNode` via `TreeNode.children` https://github.com/Textualize/textual/issues/1398
 - Added `Tree.get_node_by_id` to allow getting a node by its ID https://github.com/Textualize/textual/pull/1535
 - Added a `Tree.NodeHighlighted` message, giving a `on_tree_node_highlighted` event handler https://github.com/Textualize/textual/issues/1400
-- Added a `_inherit_component_classes` class variable to control whether or not component classes are inherited from base classes https://github.com/Textualize/textual/issues/1399
+- Added a `inherit_component_classes` subclassing parameter to control whether or not component classes are inherited from base classes https://github.com/Textualize/textual/issues/1399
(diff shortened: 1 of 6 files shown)


willmcgugan:
@rodrigogiraoserrao do you have the latest Black? You might want to configure your editor to format on save.

src/textual/dom.py
```diff
@@ -98,6 +98,9 @@ class DOMNode(MessagePump):
     # True if this node inherits the CSS from the base class.
     _inherit_css: ClassVar[bool] = True
 
+    # True if this node inherits the component classes from the base class.
+    _inherit_component_classes: ClassVar[bool] = True
```
willmcgugan, line 102: The other `inherit_*` values are set from `__init_subclass__`. We should probably do the same for this.

src/textual/dom.py
```diff
@@ -279,6 +282,25 @@ def get_path(base: Type[DOMNode]) -> str:
 
         return css_stack
 
+    def get_component_classes(self) -> set[str]:
```
willmcgugan, line 285: Any reason why this should be public?

Unless there is a good reason to expose this to the dev, we should make it "private".

tests/test_dom.py
```diff
@@ -115,6 +115,53 @@ class E(D):
...
 
+def test_component_classes_inheritance():
+    """Test if component classes are inherited properly."""
+
+    class A(DOMNode):
+        COMPONENT_CLASSES = {"a-1", "a-2"}
+
+    class B(A):
+        COMPONENT_CLASSES = {"b-1"}
+        _inherit_component_classes = False
```
davep, line 126: Given that this will be useful for people using the library, wanting to inherit from a widget and decide if they want the component classes or not, does it make sense to have the inherit control switch be a "private" value?

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Added read-only public access to the children of a `TreeNode` via `TreeNode.children` https://github.com/Textualize/textual/issues/1398
 - Added `Tree.get_node_by_id` to allow getting a node by its ID https://github.com/Textualize/textual/pull/1535
 - Added a `Tree.NodeHighlighted` message, giving a `on_tree_node_highlighted` event handler https://github.com/Textualize/textual/issues/1400
+- Added a `_inherit_component_classes` class variable to control whether or not component classes are inherited from base classes https://github.com/Textualize/textual/issues/1399
```
willmcgugan, line 17: Probably not any point in a changelog for a private variable, as it doesn't impact the typical Textual dev.

willmcgugan (changes requested):
A few requests

rodrigogiraoserrao:
> @rodrigogiraoserrao do you have the latest Black? You might want to configure your editor to format on save.

I don't know what I did to my environments 🤷‍♀️ somehow pre-commit was no longer running but I've set it up again.

tests/test_dom.py
rodrigogiraoserrao: Thanks for pointing this out! The problem is that I misunderstood how `_inherit_component_classes` should be set.

src/textual/dom.py
rodrigogiraoserrao: I made it public instead of private because `Stylesheet.apply` needs it and because this feels analogous to `get_default_css`, which is public.

davep: If this is staying "public", it might make sense to reword the docstring so that it doesn't refer to `_inherit_component_classes`.

willmcgugan: We use a more relaxed definition of public and private. Essentially `public` is something that the Textual dev may need to use. If it is only used by the Textual system then it should be `private`.

A better term, might be `internal`.

rodrigogiraoserrao: I reworded the docstring, good catch!

I don't think the average Textual dev will need to use `get_component_classes` but I did this in light of `get_default_css` because it is more or less the same level of thing.
Should I prefix both with an underscore `_`?

willmcgugan: Sure. May as well.

willmcgugan (approved):

PR #1547, opened 2023-01-11
DataTable improvements (and more)

darrenburns:
## Added

- Different cursor types: cell (pre-existing), row, column, none. If you want no cursor at all, you can use the "none" cursor.
  - Scrolling (with arrow keys) perpendicular to the cursor type will move the cursor.
  - Scrolling (with arrow keys) parallel to the cursor type will scroll as normal without shifting the cursor.
  - When "none" cursor is used, arrow keys are used just for scrolling.
- Ability to switch cursor off `table.show_cursor = False`. This will hide the cursor. The cursor state is retained when re-enabled. When you disable it, scrolling with arrow keys works as normal. When you re-enable it, the cursor will be scrolled to.
- Hover cursor will now only become visible when you interact with the table using the mouse. It'll hide again when you use the keyboard.
- Events for row/col/cell highlighting and selection.
- Improving reactive tests - some of them had redundant code or things commented out that were previously broken.
- Extract `Coord` class into own module for reuse, and rename it to `Coordinate`.

https://user-images.githubusercontent.com/5740731/212922349-8015891c-a05e-448c-9430-049b93b8cc37.mov

## Issues encountered

- [x] `show_cursor` was part of the API but not actually implemented.
- [x] When fixed columns are used, the cursor isn't visible until you move it.
- [ ] Component style for datatable--even-row doesn't apply at all (but odd works).
- [ ] Crash if number of fixed_columns is equal to number of columns (presumably when greater too).
- [x] (Possibly missing API) We need to clamp cursor position when number of fixed_columns changes, so we clamp and then assign it. Since we assign it, it triggers the validator which clamps it again, repeating the work.
- [ ] Vertical scrollbar doesn't update when `show_header` set to `False`.

- [x] Updated documentation


src/textual/widgets/_data_table.py
```diff
@@ -169,13 +132,23 @@ class DataTable(ScrollView, Generic[CellType], can_focus=True):
 
     COMPONENT_CLASSES: ClassVar[set[str]] = {
         "datatable--header",
+        "datatable--cursor-fixed",
+        "datatable--highlight-fixed",
```
darrenburns, line 136: We'll probably need to change the names of these when we do "labelled" rows/columns - the fixed/header concepts are a little conflated at the moment.

src/textual/widgets/_data_table.py
```diff
@@ -184,8 +157,10 @@ class DataTable(ScrollView, Generic[CellType], can_focus=True):
     show_cursor = Reactive(True)
     cursor_type = Reactive(CELL)
 
-    cursor_cell: Reactive[Coord] = Reactive(Coord(0, 0), repaint=False)
-    hover_cell: Reactive[Coord] = Reactive(Coord(0, 0), repaint=False)
+    cursor_cell: Reactive[Coordinate] = Reactive(
+        Coordinate(0, 0), repaint=False, always_update=True
```
darrenburns, line 161: `always_update` is required as sometimes we need to force validation via `self.cursor_cell = self.cursor_cell`.

willmcgugan (approved):
LGTM

We should think about an example that uses the data-table. One for another PR.

Issue #1201, opened 2022-11-17
Programatic scrolling should work without scrollbars (overflow hidden).


willmcgugan:
Currently if there are no scrollbars, and you call `Widget.scroll_to`, the scrolling us suppressed. This is so that moving the scroll wheel doesn't scroll when the CSS has disabled scrolling.

However, some times you may to scroll a container programatically. Suggest we need to differentiate scrolling initiated from mouse / keyboard input and programatic scrolling.

davep:
Quick test code to visually confirm the issue (using `scroll_to_widget`, which I imagine is part of this issue too?):

```python
from textual.app        import App, ComposeResult
from textual.containers import Vertical
from textual.widgets    import Header, Footer, Button
from textual.binding    import Binding

class ScrollToTester( App[ None ] ):

    CSS = """
    Vertical {
        overflow-y: hidden;
    }

    Button {
        margin: 1;
        width: 100%;
    }
    """

    BINDINGS = [
        Binding( "j", "jump", "Jump to a button off the screen" ),
    ]

    def compose( self ) -> ComposeResult:
        yield Header()
        yield Vertical(
            *[ Button( str( n ), id=f"btn-{n}" ) for n in range( 200 ) ]
        )
        yield Footer()

    def action_jump( self ) -> None:
        btn = self.query_one( "#btn-150", Button )
        self.query_one( Vertical ).scroll_to_widget( btn )
        btn.focus()

if __name__ == "__main__":
    ScrollToTester().run()
```

Pressing <kbd>j</kbd> should bring button 150 into view.

davep:
Note to self: diving into the code, in the above situation, `scroll_to` gets called *and* returns `True`; so as far as it's concerned it did scroll. Whatever needs changing, it looks like it's down in the animation code, most likely.

davep:
Tracing further, a quick fix to this is this:

```diff
diff --git a/src/textual/widget.py b/src/textual/widget.py
index 7808f1a0..67c140b7 100644
--- a/src/textual/widget.py
+++ b/src/textual/widget.py
@@ -811,8 +811,8 @@ class Widget(DOMNode):
     def watch_scroll_y(self, old_value: float, new_value: float) -> None:
         if self.show_vertical_scrollbar:
             self.vertical_scrollbar.position = round(new_value)
-            if round(old_value) != round(new_value):
-                self._refresh_scroll()
+        if round(old_value) != round(new_value):
+            self._refresh_scroll()

[...]

PR #1550, opened 2023-01-12
Allow by-code scrolling even when by-input scrolling would be denied

davep:
See #1201.

See
https://github.com/Textualize/textual/issues/1201#issuecomment-1380123660 though.

### DIFF
--- a/src/textual/widget.py
+++ b/src/textual/widget.py
@@ -803,10 +803,12 @@ class Widget(DOMNode):
             self.highlight_link_id = hover_style.link_id
 
     def watch_scroll_x(self, old_value: float, new_value: float) -> None:
+        self.horizontal_scrollbar.position = round(new_value)
         if round(old_value) != round(new_value):
             self._refresh_scroll()
 
     def watch_scroll_y(self, old_value: float, new_value: float) -> None:
+        self.vertical_scrollbar.position = round(new_value)
         if round(old_value) != round(new_value):
             self._refresh_scroll()
 
--- a/src/textual/widget.py
+++ b/src/textual/widget.py
@@ -803,14 +803,10 @@ class Widget(DOMNode):
             self.highlight_link_id = hover_style.link_id
 
     def watch_scroll_x(self, old_value: float, new_value: float) -> None:
-        if self.allow_horizontal_scroll:
-            self.horizontal_scrollbar.position = round(new_value)
         if round(old_value) != round(new_value):
             self._refresh_scroll()
 
     def watch_scroll_y(self, old_value: float, new_value: float) -> None:
-        if self.allow_vertical_scroll:
-            self.vertical_scrollbar.position = round(new_value)
         if round(old_value) != round(new_value):
             self._refresh_scroll()
 
--- a/src/textual/widget.py
+++ b/src/textual/widget.py
@@ -1349,7 +1349,7 @@ class Widget(DOMNode):
             duration (float | None, optional): Duration of animation, if animate is True and speed is None.
             easing (EasingFunction | str | None, optional): An easing method for the scrolling animation. Defaults to "None",
                 which will result in Textual choosing the default scrolling easing function.
-            force (bool,optional): Force scrolling even when prohibited by overflow styling. Defaults to `False`.
+            force (bool, optional): Force scrolling even when prohibited by overflow styling. Defaults to `False`.
 
         Returns:
             bool: True if the scroll position changed, otherwise False.
@@ -1421,7 +1421,7 @@ class Widget(DOMNode):
             duration (float | None, optional): Duration of animation, if animate is True and speed is None.
             easing (EasingFunction | str | None, optional): An easing method for the scrolling animation. Defaults to "None",
                 which will result in Textual choosing the configured default scrolling easing function.
-            force (bool,optional): Force scrolling even when prohibited by overflow styling. Defaults to `False`.
+            force (bool, optional): Force scrolling even when prohibited by overflow styling. Defaults to `False`.
 
         Returns:
             bool: True if the scroll position changed, otherwise False.
@@ -1453,7 +1453,7 @@ class Widget(DOMNode):
             duration (float | None, optional): Duration of animation, if animate is True and speed is None.
             easing (EasingFunction | str | None, optional): An easing method for the scrolling animation. Defaults to "None",
                 which will result in Textual choosing the configured default scrolling easing function.
-            force (bool,optional): Force scrolling even when prohibited by overflow styling. Defaults to `False`.
+            force (bool, optional): Force scrolling even when prohibited by overflow styling. Defaults to `False`.
 
         Returns:
             bool: True if any scrolling was done.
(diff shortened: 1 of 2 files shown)


src/textual/widget.py
```diff
@@ -1853,6 +1902,7 @@ def scroll_visible(
             top (bool, optional): Scroll to top of container. Defaults to False.
             easing (EasingFunction | str | None, optional): An easing method for the scrolling animation. Defaults to "None",
                 which will result in Textual choosing the configured default scrolling easing function.
+            force (bool,optional): Force scrolling even when prohibited by overflow styling. Defaults to `False`.
```
rodrigogiraoserrao, line 1905:
```suggestion
            force (bool, optional): Force scrolling even when prohibited by overflow styling. Defaults to `False`.
```

src/textual/widget.py
```diff
@@ -1805,6 +1851,7 @@ def scroll_to_region(
             easing (EasingFunction | str | None, optional): An easing method for the scrolling animation. Defaults to "None",
                 which will result in Textual choosing the configured default scrolling easing function.
             top (bool, optional): Scroll region to top of container. Defaults to False.
+            force (bool,optional): Force scrolling even when prohibited by overflow styling. Defaults to `False`.
```
rodrigogiraoserrao, line 1854:
```suggestion
            force (bool, optional): Force scrolling even when prohibited by overflow styling. Defaults to `False`.
```

src/textual/widget.py
```diff
@@ -1743,6 +1786,7 @@ def scroll_to_widget(
             easing (EasingFunction | str | None, optional): An easing method for the scrolling animation. Defaults to "None",
                 which will result in Textual choosing the configured default scrolling easing function.
             top (bool, optional): Scroll widget to top of container. Defaults to False.
+            force (bool,optional): Force scrolling even when prohibited by overflow styling. Defaults to `False`.
```
rodrigogiraoserrao, line 1789:
```suggestion
            force (bool, optional): Force scrolling even when prohibited by overflow styling. Defaults to `False`.
```

src/textual/widget.py
```diff
@@ -1708,6 +1748,7 @@ def scroll_page_right(
             duration (float | None, optional): Duration of animation, if animate is True and speed is None.
             easing (EasingFunction | str | None, optional): An easing method for the scrolling animation. Defaults to "None",
                 which will result in Textual choosing the configured default scrolling easing function.
+            force (bool,optional): Force scrolling even when prohibited by overflow styling. Defaults to `False`.
```
rodrigogiraoserrao, line 1751:
```suggestion
            force (bool, optional): Force scrolling even when prohibited by overflow styling. Defaults to `False`.
```

src/textual/widget.py
```diff
@@ -1677,6 +1714,7 @@ def scroll_page_left(
...
```
rodrigogiraoserrao, line 1717:
```suggestion
            force (bool, optional): Force scrolling even when prohibited by overflow styling. Defaults to `False`.
```

src/textual/widget.py
```diff
@@ -1648,6 +1682,7 @@ def scroll_page_down(
...
```
rodrigogiraoserrao, line 1685:
```suggestion
            force (bool, optional): Force scrolling even when prohibited by overflow styling. Defaults to `False`.
```

src/textual/widget.py
```diff
@@ -1590,6 +1618,7 @@ def scroll_up(
...
```
rodrigogiraoserrao, line 1621:
```suggestion
            force (bool, optional): Force scrolling even when prohibited by overflow styling. Defaults to `False`.
```

src/textual/widget.py
```diff
@@ -1561,6 +1586,7 @@ def scroll_down(
...
```
rodrigogiraoserrao, line 1589:
```suggestion
            force (bool, optional): Force scrolling even when prohibited by overflow styling. Defaults to `False`.
```

src/textual/widget.py
```diff
@@ -1532,6 +1554,7 @@ def scroll_right(
...
```
rodrigogiraoserrao, line 1557:
```suggestion
            force (bool, optional): Force scrolling even when prohibited by overflow styling. Defaults to `False`.
```

src/textual/widget.py
```diff
@@ -1503,6 +1522,7 @@ def scroll_left(
...
```
rodrigogiraoserrao, line 1525:
```suggestion
            force (bool, optional): Force scrolling even when prohibited by overflow styling. Defaults to `False`.
```

src/textual/widget.py
```diff
@@ -1471,6 +1487,7 @@ def scroll_end(
...
```
rodrigogiraoserrao, line 1490:
```suggestion
            force (bool, optional): Force scrolling even when prohibited by overflow styling. Defaults to `False`.
```

src/textual/widget.py
```diff
@@ -1416,6 +1421,7 @@ def scroll_relative(
...
```
rodrigogiraoserrao, line 1424:
```suggestion
            force (bool, optional): Force scrolling even when prohibited by overflow styling. Defaults to `False`.
```

src/textual/widget.py
```diff
@@ -1348,10 +1349,13 @@ def scroll_to(
             duration (float | None, optional): Duration of animation, if animate is True and speed is None.
             easing (EasingFunction | str | None, optional): An easing method for the scrolling animation. Defaults to "None",
                 which will result in Textual choosing the default scrolling easing function.
+            force (bool,optional): Force scrolling even when prohibited by overflow styling. Defaults to `False`.
```
rodrigogiraoserrao, line 1352:
```suggestion
            force (bool, optional): Force scrolling even when prohibited by overflow styling. Defaults to `False`.
```

There are a bunch of these, so maybe Search & Replace and then a single commit? 🤷

src/textual/widget.py
```diff
@@ -803,16 +803,16 @@ def watch_hover_style(
             self.highlight_link_id = hover_style.link_id
 
     def watch_scroll_x(self, old_value: float, new_value: float) -> None:
-        if self.show_horizontal_scrollbar:
+        if self.allow_horizontal_scroll:
             self.horizontal_scrollbar.position = round(new_value)
```
rodrigogiraoserrao, line 807: I was reading your comment in the original issue and I'm not convinced `allow_horizontal_scroll` should be used here. Note that `scroll_x` was changed, which is why we are in `watch_scroll_x`, so now that `scroll_x` was changed, we need to update the horizontal scrollbar if it is visible.

To me, scrollbar being visible != scrolling enabled, so we really want to check if we are _showing_ the scrollbar.

rodrigogiraoserrao (changes requested):

src/textual/widget.py
willmcgugan: Hasn't the logic already been done in the various scroll methods, with regards to allowing scrolling?

By the time we get here, I don't think we need to apply the logic again.

This does mean that you can circumvent scroll logic by setting the `scroll_x`  attribute, but I think I'm fine with that.

davep: Makes sense to me, I only made that change as it was something you explicitly asked for. I'll nuke that then.

davep: Done.

src/textual/widget.py
```diff
@@ -803,16 +803,12 @@ def watch_hover_style(
             self.highlight_link_id = hover_style.link_id
 
     def watch_scroll_x(self, old_value: float, new_value: float) -> None:
-        if self.show_horizontal_scrollbar:
-            self.horizontal_scrollbar.position = round(new_value)
```
willmcgugan, line 807: This like is required to change the scrollbar.

davep: I'm reading *"like"* as *"line"* there, and assuming it's in reference to updating the scrollbar position. If so, good catch. Restored.

PR #1551, opened 2023-01-12
Raise clearer exception when `none` is in a space-separated list of text styles.

### LINKED ISSUE #1420: Improve error message when value `none` is in a list of text styles
When you provide an illegal value for the CSS rule `link-style`, you get a helpful message displaying all legal values, saying that `text-style` expects a space-separated list of those.

If you set `text-style: bold italic reverse` it works fine.

If you set `text-style: none bold italic reverse` you are being silly, but there is no real reason for this to fail (I think). Yet, the error we get is 

```py
StyleSyntaxError: unable to parse 'none' as color; 'none' is not a valid color
```

To repro:

```py
from textual.app import App

class MyApp(App):
    DEFAULT_CSS = "* { text-style: bold italic none }"
```

rodrigogiraoserrao:

- ~[ ] Updated documentation~

**Note**: This PR should only be merged if we agree that the behaviour of [`rich.Style.parse`](https://github.com/Textualize/rich/blob/0c1aedac9c004d45d111c5258fe0f36ccbd88c91/rich/style.py#L498) is correct in that it fails when trying to parse a combination of styles with `"none"` in there.

So, we short-circuit the parsing when `"none"` shows up mixed with other styles.

If rich is "wrong" and it should treat a sequence of styles that contain `"none"` as the style `"none"`, then I'll PR a fix on rich instead.

This PR will close #1420


willmcgugan:
Can you change the title so it says what it is fixing?

rodrigogiraoserrao:
> Can you change the title so it says what it is fixing?

Done.

src/textual/css/_style_properties.py
```diff
@@ -891,6 +891,7 @@ def __set__(self, obj: StylesBase, style_flags: Style | str | None):
         Raises:
             StyleValueError: If the value is an invalid style flag
         """
+        print(repr(style_flags))
```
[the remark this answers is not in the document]
rodrigogiraoserrao: Yes 🙃

src/textual/css/_style_properties.py
```diff
@@ -909,7 +910,8 @@ def __set__(self, obj: StylesBase, style_flags: Style | str | None):
                             self.name, word, context="inline"
                         ),
                     )
-            style = Style.parse(style_flags)
+            # rich doesn't like "none" mixed with other styles, so short-circuit here.
+            style = Style() if "none" in words else Style.parse(style_flags)
```
willmcgugan, line 913: If I enter `none red` won't that result in an empty Style? Indeed `none` plus any combination would result in an empty style.

I think if `none` is used, then its only valid by itself. It should be an error if there are any other words.

willmcgugan (changes requested):

src/textual/css/_style_properties.py
rodrigogiraoserrao: `none red` raises an error in rich because the `rich.style.Style.parse` method starts by checking if `none` is alone. Because it isn't, parsing will fail later on because `none` isn't a recognised color.

So, we are mimicking the rich behavior and saying that `none` can only show up by itself?

willmcgugan: Yes. But with a better error.

A single word in Rich may be a theme color, which is why that happens.

Textual doesn't support Rich's themes, so the styles are a subset of what Rich does. `none` by itself has a special meaning in Textual.

rodrigogiraoserrao: Addressed 👍

src/textual/css/_style_properties.py
```diff
@@ -909,7 +910,17 @@ def __set__(self, obj: StylesBase, style_flags: Style | str | None):
                             self.name, word, context="inline"
                         ),
                     )
-            style = Style.parse(style_flags)
+            try:
+                style = Style.parse(style_flags)
+            except rich.errors.StyleSyntaxError as exc:
```
willmcgugan, line 915: No abbreviations please. Suggest `error` rather than `exc`

src/textual/css/_style_properties.py
```diff
@@ -909,7 +910,17 @@ def __set__(self, obj: StylesBase, style_flags: Style | str | None):
                             self.name, word, context="inline"
                         ),
                     )
-            style = Style.parse(style_flags)
+            try:
+                style = Style.parse(style_flags)
```
willmcgugan, line 914: I'm confused about what happens if set the style to `none`.

ie. `my_widget.styles.text_style == "none"`

AFAICT this will throw an exception. It *should* set the style to `Style.NULL`

rodrigogiraoserrao: If you set the style to `none`, it sets the style to `none` 🤷  Don't take my word for it, I just added a new test that ensures this is the case.

`Style.parse` can handle the string `"none"` (that's the first thing that `Style.parse` does – it checks if the argument is the string `"none"`.

What fails is if `style_flags` is a space-separated list of styles, one of which is `none`. When that's the case, `Style.parse` doesn't realise we are setting a null style so it proceeds to parsing the full space-separated list and then fails to acknowledge the `none` that is present.

willmcgugan: Ah. I had forgotten that `Styles.parse` has an explicit for the word "none"

willmcgugan (approved):

PR #1548, opened 2023-01-11
Improve `Input` view position calculation

rodrigogiraoserrao:
**Please review the following checklist.**

- [x] Docstrings on all new or modified functions / classes 
- ~[ ] Updated documentation~
- [x] Updated CHANGELOG.md (where appropriate)

PR #1555, opened 2023-01-12
Fix tests: await all mounts before checking rendered values

rodrigogiraoserrao:
Follow-up to #1548


rodrigogiraoserrao:
These tests have race conditions waiting to fail randomly in CI (as it _did_ happen for me in some other PR) so the safe route is to await all mounts of the input before checking that the values rendered are the correct ones.

willmcgugan (approved):
Good catch!

PR #1505, opened 2023-01-06
Ensure pretty traceback for error in Widget compose method

darrenburns:
It now fails immediately and displays a Rich traceback.
Previously it still rendered (although incorrectly) and then displayed a plain traceback on app exit.

**Please review the following checklist.**

- [x] Docstrings on all new or modified functions / classes 
- [x] Updated documentation
- [x] Updated CHANGELOG.md (where appropriate)

PR #1558, opened 2023-01-13
Reactive refactor and fixes

willmcgugan:
This PR sponsored by "How did this ever work?"

- Fixes a numbers of reactive interactions.
- Fixes ugly errors from compose and mount events.
- Adds a `compute` parameter to reactives, with a default of `True`. Set this to `False` to disable running (potentially expensive) computes.
- Implements watch methods with no args.
- Fixed issue where scrollbars widgets are needlessly created.

### DIFF
--- a/src/textual/message_pump.py
+++ b/src/textual/message_pump.py
@@ -108,6 +108,7 @@ class MessagePump(metaclass=MessagePumpMeta):
 
     @property
     def is_running(self) -> bool:
+        """bool: Check if the message pump is running (potentially processing messages)"""
         return self._running
 
     @property
@@ -326,6 +327,8 @@ class MessagePump(metaclass=MessagePumpMeta):
         try:
             await self._dispatch_message(events.Compose(sender=self))
             await self._dispatch_message(events.Mount(sender=self))
+        except Exception as error:
+            self.app._handle_exception(error)
         finally:
             # This is critical, mount may be waiting
             self._mounted_event.set()
--- a/src/textual/message_pump.py
+++ b/src/textual/message_pump.py
@@ -108,7 +108,7 @@ class MessagePump(metaclass=MessagePumpMeta):
 
     @property
     def is_running(self) -> bool:
-        """bool: Check if the message pump is running (potentially processing messages)"""
+        """bool: Is the message pump running (potentially processing messages)."""
         return self._running
 
     @property
--- a/src/textual/reactive.py
+++ b/src/textual/reactive.py
@@ -44,6 +44,7 @@ class Reactive(Generic[ReactiveType]):
         repaint (bool, optional): Perform a repaint on change. Defaults to True.
         init (bool, optional): Call watchers on initialize (post mount). Defaults to False.
         always_update (bool, optional): Call watchers even when the new value equals the old value. Defaults to False.
+        no_compute (bool, optional): Don't run compute methods when attribute is changed. Defaults to False.
     """
 
     def __init__(
@@ -54,12 +55,15 @@ class Reactive(Generic[ReactiveType]):
         repaint: bool = True,
         init: bool = False,
         always_update: bool = False,
+        no_compute: bool = False,
     ) -> None:
         self._default = default
         self._layout = layout
         self._repaint = repaint
         self._init = init
         self._always_update = always_update
+        self._no_compute = no_compute
+        self._is_compute = False
 
     @classmethod
     def init(
@@ -69,6 +73,7 @@ class Reactive(Generic[ReactiveType]):
         layout: bool = False,
         repaint: bool = True,
         always_update: bool = False,
+        no_compute: bool = False,
     ) -> Reactive:
         """A reactive variable that calls watchers and compute on initialize (post mount).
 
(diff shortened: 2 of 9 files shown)


src/textual/reactive.py
```diff
@@ -44,6 +44,7 @@ class Reactive(Generic[ReactiveType]):
         repaint (bool, optional): Perform a repaint on change. Defaults to True.
         init (bool, optional): Call watchers on initialize (post mount). Defaults to False.
         always_update (bool, optional): Call watchers even when the new value equals the old value. Defaults to False.
+        no_compute (bool, optional): Don't run compute methods when attribute is changed. Defaults to False.
```
rodrigogiraoserrao, line 47: Might just be me, but the double negative always takes some to process. Calling it `compute` and defaulting to `True` would make it easier to parse, no?

willmcgugan: Agree. Double negative was awkward.

src/textual/message_pump.py
```diff
@@ -108,6 +108,7 @@ def is_parent_active(self) -> bool:
 
     @property
     def is_running(self) -> bool:
+        """bool: Check if the message pump is running (potentially processing messages)"""
```
davep, line 111: As a property, possibly better documented as just "Is the..." - from the caller's perspective it's a value, not a call that does things.

davep (approved):

src/textual/message_pump.py
```diff
@@ -326,6 +327,8 @@ async def _pre_process(self) -> None:
         try:
             await self._dispatch_message(events.Compose(sender=self))
             await self._dispatch_message(events.Mount(sender=self))
+        except Exception as error:
```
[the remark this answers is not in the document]
darrenburns: Surprising - I opened a PR just this week to do ensure that they were immediate and Rich style tracebacks (and I did test it 😄).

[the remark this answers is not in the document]
willmcgugan: This? https://github.com/Textualize/textual/pull/1505/files

I think that handles compose. Perhaps it was a mount exception I was seeing. Forget exactly what prompted it.

src/textual/reactive.py
```diff
@@ -102,28 +121,43 @@ def var(
...
+            obj (Reactable): An object with reactive attributes.
+            name (str): name of attribute.
+        """
+        internal_name = f"_reactive_{name}"
+        if hasattr(obj, internal_name):
+            # Attribute already has a value
+            return
+        compute_method = getattr(obj, f"compute_{name}", None)
+        if compute_method is not None and self._init:
+            default = getattr(obj, f"compute_{name}")()
```
nitzan-shaked, line 139: isn't this basically `default = compute_method.__call__()` ?

nitzan-shaked: or simply `default = compute_method()` ?

src/textual/reactive.py
```diff
@@ -156,21 +190,24 @@ def __set_name__(self, owner: Type[MessageTarget], name: str) -> None:
...
+        else:
+            value = getattr(obj, self.internal_name)
         return value
 
     def __set__(self, obj: Reactable, value: ReactiveType) -> None:
         _rich_traceback_omit = True
+
+        self._initialize_reactive(obj, self.name)
         name = self.name
         current_value = getattr(obj, name)
```
nitzan-shaked, line 212: I could be wrong, and this is confusing, but:

1. Shouldn't this be `self.internal_name` rather than `name`?
2. Specifically, using `name` -- doesn't that invoke the compute method again, because you're accessing through the `__get__` in the descriptor? Isn't that against the contract? (compute is invoked only when value changes)

src/textual/reactive.py
```diff
@@ -240,43 +285,32 @@ def invoke_watcher(
...
                 compute_method = getattr(obj, f"compute_{compute}")
             except AttributeError:
                 continue
-
-            value = await invoke(compute_method)
-            setattr(obj, compute, value)
+            current_value = getattr(obj, f"_reactive_{compute}")
+            value = compute_method()
+            setattr(obj, f"_reactive_{compute}", value)
+            if value != current_value:
```
nitzan-shaked, line 312: does this need some `or always_update` or something? I'm really not sure here.

src/textual/reactive.py
```diff
@@ -240,43 +285,32 @@ def invoke_watcher(
             else:
```
nitzan-shaked, line 285: it seems like the return value isn't used at, should probably remove it.

src/textual/reactive.py
```diff
@@ -156,21 +190,24 @@ def __set_name__(self, owner: Type[MessageTarget], name: str) -> None:
...
-            # Set and return the value
-            setattr(obj, self.internal_name, default_value)
-            if self._init:
-                self._check_watchers(obj, self.name, default_value)
-            return default_value
+
+        self._initialize_reactive(obj, self.name)
+
+        value: ReactiveType
+        compute_method = getattr(self, f"compute_{self.name}", None)
```
nitzan-shaked, line 197: shouldn't this be "obj" rather than "self"?

src/textual/reactive.py
```diff
@@ -156,21 +190,24 @@ def __set_name__(self, owner: Type[MessageTarget], name: str) -> None:
```
nitzan-shaked, line 190: the setattr one line above -- seems like it's not used at all?

Issue #1573, opened 2023-01-16
Meta, Docs: Most `Binding` attribute docstrings aren't showing in the docs


davep:
Each of [the attributes of the `Binding` dataclass has a docstring](https://github.com/Textualize/textual/blob/45abae58885ff65a63fdc2c852386e55479fbe83/src/textual/binding.py#L26-L41). And yet the API documentation looks like this:

![Screenshot 2023-01-16 at 09 06 47](https://user-images.githubusercontent.com/28237/212639587-f4b2ac99-6e60-4d8d-978f-122c73528bc2.png)

It seems unlikely that this affects just `Binding`, but that's the first place I've noticed it. This issue is a reminder to go check out more of the API and see if some docstrings aren't making it through.

davep:
It's worth noting that `key_display` has a `|` in the type information, whereas all the other properties have are just a single type. If I just set that to `str` (as a test) it loses its docstring from the docs too. This looks to be an issue with the recognition of a docstring -- possibly by `mkdocstrings`?

davep:
Tidied up the worst case of this; but I think some extra investigation and documentation of the issue, and perhaps the raising of an issue with the relevant tool, is in order.

PR #1578, opened 2023-01-16
Add some extra docstrings to widgets

davep:
More of a driveby change than a coordinated attempt to hit everything. Inspired by #1573. I'll do more of these as an occasional improvement as I run into things that are missing.

### DIFF
--- a/src/textual/widgets/_input.py
+++ b/src/textual/widgets/_input.py
@@ -138,7 +138,7 @@ class Input(Widget, can_focus=True):
 
     @property
     def _cursor_offset(self) -> int:
-        """Get the cell offset of the cursor."""
+        """The cell offset of the cursor."""
         offset = self._position_to_cell(self.cursor_position)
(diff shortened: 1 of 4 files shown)


src/textual/widgets/_input.py
```diff
@@ -113,6 +113,17 @@ def __init__(
         id: str | None = None,
         classes: str | None = None,
     ) -> None:
+        """Initialise the `Input` widget.
```
willmcgugan, line 116: Does this not need to go in the class docstrings to be picked up?

davep: I don't believe so; although it is an alternative approach that's permitted by the Google standard IIRC.

Screenshot 2023-01-16 at 16 28 01

willmcgugan: Great. I think the Rich docs require the params on the class. But this is better.

willmcgugan (approved):

PR #1595, opened 2023-01-17
Fix 1196 add focus selector

rodrigogiraoserrao:

- [x] Updated documentation

The changes made in this PR embody some design decisions that were implicit in the original issue:

 - When the focused widget does not match the selector, the meaning of "previous" and "next" is "the first widget that _does_ match in the correct direction, starting from the currently focused widget".
 - If the focused widget is not `None`, it is guaranteed that it will match the selector in the argument:
   - if the direction is `0` and the selector doesn't match the currently focused widget, this will push focus to the next widget that matches.
   - if the selector doesn't match a single focusable widget, focus is set to `None`.
 - If the currently focused widget is the only widget that matches the given selector, focus will be explicitly set to that same widget.


willmcgugan (approved):
LGTM. Nice work.

PR #1599, opened 2023-01-17
Remove explicit type hints from docstrings.

rodrigogiraoserrao:
No point in having explicit type hints in docstrings because our tooling extracts them automatically. This applies the DRY principle and reduced the probability of us having wrong docstrings.

Regular expressions were used to find & replace docstrings with explicit type hints but every change was reviewed by hand.


src/textual/widgets/_list_view.py
```diff
@@ -146,9 +146,11 @@ class Highlighted(Message, bubble=True):
         """Emitted when the highlighted item changes. Highlighted item is controlled using up/down keys.
 
         Attributes:
-            item (ListItem | None): The highlighted item, if there is one highlighted.
+            item: The highlighted item, if there is one highlighted.
         """
 
+        item: ListItem | None
+
```
davep, line 153: Does doing this hoist the type up in the the attributes in the class docstring? Testing stand-alone here I'm finding that this:

```python
class MyClass:
    """Some class.

    Attributes:
       some_attr: This is some instance attribute.
    """

    def __init__( self ):
        self.some_attr: float = 0.0
```

does the trick but this:

```python
class MyClass:
    """Some class.

    Attributes:
       some_attr: This is some instance attribute.
    """

    some_attr: float

    def __init__( self ):
        self.some_attr = 0.0
```

doesn't.

davep: Also, for what it's worth, this works too, but highlights the individual instance attributes rather than makes a table:

```python
class MyClass:
    """Some class."""

    some_attr: float

    def __init__( self ):
        self.some_attr = 0.0
        """This is some instance attribute."""
```

willmcgugan:
@rodrigogiraoserrao Created some merge conflicts for you. Guess you might have to run your regexes again.

src/textual/widgets/_list_view.py
rodrigogiraoserrao: You are right. The second option isn't working... Which is weird, because I tested it before going for that type of change. I must've done something wrong when testing.

In that case, I'd say it's open for debate what we go with:

 - Option A

```py
class MyClass:
    """This is my docstring:

    Attributes:
        my_attr (float): This is what my_attr does.
    """

        self.my_attr = ...
```

 - Option B

```py
class MyClass:
    """This is my docstring:

    Attributes:
        my_attr: This is what my_attr does.
    """

        self.my_attr: float = ...
```

I recommend option B because type checkers will detect if that becomes wrong, whereas option A can become a lie without us ever noticing.
@willmcgugan what have you say?

willmcgugan: I haven't been following. Suffice to say it is essential not to break type checkers. Over that we should do whatever works best in out docs.

If you and @davep agree a way forward I'll +1 it.

davep: I'd 100% go for option B -- it's a shame it doesn't infer the type from the value assigned in the same way that a type checker would; but explicit typing is always better than no typing. Count me in for this.

src/textual/widgets/_tree.py
```diff
@@ -334,43 +334,67 @@ class NodeMessage(Generic[EventTreeDataType], Message, bubble=True):
         """Base class for events sent when something happens with a node.
 
         Attributes:
-            TreeNode[EventTreeDataType]: The node involved in the event.
+            node: The node involved in the event.
```
willmcgugan, line 337: I think we can discard this message altogether. We don't use it as a base class now, and we never explicitly send it.

willmcgugan (changes requested):

willmcgugan:
@rodrigogiraoserrao LGTM. I'm assuming you've checked that the types are shown in the docs. Just one request to remove the base class for those tree events.

src/textual/widgets/_tree.py
rodrigogiraoserrao: I don't know why I kept it 🤷 
Removed o7

willmcgugan (approved):

Issue #1588, opened 2023-01-17
Populating a `ListView` via constructor vs via inherit and `compose`


davep:
There seems to be a subtle difference in the working of a `ListView` if you create one by passing the `ListItem`s to it, vs if you create a custom `ListView` by inheriting from it and using `compose` to populate it. Take the following example code, which places both approaches side-by-side:

```python
from textual.app        import App, ComposeResult
from textual.containers import Horizontal
from textual.widgets    import Header, Footer, ListView, ListItem, Label

class CustomListView( ListView ):

    def __init__( self, choices: list[ str ] ) -> None:
        super().__init__()
        self._choices = choices

    def compose( self ) -> ComposeResult:
        """Compose the child widgets."""
        for choice in self._choices:
            yield ListItem( Label( choice ) )

class ListViewMakerApp( App[ None ] ):

    CSS = """
    ListView {
        width: 1fr;
        height: 1fr;
        border: round red;
    }
    """

    OPTIONS = [ f"This is the nifty test option {n}" for n in range( 20 ) ]

    def compose( self ) -> ComposeResult:
        yield Header()
        yield Horizontal(
            ListView(
                *[ ListItem( Label( option ) ) for option in self.OPTIONS ]
            ),
            CustomListView( self.OPTIONS )
        )
        yield Footer()

if __name__ == "__main__":
    ListViewMakerApp().run()
```

I feel the `CustomListView` would not be an unusual approach for people to take; perhaps wanting to make their own commonly-used selection list of simple values (or even very complex child values that they want easily handled). Side-by-side the `ListView`s look the same to start with:

<img width="922" alt="Screenshot 2023-01-17 at 11 10 47" src="https://user-images.githubusercontent.com/28237/212884262-8d5252a8-3318-4f72-a3de-ae5f73b5a41c.png">

Note that the non-inherited `ListView` is to the left and that, even though it doesn't have focus, the first item is highlighted. If you <kbd>tab</kbd> into it you can cursor around, etc, just fine. On the other hand notice that the right `ListView` (which is a `CustomListView`) has no obvious highlighted item and if you tab into it nothing gets highlighted.

Further to this, if you (via keyboard) focus the inherited (right hand) `ListView` and then attempt to cursor down, the following exception will be raised:

```
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ /Users/davep/develop/python/textual-sandbox/.venv/lib/python3.10/site-packages/textual/widgets/_ │
[...]

PR #1600, opened 2023-01-18
Allow populating a child class of `ListView` purely from its `compose`

davep:
See #1588. Simply put, this change will remove the problems encountered (no default highlighted item, no item gets highlighted on focus, exception thrown when attempting to cursor through the items) when inherited from a `ListView` and then populating its items purely from `compose`.

### DIFF
--- a/src/textual/widgets/_list_view.py
+++ b/src/textual/widgets/_list_view.py
@@ -50,16 +50,11 @@ class ListView(Vertical, can_focus=True, can_focus_children=False):
             classes: The CSS classes of the widget.
         """
         super().__init__(*children, name=name, id=id, classes=classes)
-        self.index = initial_index
+        self._index = initial_index
 
     def on_mount(self) -> None:
         """Ensure the ListView is fully-settled after mounting."""
-        # If someone inherits from ListView, they might be populating it
-        # from Widget.compose rather than from passing the ListItems to the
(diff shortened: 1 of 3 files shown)


rodrigogiraoserrao (approved):
+1 for “let's just tickle it again”

src/textual/widgets/_list_view.py
```diff
@@ -52,6 +52,15 @@ def __init__(
...
         self.index = initial_index
 
+    def on_mount(self) -> None:
+        """Ensure the ListView is fully-settled after mounting."""
+        # If someone inherits from ListView, they might be populating it
+        # from Widget.compose rather than from passing the ListItems to the
+        # initialisation call. Given that watch_index does all the work,
+        # let's just tickle it again after the DOM has been spun up and all
+        # children of this widget will have been mounted.
+        self.index = self.index or 0
```
darrenburns, line 62: Small thing, but the validator would run twice I think.

I think we want to do something like `self._index = initial_index` inside the init, then in `on_mount` do `self.index = self._index`?

davep: I'll give that a shot.

davep: Done: 7544448002a6c382c5cd136933bbc7cf7465330d

darrenburns (approved):

PR #1601, opened 2023-01-18
Resetting cursor/hover on clear of DataTable

darrenburns:
## Changes

- `[Row|Cell|Column]Highlighted` event will now be emitted when the table gets its first row, assuming the cursor becomes visible (i.e. not hidden and a visible cursor_type).
- Adds considerable testing around emission of events from the DataTable.
- Adds testing around cursor position handling and clearing.
- More defensive handling of hover/cursor coordinates to account for table being cleared (I think there was a potential crash with cell cursor on clear).
- Reset the cursor positions when the table is cleared.
- Fixes edge case - if you change the cursor_type from "none" to a visible type, but `show_cursor` is False, then a highlighted event is no longer raised since the cursor still isn't visible.

- [x] Updated documentation

### DIFF
--- a/src/textual/widgets/_data_table.py
+++ b/src/textual/widgets/_data_table.py
@@ -336,18 +336,12 @@ class DataTable(ScrollView, Generic[CellType], can_focus=True):
 
     def watch_cursor_type(self, old: str, new: str) -> None:
         self._set_hover_cursor(False)
-        row_index, column_index = self.cursor_cell
-
-        # Apply the highlighting to the newly relevant cells
-        if new == "cell":
-            self._highlight_cell(self.cursor_cell)
-        elif new == "row":
-            self._highlight_row(row_index)
-        elif new == "column":
-            self._highlight_column(column_index)
+        if self.show_cursor:
+            self._highlight_cursor()
 
         # Refresh cells that were previously impacted by the cursor
         # but may no longer be.
+        row_index, column_index = self.cursor_cell
         if old == "cell":
             self.refresh_cell(row_index, column_index)
         elif old == "row":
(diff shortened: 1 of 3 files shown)


src/textual/widgets/_data_table.py
```diff
@@ -14,23 +14,27 @@
...
 from ..scroll_view import ScrollView
 from ..strip import Strip
-from .._typing import Literal
 
 CursorType = Literal["cell", "row", "column", "none"]
 CELL: CursorType = "cell"
 CellType = TypeVar("CellType")
 
 
+class CellDoesntExist(Exception):
```
rodrigogiraoserrao, line 34:
```suggestion
class CellDoesNotExist(Exception):
```

For some reason, feels wrong to use a contraction in a class name.

src/textual/widgets/_data_table.py
```diff
@@ -229,9 +233,16 @@ def get_cell_value(self, coordinate: Coordinate) -> CellType:
...
+
+        Raises:
+            CellDoesntExist: If there is no cell with the given coordinate.
         """
         row, column = coordinate
-        return self.data[row][column]
+        try:
+            cell_value = self.data[row][column]
+        except KeyError:
+            raise CellDoesntExist(f"No cell exists at {coordinate!r}")
```
rodrigogiraoserrao, line 244: This feels like the perfect place for you to teach me this: won't this raise the `CellDoesn'tExist` and show a traceback saying something "KeyError was raised and during handling of that exception, `CellDoesn'tExist` was raised"? Don't we need to add a `from None` here?

src/textual/widgets/_data_table.py
```diff
@@ -293,18 +304,26 @@ def watch_cursor_cell(
...
+            return
+        else:
+            self.emit_no_wait(DataTable.CellHighlighted(self, cell_value, coordinate))
 
     def _highlight_row(self, row_index: int) -> None:
         """Apply highlighting to the row at the given index, and emit event."""
         self.refresh_row(row_index)
-        self.emit_no_wait(DataTable.RowHighlighted(self, row_index))
+        if row_index in self.data:
+            self.emit_no_wait(DataTable.RowHighlighted(self, row_index))
```
rodrigogiraoserrao, line 320: `self.refresh_row` will also check that we are trying to refresh a row that exists, so is that why you don't have the call to `refresh_row` inside the if statement?

rodrigogiraoserrao (approved):
Looks good to me, although I did leave a couple of questions!

src/textual/widgets/_data_table.py
darrenburns: Yep, that's correct.

src/textual/widgets/_data_table.py
darrenburns: Yeah, I think this is better. I'll update this.

src/textual/widgets/_data_table.py
davep: I'd second a `from None` here.

darrenburns: Added `from None`

willmcgugan (approved):

Issue #1287, opened 2022-11-25
Textual has no version dunder


jacobtomlinson:
I'm in the habit of looking for a [`__version__` dunder](https://peps.python.org/pep-0008/#module-level-dunder-names) when trying to figure out what versions of packages I have installed. I noticed today that textual doesn't have one.

```python
>>> import textual
>>> textual.__version__
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: module 'textual' has no attribute '__version__'
```

It would be nice if textual followed this convention.

davep:
While I say this as a long-time "fan" of supplying and using `__version__` (and related metadata dunders), it's my understanding that the preferred method these days (and one I started to lean into with a previous system I built) is going via [`importlib.metadata`](https://docs.python.org/3/library/importlib.metadata.html#overview):

```python
>>> from importlib.metadata import version
>>> version( "textual" )
'0.5.0'
```

oleksis:
Use textual `--version` : `ModuleNotFoundError: No module named 'click'`

```bash
➜ python -V
Python 3.10.2

➜ python -m pip install --upgrade pip

➜ pip install textual

➜ textual --version
  File "C:\temp\venv\Scripts\textual.exe\__main__.py", line 4, in <module>
  File "C:\temp\venv\lib\site-packages\textual\cli\cli.py", line 4, in <module>
    import click
ModuleNotFoundError: No module named 'click'

➜ pip install click

➜ textual --version
textual, version 0.5.0

```

[Tree widget import error](https://github.com/Textualize/textual/issues/1300#issuecomment-1329661167)

davep:
@oleksis As mentioned elsewhere, that's related to #983

jacobtomlinson:
Yeah I ran into #983 too but I'm not sure it is related to my issue here.

I totally agree that `from importlib.metadata import version; version( "textual" )` is current best practice. But `__version__` is still a pretty common expectation for Python libraries and it would take minimal effort to support (just use `importlib.version` to populate it).

willmcgugan:
The problem was that there was no tool to do this automatically. So pretty much any tool that uses `poetry` won't have a `__version__`.

IIRC I tried  `importlib.version` in `__init__.py` with Rich, but backed out because it was very slow (it enumerates all the packages)!

I believe Sebastian Ramirez was working on a Poetry plugin to do this automatically...

jacobtomlinson:
[...]

PR #1605, opened 2023-01-18
Added module level getattr to get version

willmcgugan:
Implements a `__version__`, so you can do the following:

```
>>> import textual
>>> textual.__version__
'0.10.0'
```

Fixes https://github.com/Textualize/textual/issues/1287

### DIFF
--- a/src/textual/__init__.py
+++ b/src/textual/__init__.py
@@ -1,20 +1,31 @@
 from __future__ import annotations
 
 import inspect
-from typing import Callable
 
 import rich.repr
 from rich.console import RenderableType
-
-__all__ = ["log", "panic"]
-
 
 from ._context import active_app
 from ._log import LogGroup, LogVerbosity
 from ._typing import TypeAlias
 
+__all__ = ["log", "panic"]
+
 
 LogCallable: TypeAlias = "Callable"
+
+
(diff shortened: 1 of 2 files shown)


rodrigogiraoserrao:
> Implements a `__version__`, so you can do the following:
> 
> ```
> >>> import textual
> >>> textual.__version__
> '3.10.0'
> ```
> 
> Fixes #1287

Is this the actual output of the example? How is textual in major version 3?

willmcgugan:
No, that was typed by hand.

darrenburns:
Looks correct, although I'd probably be more in favour of the backport. The pkgutils stuff is *really* slow, though. In Ward, during a single test run, something like 30% of the time was spent inside `get_distribution`. Whether it's going to be even remotely noticeable to a human, IDK. I guess not many people will be getting the version anyway.

willmcgugan:
Think I want to go with the non-backport approach for now. 3.7 is so close to EOL. If anyone complains I'm willing to reconsider!

darrenburns:
@willmcgugan @davep Just realised we already have the `importlib-metadata` dependency. It's used in the devtools CLI to get the version already, so we should probably move it into the "dev" extras or get rid of it and update the CLI to use this approach.

PR #1610, opened 2023-01-18
Fix #1607 for generic/integer properties.

### LINKED ISSUE #1607: Programmatic styles assignment doesn't refresh children widgets
Investigation of #1481 surfaced a more fundamental issue, in that programmatically changing some styles of a widget `W` doesn't refresh the children of `W` and hence doesn't update the app to its correct state.

---
**Edit**:

The styles _found_ to be suffering from this are (not necessarily a comprehensive list):

 - `grid_size_rows`
 - `grid_size_columns`
 - `grid_gutter_vertical`
 - `grid_gutter_horizontal`
 - `grid_rows`
 - `grid_columns`
 - `overflow_x`
 - `overflow_y`
 - `scrollbar_gutter`
 - `align`
 - `align_horizontal`
 - `align_vertical`

---

Consider the app below that contains a 2 x 2 grid of boxes. Clicking the app should change it so that the grid now has 10 rows (instead of 2) but nothing changes. You have to force the layout update by resizing the terminal, for example.

Notice that the method `on_click` has many different lines commented. Each of those represents an attribute that can be programmatically changed without the layout of the app updating immediately and automatically.

```py
from textual.app import App
from textual.containers import Grid


class Box(Static):
    DEFAULT_CSS = """Box { border: round green; width: 100%;  height: 100%; }"""


class MyApp(App[None]):
    DEFAULT_CSS = """
    Grid {
        grid-size: 2 2;
    }
    """

    def compose(self):
        yield Grid(
            Box(),
            Box(),
            Box(),
            Box(),
        )

    def on_click(self):
        self.query_one(Grid).styles.grid_size_rows = 10
        # self.query_one(Grid).styles.grid_size_columns = 10
        # self.query_one(Grid).styles.grid_gutter_vertical = 10
        # self.query_one(Grid).styles.grid_gutter_horizontal = 10
        # self.query_one(Grid).styles.grid_rows = "1fr 10fr"
        # self.query_one(Grid).styles.grid_columns = "1fr 10fr"


if __name__ == "__main__":
    MyApp().run()
```

rodrigogiraoserrao:

- [ ] Updated documentation

See the thorough discussion in #1607.

### DIFF
--- a/src/textual/css/_style_properties.py
+++ b/src/textual/css/_style_properties.py
@@ -60,9 +60,20 @@ PropertySetType = TypeVar("PropertySetType")
 
 
 class GenericProperty(Generic[PropertyGetType, PropertySetType]):
-    def __init__(self, default: PropertyGetType, layout: bool = False) -> None:
+    """Generic property to be inherited by other style properties.
+
+    Args:
+        default: The default value (or a factory thereof) of the property.
+        layout: Whether to refresh the node layout on value change.
+        children: Whether to refresh the node children on value change.
+    """
+
+    def __init__(
+        self, default: PropertyGetType, layout: bool = False, children: bool = False
+    ) -> None:
         self.default = default
         self.layout = layout
+        self.children = children
 
     def validate_value(self, value: object) -> PropertyGetType:
         """Validate the setter value.
@@ -88,11 +99,11 @@ class GenericProperty(Generic[PropertyGetType, PropertySetType]):
         _rich_traceback_omit = True
         if value is None:
             obj.clear_rule(self.name)
-            obj.refresh(layout=self.layout)
+            obj.refresh(layout=self.layout, children=self.children)
             return
         new_value = self.validate_value(value)
         if obj.set_rule(self.name, new_value):
-            obj.refresh(layout=self.layout)
+            obj.refresh(layout=self.layout, children=self.children)
 
 
 class IntegerProperty(GenericProperty[int, int]):
--- a/src/textual/css/styles.py
+++ b/src/textual/css/styles.py
@@ -280,10 +280,10 @@ class StylesBase(ABC):
     grid_rows = ScalarListProperty()
     grid_columns = ScalarListProperty()
 
-    grid_size_columns = IntegerProperty(default=1, layout=True)
-    grid_size_rows = IntegerProperty(default=0, layout=True)
-    grid_gutter_horizontal = IntegerProperty(default=0, layout=True)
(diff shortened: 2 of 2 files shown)


rodrigogiraoserrao:
The strategy employed here to fix these issues doesn't seem to cut it for `overflow_x`/`overflow_y`.
More investigation is needed.

willmcgugan:
@rodrigogiraoserrao What's the status of this one? WIP?

rodrigogiraoserrao:
> @rodrigogiraoserrao What's the status of this one? WIP?

It's been ready for review for ~2 weeks. Doesn't the PR show that I've requested reviews?

willmcgugan:
Looked like you were still working on it. Will review after lunch.

src/textual/css/_style_properties.py
```diff
@@ -60,9 +60,20 @@
...
+    """Descriptor that abstracts away common machinery for other style descriptors.
+
+    Args:
+        default: The default value (or a factory thereof) of the property.
+        layout: Whether to refresh the node layout on value change.
+        children: Whether to refresh the node children on value change.
+    """
+
+    def __init__(
+        self, default: PropertyGetType, layout: bool = False, children: bool = False
```
willmcgugan, line 72: I think maybe the name could use an update. I know its inherited from the refresh method, but in this context its not clear what it does. How about `refresh_children`?

src/textual/css/_style_properties.py
```diff
@@ -202,8 +213,16 @@ def __set__(
...
+    """Descriptor for lists of scalars.
+
+    Args:
+        children: Whether to refresh the node children on value change.
+        percent_unit: The dimension to which percentage scalars will be relative to.
+    """
+
+    def __init__(self, percent_unit: Unit, children: bool = False) -> None:
         self.percent_unit = percent_unit
+        self.children = children
```
willmcgugan, line 225: Ditto re `refresh_children`

willmcgugan (approved):
Looks good! Just one comment re naming, which you may take or leave.

src/textual/css/_style_properties.py
rodrigogiraoserrao: I actually like your point. Should I also change `layout` to `refresh_layout`?

willmcgugan: `layout` feels fine. It's an established term. It also works as a verb. It was just `children` felt too generic in context.

Issue #1608, opened 2023-01-18
During tests there may be deadlocks


ofek:
What Operating System are you running on? Windows

In the following snippet if you uncomment the yielding of the Header the test will block:

```python
from textual.app import App
from textual.screen import Screen
from textual.widgets import Footer, Header, Input


class MyScreen(Screen):
    def compose(self):
        # yield Header()
        yield Input()
        yield Footer()


class MyApp(App):
    def on_mount(self):
        self.install_screen(MyScreen(), 'myscreen')
        self.push_screen('myscreen')


async def test():
    app = MyApp()
    async with app.run_test():
        raise Exception('never raised')

        assert True
```

ofek:
Another example:

```python
from textual.app import App
from textual.containers import Container
from textual.screen import Screen
from textual.widget import Widget
from textual.widgets import Footer


class MyWidget(Widget):
    def compose(self):
        yield Footer()


class MyScreen(Screen):
    def compose(self):
        yield Container(MyWidget())


class MyApp(App):
    def on_mount(self):
        self.install_screen(MyScreen(), 'myscreen')
        self.push_screen('myscreen')


async def test():
    app = MyApp()
    async with app.run_test():
        raise Exception('never raised')

        assert True
```

davep:
Installed as a unit test in Textual, the most minimal version of this I can get to hang (although I can <kbd>Ctrl</kbd>+<kbd>C</kbd> it) is this:

```python
from textual.app import App
from textual.screen import Screen
from textual.widget import Widget


class MyWidget(Widget):

    def compose(self):
        yield Widget()


class MyScreen(Screen):

    def compose(self):
        yield MyWidget()


class MyApp(App):

    def on_mount(self):
        self.install_screen(MyScreen(), 'myscreen')
        self.push_screen('myscreen')


async def test_this():
    app = MyApp()
    async with app.run_test():
        raise Exception('never raised')
        assert True
```

If I make `MyApp.on_mount` `async` and `await` the installation and the push of the screen:

```python
from textual.app import App
from textual.screen import Screen
from textual.widget import Widget


class MyWidget(Widget):

    def compose(self):
        yield Widget()


class MyScreen(Screen):

    def compose(self):
        yield MyWidget()


class MyApp(App):

    async def on_mount(self):
        await self.install_screen(MyScreen(), 'myscreen')
        await self.push_screen('myscreen')


async def test_this():
    app = MyApp()
    async with app.run_test():
        raise Exception('never raised')
[...]

PR #1615, opened 2023-01-19
Fix for test freeze

willmcgugan:
- Fixes https://github.com/Textualize/textual/issues/1608
- Added names to tasks which is very handy for debugging
- Adds textual._asyncio to provide a compatibility layer (create_task doesn't accept name arg on 3.7)

Essentially the task to mount widgets occurred after the app was asked to shut down.

### DIFF
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -859,7 +859,7 @@ class App(Generic[ReturnType], DOMNode):
             )
 
         # Launch the app in the "background"
-        app_task = asyncio.create_task(run_app(app))
+        app_task = asyncio.create_task(run_app(app), name=f"run_test {app}")
 
         # Wait until the app has performed all startup routines.
         await app_ready_event.wait()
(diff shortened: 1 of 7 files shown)


davep (approved):

src/textual/app.py
```diff
@@ -1668,6 +1671,11 @@ def _register(
                 if widget.children:
                     self._register(widget, *widget.children)
                 apply_stylesheet(widget)
+
+        if not self._running:
```
rodrigogiraoserrao, line 1675: I get that if the app is not running we don't want to return widgets to be awaited... but if I am in the method that registers widgets, isn't this a weird place to check if the app is running, right before returning..?

willmcgugan: It's mainly for the benefit of tests which don't run an app. We still want to check for duplicate IDs and confirm the DOM state has been set up.

PR #1619, opened 2023-01-19
Request that people include `textual diagnose` output in bug reports

davep:
Adds a wee prompt to a bug report issue to ask folk to run a `textual diagnose` and include it in the issue.


davep:
Closing as @willmcgugan had the same idea at the same time. 🤣

Issue #1620, opened 2023-01-19
Add a `text` property to Strip


willmcgugan:
I think the Strip class needs a `text` property which returns the combined text of all the Segments.

This will be useful for testing / debugging.

PR #1621, opened 2023-01-19
Strip text

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/1620

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## Unreleased
+
+### Added
+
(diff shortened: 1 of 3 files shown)


darrenburns (approved):

davep (approved):

ofek:
Nice! Is this how I would do it? I'm getting an empty string:

```python
''.join(segment.text for segment in text_log.render_lines(terminal_region))
```

willmcgugan:
Try 

```python
''.join(line.text for line in text_log.lines)
```

That will give you the content prior to rendering. You want to avoid rendering the lines as the output depends on the dimensions of the terminal.

willmcgugan:
If it is for testing though. You might want to consider snapshot tests. There are examples in our tests. At some point we will make those more easily accessible for other projects. We've found them to be fantastic for catching render related bugs.

ofek:
That works now thank you! I will look into snapshot tests, that sounds very useful to be able to track changes to actual images 😮

Issue #1622, opened 2023-01-20
textual 0.10.0 renders a lonely "p" on the screen before entering the app


kratsg:
`textual diagnose` crashes for me with 

```
$ textual diagnose
Traceback (most recent call last):
  File "/Users/kratsg/.pyenv/versions/itkdb-browser/bin/textual", line 8, in <module>
    sys.exit(run())
  File "/Users/kratsg/.pyenv/versions/itkdb-browser/lib/python3.8/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/Users/kratsg/.pyenv/versions/itkdb-browser/lib/python3.8/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/Users/kratsg/.pyenv/versions/itkdb-browser/lib/python3.8/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/kratsg/.pyenv/versions/itkdb-browser/lib/python3.8/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/kratsg/.pyenv/versions/itkdb-browser/lib/python3.8/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/Users/kratsg/.pyenv/versions/itkdb-browser/lib/python3.8/site-packages/textual/cli/cli.py", line 139, in run_diagnose
    from textual.cli.tools.diagnose import diagnose
  File "/Users/kratsg/.pyenv/versions/itkdb-browser/lib/python3.8/site-packages/textual/cli/tools/diagnose.py", line 12, in <module>
    def _section(title: str, values: dict[str, str]) -> None:
TypeError: 'type' object is not subscriptable
```

I'm uploading a video (set playback speed to 0.5 or 0.75 to definitely see it)

https://user-images.githubusercontent.com/761483/213593211-c1092a83-e2da-4e5a-90b2-d64772bbfce1.mp4

showing what happens when I run `textual colors`. There's a brief moment when there's a lonely `p` on the 2nd line of the terminal screen before going into the app. Any ideas what this is? I thought it was a problem with my app, but it turns out that this doesn't show up on 0.1.17 to 0.2.0 (I suspect it made it's way in after that?).

davep:
[...]

PR #1623, opened 2023-01-20
Import type annotations from the future for `textual diagnose`

davep:
I mean, I guess flat out crashing on older versions of Python is *a* way of reporting the Python version. Kinda?

See the side issue reported in #1622.


willmcgugan (approved):

willmcgugan:
Thanks. I'll do an interim release.

PR #1627, opened 2023-01-20
Changing overflow programmatically updates layout

rodrigogiraoserrao:
This PR proposes a fix for #1616.

*Disclaimer*: The comments in the linked issue explain why I think this is more of a patch rather a proper fix.

A proper fix would be to add a call to `self._refresh_scrollbars` in the beginning of the function I modified, instead of adding the two reassignments to `show_???_scrollbar`.
However, I went with this patch instead of the full function call because some ad-hoc experiments seem to suggest that the function `_get_scrollable_region` is called a lot and adding that extra function call, that most of the time isn't needed, might add too much overhead.

However, I don't know much about performance in Python nor am I familiar with the performance profile of Textual, so this might be a gross misjudgement.

Even so, the two reassignments run as often as the call to `self._refresh_scrollbars` would run, and most of the time they would be useless.
That is why I think an _even better approach_ would be to implement a style property specifically for the two overflow attributes and have it so that when the overflow is _set_ and _changed_, the property notifies its node that it needs to refresh the scrollbars.
Sadly, I can't quite see how this could be done, given that the implementation of `Widget` and the implementation of `Styles` seem to be quite independent (and rightfully so).


src/textual/widget.py
```diff
@@ -908,7 +907,7 @@ def _refresh_scrollbars(self) -> None:
         show_horizontal = self.show_horizontal_scrollbar
         if overflow_x == "hidden":
             show_horizontal = False
-        if overflow_x == "scroll":
+        elif overflow_x == "scroll":
```
willmcgugan, line 910: Good catch.

willmcgugan (changes requested):
I think you've identified what the issue is, but I don't think the proposed fix is in the right place.

The change introduces additional logic and state in to deciding if the scrollbars should be shown. I think we can keep the current logic (which is simpler) by calling `_refresh_scrollbars` when the style is changed, i.e. in the property.

Suggest we refresh the scrollbar state when  `overflow_x` and `overflow_y` properties are modified. This will likely require a custom descriptor based on StringEnumProperty

willmcgugan (approved):

PR #1629, opened 2023-01-20
Skip FAQtory on new issues by Textualize members

### LINKED ISSUE #1617: Meta: Don't run FAQtory if you're a member of Textualize
FAQtory is, without question, awesome. On the other hand there seems little reason to FAQ-check every issue that we (Textualize folk) raise. Given this, look into modifying [the handling of a new issue](https://github.com/Textualize/textual/blob/main/.github/workflows/new_issue.yml) such that it doesn't get run if it's one of us.

I suspect it could be done with a combination of [`github.actor`](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context), `if:`, and [`contains`](https://docs.github.com/en/actions/learn-github-actions/expressions#contains). I've not tried it yet but I'm thinking:

```
if: ${{ !contains( "user1,user2,user3,user3", github.actor ) }}
```

or something like that (I've no real experience with GitHub actions, so that's a half guess from scanning the docs).

rodrigogiraoserrao:
Fix #1617 (excellent idea).

Did a couple of tests in a private repo. This does look like the correct syntax.

### DIFF
--- a/.github/workflows/new_issue.yml
+++ b/.github/workflows/new_issue.yml
@@ -4,6 +4,7 @@ on:
     types: [opened]
 jobs:
   add-comment:
+    if: ${{ !contains( 'willmcgugan,darrenburns,davep,rodrigogiraoserrao', github.actor ) }}
     runs-on: ubuntu-latest
     permissions:
       issues: write


davep (approved):
Thanks for that. Saved me a job. 😁

willmcgugan (approved):

willmcgugan:
Nice, thanks

Issue #1632, opened 2023-01-21
textual push_screen/pop_screen breaks app functionality (for example, with ListView)


kratsg:
code

```python
from textual.app import App, ComposeResult
from textual.containers import Horizontal, Vertical
from textual.screen import Screen
from textual.widgets import Button, Footer, Header, Label, ListItem, ListView, Static

class LoginScreen(Screen):

    def on_button_pressed(self, event: Button.Pressed) -> None:
        """Event handler called when login button is pressed."""
        button_id = event.button.id
        if button_id == "login":
            self.app.login()
            event.stop()

    def compose(self) -> ComposeResult:
        yield Horizontal(
            Vertical(
                Button("Login", id="login", variant="primary")
            )
        )


class MyList(ListView):
    """A widget to display a list of items."""
    def load_items(self) -> None:
        self.clear()
        for item in range(10):
            print(f'{item}')
            self.append(ListItem(Label(f'test {item}')))


class Browser(App):
    """App"""

    BINDINGS = [("q", "exit", "Quit"), ("d", "toggle_dark", "Toggle dark mode")]

    DEFAULT_CSS = """
    Browser > Static {
        content-align: center middle;
        background: crimson;
        border: solid darkred;
        height: 1fr;
    }

    .column {
        width: 1fr;
    }
    """

    def __init__(self) -> None:
        super().__init__()
        self.my_list = MyList(classes="column")

    def action_toggle_dark(self) -> None:
        """An action to toggle dark mode."""
        self.dark = not self.dark

    def action_exit(self) -> None:
        """An action to exit."""
        self.exit()

    def login(self) -> None:
        """Called when the LoginScreen has logged in."""
        self.pop_screen()
        self.my_list.load_items()

    def on_mount(self) -> None:
        """Call after entering application mode."""
        self.push_screen(LoginScreen())

    def compose(self) -> ComposeResult:
        """Call to compose the app"""
        yield Header()
        yield Footer()
        yield Horizontal(
            self.my_list, Vertical(Static("column 2"), classes="column")
        )
```

If you run this app, you'll see some behavior that breaks:
- bindings don't work (`d` for dark mode, `q` for quit)
- `ListView` isn't updating when `ListView.append` is called (but you can confirm via `textual console` and `textual run --dev script.py` that the `print` statements are executing, so the `load_items()` is being called)
- `ListView` behavior also seems broken (clicking, arrow keys, highlighting all do not work)
[...]

PR #1636, opened 2023-01-23
fix stuck screen

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/1632

The fix was straightforward, but finding it required adding a few tools that are worth keeping.

A notable change is that if you mount anything without async, it is guaranteed to complete before the next message is processed. I think this will reduce race conditions where a dev doesn't await a mount.

- Fixes a stuck screen (deadlock on DOM lock)
- Adds a warning in debug mode when a callback takes too long to return
- Adds a `call_next` function which invokes a callback immediately after processing messages
- Forces awaiting of non-async mounts on the next message


src/textual/message_pump.py
```diff
@@ -367,35 +381,52 @@ async def _process_messages_loop(self) -> None:
                 except MessagePumpClosed:
                     break
 
+            self._active_message = message
```
willmcgugan, line 384: Purely for debugging.  If a handler is stuck, we can see what message is being processed.

davep:
Quickly testing the original issue here, 0.10.1 vs this PR and it looks good in terms if impact.

src/textual/widget.py
```diff
@@ -91,6 +91,10 @@ def __init__(self, parent: Widget, widgets: Sequence[Widget]) -> None:
         self._parent = parent
         self._widgets = widgets
 
+    async def __call__(self) -> None:
```
rodrigogiraoserrao, line 94: Is this so we can use `invoke` to await this?

rodrigogiraoserrao (approved):
The changes make sense to me.
Async isn't my cup of tea yet, though.

davep (approved):

willmcgugan:
> The changes make sense to me. Async isn't my cup of tea yet, though.

It's an acquired taste. Like blue cheese.

src/textual/widget.py
willmcgugan: precisely.

Issue #1589, opened 2023-01-17
Refactor widgets imports to allow importing of related objects


willmcgugan:
We have a module `__getattr__` in `widgets` which allows us to have a nice namespace for widgets.

For instance:

```python
from textual.widgets import DataTable
```

The problem with this is that some widgets have related objects that are required if you wish to subclass the widgets. Currently the only way to import these objects is with something like the following:

```python
from textual.widgets._data_table import Row
```

Some highly-coupled objects, such as messages and events. can go inside the widget class itself. For instance we have `DirectoryTree.FileSelected` but there remains some objects and symbols for which that doesn't work.

The proposed solution is for the widget classes to drop their underscore prefix, so you could import a related object like this:

```python
from textual.widgets.data_table import Row
```

This inevitably means that the `DataTable` has two imports, which can be problematic. To solve this we can prefix the class name with an underscore, which we drop in the lazy import machinary. Which means that the following would work in tandem:

```python
from textual.widgets import DataTable
from textual.widgets.data_table import Row
```

But the following will *not* work.

```python
from textual.widgets.data_table import DataTable
```

willmcgugan:
@darrenburns to comment

darrenburns:
@willmcgugan This looks exactly like what I had in mind.

davep:
Taking a look at this, I noticed that this would have a bit of a knock-on effect on type-hinting and the like within Textual itself, and would also longer-term result in a greater maintenance load when it comes to adding and updating widgets. In an "offline" chat with @willmcgugan and @darrenburns I've suggested and we've decided to go with a slightly different approach.

We'll keep everything as it is right now, and the "support classes" that are needed by developers using Textual will be imported and reexported from non-underscore-prefixed files that live parallel to their widget counterpart. So, to support this example:

```python
from textual.widgets import DataTable
from textual.widgets.data_table import Row
```

`widgets/_data_table.py` remains as it is, and a `data_table.py` is created which simply imports `Row` and makes it available. This should keep the current general widget lazy-loading approach while also making it easy and clear to declare "public support classes" for widgets.

davep:
Testing this, an example of the above in action:

```python
>>> from textual.widgets.data_table import DataTable
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: cannot import name 'DataTable' from 'textual.widgets.data_table' (/Users/davep/develop/python/textual/src/textual/widgets/data_table.py)
[...]

PR #1637, opened 2023-01-23
Support easier widget "support classes" importing

davep:
This PR implements the suggested solution for the requirement detailed in #1589. Quite simply, for those widgets that have non-widget support classes that the using-developer needs access to, have a parallel non-under-prefixed version of the code that imports and "reexports" the support classes.

As such, all this PR does is implement this approach for the widgets that seen to need it. The only one that required it at the moment was `DataTable`, so that `Column` and `Row` can be seen by the calling code. `TreeNode` already had an approach to this but for the sake of completeness I've added parallel support for this newer convention.

We may want to think about deprecating `widgets.TreeNode`.


rodrigogiraoserrao (approved):

willmcgugan (changes requested):
Could you remove references to TreeNode from the lazy-loading "widgets" while you are here.

davep:
> Could you remove references to TreeNode from the lazy-loading "widgets" while you are here.

Yup, happy to. Was holding that for a followup PR but if we're happy for the breaking change now that's my preference too. :-)

davep:
`TreeNode` now modified to follow this PR: https://github.com/Textualize/textual/pull/1637/commits/4adfe69ec90499a50a966ae655c05276f14a6acf

rodrigogiraoserrao (approved):

PR #1638, opened 2023-01-23
DataTable refactor - Adds cell updating and sorting functionality

darrenburns:
Rows and columns in the DataTable have been decoupled from their current visual location.

## API Changes

-  A `sort` method was added to the `DataTable`, allowing you to supply one or more column keys. Rows will be re-ordered using the provided keys.
- A new method `update_cell(row_key, column_key, value, update_width)` was added to the `DataTable`, which can be used to update the value of a cell identified by `(row_key, column_key)`.
- A similar method `update_coordinate(coordinate, value, update_width)` was added which can be used to update the cell which is currently present at `coordinate`.
- `add_row` and `add_column` have been updated to accept a `key` argument. If this `key` is not supplied, the DataTable will generate a key for you. This key will be returned from calls to `add_row` and `add_column` and can be used to reference the row and column regardless of where it’s currently located on the screen. This means that if you sort the rows in your table, you can always use the key of that row to reference it, even if its coordinate/index has changed.
- Similarly, `add_rows` and `add_columns` have been updated to return `list[RowKey]` and `list[ColumnKey]` respectively.
- `get_cell_value` has been renamed to `get_value_at(coordinate)` - this seems more appropriate given our new distinction between spatial coordinates and cells identified by row/column keys.
- A new version of `get_cell_value` exists which returns the value of a cell given a `RowKey` and `ColumnKey`.
- `cursor_cell` was renamed to `cursor_coordinate` since cells and coordinates are no longer the same things. It was renamed to make that more explicit.
    - The method `validate_cursor_cell` was renamed to `validate_cursor_coordinate`.
    - The method `watch_cursor_cell` was renamed to `watch_cursor_coordinate`.
- `hover_cell` renamed to `hover_coordinate` for similar reasons.
    - The method `validate_hover_cell` was renamed to `validate_hover_coordinate`.
- `row_count` is now a property. Previously it was an attribute that was manually kept in sync.
- Public attributes `data`, `columns` and `rows` are re-structured to all be dicts keyed by row and column keys. 
  - Ordering can no longer be relied upon when interacting with these structures. These structures store data but have no say in where it is located on the screen.
- New public properties `ordered_rows` and `ordered_columns` have been added, which return `list[Row]` and `list[Column]` respectively. The ordering of these lists reflects the positioning of the rows and columns you see on screen.
- Row/column/cell keys are accessible as attributes on the events emitted from the DataTable.
- Public method `refresh_cell` was changed to `refresh_coordinate(coordinate)`. (Although I'm not sure what the use case for this being public is).
- New public methods for validating coordinates are within the bounds of the DataTable and used extensively internally. Allows you to say "can I move my cursor to here without it being clamped?" before moving it.
- Fixes https://github.com/Textualize/textual/issues/1723
- Makes the filter module public.

- [ ] Updated documentation


src/textual/widgets/_data_table.py
```diff
@@ -26,40 +29,105 @@
...
+
+    def __hash__(self):
+        # If a string is supplied, we use the hash of the string. If no string was
+        # supplied, we use the default hash to ensure uniqueness amongst instances.
+        return hash(self.value) if self.value is not None else id(self)
+
+    def __eq__(self, other: object) -> bool:
+        # Strings will match Keys containing the same string value.
+        # Otherwise, you'll need to supply the exact same key object.
+        try:
```
willmcgugan, line 69: Comparing the hashes can lead to false positives. It might also make objects compare equal when they are different types, ie. `hash("") == hash(0)`

I think we will need to check it is a valid type to compare it to, and return `NotImplement` if it isn't.

src/textual/widgets/_data_table.py
```diff
@@ -26,40 +29,105 @@
...
 
 class CellDoesNotExist(Exception):
+    """Raised when the user supplies coordinates or cell keys which
+    do not exist in the DataTable."""
+
+    pass
+
+
+@functools.total_ordering
+class StringKey:
```
willmcgugan, line 51: Can we add a rich repr here? Would be useful for debugging.

src/textual/widgets/_data_table.py
```diff
@@ -569,77 +878,93 @@ def clear(self, columns: bool = False) -> None:
...
+            key: A key which uniquely identifies this column.
+                If None, it will be generated for you.
 
-        content_width = measure(self.app.console, text_label, 1)
+        Returns:
+            Uniquely identifies this column. Can be used to retrieve this column
+                regardless of its current location in the DataTable (it could have moved
+                after being added due to sorting/insertion/deletion of other columns).
+        """
+        column_key = ColumnKey(key)
```
willmcgugan, line 908: Should we check for duplicates here? I can imagine users forgetting and making two calls with the same `key`.

src/textual/widgets/_data_table.py
```diff
@@ -569,77 +878,93 @@ def clear(self, columns: bool = False) -> None:
...
-        self.rows[row_index] = Row(row_index, height, self._line_no)
-
-        for line_no in range(height):
-            self._y_offsets.append((row_index, line_no))
-
-        self.row_count += 1
-        self._line_no += height
-
-        self._new_rows.add(row_index)
+        row_key = RowKey(key)
```
willmcgugan, line 952: Also check for duplicate row keys?

willmcgugan (commented):
Just some comments to consider.

Pretty sure this will be good to merge early next week.

ofek:
I think there is a bug where if a cell's initial value is an empty string then the cell can no longer be updated after row creation:

```python
import asyncio

from textual.app import App
from textual.widgets import DataTable, Footer, Header, Switch

BACKGROUND_TASKS = []

class Listing(DataTable):
    def on_mount(self):
        super().on_mount()

        self.add_column('', key='status')
        self.add_column('', key='title')
        self.show_header = False
        self.cursor_type = 'row'

        self.call_after_refresh(lambda: BACKGROUND_TASKS.append(asyncio.create_task(self.__on_mount())))

    async def __on_mount(self):
        # self.add_row('something', 'title1', key='foo')
        self.add_row('', 'title1', key='foo')

class MyApp(App):
    def compose(self):
        yield Header()
        yield Switch()
        yield Listing()
        yield Footer()

    async def on_switch_changed(self, event: Switch.Changed):
        listing = self.query_one(Listing)
        listing.update_cell('foo', 'status', '✓' if event.value else '')

if __name__ == '__main__':
    app = MyApp()
    app.run()
```

darrenburns:
@ofek Since all the cells in that column are the empty string (including the label), the width will be calculated as 0. The width of a column will only be recomputed if you pass `update_width=True` to your call to `update_cell`.

src/textual/widgets/_data_table.py
```diff
@@ -26,40 +29,114 @@
...
+class DuplicateKey(Exception):
+    """Raised when the RowKey or ColumnKey provided already refers to
+    an existing row or column in the DataTable. Keys must be unique."""
+
+
+@functools.total_ordering
+class StringKey:
+    """An object used as a key in a mapping. It can optionally wrap a string,
+    and lookups into a map using the object behave the same as lookups using
+    the string itself."""
```
davep, line 57: May want reformatting to confirm with [our docstring guidelines](https://github.com/Textualize/textual/wiki/Documentation-standards#class).

src/textual/widgets/_data_table.py
```diff
@@ -26,40 +29,114 @@
...
+class RowKey(StringKey):
+    """Uniquely identifies a row in the DataTable. Even if the visual location
+    of the row changes due to sorting or other modifications, a key will always
+    refer to the same row."""
+
+
+class ColumnKey(StringKey):
+    """Uniquely identifies a column in the DataTable. Even if the visual location
+    of the column changes due to sorting or other modifications, a key will always
+    refer to the same column."""
```
davep, line 98: May want reformatting to confirm with [our docstring guidelines](https://github.com/Textualize/textual/wiki/Documentation-standards#class).

src/textual/widgets/_data_table.py
```diff
@@ -26,40 +29,114 @@
...
+
+
+class CellKey(NamedTuple):
+    """A unique identifier for a cell in the DataTable. Even if the cell changes
+    visual location (i.e. moves to a different coordinate in the table), this key
+    can still be used to retrieve it, regardless of where it currently is."""
```
davep, line 104: May want reformatting to confirm with [our docstring guidelines](https://github.com/Textualize/textual/wiki/Documentation-standards#class).

src/textual/widgets/_data_table.py
```diff
@@ -182,37 +257,43 @@ class DataTable(ScrollView, Generic[CellType], can_focus=True):
...
-        It's only relevant when the `cursor_type` is `"cell"`.
-        It's also posted when the cell cursor is re-enabled (by setting `show_cursor=True`),
-        and when the cursor type is changed to `"cell"`. Can be handled using
-        `on_data_table_cell_highlighted` in a subclass of `DataTable` or in a parent
-        widget in the DOM.
+        """Posted when the cursor moves to highlight a new cell. It's only relevant
+        when the `cursor_type` is `"cell"`. It's also posted when the cell cursor is
+        re-enabled (by setting `show_cursor=True`), and when the cursor type is
+        changed to `"cell"`. Can be handled using `on_data_table_cell_highlighted` in
+        a subclass of `DataTable` or in a parent widget in the DOM.
```
davep, line 272: May want reformatting to confirm with [our docstring guidelines](https://github.com/Textualize/textual/wiki/Documentation-standards#class).

src/textual/widgets/_data_table.py
```diff
@@ -223,36 +304,47 @@ class CellSelected(Message, bubble=True):
...
             yield "coordinate", self.coordinate
+            yield "cell_key", self.cell_key
 
     class RowHighlighted(Message, bubble=True):
         """Posted when a row is highlighted. This message is only posted when the
-        `cursor_type` is set to `"row"`. Can be handled using `on_data_table_row_highlighted`
-        in a subclass of `DataTable` or in a parent widget in the DOM.
+        `cursor_type` is set to `"row"`. Can be handled using
+        `on_data_table_row_highlighted` in a subclass of `DataTable` or in a parent
+        widget in the DOM.
```
davep, line 332: May want reformatting to confirm with [our docstring guidelines](https://github.com/Textualize/textual/wiki/Documentation-standards#class).

src/textual/widgets/_data_table.py
```diff
@@ -322,82 +427,233 @@ def __init__(
         classes: str | None = None,
     ) -> None:
         super().__init__(name=name, id=id, classes=classes)
+        self.data: dict[RowKey, dict[ColumnKey, CellType]] = {}
+        """Contains the cells of the table, indexed by row key and column key.
+        The final positioning of a cell on screen cannot be determined solely by this
+        structure. Instead, we must check _row_locations and _column_locations to find
+        where each cell currently resides in space."""
```
davep, line 434: May want reformatting to confirm with [our docstring guidelines](https://github.com/Textualize/textual/wiki/Documentation-standards#class).

src/textual/widgets/_data_table.py
```diff
@@ -322,82 +427,233 @@ def __init__(
...
-            tuple[int, int, Style, bool, bool], SegmentLines
-        ]
-        self._cell_render_cache = LRUCache(10000)
-        self._line_cache: LRUCache[tuple[int, int, int, int, int, int, Style], Strip]
-        self._line_cache = LRUCache(1000)
+            RowCacheKey, tuple[SegmentLines, SegmentLines]
+        ] = LRUCache(1000)
+        """For each row (a row can have a height of multiple lines), we maintain a
+        cache of the fixed and scrollable lines within that row to minimise how often
+        we need to re-render it. """
```
davep, line 455: May want reformatting to confirm with [our docstring guidelines](https://github.com/Textualize/textual/wiki/Documentation-standards#class).

src/textual/widgets/_data_table.py
```diff
@@ -322,82 +427,233 @@ def __init__(
...
+    def update_cell(
+        self,
+        row_key: RowKey | str,
+        column_key: ColumnKey | str,
+        value: CellType,
+        *,
+        update_width: bool = False,
+    ) -> None:
+        """Update the content inside the cell with the specified row key
+        and column key.
```
davep, line 546: May want reformatting to confirm with [our docstring guidelines](https://github.com/Textualize/textual/wiki/Documentation-standards#class).

src/textual/widgets/_data_table.py
```diff
@@ -408,72 +664,103 @@ def watch_show_cursor(self, show_cursor: bool) -> None:
...
             self.post_message_no_wait(
-                DataTable.CellHighlighted(self, cell_value, coordinate)
+                DataTable.CellHighlighted(
+                    self, cell_value, coordinate=coordinate, cell_key=cell_key
+                )
             )
 
+    def coordinate_to_cell_key(self, coordinate: Coordinate) -> CellKey:
+        """Return the key for the cell currently occupying this coordinate in the
+        DataTable
```
davep, line 721: May want reformatting to confirm with [our docstring guidelines](https://github.com/Textualize/textual/wiki/Documentation-standards#class).

src/textual/widgets/_data_table.py
```diff
@@ -650,33 +995,73 @@ def add_row(self, *cells: CellType, height: int = 1) -> None:
...
+            row_key = self.add_row(*row)
+            row_keys.append(row_key)
+        return row_keys
 
     def on_idle(self) -> None:
+        """Runs when the message pump is empty. We use this for
+        some expensive calculations like re-computing dimensions of the
+        whole DataTable and re-computing column widths after some cells
+        have been updated. This is more efficient in the case of high
+        frequency updates, ensuring we only do expensive computations once."""
```
davep, line 1040: May want reformatting to confirm with [our docstring guidelines](https://github.com/Textualize/textual/wiki/Documentation-standards#class).

src/textual/widgets/_data_table.py
```diff
@@ -712,29 +1097,99 @@ def _refresh_region(self, region: Region) -> None:
...
+            column_index: The column index to check.
+
+        Returns:
+            True if the column index is within the bounds of the table.
+        """
+        return 0 <= column_index < len(self.columns)
+
+    def is_valid_coordinate(self, coordinate: Coordinate) -> bool:
+        """Return a boolean indicating whether the given coordinate is within table
+        bounds.
```
davep, line 1124: May want reformatting to confirm with [our docstring guidelines](https://github.com/Textualize/textual/wiki/Documentation-standards#class).

src/textual/widgets/_data_table.py
darrenburns: I'm not going to bother doing this for the internal things. They're mostly just notes for us since the class is quite complex. I'll update the public facing stuff though.

src/textual/widgets/_data_table.py
```diff
@@ -26,40 +29,125 @@
...
+    def __hash__(self):
+        # If a string is supplied, we use the hash of the string. If no string was
+        # supplied, we use the default hash to ensure uniqueness amongst instances.
+        return hash(self.value) if self.value is not None else id(self)
+
+    def __eq__(self, other: object) -> bool:
+        # Strings will match Keys containing the same string value.
+        # Otherwise, you'll need to supply the exact same key object.
+        if isinstance(other, (str, StringKey)):
+            return hash(self) == hash(other)
```
willmcgugan, line 79: There's a tiny, but non-zero chance that comparing hashes may return false positives.

I think we need to use string comparison if both sides have a str.

Something like this: 

```suggestion
        if isinstance(other, str):
            return self.value == other
        elif isinstance(other, StringKey):
            if self.value is not None and other.value is not None:
                return self.value == other.value
            else:
                return hash(self) == hash(other)               
```

src/textual/widgets/_data_table.py
```diff
@@ -182,131 +268,172 @@ class DataTable(ScrollView, Generic[CellType], can_focus=True):
...
-            self, sender: DataTable, value: CellType, coordinate: Coordinate
+            self,
+            sender: DataTable,
+            value: CellType,
+            coordinate: Coordinate,
+            cell_key: CellKey,
         ) -> None:
             self.value: CellType = value
             self.coordinate: Coordinate = coordinate
+            self.cell_key: CellKey = cell_key
```
willmcgugan, line 302: Can you add docstrings for these attributes (and other messages). Just a string underneath the assignment, so the docs can pull it out.

src/textual/widgets/_data_table.py
```diff
@@ -485,81 +792,119 @@ def watch_cursor_type(self, old: str, new: str) -> None:
...
         if cursor_type == "cell":
-            self._highlight_cell(self.cursor_cell)
+            self._highlight_coordinate(self.cursor_coordinate)
         elif cursor_type == "row":
             self._highlight_row(row_index)
         elif cursor_type == "column":
             self._highlight_column(column_index)
 
-    def _update_dimensions(self, new_rows: Iterable[int]) -> None:
+    def _update_column_widths(self, updated_cells: set[CellKey]) -> None:
```
willmcgugan, line 820: Docstring por favor

willmcgugan (changes requested):
Nice work. Just a few minor requests.

PR #1639, opened 2023-01-23
Remove some parameter types that got missed on `Widget.move_child`

davep:
Rogue colon likely threw off whatever regexp was being used to hunt the examples down.


davep:
> Woops, thanks for catching this!

Easily missed given the format of the existing text was wrong. :-)

Issue #1430, opened 2022-12-22
[Tree] Provide some sort of `expand/collapse_all` facility for `TreeNode`


davep:
It's currently possible to [`expand`](https://textual.textualize.io/api/tree_node/#textual.widgets._tree.TreeNode.expand) a `TreeNode`, but that only expands the direct children. It would be useful if it had an optional parameter, or another method akin to `expand_all`, so that every expandable  descendent node is expanded.

willmcgugan:
Also `collapse_all`

PR #1644, opened 2023-01-23
Add support for a method of expanding/collapsing all tree nodes from a given node down

davep:
Adds optional `expand_all`, `collapse_all` and `toggle_all` parameters to the `expand`, `collapse` and `toggle` methods of `TreeNode`. In doing so I move the bulk of the work of those methods inside "private" methods that do the setting of the expanded status (and in turn calling on any children to do the same) without ever invalidating the display of the owning `Tree`. The "public" methods then simply call on those "private" methods and then invalidate the `Tree` (the intention being that there's no need to get the tree to redraw until after all the nodes have done their thing_.

See #1430.


rodrigogiraoserrao (approved):
The changes look good to me.
I have two questions, though:

 1. I'd expect the `collapse_all` default value to also be `False` so that, by default, collapsing a node will not mess with the collapsed/expanded state of children. E.g., like VS Code does in the built-in file explorer.
 2. What's the `self._tree._invalidate()`? Is there a "local" version of that that we can call when we only change the state of a single node, i.e. when the recursive calls are skipped?

davep:
> 1. I'd expect the `collapse_all` default value to also be `False` so that, by default, collapsing a node will not mess with the collapsed/expanded state of children. E.g., like VS Code does in the built-in file explorer.

Not sure that's a question. :-P

But, yes, you're right, that (and toggle) is a typo, so good catch.

> 2. What's the `self._tree._invalidate()`? Is there a "local" version of that that we can call when we only change the state of a single node, i.e. when the recursive calls are skipped?

I'm not sure I understand this question though, could you elaborate?

rodrigogiraoserrao:
> Not sure that's a question. :-P

It's not, I started typing with one formulation in mind and ended up writing something different 😆 

> I'm not sure I understand this question though, could you elaborate?

I'm assuming the call to `invalidate` is to invalidate some cache or to flag the need for redrawing the tree, or something of the sorts. When there are no recursive updates, can you not make a similar call that just says "hey, invalidate just this node"? Do you need to have the same `invalidate` call regardless of whether you updated one node or 1000 nodes?

darrenburns:
The "toggle children" behaviour seems weird to me - if I have some children open those will be closed, and the children that are closed will be open?

Would it make more sense to just toggle the parent, then walk the children and make sure their state _matches_ that of the parent?

davep:
> When there are no recursive updates, can you not make a similar call that just says "hey, invalidate just this node"? Do you need to have the same `invalidate` call regardless of whether you updated one node or 1000 nodes?

Sorry, I'm still not sure I'm following the question. `self._tree._invalidate` invalidates the tree, not the node, so it only needs to be done the once no matter if you change the expanded state of a single node or 1000s of nodes.

davep:
Taking this back to a draft PR for a wee while as I can see there's questions about user experience and expectation to address.

davep:
Right, a36583612c78d21fe8bbd2cce2bb6da2caf70fe8 takes the toggle from being "technically correct" to, erm... actually useful? 😄

rodrigogiraoserrao (approved):
Nice adding the `*` to force the `x_all` parameters to be keyword only.

src/textual/widgets/_tree.py
```diff
@@ -159,24 +159,60 @@ def allow_expand(self, allow_expand: bool) -> None:
...
+        Args:
+            expand_all: If `True` expand all offspring at all depths.
+        """
         self._expanded = True
         self._updates += 1
+        if expand_all:
+            for child in self.children:
+                child._expand(expand_all)
+
+    def expand(self, *, expand_all: bool = False) -> None:
```
willmcgugan, line 174: There's some redundancy in the parameter name here `exapand_all`, and friends.

At the moment to expand all nodes you would do this:

```python
node.expand(expand_all=True)
```

Two mentions of "expand". But its not like there is any ambiguity re what the "all" refers to.

How about this?

```python
node.expand(all=True)
```

davep:
Todo: we've decided to break these out into different methods; `expand` vs `expand_all`, etc.

willmcgugan (changes requested):
Looks great. Could use tests.

davep:
Cool, now we're happy with this I'll add them.

davep:
Tests for all the `TreeNode` expand/collapse/toggle methods now added.

Issue #1634, opened 2023-01-21
Removing a widget from an action, when it initiated the action, causes the app to hang


davep:
The following code may be a little convoluted in that it's using `@click` from inside a renderable, this started out as an investigation as to why dependency links in [PISpy](https://github.com/davep/pispy) were hanging the application after upgrading Textual to 0.10.1 -- when I get a moment I'll see if I can reduce it the the smallest recreation possible.

Anyway, given this:

```python
from textual.app        import App, ComposeResult
from textual.containers import Vertical
from textual.widgets    import Header, Footer, Label

class Output( Vertical ):

    async def clear( self ) -> None:
        await self.query( "*" ).remove()

    async def populate( self ) -> None:
        await self.clear()
        await self.mount( *[
            Label( f"[@click=app.populate]Repopulate this widget[/] {n}")
            for n in range( 20 )
        ] )

    async def on_mount( self ) -> None:
        await self.populate()

class AsyncClickApp( App[ None ] ):

    def compose( self ) -> ComposeResult:
        yield Header()
        yield Output()
        yield Footer()

    async def action_populate( self ) -> None:
        await self.query_one( Output ).populate()

if __name__ == "__main__":
    AsyncClickApp().run()
```

Clicking on any of the links causes the application to hang with 0.10.1. If I run this with 0.9.1 it works fine, clearing the `Vertical` of its content and repopulating it.

PR #1653, opened 2023-01-24
Rhymes with "Bed rock"

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/1634

1. Clicking on a link happens in response to a Click event on Label.
2. Textual runs an action which tries to clear the label
3. The label waits for all the labels to close, but it never will because the label is waiting for itself to close.

- There is now an `active_message_pump` which is set the widget currently handling an event. The `_close_messages` functions uses this to avoid waiting on a message pump that will never return,
- Added a `Widget.mount_all` method. I find the pattern of `self.mount(*[foo for foo in bar])` to be too ugly.

### DIFF
--- a/src/textual/widget.py
+++ b/src/textual/widget.py
@@ -618,6 +618,33 @@ class Widget(DOMNode):
         self.call_next(await_mount)
         return await_mount
 
+    def mount_all(
+        self,
+        widgets: Iterable[Widget],
+        *,
+        before: int | str | Widget | None = None,
+        after: int | str | Widget | None = None,
+    ) -> AwaitMount:
+        """Mount widgets from an iterable.
+
+        Args:
+            widgets: An iterable of widgets.
+            before: Optional location to mount before.
+            after: Optional location to mount after.
+
+        Returns:
+            An awaitable object that waits for widgets to be mounted.
+
+        Raises:
+            MountError: If there is a problem with the mount request.
+
+        Note:
+            Only one of ``before`` or ``after`` can be provided. If both are
+            provided a ``MountError`` will be raised.
+        """
+        await_mount = self.mount(*widgets, before=before, after=after)
+        return await_mount
+
     def move_child(
         self,
         child: int | Widget,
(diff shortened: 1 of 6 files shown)


src/textual/widget.py
```diff
@@ -618,6 +618,33 @@ def mount(
...
+        *,
+        before: int | str | Widget | None = None,
+        after: int | str | Widget | None = None,
+    ) -> AwaitMount:
+        """Mount widgets from an iterable.
+
+        Args:
+            widgets: An iterable of widgets.
+            before: Optional location to mount before.
+            after: Optional location to mount after.
```
davep, line 633: Wee reminder for when this goes non-draft, yesterday I expanded the documentation for `before`/`after` so this may want to follow that pattern: https://github.com/Textualize/textual/blob/2662e6b2923d0f5aad8abff1281d463036e7d3d0/src/textual/widget.py#L564-L569

Also you've reminded me that I *didn't* update `App.mount_all` with this (I'll do that).

rodrigogiraoserrao (approved):
As far as my insight goes, looks good to me.

tests/test_widget.py
```diff
@@ -157,3 +156,29 @@ def test_widget_mount_ids_must_be_unique_mounting_multiple_calls(parent):
...
+        async def action_remove_all(self) -> None:
+            await self.query_one(Container).clear()
+            self.exit(123)
+
+    app = RemoveApp()
+    async with app.run_test() as pilot:
+        await pilot.press("r")
+        await pilot.pause()
+    assert app.return_value == 123
+    assert True
```
darrenburns, line 184:
```suggestion
```

rodrigogiraoserrao: I mean, it's not wrong 😆

darrenburns (approved):

tests/test_widget.py
davep: It's a very common cosmic ray detection system. If it fails you know a cosmic ray was detected.

davep (approved):

tests/test_widget.py
willmcgugan: My bad. A placeholder while I was finding something to assert on.

PR #1655, opened 2023-01-24
[WiP] Add support for suspending/resuming a Textual app, both from code, and with Ctrl-Z support on Un*x

davep:
A PR that draws in #1541 from @JoshKarpel and then uses it to support allowing a Textual app to react to `SIGTSTP` (and so <kbd>Ctrl</kbd>+<kbd>Z</kbd> and a subsequent `fg` (assuming a shell with such a command).

I want to do a bunch more testing first but I think this is working pretty well. Tested fine on macOS, need to test on GNU/Linux. For now <kbd>Ctrl</kbd>+<kbd>Z</kbd> is (hopefully obviously) a nop on Windows.


src/textual/app.py
```diff
@@ -979,6 +1001,28 @@ async def run_app() -> None:
             event_loop.run_until_complete(run_app())
         return self.return_value
 
+    @contextmanager
+    def suspend(self) -> Iterator[None]:
```
davep, line 1005: Very minor thing this, but I wonder if this would be better called `suspended` rather than `suspend`:

```py
with self.suspended():
   ...
```

sort of reads a little better, ish, kinda? Maybe?

@JoshKarpel, @willmcgugan, any preferences?

willmcgugan: suspend fits a convention of methods starting with verbs. Suspended reads like a property to me.

davep:
At the moment, in its current state, unit testing doesn't seem to manage to complete at all. I also need to update some unit tests relating to default app key bindings given that we've added a new one.

davep:
Fixed the binding inheritance tests so that they they take into account the fact that <kbd>Ctrl</kbd>+<kbd>Z</kbd> is now a hard-coded `App` binding.

Next up: `test_suspend.py` just plain flat hangs and has to be killed; I think this is also what's causing CI to get stuck endlessly too.

davep:
Just reducing the test down to:

```python
from textual.app import App


async def test_suspend() -> None:
    async with App().run_test() as pilot:
        with pilot.app.suspend():
            pass
```

still has it just hanging and doing nothing (on macOS, Textual `main`).

I think a bit of manual testing might be in order here, at least some testing of this outside of `pytest` just to be sure it works fine (using `suspend` in a <kbd>Ctrl</kbd>+<kbd>Z</kbd> context is just fine -- I can suspend and resume the Textual demo in exactly the way you'd expect).

davep:
Simple test of `App.suspend`:

```python
import os

from textual.app     import App, ComposeResult
from textual.widgets import Header, Footer, Label

class ShellApp( App[ None ] ):

    CSS = """
    Screen {
        align: center middle;
    }
    """

    BINDINGS = [
        ( "s", "shell", "Drop to a shell" )
    ]

        yield Header()
        yield Label( "Press S to drop to a shell" )
        yield Footer()

    def action_shell( self ) -> None:
        """Drop to a shell"""
        with self.suspend():
            os.system( os.environ.get( "SHELL", "/bin/sh" ) )

if __name__ == "__main__":
    ShellApp().run()
```

this works perfectly fine. Whatever the issue is when running the tests, it would suggest it's one of pytest itself or something to do with the pilot.

davep:
Testing some more under `pytest`, using the highly-scientific approach of running:

```sh
$ poetry run pytest -vv --full-trace tests/test_suspend.py
```

waiting varying amounts of time, then hitting <kbd>Ctrl</kbd>+<kbd>C</kbd>, it looks like when running under the `Pilot` it gets stuck [in this code](https://github.com/davep/textual/blob/9bcda041f06646285f4d77df9be7e6f67fb20e02/src/textual/app.py#L218-L223); specifically on this line, as best as I can tell:

```python
self._flushed.wait()
```

Again, this only seems to be in testing.

To further eliminate things, if I leave `pytest` out of the equation and just try running it standalone:

```python
import sys
import asyncio

from textual.app import App

async def test_suspend() -> None:
    async with App().run_test() as pilot:
        with pilot.app.suspend():
            print("foo", end="", flush=True)
            print("bar", file=sys.stderr, end="", flush=True)

asyncio.run(test_suspend())
```

by running:

```sh
$ poetry run python tests/test_suspend.py
```

I still get the hang and it appears to be in the same place.

Current thinking then is that `Pilot` and `App.suspend` don't place nice.

@JoshKarpel -- were you running that test just fine before now? How about now?

JoshKarpel:
Ick - that had worked during my local testing, but it's weird threaded code and it's definitely possible I missed something or that there are timing issues that worked on my laptop but not in CI/other people's laptops. I'm not sure it ever passed in CI - there might have been merge conflicts at the time that stopped it from running.

See also my comment here https://github.com/Textualize/textual/pull/1541#discussion_r1080746042 ; I am not really sure what the right way to go is. I think the key thing is that you need to make sure the write queue is drained before `yield`ing in `App.suspend()`.

willmcgugan:
I think this one might require a bit more discussion. I don't have it clear in my head how Textual should handle this in a predictable way. @davep not sure how far you have got in to this, but you may want to park it until we've had a chat about it.

src/textual/app.py
```diff
@@ -156,6 +163,10 @@ def __init__(self) -> None:
         self._queue: Queue[str | None] = Queue(MAX_QUEUED_WRITES)
         self._file = sys.__stdout__
 
+        self._flushed = threading.Event()
```
willmcgugan, line 166: I suspect this is not the right approach.

At some point we need to tell the app to not make any more writes. With that in place, we can stop and join the thread. That way we can be certain there are no writes while we "flush".

davep:
> @davep not sure how far you have got in to this, but you may want to park it until we've had a chat about it.

At the moment I'm at a point where it all works really well (or at least appears to), except via the Pilot. Keyboard suspend with `fg` resume, and the `suspend()` context, are working really well for me.

But, yeah, I've kept this as a WiP draft for now as I feel it needs a few more eyes/ideas.

src/textual/app.py
dbieber: https://peps.python.org/pep-0343/#examples suggests the names suspended and paused over suspend and pause since these are contextmanagers that take an action in enter and undo it during exit. (This is consistent w/ my intuition / preference too.)

davep:
Closing in favour of #4064.

PR #1656, opened 2023-01-24
Explain the significance of the before/after types (redux)

davep:
Follows on from https://github.com/Textualize/textual/pull/1641 where I did the update for mounting via Widget but forgot about App.


rodrigogiraoserrao (approved):
Ship it! 🚢

PR #1659, opened 2023-01-24
Keep track of scheduled animations

### LINKED ISSUE #1372: Widgets styled with a `:hover` can sometimes end up with their styles sticking
It's not 100% clear to me yet exactly what happens, but this can be seen with 5x5 from our examples, where a `:hover` style can get stuck after the mouse has left the widget. See here for example:

Screenshot 2022-12-15 at 16 05 58

In the image 3 cells have the hover style (the background slightly lighter than the background for the other "off" cells). One of them, at the time of the screenshot, had the mouse over it, the other 2 didn't.

This *seems* to happen if you're a bit quick with mouse movements.

Needs some investigation and a proper route to reproduction found.

### LINKED ISSUE #1658: No way to wait for all animations (current and scheduled) to be done.
The coroutine `Animator.wait_for_idle` will wait for animations that are in effect to be done.
However, it does not take into account transitions that are scheduled.

It would be nice if we could await for _all_ transitions, both the ones that are happening and the ones that are scheduled to happen in the future.

This will come in handy for testing and _might_ also be useful for users in apps.

rodrigogiraoserrao:
- We keep track of animations that are scheduled but not running yet.
 - We add the ability to wait for current and scheduled animations.
 - We fix #1372. 
 - We fix #1658. 
 - We add regression tests for 1372 plus a couple of other tests for animation features.

### DIFF
--- a/src/textual/_animator.py
+++ b/src/textual/_animator.py
@@ -190,11 +190,15 @@ class Animator:
             callback=self,
             pause=True,
         )
+        # Flag if no animations are currently taking place.
         self._idle_event = asyncio.Event()
+        # Flag if no animations are currently taking place and none are scheduled.
+        self._fully_idle_event = asyncio.Event()
 
     async def start(self) -> None:
         """Start the animator task."""
         self._idle_event.set()
+        self._fully_idle_event.set()
         self._timer.start()
 
     async def stop(self) -> None:
@@ -205,6 +209,7 @@ class Animator:
             pass
         finally:
             self._idle_event.set()
+            self._fully_idle_event.set()
 
     def bind(self, obj: object) -> BoundAnimator:
         """Bind the animator to a given object."""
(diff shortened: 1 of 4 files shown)


src/textual/_animator.py
```diff
@@ -368,3 +396,7 @@ def _get_time(self) -> float:
     async def wait_for_idle(self) -> None:
         """Wait for any animations to complete."""
         await self._idle_event.wait()
+
+    async def wait_for_fully_idle(self) -> None:
```
willmcgugan, line 400: I would rename this. Suggest `wait_until_complete`. Otherwise its not clear what idle means. I would take idle to mean "doing nothing right now".

willmcgugan (changes requested):
Nice work! Just one request to rename a method.

src/textual/_animator.py
rodrigogiraoserrao: Fair point. How about `wait_until_done`?

willmcgugan: We use the term `complete` in reference to animations, so I think it makes sense to stick to that.

rodrigogiraoserrao: Sorted in d91a10b

willmcgugan (approved):

Issue #1657, opened 2023-01-24
Text is pasted twice


nitzan-shaked:
I suspect this has to do with `events.Paste` now bubbling? Not sure, but either way text gets pasted twice into `Input` widgets.
Adding `event.stop()` at the bottom of `Input.on_paste` solves this.

I can't be the only one, right?...

PR #1660, opened 2023-01-24
Fix double-paste into `Input`

davep:
See #1657.


nitzan-shaked:
lgtm :) thanks for fixing. I'd approve if I could :)

PR #1664, opened 2023-01-25
don't request sync on macOS app

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/1622

I think the terminal *should* ignore this escape sequence. It isn't so we will have to workaround it.

### DIFF
--- a/src/textual/drivers/linux_driver.py
+++ b/src/textual/drivers/linux_driver.py
@@ -145,7 +145,11 @@ class LinuxDriver(Driver):
         self._request_terminal_sync_mode_support()
         self._enable_bracketed_paste()
 
-    def _request_terminal_sync_mode_support(self):
+    def _request_terminal_sync_mode_support(self) -> None:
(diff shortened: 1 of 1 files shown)


davep (approved):

src/textual/drivers/linux_driver.py
```diff
@@ -146,8 +146,9 @@ def on_terminal_resize(signum, stack) -> None:
         self._enable_bracketed_paste()
 
     def _request_terminal_sync_mode_support(self):
-        self.console.file.write("\033[?2026$p")
-        self.console.file.flush()
+        if self.console._environ.get("TERM_PROGRAM", "") != "Apple_Terminal":
```
darrenburns, line 149: Could you add a comment here saying why this code is here? It's the sort of thing someone else could look at 6 months from now and not understand.

willmcgugan: Yeah, fair enough.

willmcgugan: Done

PR #1672, opened 2023-01-26
update to intro page, fix for API docs

willmcgugan:
- Updated the wording of the intro.
- Fixes some broken API docs.

Mainly to change the messaging from "build TUIs"  to "Build UIs that happen to run in the terminal"

Screenshot 2023-01-26 at 12 24 52

### DIFF
--- a/docs/widgets/tree.md
+++ b/docs/widgets/tree.md
@@ -21,7 +21,7 @@ The example below creates a simple tree.
     --8<-- "docs/examples/widgets/tree.py"
     ```
 
-A each tree widget has a "root" attribute which is an instance of a [TreeNode][textual.widgets.TreeNode]. Call [add()][textual.widgets.TreeNode.add] or [add_leaf()][textual.widgets.TreeNode.add_leaf] to add new nodes underneath the root. Both these methods return a TreeNode for the child, so you can add more levels.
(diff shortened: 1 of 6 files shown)


docs/widgets/tree.md
```diff
@@ -21,7 +21,7 @@ The example below creates a simple tree.
     --8<-- "docs/examples/widgets/tree.py"
     ```
 
-A each tree widget has a "root" attribute which is an instance of a [TreeNode][textual.widgets.TreeNode]. Call [add()][textual.widgets.TreeNode.add] or [add_leaf()][textual.widgets.TreeNode.add_leaf] to add new nodes underneath the root. Both these methods return a TreeNode for the child, so you can add more levels.
+A each tree widget has a "root" attribute which is an instance of a [TreeNode][textual.widgets.tree.TreeNode]. Call [add()][textual.widgets.tree.TreeNode.add] or [add_leaf()][textual.widgets.tree,TreeNode.add_leaf] to add new nodes underneath the root. Both these methods return a TreeNode for the child, so you can add more levels.
```
davep, line 24: Drop the "A" the start and start "Each"? Or the other way round.

davep (approved):

PR #1689, opened 2023-01-30
Scroll sensitivity

willmcgugan:
- Adds scroll_sensitivity_x and scroll_sensitivity_y to change the number of lines to move the scroll position by when using the scroll wheel and trackpad
- Adds horizontal scrolling, with shift or ctrl + scroll wheel. Horizontal scrolling uses shift + mouse wheel on macOS, but for some reason ITerm won't send mouse wheel events when shift is pressed. Kitty does. iTerm bug? Don't know, but for now both ctrl and shift will scroll horizontally.

I've doubled the scroll speed which makes the wheel somewhat usable, and the trackpad remains usable if a little faster.

There is no way to know which is being used, which means we need to find settings that works with both.

**Please check this out and play with scrolling**. I'm concerned that it may work for me, and break others.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Added the coroutines `Animator.wait_until_complete` and `pilot.wait_for_scheduled_animations` that allow waiting for all current and scheduled animations https://github.com/Textualize/textual/issues/1658
 - Added the method `Animator.is_being_animated` that checks if an attribute of an object is being animated or is scheduled for animation
-- Added App.scroll_sensitivity to adjust how many lines the scroll wheel moves the scroll position https://github.com/orgs/Textualize/projects/5?pane=issue&itemId=12998591
+- Added App.scroll_sensitivity_x and App.scroll_sensitivity_y to adjust how many lines the scroll wheel moves the scroll position https://github.com/orgs/Textualize/projects/5?pane=issue&itemId=12998591
 - Added Shift+scroll wheel and ctrl+scroll wheen to scroll horizontally
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Added the coroutines `Animator.wait_until_complete` and `pilot.wait_for_scheduled_animations` that allow waiting for all current and scheduled animations https://github.com/Textualize/textual/issues/1658
 - Added the method `Animator.is_being_animated` that checks if an attribute of an object is being animated or is scheduled for animation
+- Added App.scroll_sensitivity to adjust how many lines the scroll wheel moves the scroll position https://github.com/orgs/Textualize/projects/5?pane=issue&itemId=12998591
+- Added Shift+scroll wheel and ctrl+scroll wheen to scroll horizontally
 
 ### Changed
 
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -369,6 +369,8 @@ class App(Generic[ReturnType], DOMNode):
         self.css_path = css_paths
         self._registry: WeakSet[DOMNode] = WeakSet()
 
+        self.scroll_sensitivity: float = 3
+
         self._installed_screens: WeakValueDictionary[
             str, Screen | Callable[[], Screen]
         ] = WeakValueDictionary()
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -369,7 +369,8 @@ class App(Generic[ReturnType], DOMNode):
         self.css_path = css_paths
         self._registry: WeakSet[DOMNode] = WeakSet()
 
-        self.scroll_sensitivity: float = 3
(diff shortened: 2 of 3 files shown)


src/textual/app.py
```diff
@@ -369,6 +369,11 @@ def __init__(
         self.css_path = css_paths
         self._registry: WeakSet[DOMNode] = WeakSet()
 
+        self.scroll_sensitivity_x: float = 4
```
willmcgugan, line 372: Scrolling in x dimension is double to make the perceived distance the same in both directions (given cells are typically twice as tall as they are wide).

src/textual/app.py
```diff
@@ -369,6 +369,11 @@ def __init__(
         self.css_path = css_paths
         self._registry: WeakSet[DOMNode] = WeakSet()
 
+        self.scroll_sensitivity_x: float = 4.0
+        """Number of lines to scroll in the X direction with wheel or trackpad."""
```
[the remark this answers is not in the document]
willmcgugan: I suppose its cells in X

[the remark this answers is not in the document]
willmcgugan: No, you're right, columns!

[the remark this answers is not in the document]
davep: Split the difference: vertical lines. ;-)

src/textual/app.py
rodrigogiraoserrao: Isn't that a good `# comment` in the original source?

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Added the coroutines `Animator.wait_until_complete` and `pilot.wait_for_scheduled_animations` that allow waiting for all current and scheduled animations https://github.com/Textualize/textual/issues/1658
 - Added the method `Animator.is_being_animated` that checks if an attribute of an object is being animated or is scheduled for animation
+- Added App.scroll_sensitivity_x and App.scroll_sensitivity_y to adjust how many lines the scroll wheel moves the scroll position https://github.com/orgs/Textualize/projects/5?pane=issue&itemId=12998591
+- Added Shift+scroll wheel and ctrl+scroll wheen to scroll horizontally
```
rodrigogiraoserrao, line 15:
```suggestion
- Added Shift+scroll wheel and ctrl+scroll wheel to scroll horizontally
```

rodrigogiraoserrao (changes requested):

src/textual/widget.py
```diff
@@ -2480,15 +2480,25 @@ def _on_descendant_focus(self, event: events.DescendantBlur) -> None:
...
-        if self.allow_vertical_scroll:
-            if self.scroll_down(animate=False):
-                event.stop()
-
-    def _on_mouse_scroll_up(self, event) -> None:
-        if self.allow_vertical_scroll:
-            if self.scroll_up(animate=False):
-                event.stop()
+    def _on_mouse_scroll_down(self, event: events.MouseScrollDown) -> None:
+        if event.ctrl or event.shift:
```
darrenburns, line 2484: Do we want to steal both control and shift here? I forget which one is usually used for this kind of behaviour.

If we steal both, users wouldn't be able to do something like "hold shift and use mouse wheel to zoom instead of scroll", would they?

davep (approved):
> **Pease check this out and play with scrolling**. I'm concerned that it may work for me, and break others.

Tested on macOS with both a Magic Trackpad and a Logitech M575 trackball. In this version scrolling with the trackpad is obviously faster, but I find it just as comfortable to control; starting and stopping and getting to where I want to be vertically feels about the same (testing with the Textual demo). With the scroll wheel on the M575, while still a bit of a snooze-fest when it comes to scrolling, it's obviously far better than before this change (I had two windows open side-by-side, without and with this change, to compare).

Hopefully this'll be the bedrock for a method of letting people configure it "just so", in which case I see this as really handy and, from my experience, it doesn't impact my current use of Textual in iTerm on macOS.

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Added the coroutines `Animator.wait_until_complete` and `pilot.wait_for_scheduled_animations` that allow waiting for all current and scheduled animations https://github.com/Textualize/textual/issues/1658
 - Added the method `Animator.is_being_animated` that checks if an attribute of an object is being animated or is scheduled for animation
+- Added App.scroll_sensitivity_x and App.scroll_sensitivity_y to adjust how many lines the scroll wheel moves the scroll position https://github.com/orgs/Textualize/projects/5?pane=issue&itemId=12998591
```
davep, line 14: That URL can only be seen by members of the organisation.

src/textual/widget.py
willmcgugan: Ideally it would be shift+mouse wheel only, but the problem is that iTerm doesn't report shift when pressed with the mouse wheel. Kitty does. I added both shift and ctrl so there is at least one way of horizontal scrolling.

CHANGELOG.md
willmcgugan: Fixed

PR #1690, opened 2023-01-30
Fix render.py

rodrigogiraoserrao:


willmcgugan (approved):

rodrigogiraoserrao:
Closed this PR because I was trying to fix some issues that stemmed from merging #1686 but there is a snapshot app missing (`label_wrap.py`) so it has to be Will to fix it.

PR #1691, opened 2023-01-30
fix typing

willmcgugan:
No idea how screwed this up, but this will get CI building again


rodrigogiraoserrao:
If I understand correctly, you needed to _add and commit_ the app `label_wrap.py` that you wrote as a test for the original PR (and that was already added to the snapshot history), and not delete the test altogether.

willmcgugan:
@rodrigogiraoserrao That's exactly what I would do if I had that file!

rodrigogiraoserrao (approved):

rodrigogiraoserrao:
Oh ok. Let me know if you need me to come up with a snapshot test to cover for the original bug the snapshot test was targeting.

PR #1697, opened 2023-01-30
scroll acceleration experiment

willmcgugan:
Adds scroll acceleration, so the scroll position moves in greater increments when the wheel or trackpad is moved _faster_. Conversely slower motions move the scroll position slower, at a line at a time.

The feels better to me. Wheel and trackpad feel nearly right. But would like feedback...

### DIFF
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -372,9 +372,9 @@ class App(Generic[ReturnType], DOMNode):
 
         # Sensitivity on X is double the sensitivity on Y to account for
         # cells being twice as tall as wide
-        self.scroll_sensitivity_x: float = 4.0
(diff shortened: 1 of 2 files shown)


davep:
> The feels better to me. Wheel and trackpad feel nearly right. But would like feedback...

To mirror what I said on Slack: testing this here (macOS 13.2, 2019 MacBook Pro (Intel), iTerm, Magic Trackpad and a Logitech M575 trackball), I'm finding that this change:

- Makes trackpad movement very jumpy; whereas before scrolling has always been butter-smooth and precise, this feels very ragged and tricky to land where I want the scroll to land.
- Scroll wheel movement on the trackball seems to have massively slowed down with this change. I've not tested side-by-side yet but it feels slower than before #1689

Issue #1666, opened 2023-01-25
Review how the `Paste` event is handled and bubbled


davep:
Using the code below... Review how the `Paste` event is handled and bubbled in an application. Given the code below, ideally, if one were to run the app up and not focus anything, a paste would result in either/or the `Screen` or `App` getting a `Paste` and any `on_paste` event they have being able to handle it. At the moment no event appears at the `on_paste` for either.

```python
from textual.app        import App, ComposeResult
from textual.containers import Vertical
from textual.screen     import Screen
from textual.widgets    import Header, Footer, TextLog, Button
from textual.events     import Paste
from textual.binding    import Binding

class PasteButton( Button ):

    def on_paste( self, event: Paste ) -> None:
        self.app.query_one( TextLog ).write(
            f"Button: {event}"
        )

class PasteVertical( Vertical ):

    def compose( self ) -> ComposeResult:
        """Compose the child widgets."""
        yield PasteButton( "This can get focus" )

    def on_paste( self, event: Paste ) -> None:
        self.app.query_one( TextLog ).write(
            f"Vertical: {event}"
        )

class PasteTextLog( TextLog ):

    def on_paste( self, event: Paste ) -> None:
        self.app.query_one( TextLog ).write(
            f"TextLog: {event}"
        )

class PasteScreen( Screen ):

    def on_paste( self, event: Paste ) -> None:
        self.query_one( TextLog ).write(
            f"Screen: {event}"
        )

    def compose( self ) -> ComposeResult:
        yield Header()
        yield PasteVertical()
        yield PasteTextLog()
        yield Footer()

class PasteExplorer( App[ None ] ):

    BINDINGS = [
        Binding( "c", "clear", "Clear Log" ),
    ]

    SCREENS = {
        "main": PasteScreen
    }

    CSS = """
    *:focus {
        border: double green;
    }

    Vertical {
        height: 1fr;
        border: round yellow;
    }

    TextLog {
        height: 80%;
        border: round red;
    }
    """

    def on_paste( self, event: Paste ) -> None:
        self.query_one( TextLog ).write(
            f"App: {event}"
        )

    def on_mount( self ) -> None:
        self.push_screen( "main" )

    def action_clear( self ) -> None:
        """Clear"""
        self.query_one( TextLog ).clear()

if __name__ == "__main__":
    PasteExplorer().run()

### paste_explorer.py ends here

```

willmcgugan:
@davep Stealing this

PR #1698, opened 2023-01-31
Fix paste and test

willmcgugan:
Fix issue with paste event that prevented the app from handling it.

Fixes https://github.com/Textualize/textual/issues/1666


rodrigogiraoserrao (approved):
The `event.is_forwarded` flag is the flag that takes care of making sure the app directs the events to the correct place, let them bubble up once, and then get rid of them?

willmcgugan:
@rodrigogiraoserrao Exactly.

Issue #1693, opened 2023-01-30
An `Input` with `width: auto` tends to scroll the first character of its `value` off the left of its display


davep:
Given the following code:

```python
from textual.app        import App, ComposeResult
from textual.containers import Vertical
from textual.widgets    import Header, Footer, Label, Input

class InputWidthAutoApp( App[ None ] ):

    CSS = """
    Input.fixed {
        width: 30;
    }
    Input.auto {
        width: auto;
    }
    Input.pcent {
        width: 20%;
    }
    """

    def compose( self ) -> ComposeResult:
        yield Header()
        yield Vertical(
            Label( "Default width:" ), Input( placeholder="This has default width" ),
            Label( "Fixed width:" ), Input( placeholder="This has a fixed width", classes="fixed" ),
            Label( "Auto width:" ), Input( placeholder="This has auto width", classes="auto" ),
            Label( "%age width:" ), Input( placeholder="This has %age width", classes="pcent" )
        )
        yield Footer()

if __name__ == "__main__":
    InputWidthAutoApp().run()
```

All of the inputs will show the content as you'd expect, with the exception of the one that is `width: auto`; where the first character will scroll off the left. It's still there and can be scrolled back to. For example, typing the same word into each `Input`:

![Screenshot 2023-01-30 at 16 21 40](https://user-images.githubusercontent.com/28237/215534215-020582ae-56cc-40fe-923a-51d853362cf6.png)

PR #1699, opened 2023-01-31
Auto width inputs

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/1693

Fixes a glitch where auto-width inputs were scrolled off by one character. 

https://user-images.githubusercontent.com/554369/215745942-bb0adaeb-8b94-40f6-a8ce-1827d7c54602.mov

### DIFF
--- a/src/textual/widgets/_input.py
+++ b/src/textual/widgets/_input.py
@@ -255,6 +255,7 @@ class Input(Widget, can_focus=True):
         return self._position_to_cell(len(self.value)) + 1
 
     def render(self) -> RenderableType:
+        self.view_position = self.view_position
         if not self.value:
             placeholder = Text(self.placeholder, justify="left")
             placeholder.stylize(self.get_component_rich_style("input--placeholder"))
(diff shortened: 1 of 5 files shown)


src/textual/widgets/_input.py
```diff
@@ -255,6 +255,7 @@ def cursor_width(self) -> int:
         return self._position_to_cell(len(self.value)) + 1
 
     def render(self) -> RenderableType:
+        self.view_position = self.view_position
```
willmcgugan, line 258: Forces validation again. This is required as the input may have changed size since last render.

rodrigogiraoserrao: How the heck does one conclude the issue was here?

rodrigogiraoserrao (changes requested):
Don't forget to add and commit the snapshot app `auto_width_input.py` 😆

src/textual/widgets/_input.py
davep: > Forces validation again. This is required as the input may have changed size since last render.

Given that, to the casual reader, the change might look like a nop; perhaps this should be a comment in the code?

willmcgugan: @rodrigogiraoserrao The only difference with width:auto is that the size of the input changes after text is appended, leaving the view position in the wrong place. The `validate_view_position` should always keep the view position in the right place, so I figured i justt wasn't being called. 

@davep It's somewhat of an established pattern now. Not keen on cut and pasting the same comment everywhere.

davep (approved):

rodrigogiraoserrao (approved):

Issue #1471, opened 2023-01-03
[Tree] Consider adding a "node is selected but widget doesn't have focus" styling to `Tree`


davep:
Coming from https://github.com/Textualize/textual/discussions/1375 -- it would be helpful if there was a way to style a `Tree` such that a selected node retains some sort of highlight, even when the `Tree` itself doesn't have focus. It'll be common that a `Tree` will drive the population of some other widget, which might be capable of being focused, and when the user tabs into that widget it would make sense that the `Tree` is still capable of making it clear which node is the selected one.

nitzan-shaked:
I do this today with:

```
#my-tree:focus > .tree--cursor {
    background: $secondary-lighten-1;
}
```

isn't that the intended way?

davep:
In my use of `Tree`, at least with Textual v0.9.1, when the `Tree` loses focus the selected node isn't highlighted in any way. Consider this (deliberately) garish take on the above for my AST viewer widget:

```diff
diff --git a/textual_astview/widgets/astview.py b/textual_astview/widgets/astview.py
index 8389086..6007a6a 100644
--- a/textual_astview/widgets/astview.py
+++ b/textual_astview/widgets/astview.py
@@ -38,6 +38,14 @@ class ASTView( Tree[ Any ] ):
         width: 1fr;
     }

+    ASTView > .tree--cursor {
+        background: red;
+    }
+
+    ASTView:focus > .tree--cursor {
+        background: green;
+    }
+
     .astview--def-name {
         text-style: italic;
         color: #888;
```

Tree with focus:

![Screenshot 2023-01-03 at 12 10 35](https://user-images.githubusercontent.com/28237/210354871-0b85f9f9-6995-4064-9eba-e2c017843834.png)

Tree without focus:

![Screenshot 2023-01-03 at 12 10 40](https://user-images.githubusercontent.com/28237/210354905-fcf31a7d-a74a-48bf-ac79-219201a88580.png)

But even if that did work I feel there's a good case to be made for this to be built into the way that `Tree` works as it would be in keeping with other similar frameworks down (CHUI and GUI) down the years.

mathman1618:
+1 for this.  If another widget causes a new `Tree` node to be selected (via firing an `Event`, say) and the `Tree` is not in focus then the selected node may (will?) scroll into view but it will not be highlighted or otherwise visually indicated.  I think the current behavior is not intuitive ("Why did my tree suddenly scroll when it isn't selected?").

[...]

PR #1703, opened 2023-01-31
Lowkey tree selection

davep:
See #1471. It seems useful and important for a Tree to still have a cursor visible even when it doesn't have focus -- that ways someone can build a UI where the user can tab away from the tree to something that relates to the node, and still see which node is in play.


willmcgugan:
I wonder if we should at some point define some css variables for things like this. Rather than `$secondary-darken` we would have `$cursor-blurred`.

davep:
> I wonder if we should at some point define some css variables for things like this. Rather than `$secondary-darken` we would have `$cursor-blurred`.

This sounds like an excellent idea; issue created.

PR #1706, opened 2023-01-31
Document setting (sub) title in the app guide

### LINKED ISSUE #1577: Document setting title and sub title
We need to cover how to set the title and sub title on an application. This includes the `TITLE` and `SUB_TITLE` class vars, but also the instance attributes `title` and `sub_title`.

This probably belongs in the "App Basics" page in the guide, near the top.

rodrigogiraoserrao:
Fix #1577

### DIFF
--- a/docs/guide/app.md
+++ b/docs/guide/app.md
@@ -38,6 +38,7 @@ If you hit ++ctrl+c++ Textual will exit application mode and return you to the c
 
     A side effect of application mode is that you may no longer be able to select and copy text in the usual way. Terminals typically offer a way to bypass this limit with a key modifier. On iTerm you can select text if you hold the ++option++ key. See the documentation for your terminal software for how to select text in application mode.
 
+
 ## Events
 
 Textual has an event system you can use to respond to key presses, mouse actions, and internal state changes. Event handlers are methods prefixed with `on_` followed by the name of the event.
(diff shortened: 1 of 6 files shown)


davep (approved):

docs/guide/app.md
```diff
@@ -147,7 +148,39 @@ The addition of `[str]` tells mypy that `run()` is expected to return a string.
 
     Type annotations are entirely optional (but recommended) with Textual.
 
-## CSS
+## Customising your app
```
willmcgugan, line 151: Not sure this hierarchy is required. Can we put the following section under "App Class" after the run method.

docs/guide/app.md
```diff
@@ -147,7 +148,39 @@ The addition of `[str]` tells mypy that `run()` is expected to return a string.
...
+
+### Setting the title and subtitle
+
+By default, the title of your application matches the name of the application class and the subtitle is empty.
```
willmcgugan, line 155: It's perhaps not clear to the reader yet what "title" and "subtitle" are. Suggest an introduction something along these line:

"Textual apps have a "title" attribute which is the name of your app, and a "subtitle" attribute which is additional title text that indicates the context of your app. Both values are will be displayed in the apps Header (if there is one)."

willmcgugan (changes requested):
Examples are great. The text could use a little revision.

docs/guide/app.md
willmcgugan: Come to think of it. If you are mentioning compose, it likely should come later.

I think the "Title and Subtitle" should come after CSS. On a top-level heading of its own.

Just noticed another unrelated issue. "Exiting" is under Widgets. But that feels wrong. Could you promote Exiting to a top header.

rodrigogiraoserrao: Addressed in 476eca7.

docs/guide/app.md
```diff
@@ -182,6 +185,41 @@ Here's the question app with classvar CSS:
...
+
+For example, the application shown below changes its title and subtitle as soon as the application is instantiated.
+
+```py title="question_title02.py" hl_lines="19-21"
+--8<-- "docs/examples/app/question_title02.py"
+```
+
+If you run the app shown above and if you press a key, the title and subtitle update accordingly.
+
+For example, if you press `t`, your application will look as shown below:
```
willmcgugan, line 217: We have an extension that renders keys, try replacing the key with `++t++`

willmcgugan (approved):

docs/guide/app.md
rodrigogiraoserrao: Thanks for the reminder o7

Issue #1437, opened 2022-12-24
[Tree] Allow for the complete repopulation of a `Tree`.


davep:
Slightly related to #1396 in that to do it now the ability to change a `TreeNode.label` would be required. At the moment it is possible to use [`Tree.clear`](https://textual.textualize.io/api/tree/#textual.widgets._tree.Tree.clear) to clear all the nodes *under the root node*, but there appears to be no public interface for totally clearing a `Tree` and repopulating it from scratch -- other that performing a `remove` on the widget itself and then a `mount` of a fresh instance.

A use case for this would be if the `Tree` is being used to view the content of a file in a structured way, and you want the user to be able to load in a new file.

willmcgugan:
Maybe add a root label option to `clear()`

PR #1709, opened 2023-02-01
Allow setting a new label when performing a clear on a Tree

davep:
See #1437 for background. While it would be ideal to allow for the complete emptying of a Tree, the root node is required (and it's part of the construction of a Tree). So, here, when clearing the Tree we optionally allow for a new label to be given.

Ideally we'll also allow for fresh data to be provided too; but there's a wrinkle there in knowing the difference between the data being None, and no data being provided (so the current root's data being carried over). Following the method of defaulting used in __init__ would cause problems. As such, rather than roll all of this into one commit, this goes with the basic requirement and the solution for data will follow.

Note this also starts some unit tests for the clearing of a Tree.

### DIFF
--- a/src/textual/widgets/_tree.py
+++ b/src/textual/widgets/_tree.py
@@ -560,12 +560,17 @@ class Tree(Generic[TreeDataType], ScrollView, can_focus=True):
         label = self.render_label(node, NULL_STYLE, NULL_STYLE)
         return label.cell_len
 
-    def clear(self) -> None:
-        """Clear all nodes under root."""
+    def clear(self, label: TextType | None = None) -> None:
+        """Clear all nodes under root.
+
+        Args:
+            label: An optional new label for the root node. If not provided
+                the current root node's label will be used.
+        """
         self._line_cache.clear()
         self._tree_lines_cached = None
         self._current_id = 0
-        root_label = self.root._label
+        root_label = self.root._label if label is None else label
         root_data = self.root.data
         self.root = TreeNode(
             self,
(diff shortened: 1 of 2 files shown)


rodrigogiraoserrao:
I think the behaviour we are implementing makes sense _but_ belongs in a new method, perhaps called `reset` or `repopulate`.
A method `clear` should limit itself to clearing things.

If you pay me to wash your car, you don't expect to come back and see your car painted fresh with a different colour!

src/textual/widgets/_tree.py
```diff
@@ -560,13 +560,20 @@ def get_label_width(self, node: TreeNode[TreeDataType]) -> int:
         label = self.render_label(node, NULL_STYLE, NULL_STYLE)
         return label.cell_len
 
-    def clear(self) -> None:
-        """Clear all nodes under root."""
+    def clear(self, label: TextType | None = None, data: TreeDataType = ...) -> None:
```
rodrigogiraoserrao, line 563: I get the limitation of not being able to clear the root and I think this is so "exceptional" that it should be made obvious from the name of the method, e.g., by renaming it to `clear_children` or `clear_under_root`.

willmcgugan (changes requested):
I like @rodrigogiraoserrao 's idea to have a `clear` method and a `reset`. Where clear did the bare minimum, but `reset` adds label and data. Feels cleaner.

It means that `reset` wouldn't need the ellipsis sentinel.

davep:
Sold.

willmcgugan (approved):

PR #1711, opened 2023-02-01
Make `textual keys` work on older supported Pythons

davep:
`keys.py` wasn't importing `annotations` from the `__future__`.


darrenburns (approved):

willmcgugan:
Thanks. We should really get these apps in to the tests. A snapshot test would at least check it runs on 3.7

PR #1730, opened 2023-02-06
Typing update

willmcgugan:
A while back we created `_typing.py` which had the various additions to the typing module since Python3.7. These were conditional imports depending on `sys.version_info` so we would import from `typing` if available, or `typing_extensions` if not.

Unfortunately PyLance didn't like this, and we (I) lost nice typing warning in the editor.

This change makes `typing_extensions` a requirement even for modern Python's. Imports of Literal, TypeAlias, Protocol etc come directly from there. This appears to keep PyLance happy.


src/textual/css/constants.py
```diff
@@ -2,14 +2,14 @@
 import typing
 
 from ..geometry import Spacing
-from .._typing import Final
 
 if typing.TYPE_CHECKING:
+    from typing_extensions import Final
     from .types import EdgeType
```
davep, line 8: `EdgeType` isn't needed here any more.

src/textual/_animator.py
```diff
@@ -5,19 +5,22 @@
...
+    runtime_checkable,
+)
 
 from . import _clock
 from ._callback import invoke
 from ._easing import DEFAULT_EASING, EASING
 from ._types import CallbackType
 from .timer import Timer
 
-if sys.version_info >= (3, 8):
```
davep, line 16: With the remove of this code the `import sys` higher up isn't needed any more.

src/textual/_animator.py
```diff
@@ -5,19 +5,22 @@
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from functools import partial
-from typing import TYPE_CHECKING, Any, Callable, TypeVar
+
+from typing import Callable
+from typing_extensions import (
```
rodrigogiraoserrao: Shouldn't we import as much as possible from `typing`? E.g., `TYPE_CHECKING` and `Any` are definitely available in `typing`, right?

src/textual/_asyncio.py
```diff
@@ -16,7 +16,7 @@
 else:
     import asyncio
     from asyncio import create_task as _create_task
-    from typing import Awaitable
+    from typing_extensions import Awaitable
```
rodrigogiraoserrao, line 19: Why the change, here? The docs show that `typing.Awaitable` exists since Python 3.5 and it evens says (if I'm reading it right) that it has been deprecated in favour of `collections.abc.Awaitable`. ([source](https://docs.python.org/3/library/typing.html#typing.Awaitable))

examples/five_by_five.py
```diff
@@ -1,24 +1,21 @@
 """Simple version of 5x5, developed for/with Textual."""
 
 from pathlib import Path
-from typing import cast
-import sys
+from typing_extensions import TYPE_CHECKING, cast
```
davep: Testing this with Python 3.11.1, attempting to import `cast` from `typing_extensions` gives me this:

```
Traceback (most recent call last):
  File "/Users/davep/temp/textual/examples/five_by_five.py", line 4, in <module>
    from typing_extensions import TYPE_CHECKING, cast
ImportError: cannot import name 'cast' from 'typing_extensions' (/Users/davep/temp/textual/.venv/lib/python3.11/site-packages/typing_extensions.py)
```

src/textual/_resolve.py
```diff
@@ -1,9 +1,10 @@
 from __future__ import annotations
 
-import sys
 from fractions import Fraction
 from itertools import accumulate
-from typing import cast, Sequence, TYPE_CHECKING
+from typing import Sequence, cast, TYPE_CHECKING
+
+from typing_extensions import Literal
```
rodrigogiraoserrao, line 7: In this PR, I've seen that sometimes you put the "`from typing_extensions import ...`" inside an `if TYPE_CHECKING` and sometimes you don't. What determines which one to go for?

src/textual/_types.py
```diff
@@ -1,13 +1,12 @@
-from typing import Awaitable, Callable, List, TYPE_CHECKING, Union
+from typing import Callable, List, Union
```
rodrigogiraoserrao, line 1: Join the two lines together, no?

src/textual/_types.py
```diff
@@ -1,13 +1,12 @@
...
+from typing import Awaitable, TYPE_CHECKING
```
rodrigogiraoserrao, line 2: Here you import `Awaitable` from `typing` but you've imported it from `typing_extensions` above. (I left a comment there.)

src/textual/css/styles.py
```diff
@@ -4,13 +4,15 @@
 from dataclasses import dataclass, field
 from functools import lru_cache
 from operator import attrgetter
-from typing import TYPE_CHECKING, Any, Iterable, NamedTuple, cast
+from typing import Iterable, cast
+from typing import TYPE_CHECKING, Any, NamedTuple
```
rodrigogiraoserrao, line 8: Join them together?

src/textual/_types.py
```diff
@@ -1,13 +1,12 @@
...
 
-from rich.segment import Segment
```
davep, line 3: Is this removal on purpose? There's reference to `Segment` below and I get a type error for this.

src/textual/_animator.py
rodrigogiraoserrao: I see this has been fixed meanwhile.

src/textual/css/_help_text.py
```diff
@@ -3,10 +3,11 @@
 from dataclasses import dataclass
 from typing import Iterable, Sequence
 
-from textual._typing import Literal
-from textual.color import ColorParseError
-from textual.css._help_renderables import Example, Bullet, HelpText
-from textual.css.constants import (
+from typing_extensions import Literal
```
davep, line 6: Not an issue with these changes, but while scanning through this file, and while we're in typing mode, I noticed a couple of instances of `thing: type = None` in parameters where it should be `thing: type | None=None`. Might be worth tweaking them while we're here. (if not note to self to do this).

davep (approved):
pyright under Emacs is a lot less annoyed with the code after this, so that's a win there tool. Handful of suggested changes but nothing that's a showstopper.

src/textual/_asyncio.py
willmcgugan: The result of some automated search and replacing.

src/textual/css/_help_text.py
willmcgugan: Done.

src/textual/_types.py
willmcgugan: Fixed

src/textual/_resolve.py
willmcgugan: It's not likely intentional, and more likely a side effect of mass editing files. There's not much difference either way, unless the object is used at runtime, and then it can't go inside a `TYPECHECKING` conditional.

FollieHiyuki:
Sorry for busting in a closed PR!

I'm the maintainer of `py3-textual` package on AlpineLinux. As the package is always built with the latest version of Python at the time, we want to minimize useless dependencies like `typing-extentions` in the tool chain.

Since it's mentioned that having `sys.version_info` check for various `typing` imports is tedious, would it be nice for rolling distro like us to have a global option to easily opt out of `typing-extentions` entirely?

willmcgugan:
> would it be nice for rolling distro like us to have a global option to easily opt out of typing-extentions entirely?

How would that global option work?

FollieHiyuki:
My stupid way is just putting this everywhere, so the decision is in the hands of the user (setting the environment variable).
```python
if I_USE_THE_LATEST_PYTHON_VERSION:
    from typing import ...
else:
    from typing_extensions import ...
```
or making all the `import typing/typing_extensions ...` statements generated at build time conditionally.

It sucks, but I'm really hesitant to making suggestion here. This PR indicates that you already had a well-thought conditional import in place, but intentionally removed it.

PR #1737, opened 2023-02-07
typing fixes

willmcgugan:
Reduces typing errors from 172 to 151

Leaving the rest to @rodrigogiraoserrao


docs/guide/widgets.md
```diff
@@ -371,11 +371,11 @@ We can perform this conversion by adding `self.scroll_offset` to `event.offset`.
...
-The `get_square_region` function calculates a [Region][textual.geometry.Region] object for each square and uses them as a positional argument in a call to [refresh][textual.widget.Widget.refresh]. Passing Regions to `refresh` tells Textual to update only the cells underneath those regions, and not the entire region.
+The `get_square_region` function calculates a [Region][textual.geometry.Region] object for each square and uses them as a positional argument in a call to [refresh][textual.widget.Widget.refresh]. Passing Region objects to `refresh` tells Textual to update only the cells underneath those regions, and not the entire widget.
 
 !!! note
 
-    Textual is smart about performing updates. If you refresh multiple regions (even if they overlap), Textual will combine them in to as few non-overlapping regions as possible.
+    Textual is smart about performing updates. If you refresh multiple regions, Textual will combine them in to as few non-overlapping regions as possible.
```
davep, line 378: "in to" should likely be "into".

rodrigogiraoserrao, line 378:
```suggestion
    Textual is smart about performing updates. If you refresh multiple regions, Textual will combine them into as few non-overlapping regions as possible.
```

davep (approved):
Spotted one pre-existing typo but otherwise looks good.

src/textual/color.py
```diff
@@ -447,10 +449,10 @@ def parse(cls, color_text: str | Color) -> Color:
         ) = color_match.groups()
 
         if rgb_hex_triple is not None:
-            r, g, b = rgb_hex_triple
+            r, g, b = rgb_hex_triple  # type: ignore[misc]
```
rodrigogiraoserrao, line 452: Is this because MyPy has no way of knowing that the regex match group can be unpacked into this?

rodrigogiraoserrao (changes requested):

rodrigogiraoserrao (approved):

src/textual/color.py
willmcgugan: It's because Mypy doesn't approve of unpacking strings. Somebody decided that was likely to be an error. Not that I've ever done it accidentally.

PR #1738, opened 2023-02-07
Change emit to post to self.

rodrigogiraoserrao:
The point of this PR is that it makes it simpler to subclass widgets because you can capture messages that would be sent by the parent and repurpose them.

E.g., in the app below, the subclass of `Input` no longer emits `Input.Submitted` events, but rather `MyInput.MyMessage` events.

<details>
<summary>Code</summary>

```py
from textual.message import Message
from textual.widgets import Input, TextLog


class MyInput(Input):
    class MyMessage(Message):
        pass

    def on_input_submitted(self, event: Input.Submitted):
        event.stop()
        self.emit_no_wait(self.MyMessage(self))


class MyApp(App[None]):
        yield MyInput()
        yield TextLog()

    def on_input_submitted(self) -> None:
        self.query_one(TextLog).write("input")

    def on_my_input_my_message(self) -> None:
        self.query_one(TextLog).write("mymessage")


app = MyApp()


if __name__ == "__main__":
    app.run()
```

</details>

### DIFF
--- a/docs/blog/posts/on-dog-food-the-original-metaverse-and-not-being-bored.md
+++ b/docs/blog/posts/on-dog-food-the-original-metaverse-and-not-being-bored.md
@@ -288,7 +288,7 @@ So, thanks to this bit of code in my `Activity` widget...
             parent.move_child(
                 self, before=parent.children.index( self ) - 1
             )
-            self.emit_no_wait( self.Moved( self ) )
+            self.post_messa_no_wait( self.Moved( self ) )
             self.scroll_visible( top=True )
 ```
 
--- a/docs/blog/posts/on-dog-food-the-original-metaverse-and-not-being-bored.md
+++ b/docs/blog/posts/on-dog-food-the-original-metaverse-and-not-being-bored.md
@@ -288,7 +288,7 @@ So, thanks to this bit of code in my `Activity` widget...
             parent.move_child(
                 self, before=parent.children.index( self ) - 1
             )
-            self.post_messa_no_wait( self.Moved( self ) )
+            self.post_message_no_wait( self.Moved( self ) )
             self.scroll_visible( top=True )
 ```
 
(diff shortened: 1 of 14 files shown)


willmcgugan:
It occurs to me that `emit` and `emit_no_wait` are essentially the same as `post_message` and `post_message_no_wait`. Which is a strange state of affairs.

I'm wondering if there is a legitimate need for something which posts to a parent, or we just drop the concept altogether in favor of `post_message`.

rodrigogiraoserrao:
> I'm wondering if there is a legitimate need for something which posts to a parent, or we just drop the concept altogether in favor of `post_message`.

In my mind, I thought that `emit` would make a message go up the DOM and that `post` would send it to a specific widget without bubbling, but now it seems that it is the message itself that either bubbles or not, and that has nothing to do with the method that is used?

rodrigogiraoserrao:
I also think it makes sense to have a shortcut for a widget to post a message to itself, but perhaps it'd be better named `post_to_self` instead of `emit`...

rodrigogiraoserrao:
> But `self.post_message` already does post to self.

Yeah, you are right 🤦 
In that case, to me, it makes sense to get rid of `emit` altogether (or of `post_message` if we prefer the name `emit`).
Let me know and I'll pull the trigger.

willmcgugan:
Give it a go.

I feel sad to get rid of `emit` for some reason. But one method is simpler than 2.

rodrigogiraoserrao:
> Give it a go.
> 
> I feel sad to get rid of `emit` for some reason. But one method is simpler than 2.

Done in bd719e0.

docs/blog/posts/on-dog-food-the-original-metaverse-and-not-being-bored.md
```diff
@@ -288,7 +288,7 @@ So, thanks to this bit of code in my `Activity` widget...
             parent.move_child(
                 self, before=parent.children.index( self ) - 1
             )
-            self.emit_no_wait( self.Moved( self ) )
+            self.post_messa_no_wait( self.Moved( self ) )
```
willmcgugan, line 291: Dear diary, today I found a typo in Rodrigo's code. Today was a good day.

willmcgugan (changes requested):
Just one tiny error. Have you searched the entire code to make sure there aren't any `emit`s hiding in docs or examples?

docs/blog/posts/on-dog-food-the-original-metaverse-and-not-being-bored.md
rodrigogiraoserrao: Fixed in c39b23f.

rodrigogiraoserrao:
> Have you searched the entire code to make sure there aren't any `emit`s hiding in docs or examples?

Yes, I've searched the codebase and the documentation (and the examples wherein).

willmcgugan (approved):

PR #1741, opened 2023-02-08
Revert retrofitting of example code in an old blog post

davep:
Makes sense to update all the docs to reflect the work done in #1738 but I feel it doesn't quite make sense to retrofit this into an old blog post -- especially if the code it is referring to was like that at the time and likely still will be for a wee while after this gets republished.

### DIFF
--- a/docs/blog/posts/on-dog-food-the-original-metaverse-and-not-being-bored.md
+++ b/docs/blog/posts/on-dog-food-the-original-metaverse-and-not-being-bored.md
@@ -288,7 +288,7 @@ So, thanks to this bit of code in my `Activity` widget...
             parent.move_child(
                 self, before=parent.children.index( self ) - 1
             )
-            self.post_message_no_wait( self.Moved( self ) )
+            self.emit_no_wait( self.Moved( self ) )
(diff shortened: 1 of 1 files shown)


rodrigogiraoserrao:
Apologies! I thought it made sense to do the change but if the blog post refers explicitly the textual version it was written for, I think your reversal makes sense.

rodrigogiraoserrao (approved):

willmcgugan (approved):
Makes sense to keep a historical record in the blog.

willmcgugan:
In the unlikely event that confuses people, I suggest we add some kind of admonition. Like "This has changed since the this post was published".

davep:
> Apologies! I thought it made sense to do the change but if the blog post refers explicitly the textual version it was written for, I think your reversal makes sense.

Ahh, no apologies needed; it's an either/or call there I'd say. Your call on it made perfect sense. :-)

davep:
> In the unlikely event that confuses people, I suggest we add some kind of admonition. Like "This has changed since the this post was published". 

I like this. I'll likely do that once it's all settled.

PR #1747, opened 2023-02-09
Reactive polish

willmcgugan:
- Remove no longer used symbols (remenant of a previous version)
- Remove `Reactive.init` as it was barely used. `reactive(init=False)` should be used there. It wasn't documented anywhere.
- General polish.

### DIFF
--- a/src/textual/pilot.py
+++ b/src/textual/pilot.py
@@ -53,6 +53,7 @@ class Pilot(Generic[ReturnType]):
     async def wait_for_scheduled_animations(self) -> None:
         """Wait for any current and scheduled animations to complete."""
         await self._app.animator.wait_until_complete()
+        await wait_for_idle(0)
 
     async def exit(self, result: ReturnType) -> None:
         """Exit the app with the given result.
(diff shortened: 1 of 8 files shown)


src/textual/pilot.py
```diff
@@ -53,6 +53,7 @@ async def wait_for_animation(self) -> None:
     async def wait_for_scheduled_animations(self) -> None:
         """Wait for any current and scheduled animations to complete."""
         await self._app.animator.wait_until_complete()
+        await wait_for_idle(0)
```
willmcgugan, line 56: Unrelated to reactive stuff. This test was flaky. I think because even if the animator is done, there may still be pending messages.

rodrigogiraoserrao (approved):
So, the standard way going forward is to import and use `reactive` instead of `Reactive`?

willmcgugan:
Yeah, apps would typically import `reactive` and / or `var`

rodrigogiraoserrao:
> Yeah, apps would typically import `reactive` and / or `var`

But not widgets, for example? `widget.py`, `_footer.py`, `_header.py`, etc, use a bunch of `Reactive`s.

willmcgugan:
They should probably be updated. `Reactive` will still exist.

PR #1772, opened 2023-02-12
Fixes and polish

willmcgugan:
- Docstrings
- Typing fixes
- Very minor fixes

### DIFF
--- a/src/textual/_wait.py
+++ b/src/textual/_wait.py
@@ -2,7 +2,7 @@ from asyncio import sleep
 from time import monotonic, process_time
 
 SLEEP_GRANULARITY: float = 1 / 50
-SLEEP_IDLE: float = SLEEP_GRANULARITY / 2.0
+SLEEP_IDLE: float = SLEEP_GRANULARITY / 10.0
 
 
 async def wait_for_idle(
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -1266,8 +1266,9 @@ class App(Generic[ReturnType], DOMNode):
         return AwaitMount(self.screen, [])
 
     def install_screen(self, screen: Screen, name: str | None = None) -> AwaitMount:
-        """Install a screen. Installing a screen prevents Textual from destroying it when it
-        is no longer on the screen stack.
+        """Install a screen.
+
+        Installing a screen prevents Textual from destroying it when it is no longer on the screen stack.
 
         Args:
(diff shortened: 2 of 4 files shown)


src/textual/app.py
```diff
@@ -209,7 +209,11 @@ def stop(self) -> None:
         self.join()
 
 
-CSSPathType = Union[str, PurePath, List[Union[str, PurePath]], None]
+CSSPathType = Union[
```
willmcgugan, line 212: Removed the None from the type, as its not really part of the type.

src/textual/_wait.py
```diff
@@ -2,7 +2,7 @@
 from time import monotonic, process_time
 
 SLEEP_GRANULARITY: float = 1 / 50
-SLEEP_IDLE: float = SLEEP_GRANULARITY / 2.0
+SLEEP_IDLE: float = SLEEP_GRANULARITY / 10.0
```
willmcgugan, line 5: This will make `wait_for_idle` potentially sleep a little more, and hopefully reduce the flaky tests even further.

src/textual/app.py
```diff
@@ -348,14 +352,17 @@ def __init__(
...
+            css_paths: List[PurePath]
             if isinstance(css_path, str):
                 css_paths = [Path(css_path)]
             elif isinstance(css_path, PurePath):
                 css_paths = [css_path]
             elif isinstance(css_path, list):
                 css_paths = []
                 for path in css_path:
-                    css_paths.append(Path(path) if isinstance(path, str) else path)
+                    css_paths.append(
```
davep, line 363: I wonder if some of the code around here could be simplified a wee bit given that it's safe to `Path` a `Path`.

src/textual/app.py
```diff
@@ -1259,7 +1266,8 @@ def switch_screen(self, screen: Screen | str) -> AwaitMount:
         return AwaitMount(self.screen, [])
 
     def install_screen(self, screen: Screen, name: str | None = None) -> AwaitMount:
-        """Install a screen.
+        """Install a screen. Installing a screen prevents Textual from destroying it when it
```
davep, line 1269: Perhaps break this up into a couple of lines [to stick to the convention we've adopted to keep the first line a stand-alone description](https://github.com/Textualize/textual/wiki/Documentation-standards#function--method)?

davep (approved):

Issue #1777, opened 2023-02-13
Flesh out API reference for screens.


willmcgugan:
The screens API methods on App are a little too terse. Devs that read the API without the docs are making incorrect assumptions. We should flesh these out a little more. Particularily around the concept of installing and uninstalling screens,

Issue #1776, opened 2023-02-13
App children should only contain the current screen


willmcgugan:
Currently the DOM will contain every Screen on the stack. This means that doing `app.query` considers all screens and may produce unexpected results. 

We should restrict queries to only the current screen.

I think the best way to do this would be for `App` to have a custom `NodeList` object which only ever contains the current screen.

Issue #1775, opened 2023-02-13
install_screen shouldn't add screen to the DOM


willmcgugan:
Calling `install_screen` should add the screen to `_install_screens` but it should not add it to the DOM.

Screens should only be added to the DOM explicitly with `push_screen` or `switch_screen`

PR #1778, opened 2023-02-13
implements screens view

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/1775
Fixes https://github.com/Textualize/textual/issues/1776
Fixes https://github.com/Textualize/textual/issues/1777

This update is mainly for the following:

- Prevents `install_screen` from also mounting the screen.
- Restricts app CSS to a single (current) screen

There are a few other noteworthy changes in the CHANGELOG


tests/test_widget_child_moving.py
```diff
@@ -59,9 +59,9 @@ async def test_widget_move_child() -> None:
             container = Widget(*widgets)
             await pilot.app.mount(container)
             container.move_child(child, before=target)
-            assert container.children[0].id == "widget-1"
-            assert container.children[1].id == "widget-0"
-            assert container.children[2].id == "widget-2"
+            assert container._nodes[0].id == "widget-1"
+            assert container._nodes[1].id == "widget-0"
+            assert container._nodes[2].id == "widget-2"
```
davep, line 64: I see a lot of these in the tests, so picking one at random to ask the question: why have all the tests been changed to use `_nodes` rather than `children`? I see this in the internal code too and I'm assuming there it's just to leave out one level of indirection, just to make the repeated calls a wee bit faster, but if (as it seems to me) `children` is a thin wrapper around `_nodes`, as a public interface, shouldn't the tests be testing the public interface?

willmcgugan: `children` was renamed to `_nodes` in this PR. So these are testing the same things as before.

 I probably should have additional tests for the new `children` attribute.

davep (approved):

PR #1785, opened 2023-02-14
Promote disabled to `Widget` level

davep:
Currently the only interactive widget in Textual that supports `disabled` is the `Button`. Really this should be something that can be set on any interactive widget (`Input`, `Tree`, `DataTable`, etc...). Moreover, ideally it should be something that can be set on any widget and it and all children will be disabled.

This PR seeks to implement this. Key changes include:

- The `disabled` property and related code is removed from `Button`.
- A new `disabled` property is added to `Widget`.
- A new `:enabled` pseudoclass is added to CSS.
- A new `:disabled` pseudoclass is added to CSS.
- A new `_self_or_ancestors_disabled` internal property is added to `Widget`.
- A new `focusable` property is added to `Widget` -- to be `focusable` a widget needs to have `can_focus` set to `True` *and* `_self_or_ancestors_disabled` needs to be `False`.
- A new `watch_disabled` watcher is added to `Widget`, which updates the styles of the current widget and all of its children.
- The `focus_chain` is now built from widgets that are `focusable` rather than those where `can_focus` is `True`.
- `set_focus` has been changed to only allow focus to be set if `focusable` is `True` (before it was if `can_focus` is `True`).
- The forwarding of events now takes `_self_or_ancestors_disabled` state into account when dealing with mouse events (in other words, if a widget is in any way disabled, it won't receive mouse events).
- A default `*:disabled` style is added to `App.DEFAULT_CSS` which applies disabled styling to all disabled widgets.
- A `disabled` keyword argument has been added to all the main "interactive" widgets.

---
If you need a standalone test application for this [see this one here in my sandbox](https://github.com/davep/textual-sandbox/blob/main/disable_tester.py).


src/textual/widget.py
```diff
@@ -2080,6 +2098,7 @@ def get_pseudo_classes(self) -> Iterable[str]:
             Names of the pseudo classes.
 
         """
+        yield "disabled" if self._self_or_ancestors_disabled else "enabled"
```
willmcgugan, line 2101: For the sake of micro-optimization I would unroll this code. I.e. make fewer implicit function calls.

src/textual/widget.py
```diff
@@ -2127,11 +2146,16 @@ def post_render(self, renderable: RenderableType) -> ConsoleRenderable:
...
+            self._update_styles()
 
     def watch_has_focus(self, value: bool) -> None:
         """Update from CSS if has focus state changes."""
-        self.app.update_styles(self)
+        self._update_styles()
+
+    def watch_disabled(self) -> None:
+        """Update the styles of the widget and its children when disabled is toggled."""
+        for node in self.walk_children(with_self=True):
```
willmcgugan, line 2157: I don't think this is necessary as `app.update_styles` will also update styles for its children.

willmcgugan (commented):
Looks good, some minor requests

src/textual/widget.py
rodrigogiraoserrao: What does this mean? Turn the conditional expression into the `if` below?

```py
if self._self_or_ancestors_disabled:
    yield "disabled"
else:
    yield "enabled"
```

If so, why is that a micro-optimisation?

willmcgugan: By unrolling I mean copy the code from `self._self_or_ancestors_disabled` to here rather than the property, and repeat so that there are fewer calls.

Function calls in Python are relatively expensive, and this method is called a lot.

rodrigogiraoserrao: Thanks for explaining.

src/textual/widget.py
```diff
@@ -1180,6 +1180,8 @@ def virtual_region_with_margin(self) -> Region:
     @property
     def _self_or_ancestors_disabled(self) -> bool:
         """Is this widget or any of its ancestors disabled?"""
+        # NOTE: Please see the copy of this code in get_pseudo_classes. I
```
rodrigogiraoserrao, line 1183:
```suggestion
        # NOTE: Please see the copy of this code in get_pseudo_classes. If
```

src/textual/widget.py
```diff
@@ -2098,7 +2100,16 @@ def get_pseudo_classes(self) -> Iterable[str]:
...
+        # NOTE: The heart of this yield is a direct copy of
+        # _self_or_ancestors_disabled. Because this method is called so
+        # much, here we save one function call as a very small but long-term
+        # useful optimisation. If _self_or_ancestors_disabled ever changes,
+        # be sure to reflect that change here!
+        yield "disabled" if any(
+            node.disabled
+            for node in self.ancestors_with_self
+            if isinstance(node, Widget)
+        ) else "enabled"
```
rodrigogiraoserrao, line 2112: @willmcgugan would something like `any(getattr(node, "disabled", False) for node in self.ancestors_with_self)` be potentially any faster? It's still one function call but now we don't use the dotted attribute access...
I'm trying to figure out these micro-optimisation heuristics.

davep: My gut reaction is that the attribute access makes more sense.

```py
import dis

def via_prop():
    return ( 1 ).__class__

def via_getattr():
    return getattr( 1, "__class__", "unknown" )

print( "via_prop" )
print( "=" * 40 )
dis.dis( via_prop )

print( "via_getattr" )
print( "=" * 40 )
dis.dis( via_getattr )
```

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+### Added
+
+- Added `Widget.disabled` https://github.com/Textualize/textual/pull/1785
```
rodrigogiraoserrao, line 12: Is it worth adding something under "Changed" about `Button.disabled` or do you think it makes no sense?

davep: I don't think it really makes sense because there is still a `Button.disabled`.

src/textual/widgets/_button.py
```diff
@@ -39,11 +39,6 @@ class Button(Static, can_focus=True):
         text-style: bold;
     }
 
-    Button.-disabled {
-        opacity: 0.4;
-        text-opacity: 0.7;
-    }
-
```
rodrigogiraoserrao, line 46: I'd expect to either rewrite this rule with the new pseudoclass _or_ to define the default styling `*:disabled` under `App.DEFAULT_CSS` to use these values.
With this change plus the default CSS under `App`, buttons are now styled differently when disabled, aren't they?

davep: There is a very subtle difference in the disabled style of `Button`, yes. @willmcgugan ask that they be lifted ever so slightly, along with everything else that can now be disabled. That is... the `Button` style was hoisted up to `App`, applied to `*` with `:disabled`, and also tweaked ever so slightly.

src/textual/widget.py
```diff
@@ -2421,6 +2455,18 @@ def release_mouse(self) -> None:
         """
         self.app.capture_mouse(None)
 
+    def check_message_enabled(self, message: Message) -> bool:
```
rodrigogiraoserrao, line 2458: `MessagePump.check_message_enabled` is missing a docstring and so is this method, so we get no docstring here.
Not sure if you feel like adding a docstring to `MessagePump` _but_ I think it's worth adding a docstring here.

src/textual/widgets/_data_table.py
```diff
@@ -473,8 +473,9 @@ def __init__(
         name: str | None = None,
         id: str | None = None,
         classes: str | None = None,
+        disabled: bool = False,
```
rodrigogiraoserrao, line 476: Other widgets received a docstring line about this new attribute but I see none was added for the `DataTable`. Is it because `DataTable` is missing the docstring for its attributes?

src/textual/widgets/_static.py
```diff
@@ -56,8 +56,9 @@ def __init__(
         name: str | None = None,
         id: str | None = None,
         classes: str | None = None,
+        disabled: bool = False,
```
rodrigogiraoserrao, line 59: Ditto here about the docstring.

src/textual/widgets/_text_log.py
```diff
@@ -43,8 +43,9 @@ def __init__(
         name: str | None = None,
         id: str | None = None,
         classes: str | None = None,
+        disabled: bool = False,
```
rodrigogiraoserrao, line 46: Ditto here about the docstring.

src/textual/widgets/_tree.py
```diff
@@ -473,8 +473,9 @@ def __init__(
         name: str | None = None,
         id: str | None = None,
         classes: str | None = None,
+        disabled: bool = False,
```
rodrigogiraoserrao, line 476: Ditto here about the docstring.
(We may end up realising it suffices to document once in `Widget`. I don't know by heart.)

tests/snapshot_tests/snapshot_apps/disable_widgets.py
```diff
@@ -0,0 +1,84 @@
...
+        height: 2;
+    }
+
+    Markdown, MarkdownViewer {
+        height: 1fr;
+    }
+    """
+
+    @property
+    def data_table(self) -> DataTable:
```
rodrigogiraoserrao, line 35: This is an interesting use of `@property`! In an actual app, this would return a new `DataTable` each time you access it, right? So was this a "quick & dirty" way of separating the creation of the widgets from their `yield`ing in `compose` or am I missing something?

tests/test_disabled.py
```diff
@@ -0,0 +1,84 @@
...
+    Markdown,
+    MarkdownViewer,
+    Switch,
+    TextLog,
+    Tree,
+)
+
+
+class DisableApp(App[None]):
+    """Application for testing Widget.disable."""
```
rodrigogiraoserrao, line 20:
```suggestion
    """Application for testing Widget.disabled."""
```

rodrigogiraoserrao (commented):
I added a bunch of questions but I found no blockers.

src/textual/widgets/_data_table.py
davep: Yeah, I think `DataTable` might need a last round of docstring review so I wasn't going to make it part of this PR.

src/textual/widgets/_tree.py
davep: > We may end up realising it suffices to document once in Widget. 

Yeah, that's a tricky one isn't it? Often the parameters can just be "inherited", as it were, and I've see autodoc systems that take that into account. On the other hand some of the widgets "inherit" almost all of the parameters but add one or two of their own, which suggests needing to repeat everything. :-/

davep: Oh, and yeah, `Tree` seems to be lacking `__init__` documentation on the whole, so I'll leave that to a followup PR as I don't really like turning a focused PR into a more general one if I can help it.

tests/snapshot_tests/snapshot_apps/disable_widgets.py
davep: > In an actual app, this would return a new DataTable each time you access it, right?

Correct. It's just a way of keeping the main `compose` tidy. Think of it more as a "macro" than a property for the purposes of generating the snapshot. It'd be a horrible idea for a proper app.

src/textual/widgets/_text_log.py
davep: Same as with `Tree` here. I'm loathe to turn a PR about adding `disabled` into a general "add more docs" PR too. Once this one is in I think it'll make sense to do another "hunt down missing docstrings" roundup.

src/textual/widget.py
willmcgugan: @rodrigogiraoserrao `getattr` is a function itself, and a global which tends to be slow. Attribute is typically faster than globals.

@davep You can keep unrolling here. Unroll `ancestors_with_self`. There's some code a few lines down which does something similar.

davep:
Re: the docstrings that are missing in total, I've added #1811.

src/textual/widget.py
davep: I think I'll also leave this for a subsequent round of "add more documentation", rather than make this PR about adding previously-missing docs.

davep: Added to #1811 as a reminder.

src/textual/widget.py
davep: > You can keep unrolling here. Unroll ancestors_with_self. 

Done.

src/textual/widget.py
rodrigogiraoserrao: Right, but the point was that even if you document the super class's method, this method probably needs an extended docstring either way, no?

rodrigogiraoserrao (approved):

src/textual/widget.py
davep: I think I'd need to see the proper documentation for the super-method to decide of the additional test that happens in `Widget` is in or out of the scope, and either way I'd need to see the documentation for the super-method to know how this one should best be documented (hence a note to document it here too in #1811).

willmcgugan (approved):

PR #1786, opened 2023-02-14
DataTable - make `data` attribute private & expand APIs for reading data

darrenburns:

- [ ] Updated documentation


src/textual/widgets/_data_table.py
```diff
@@ -655,6 +657,84 @@ def get_cell_at(self, coordinate: Coordinate) -> CellType:
...
+            A list of the values contained within the row.
+
+        Raises:
+            RowDoesNotExist: When there is no row corresponding to the key.
+        """
+        if row_key not in self._row_locations:
+            raise RowDoesNotExist(f"Row key {row_key!r} is not valid.")
+        cell_mapping: dict[ColumnKey, CellType] = self._data.get(row_key, {})
+        ordered_row: list[CellType] = []
+        for column in self.ordered_columns:
```
willmcgugan, line 676: Could we do this in a list expression? Highly unlikely there will be enough columns to make a difference, but you never know.

src/textual/widgets/_data_table.py
```diff
@@ -655,6 +657,84 @@ def get_cell_at(self, coordinate: Coordinate) -> CellType:
...
+        Returns:
+            A generator which yields the cells in the column.
+
+        Raises:
+            ColumnDoesNotExist: If there is no column corresponding to the key.
+        """
+        if column_key not in self._column_locations:
+            raise ColumnDoesNotExist(f"Column key {column_key!r} is not valid.")
+
+        for row_metadata in self.ordered_rows:
```
willmcgugan, line 715: This may be expensive. I list expression would be worthwhile, and that `self._data` could be pulled outside of the loop.

willmcgugan (changes requested):

willmcgugan (approved):

Issue #1729, opened 2023-02-06
Docstrings for actions


willmcgugan:
Some action method (i.e. methods that start with `action_`) don't have docstrings, or insufficient docstrings. We should add docstrings to all action methods and review existing actions for more detail.

PR #1790, opened 2023-02-14
Tidy up various docstrings in ListView

davep:
Adding some actions docstrings (#1729) and also tweaking a couple of others to better conform with our docstring standard.

### DIFF
--- a/src/textual/widgets/_list_view.py
+++ b/src/textual/widgets/_list_view.py
@@ -13,8 +13,10 @@ from textual.widgets._list_item import ListItem
 
 
 class ListView(Vertical, can_focus=True, can_focus_children=False):
-    """Displays a vertical list of `ListItem`s which can be highlighted
-    and selected using the mouse or keyboard.
+    """A vertical list view widget.
+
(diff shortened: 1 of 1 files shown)


src/textual/widgets/_list_view.py
```diff
@@ -155,13 +162,16 @@ def clear(self) -> AwaitRemove:
...
+        """Select the current item in the list."""
         selected_child = self.highlighted_child
         self.post_message_no_wait(self.Selected(self, selected_child))
 
     def action_cursor_down(self) -> None:
+        """Select the next item in the list."""
         self.index += 1
 
     def action_cursor_up(self) -> None:
+        """Select the previous item in the list."""
```
darrenburns, line 174: These docstrings are a little misleading. Pressing up/down doesn't select the previous/next items, it highlights them.

darrenburns (changes requested):
One minor change requested.

src/textual/widgets/_list_view.py
davep: Good catch, poor wording on my part.

darrenburns (approved):

PR #1821, opened 2023-02-17
Mark the Markdown viewer as done on the roadmap

davep:
Also add a couple of sub-items that are intended to be added.


rodrigogiraoserrao (approved):
Looks good to me. I'm particularly fond of the fact that you ordered the two sub-items alphabetically. There was only a 50% chance of getting it right if you did it randomly!

willmcgugan (approved):

PR #1823, opened 2023-02-17
Optimize scroll with a Spatial Map

willmcgugan:
- Implements a "SpatialMap" to efficiently narrow down which widgets *may* be visible.
- Adds a new scrolling mechanism, so that scrolling can use this newer fast-path.
- Changes page up / down etc to add to the current scroll position, to prevent scrolling to far when holding down the button.


CHANGELOG.md
```diff
@@ -5,7 +5,18 @@ All notable changes to this project will be documented in this file.
...
+### Changed
+
+- Scrolling by page now adds to current position.
+
+### Removed
+
+- Removed `screen.visible_widgets` and `screen.widgets`
+
 ## [0.11.1] - 2023-02-17
+>>>>>>> main
```
davep, line 19: Bit of a merge conflict hangover here by the looks of things.

src/textual/_compositor.py
```diff
@@ -322,7 +376,7 @@ def visible_widgets(self) -> dict[Widget, tuple[Region, Region]]:
         return self._visible_widgets
 
     def _arrange_root(
-        self, root: Widget, size: Size
+        self, root: Widget, size: Size, visible_only: bool = True
     ) -> tuple[CompositorMap, set[Widget]]:
         """Arrange a widgets children based on its layout attribute.
```
rodrigogiraoserrao, line 381:
```suggestion
        """Arrange a widget's children based on its layout attribute.
```

src/textual/_spatial_map.py
```diff
@@ -0,0 +1,94 @@
...
+from typing import Generic, Iterable, TypeVar
+
+from .geometry import Region
+
+ValueType = TypeVar("ValueType")
+
+
+class SpatialMap(Generic[ValueType]):
+    """A spatial map allows for data to be associated with a rectangular regions
+    in Euclidean space, and efficiently queried.
```
rodrigogiraoserrao, line 14: This could be improved because this docstring doesn't do justice to what you explained directly to me. 

My suggestion would be to consider writing a first, short sentence that gives a very high-level description of what this class is/represents, e.g., `"Provide fast spatial querying for data arranged in a Euclidean space."`.

Then, elaborate on what you mean by "allows for data to be associated with a rectangular region in Euclidean space". What type of association are you talking about?

On top of that, this is a "private" module, which means it is highly unlikely that we will write further documentation for this other than the one we leave in the docstrings, so it is a good idea to make it as comprehensive as possible.

Also, is this related/similar to [quadtrees](https://en.wikipedia.org/wiki/Quadtree)?

src/textual/_spatial_map.py
```diff
@@ -0,0 +1,94 @@
...
+            grid_width: Width of a grid square.
+            grid_height: Height of a grid square.
+        """
+        self._grid_size = (grid_width, grid_height)
+        self.total_region = Region()
+        self._map: defaultdict[tuple[int, int], list[ValueType]] = defaultdict(list)
+        self._fixed: list[ValueType] = []
+
+    def _region_to_grid(self, region: Region) -> Iterable[tuple[int, int]]:
+        """Get the grid squares under a region.
```
rodrigogiraoserrao, line 31: Are these the grid squares that are completely within the region or the ones that overlap with it?

src/textual/_spatial_map.py
```diff
@@ -0,0 +1,94 @@
...
+
+        Args:
+            region: A region.
+
+        Returns:
+            Iterable of grid squares (tuple of 2 values).
```
rodrigogiraoserrao, line 37: These "grid squares" that are returned... Is this actually the position of the squares in the grid? The return value looks like the "coordinates of the grid squares under/intersecting with the region".

src/textual/_spatial_map.py
```diff
@@ -0,0 +1,94 @@
...
+
+        return product(
+            range(x1 // grid_width, x2 // grid_width + 1),
+            range(y1 // grid_height, y2 // grid_height + 1),
+        )
+
+    def insert(
+        self, regions_and_values: Iterable[tuple[Region, bool, ValueType]]
+    ) -> None:
+        """Insert values in to the Spatial map.
```
rodrigogiraoserrao, line 54:
```suggestion
        """Insert values into the Spatial map.
```

src/textual/_spatial_map.py
```diff
@@ -0,0 +1,94 @@
...
+        y2 = y1 + height - 1
+        grid_width, grid_height = self._grid_size
+
+        return product(
+            range(x1 // grid_width, x2 // grid_width + 1),
+            range(y1 // grid_height, y2 // grid_height + 1),
+        )
+
+    def insert(
+        self, regions_and_values: Iterable[tuple[Region, bool, ValueType]]
```
rodrigogiraoserrao, line 52: cf. the docstring section in `Args`. Might it be worth creating a type alias for this?

src/textual/_spatial_map.py
```diff
@@ -0,0 +1,94 @@
...
+        Args:
+            grid_width: Width of a grid square.
+            grid_height: Height of a grid square.
+        """
+        self._grid_size = (grid_width, grid_height)
+        self.total_region = Region()
+        self._map: defaultdict[tuple[int, int], list[ValueType]] = defaultdict(list)
+        self._fixed: list[ValueType] = []
+
+    def _region_to_grid(self, region: Region) -> Iterable[tuple[int, int]]:
```
rodrigogiraoserrao, line 30: Might it be worth creating a type alias for `Coordinate` or `Position` or `Coord2D` or something of the sort? I think such an alias gives _much_ more context than just `tuple[int, int]`.

src/textual/_spatial_map.py
```diff
@@ -0,0 +1,94 @@
...
+            range(y1 // grid_height, y2 // grid_height + 1),
+        )
+
+    def insert(
+        self, regions_and_values: Iterable[tuple[Region, bool, ValueType]]
+    ) -> None:
+        """Insert values in to the Spatial map.
+
+        Values are associated with their region in Euclidean space, and a boolean that
+        indicates fixed regions. Fixed regions don't scroll and are always visible.
```
rodrigogiraoserrao, line 57: So, if fixed regions don't scroll and are always visible, what does that actually mean in practice? Maybe this behaviour needs documenting here and/or under the class docstring?

src/textual/_spatial_map.py
```diff
@@ -0,0 +1,94 @@
...
+        for region, fixed, value in regions_and_values:
+            total_region = total_region.union(region)
+            if fixed:
+                append_fixed(value)
+            else:
+                for grid in _region_to_grid(region):
+                    get_grid_list(grid).append(value)
+        self.total_region = total_region
+
+    def get_values_in_region(self, region: Region) -> list[ValueType]:
```
rodrigogiraoserrao, line 75: Do we want to commit to returning specifically a `list`? Seems oddly specific!
It may follow naturally from the current implementation, but are there any advantages in ensuring this is a `list`?

src/textual/_spatial_map.py
```diff
@@ -0,0 +1,94 @@
...
+        """Get a set of values that are under a given region.
```
rodrigogiraoserrao, line 76: I would suggest something like what you can see below:

```suggestion
        """Get a superset of all the values that intersect with a given region.
```

The fact that it may return false positives sounds too important to not be covered in the first sentence.

src/textual/_spatial_map.py
```diff
@@ -0,0 +1,94 @@
...
+        Returns:
+            Values under the region.
+        """
+        results: list[ValueType] = self._fixed.copy()
+        add_results = results.extend
+        get_grid_values = self._map.get
+        for grid in self._region_to_grid(region):
+            grid_values = get_grid_values(grid)
+            if grid_values is not None:
+                add_results(grid_values)
```
rodrigogiraoserrao, line 92: I thought the spatial map was a big rectangle divided into sub-rectangles (or specifically, squares?) by means of a grid. Thus, I find the name `grid` misleading. The variable `grid` seems to refer to the whole thing instead of just a single `grid_square` or `grid_cell` or `cell` or `square`, which seems to be what it actually refers to?

If `._region_to_grid` is typed with an alias, this may warrant another name, like `coord` or `grid_coord`.

src/textual/_spatial_map.py
willmcgugan: Yeah, I've renamed it to reflect that.

src/textual/messages.py
```diff
@@ -45,12 +45,24 @@ def can_replace(self, message: Message) -> bool:
...
+class UpdateScroll(Message, verbose=True):
+    """Sent by Textual when a scroll update is required."""
+
+    def can_replace(self, message: Message) -> bool:
+        return isinstance(message, UpdateScroll)
+
+
 @rich.repr.auto
 class InvokeLater(Message, verbose=True, bubble=False):
+    """Sent by Textual to invoke a callback."""
```
rodrigogiraoserrao, line 64: Take a look at how `Button.Pressed` is documented, for example, and consider following that format. That will make it a breeze to generate the online documentation for the messages.

https://github.com/Textualize/textual/blob/main/src/textual/widgets/_button.py#L162-L170

willmcgugan: I'll come back to that. These should really be private. Or at least something to indicated to devs you shouldn't touch them.

src/textual/_spatial_map.py
willmcgugan: Ok, `grid_coordinate` it is.

tests/test_spatial_map.py
```diff
@@ -0,0 +1,56 @@
...
+            (Region(10, 5, 5, 5), False, "foo"),
+            (Region(5, 20, 5, 5), False, "bar"),
+        ]
+    )
+
+    assert spatial_map.get_values_in_region(Region(0, 0, 10, 5)) == ["foo"]
+    assert spatial_map.get_values_in_region(Region(0, 1, 10, 5)) == ["foo"]
+    assert spatial_map.get_values_in_region(Region(0, 10, 10, 5)) == []
+    assert spatial_map.get_values_in_region(Region(0, 20, 10, 5)) == ["bar"]
+    assert spatial_map.get_values_in_region(Region(5, 5, 50, 50)) == ["foo", "bar"]
```
rodrigogiraoserrao, line 56: I would extend these tests by inserting more data with complex data arrangements, e.g., with intersections or regions contained within others. Possibly in a new test.

rodrigogiraoserrao, line 56: We're missing tests with fixed regions.

rodrigogiraoserrao (commented):
I don't feel confident enough regarding any of the changes other than the implementation of the spatial map. I left a series of comments that address the feeling one of us will have when they go debug something six months from now.

src/textual/_spatial_map.py
willmcgugan: It doesn't need to be a list, but the return type complements another value used in the caller context which is a list.

Since this isn't a base class or a protocol, and internal anyway, there probably isn't much to be gained from restricting the behaviour of the return type.

src/textual/_spatial_map.py
willmcgugan: It's fixed like CSS "fixed". Or in Textual terms when you dock, and things stay where they are when you scroll.

src/textual/_spatial_map.py
willmcgugan: I wouldn't have bothered personally. Sometimes indirection makes things harder to read IMHO. But maybe this is not one of those times. I've added a `GridCoordinate` type alias.

src/textual/_spatial_map.py
willmcgugan: The indirection means you would have to go look for the type alias, and since its only used once, I don't think it is worth it...

tests/test_spatial_map.py
willmcgugan: Added some tests with fixed.

willmcgugan: It wouldn't actually go down any new code paths. Not disavowing tests, but there is little benefit in adding more now. And the snapshots will run this code multiple times.

src/textual/_spatial_map.py
```diff
@@ -24,17 +33,17 @@ def __init__(self, grid_width: int = 100, grid_height: int = 20) -> None:
         """
         self._grid_size = (grid_width, grid_height)
         self.total_region = Region()
-        self._map: defaultdict[tuple[int, int], list[ValueType]] = defaultdict(list)
+        self._map: defaultdict[GridCoordinate, list[ValueType]] = defaultdict(list)
         self._fixed: list[ValueType] = []
 
-    def _region_to_grid(self, region: Region) -> Iterable[tuple[int, int]]:
+    def _region_to_grid_coordinate(self, region: Region) -> Iterable[GridCoordinate]:
```
rodrigogiraoserrao, line 39: Plural? As I understand it, multiple coordinates can be returned, right?

```suggestion
    def _region_to_grid_coordinates(self, region: Region) -> Iterable[GridCoordinate]:
```

src/textual/screen.py
```diff
@@ -9,6 +9,7 @@
 from . import errors, events, messages
 from ._callback import invoke
 from ._compositor import Compositor, MapGeometry
+from ._profile import timer
```
rodrigogiraoserrao, line 12: Did this leak from when you were profiling this?

src/textual/widgets/_placeholder.py
```diff
@@ -61,10 +62,10 @@ class Placeholder(Widget):
         overflow: hidden;
         color: $text;
     }
-
     Placeholder.-text {
         padding: 1;
     }
+
```
rodrigogiraoserrao, line 68: Why add a hanging empty line? Was this an artifact of some automated tool?

I've seen many empty lines hanging elsewhere, e.g., at the end of docstrings, and I don't get where they come from. 🤷

src/textual/widgets/_placeholder.py
```diff
@@ -147,6 +148,6 @@ def on_click(self) -> None:
 
     def on_resize(self, event: events.Resize) -> None:
         """Update the placeholder "size" variant with the new placeholder size."""
-        self._renderables["size"] = self._SIZE_RENDER_TEMPLATE.format(*self.size)
+        self._renderables["size"] = self._SIZE_RENDER_TEMPLATE.format(*event.size)
         if self.variant == "size":
-            self.refresh(layout=True)
+            self.refresh(layout=False)
```
rodrigogiraoserrao, line 153: This is interesting! IIRC, when I tried implementing the placeholder, it wouldn't work properly if I didn't ask it to do a layout as well. I probably got it wrong. 😛

rodrigogiraoserrao (approved):

src/textual/widgets/_placeholder.py
willmcgugan: Probably just the result of an edit that was backed out of. Black will add new lines in some places.

src/textual/widgets/_placeholder.py
willmcgugan: Doesn't seem to be necessary now. Possibly it was a workaround for a bug which is now fixed.

src/textual/_layout.py
```diff
@@ -11,7 +13,52 @@
     from .widget import Widget
 
 ArrangeResult: TypeAlias = "tuple[list[WidgetPlacement], set[Widget]]"
-DockArrangeResult: TypeAlias = "tuple[list[WidgetPlacement], set[Widget], Spacing]"
+
+
+@dataclass
+class DockArrangeResult:
+    placements: list[WidgetPlacement]
```
davep, line 20: Docstrings for the class and the fields?

src/textual/messages.py
```diff
@@ -45,12 +45,24 @@ def can_replace(self, message: Message) -> bool:
...
+
     def can_replace(self, message: Message) -> bool:
         return isinstance(message, Layout)
 
 
+@rich.repr.auto
+class UpdateScroll(Message, verbose=True):
+    """Sent by Textual when a scroll update is required."""
+
+    def can_replace(self, message: Message) -> bool:
```
davep, line 58: Missing docstring.

src/textual/screen.py
```diff
@@ -484,6 +511,12 @@ async def _on_layout(self, message: messages.Layout) -> None:
         self._layout_required = True
         self.check_idle()
 
+    async def _on_update_scroll(self, message: messages.UpdateScroll) -> None:
```
davep, line 513: Missing docstring.

src/textual/_spatial_map.py
willmcgugan: Overlap

davep (approved):

Issue #1822, opened 2023-02-17
Fix typing for MyPy


rodrigogiraoserrao:
With the API more stable, we should get Textual passing Mypy.

There are currently 148 errors, but many of them have the same root cause.

It may require a few passes to get all of them. Suggest spending a day on fixing typing, and we will review the situation after.

rodrigogiraoserrao:
Follow-up: see if my suggestion in #1813 breaks typing.

rodrigogiraoserrao:
_Edit_: By the time I picked this up, there were 85 errors.

@darrenburns two issues read:
```
src/textual/_two_way_dict.py:41: error: Incompatible return value type (got "Optional[Value]", expected "Value")  [return-value]
src/textual/_two_way_dict.py:52: error: Incompatible return value type (got "Optional[Key]", expected "Key")  [return-value]
```

To guide my fix, I need to know if the rationale here is that the methods `.get` and `.get_key` never fail?
Or are they supposed to be mirroring `dict.get`? In which case, we probably need to restore the `default: T | None = None` annotation and fix typing in the correct way.

rodrigogiraoserrao:
These have been axed down to around 30 in Python 3.7.
Most of them come from the widget Data Table and interactions with the `TwoWayDict` implemented in `_two_way_dict`. The PR above contains a possible approach but this needs to be validated with @darrenburns.

willmcgugan:
Typing is an ongoing thing. We will have another pass soon.

PR #1831, opened 2023-02-17
Make mypy happy

rodrigogiraoserrao:
This is a draft because I'm still axing issues, but each commit corresponds to one fix (either a single problem or tightly related problems).

Feel free to review each commit individually and let me know if/where I dropped the ball.


rodrigogiraoserrao:
eeeac1a fixes the typing of `fromkeys` with

```py
fromkeys = cast("Callable[[list[int]], dict[int, Strip | None]]", dict.fromkeys)
chops: list[dict[int, Strip | None]]
chops = [fromkeys(cut_set[:-1]) for cut_set in cuts]
```

I feel like a purer solution would be to use a dictionary comprehension:

```py
chops: list[dict[int, Strip | None]]
chops = [{key: None for key in cut_set[:-1]} for cut_set in cuts]
```

This seems to be around 10%-15% slower:

```py
>>> timeit.repeat("{k: None for k in range(5)}")
[0.2364686660002917, 0.2092778330006695, 0.20847449999928358, 0.20909275000121852, 0.20904904100098065]
>>> timeit.repeat("dict.fromkeys(range(5))")
[0.20496962499964866, 0.1807063329997618, 0.1800928329994349, 0.1798485420003999, 0.17970204199991713]
>>> timeit.repeat("{k: None for k in range(500)}", number=10_000)
[0.1311742920006509, 0.10960308299945609, 0.10958245799884025, 0.11009866600034002, 0.10938712500137626]
>>> timeit.repeat("dict.fromkeys(range(500))", number=10_000)
[0.112881875000312, 0.08676441699935822, 0.0867214169993531, 0.08729300000049989, 0.08724679199985985]
```

rodrigogiraoserrao:
TODO: Go back to 0ae4633 and see if I should raise an error if `token is None`. Given the context in which we're in (because `token.name == "variable_name"` in the outer scope), we probably can't get away without raising an error if there are no tokens left.

src/textual/_compositor.py
```diff
@@ -45,12 +45,23 @@ class ReflowResult(NamedTuple):
...
 
-    region: Region  # The (screen) region occupied by the widget
-    order: tuple[tuple[int, ...], ...]  # A tuple of ints defining the painting order
-    clip: Region  # A region to clip the widget by (if a Widget is within a container)
-    virtual_size: Size  # The virtual size  (scrollable region) of a widget if it is a container
-    container_size: Size  # The container size (area not occupied by scrollbars)
-    virtual_region: Region  # The region relative to the container (but not necessarily visible)
+    region: Region
+    """The (screen) region occupied by the widget."""
+    order: tuple[tuple[int, int, int], ...]
```
willmcgugan, line 50: Is this correct? I have a feeling the outer tuple can contain both tuples and ints.

rodrigogiraoserrao: That _became_ correct. The commit that introduced that change also tweaked a couple of other lines of code to make sure that was correct and it includes a commit message that explains why that is a good fix. (From my point of view.)

rodrigogiraoserrao:
5 issues remain. 3 of them depend on the resolution of #1870.

rodrigogiraoserrao:
On Python 3.7, 9 issues remain.
On Python 3.11, only 2.

src/textual/_compositor.py
```diff
@@ -419,19 +430,23 @@ def add_widget(
...
             clip: Region,
             visible: bool,
-            _MapGeometry=MapGeometry,
+            _MapGeometry: type[MapGeometry] = MapGeometry,
         ) -> None:
             """Called recursively to place a widget and its children in the map.
 
             Args:
                 widget: The widget to add.
+                virtual_region: TODO.
```
willmcgugan, line 443: Suggest "The Widget region relative to it's container."

src/textual/actions.py
```diff
@@ -3,6 +3,11 @@
 import ast
 import re
 
+from typing_extensions import Any, TypeAlias
+
+Action: TypeAlias = "tuple[str, tuple[Any, ...]]"
```
willmcgugan, line 8: I'm ambivalent about the name "Action". It's not an action per se, but the result of parsing an encoded action. `"ActionParseResult"` would be more apt, if verbose.

src/textual/app.py
```diff
@@ -1608,9 +1608,8 @@ async def invoke_ready_callback() -> None:
                                 with redirect_stdout(redirector):  # type: ignore
                                     await run_process_messages()
                         else:
-                            null_file = _NullFile()
-                            with redirect_stderr(null_file):
-                                with redirect_stdout(null_file):
+                            with redirect_stderr(None):
```
willmcgugan, line 1611: I was surprised by this! I can't find it documented that redirecting to `None` works.

src/textual/app.py
```diff
@@ -1735,13 +1734,16 @@ def _register(
         if not widgets:
             return []
 
-        new_widgets = list(widgets)
+        widget_list = list(widgets)
```
willmcgugan, line 1737: This isn't great. But the original was problematic.

It is doing a completely superfluous conversion from tuple to list. Suggest that we don't need `widget_list`, and that `new_widgets` should be either `reversed(widgets)` or just `widgets`. And type as an iterable.

src/textual/app.py
```diff
@@ -1801,6 +1803,7 @@ def is_mounted(self, widget: Widget) -> bool:
     async def _close_all(self) -> None:
         """Close all message pumps."""
 
+        screen: Screen | Callable[[], Screen]  # Explicit type to reuse variable below.
```
willmcgugan, line 1806: I'm surprised this doesn't blow up. `self._screen_stack` is a list of `Screens` only.

`self.SCREENS` can also contain callables, so I think that should be a new variable and typed differently.

src/textual/css/styles.py
```diff
@@ -581,7 +584,7 @@ def partial_rich_style(self) -> Style:
 @dataclass
 class Styles(StylesBase):
     node: DOMNode | None = None
-    _rules: RulesMap = field(default_factory=dict)
+    _rules: RulesMap = field(default_factory=RulesMap.__call__)
```
willmcgugan, line 587: Why is the `__call__` necessary here?

src/textual/demo.py
```diff
@@ -189,6 +190,14 @@
 """
 
 
+class Note(Message):
```
willmcgugan, line 193: It's a clever idea to have a message that prints a note. But I think this was to work around that `self.app` returns a base App?

If that's the case. I think we should probably revert to the original, as incentive to come up with a better solution for typing `self.app`

src/textual/keys.py
```diff
@@ -197,6 +196,18 @@ class Keys(str, Enum):
     ShiftControlEnd = ControlShiftEnd
 
 
+# We want to make sure that mypy knows that the values of Keys will always be strings.
```
willmcgugan, line 199: I don't know what is going on here.

Can you explain it like I'm 5?

src/textual/widgets/_data_table.py
```diff
@@ -64,18 +64,37 @@ class DuplicateKey(Exception):
     an existing row or column in the DataTable. Keys must be unique."""
 
 
+StringKeySubclass = TypeVar("StringKeySubclass", bound="StringKey")
```
willmcgugan, line 67: @darrenburns Could you look over Rodrigo's changes in this file please?

src/textual/widgets/_data_table.py
```diff
@@ -138,6 +157,53 @@ def __rich_repr__(self):
         yield "column_key", self.column_key
 
 
+class KeysToIndices(TwoWayDict[StringKeySubclass, int]):
```
willmcgugan, line 160: What is the purpose of this class over a plain `TwoWayDict`? Is it to generate a more specific exception? Was there a specific case were the DataTable was generating a non specific exception?

willmcgugan (commented):
A few comments for your consideration.

You may have gone a little bit beyond the remit of "fix typing errors". Might have been better to create issues for the larger problems you identified rather than fix them here.

But great stuff, you solved a number of irritating typing issues we've had for a while.

src/textual/_compositor.py
```diff
@@ -819,10 +835,7 @@ def render(self, full: bool = False) -> RenderableType | None:
         cuts = self.cuts
 
         # dict.fromkeys is a callable which takes a list of ints returns a dict which maps ints on to a list of Segments or None.
-        fromkeys = cast(
-            "Callable[[list[int]], dict[int, list[Segment] | None]]", dict.fromkeys
-        )
-        # A mapping of cut index to a list of segments for each line
+        fromkeys = cast("Callable[[list[int]], dict[int, Strip | None]]", dict.fromkeys)
```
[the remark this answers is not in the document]
rodrigogiraoserrao: The diff isn't great, but the comment is correct: it's "# dict.fromkeys is a callable [...]" and is the old line 821 / the new line 837...

src/textual/actions.py
rodrigogiraoserrao: I agree.

src/textual/app.py
rodrigogiraoserrao: Now that you mention it, me neither...
The type hints gave it away. It was complaining about the argument type and it said that "Type _NullFile" cannot be assigned to type "IO[str] | None", from which I inferred `None` would redirect to the void.

src/textual/app.py
rodrigogiraoserrao: You are absolutely right! That is the absolute most obvious thing to do and I was sure I had tried typing the result of `reversed` as an iterable and it had complained...
But I just did it correctly, so I must have tried the wrong thing!

src/textual/app.py
rodrigogiraoserrao: I don't understand why you think it would blow up! 🤷 
But I've separated the two variables.

src/textual/css/styles.py
rodrigogiraoserrao: mypy sees `RulesMap` as `Type[RulesMap]` instead of `() -> RulesMap`, which is what we want.
The error with `RulesMap` only:
```
src/textual/css/styles.py:587: error: Argument "default_factory" to "field" has incompatible type "Type[RulesMap]"; expected "Callable[[], RulesMap]"
```

I can replace `RulesMap.__call__` with `cast(Callable[[], RulesMap], RulesMap)` but I don't like `cast` very much and that is why I went with `RulesMap.__call__`. I can also add an explicit comment like `# mypy isn't happy with just RulesMap`.

src/textual/demo.py
rodrigogiraoserrao: I don't understand, sorry!

src/textual/keys.py
rodrigogiraoserrao: Fast-forward: This was the very first typing issue I tried to fix (I tackled them in alphabetical order 😅 ) and I couldn't figure out anything better.

TL;DR: the typing of Enum.value in `enum.pyi` is very meta because of the way the attribute `value` is implemented. When trying to solve this issue, I focused too much on what the typing of `value` looked like in `enum.pyi` and forgot to think about the actual implementation of `value` in `enum.py`.
Taking a look at it now inspired a simpler solution.

src/textual/widgets/_data_table.py
rodrigogiraoserrao: `TwoWayDict` was lenient and allowed returning `None`s from its `get` methods.
When that was correctly typed, many typing errors appeared because the code does not contemplate the possibility of getting `None`s back from its internal bidirectional dictionaries.

I started tackling those one by one, doing a runtime check for `None` and raising the appropriate error.

As I did that, I realised the code was getting bloated with all those checks.
So, I decided to lift those checks to the specialised bidirectional dictionary `KeysToIndices` that is responsible for raising the appropriate errors when fetching a key or a value fails.
(Which the semantics of the data table code imply should never happen.)

I thought this was a good solution but I can be convinced otherwise.

If there are no plans to use the general bidirectional dictionary elsewhere, we can merge `TwoWayDict` and `KeysToIndices` into a single class.

src/textual/app.py
willmcgugan: If `screen` really was a callable on line L1809, then I would expect that attribute to not exist. But there are no callables on the screen_stack.

src/textual/css/styles.py
willmcgugan: I guess Mypy doesn't understand that types are also callables.

would `default_factory = lambda: RulesMap()` be preferable here?

src/textual/demo.py
willmcgugan: I'm guess you did this because `self.app.add_note` was complaining that there was no `add_note` on App? I think we should probably attempt to solve that issue first, as it crops up everywhere.

src/textual/keys.py
willmcgugan: It doesn't feel like the right way to do things. The fact that you have to ignore the redefinition on 205 makes me suspicious.

Think it might be better to revert this one, and we will tackle it in the future when I have time to grok it.

src/textual/widgets/_data_table.py
willmcgugan: I have a suspicion that the DataTable code checks for these conditions, so the TWD never returns None at runtime.

Perhaps the `get` and `get_key` should just use `[]` syntax.

@darrenburns Can you confirm?

src/textual/keys.py
rodrigogiraoserrao: Sorry, I forgot to commit and push the newest fix that looks much more reasonable.

src/textual/css/styles.py
rodrigogiraoserrao: It satisfies mypy and I think it looks better, so I replaced it.

src/textual/widgets/_data_table.py
rodrigogiraoserrao: I just merged main and I have a couple of failing tests because I'm getting `None`s from `get`/`get_key`.

willmcgugan:
@rodrigogiraoserrao I'd like to get this merged soon. Could you revert the changes to DataTable please. We can consider the DT stuff for a future update, but I'd like to have Darren look over them when he's back in the office (he's sick ATM).

rodrigogiraoserrao:
> @rodrigogiraoserrao I'd like to get this merged soon. Could you revert the changes to DataTable please. We can consider the DT stuff for a future update, but I'd like to have Darren look over them when he's back in the office (he's sick ATM).

Changes reverted.

rodrigogiraoserrao:
This PR will not knock down all existing typing issues.

Should we leave issue #1822 (Fix typing) open or should I open a new issue that is more specific?

willmcgugan:
Thanks. I think we can leave the Fix Typing PR open. It will be an ongoing thing.

rodrigogiraoserrao:
> Thanks. I think we can leave the Fix Typing PR open. It will be an ongoing thing.

Ok. In that case, I just need someone to approve this PR. (Maybe this one should be squash-merged.)

willmcgugan (approved):

Issue #1807, opened 2023-02-16
Markdown widget, list before header, line across multiple lines


Paul-Aime:
Thanks for the amazing markdown support.

I noticed 2 small bugs when trying it:

- A line split across multiple lines will not be displayed with spaces in between words ending/starting line (it does in `rich --markdown`)
- Some padding is missing in between a list and ~an header level 3~ following line.

Screenshot:

(Also the screenshot automatically generated gives bad borders and headers' underlines but it looks good in my terminal).

![markdownapp_2023-02-16T09_20_29_145944](https://user-images.githubusercontent.com/36678697/219308212-ed3406f7-d26f-4962-b1cc-851b3205c6e6.svg)

Code for the app:

```python
#! /home/paul/.local/opt/pipx/venvs/textual/bin/python

import sys
from textual.app import App, ComposeResult
from textual.widgets import Markdown


class MarkdownApp(App):
    fpath = sys.argv[1]

    def compose(self) -> ComposeResult:
        with open(self.fpath) as f:
            text = f.read()
        yield Markdown(markdown=text)


if __name__ == "__main__":
    app = MarkdownApp()
    app.run()
```

The input file used:

````markdown
# Header 1

## Header 2

- A list before
- an header 3
- is NOT OK

### Header 3 (1)

Line written
on multiple
lines.

A line before an header 3 is OK

### Header 3 (2)

```python
print("Code block before Header 3 is OK")
```

### Header 3 (3)

THE END.
````

`textual diagnose`:

```
# Textual Diagnostics

## Versions

| Name    | Value  |
|---------|--------|
| Textual | 0.11.0 |
| Rich    | 13.3.1 |

## Python

| Name           | Value                                               |
|----------------|-----------------------------------------------------|
| Version        | 3.8.10                                              |
| Implementation | CPython                                             |
| Compiler       | GCC 9.4.0                                           |
| Executable     | /home/paul/.local/opt/pipx/venvs/textual/bin/python |

## Operating System

| Name    | Value                                       |
|---------|---------------------------------------------|
| System  | Linux                                       |
| Release | 5.14.0-1057-oem                             |
| Version | #64-Ubuntu SMP Mon Jan 23 17:02:19 UTC 2023 |

## Terminal

| Name                 | Value       |
|----------------------|-------------|
| Terminal Application | Kitty       |
| TERM                 | xterm-kitty |
| COLORTERM            | truecolor   |
| FORCE_COLOR          | *Not set*   |
[...]

PR #1832, opened 2023-02-18
Batch update

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/1807

Adds "batch updates" to app and Markdown improvements

The basic concept of batch updates is that you can defer repaint / layout while updating the screen. The reasons are two-fold:

1. It does less work by replacing multiple operations with one.
2. It reduces flicker because the screen doesn't update until you are finished.

In Markdown I've improved handling of lists, soft breaks, and added horizontal rules.


rodrigogiraoserrao (approved):

src/textual/app.py
```diff
@@ -426,6 +433,30 @@ def children(self) -> Sequence["Widget"]:
         except ScreenError:
             return ()
 
+    @contextmanager
+    def batch_update(self) -> Generator[None, None, None]:
+        """Suspend all repaints until the end of the batch."""
```
davep, line 438: Perhaps an `Example:` or similar thing in the docstring here to remind folk how to call this?

davep (approved):

src/textual/app.py
willmcgugan: Going to leave that for now. Want to figure out best practices before we flesh out docs.

PR #1833, opened 2023-02-18
Scroll thumb

willmcgugan:
Fixes the issue where the scroll thumb would appear to go under the window.

The solution and the problem is hard to put in to words. I can whiteboard it if anyone is interested.

https://user-images.githubusercontent.com/554369/219872768-502e4fe4-44c9-4c70-b513-f5f6b5caa3e7.mp4

### DIFF
--- a/src/textual/scrollbar.py
+++ b/src/textual/scrollbar.py
@@ -378,17 +378,3 @@ class ScrollBarCorner(Widget):
         styles = self.parent.styles
         color = styles.scrollbar_corner_color
         return Blank(color)
-
-
-if __name__ == "__main__":
-    from rich import print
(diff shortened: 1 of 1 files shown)


rodrigogiraoserrao (commented):
I don't understand the fix. I was expecting some logic to set the thumb size to a minimum value when it falls below the relevant threshold, but you seem to be increasing virtual sizes instead.
What do "virtual sizes" typically refer to?

willmcgugan:
Virtual size is typically the number of scrollable lines in the window.

It's hard to describe. Can do a whiteboard session tomorrow.

davep (approved):

PR #1841, opened 2023-02-19
Reactive get

willmcgugan:
- Fix that compute methods weren't being called on demand.
- Micro-optimizations on the descriptor getter. These are called a attribute access, so any savings add up.

### DIFF
--- a/src/textual/reactive.py
+++ b/src/textual/reactive.py
@@ -143,23 +143,25 @@ class Reactive(Generic[ReactiveType]):
         self.name = name
         # The internal name where the attribute's value is stored
         self.internal_name = f"_reactive_{name}"
+        self.compute_name = f"compute_{name}"
         default = self._default
         setattr(owner, f"_default_{name}", default)
 
     def __get__(self, obj: Reactable, obj_type: type[object]) -> ReactiveType:
-        _rich_traceback_omit = True
-
-        self._initialize_reactive(obj, self.name)
+        internal_name = self.internal_name
+        if not hasattr(obj, internal_name):
+            self._initialize_reactive(obj, self.name)
 
         value: ReactiveType
-        compute_method = getattr(self, f"compute_{self.name}", None)
+        compute_method = getattr(obj, self.compute_name, None)
         if compute_method is not None:
-            old_value = getattr(obj, self.internal_name)
-            value = getattr(obj, f"compute_{self.name}")()
-            setattr(obj, self.internal_name, value)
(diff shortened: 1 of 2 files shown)


src/textual/reactive.py
```diff
@@ -143,23 +143,25 @@ def __set_name__(self, owner: Type[MessageTarget], name: str) -> None:
...
+        self.compute_name = f"compute_{name}"
         default = self._default
         setattr(owner, f"_default_{name}", default)
 
     def __get__(self, obj: Reactable, obj_type: type[object]) -> ReactiveType:
-        _rich_traceback_omit = True
-
-        self._initialize_reactive(obj, self.name)
+        internal_name = self.internal_name
+        if not hasattr(obj, internal_name):
```
willmcgugan, line 152: Prevent a function call on the 90% case where the object has been initialized.

tests/test_reactive.py
```diff
@@ -328,6 +328,33 @@ class Tertiary(Secondary):
...
+        def compute_count_double(self) -> int:
+            return self.start + self.count * 2
+
+    app = ComputeApp()
+
+    async with app.run_test():
+        assert app.count_double == 0
+        app.count = 1
+        assert app.count_double == 2
+        assert app.count_double == 2
```
rodrigogiraoserrao, line 351: Is this duplication on purpose?

tests/test_reactive.py
```diff
@@ -347,7 +374,9 @@ def watch_show_ac(self, show_ac: bool) -> None:
 
     app = Calculator()
 
-    async with app.run_test() as pilot:
+    # Referencing the value calls compute
```
rodrigogiraoserrao, line 377: If referencing calls compute methods, why do we need to call compute methods when other reactive variables are changed if those changes will only be observed when we reference the reactive attribute, which will then re-trigger a call to its compute method anyway?

tests/test_reactive.py
willmcgugan: probably a cut and paste error

tests/test_reactive.py
willmcgugan: The main reason for `compute` method existing is that you can watch them, and have a `watch` method called when the result might change. Otherwise, you may as well define a property.

davep (approved):

tests/test_reactive.py
rodrigogiraoserrao: Of course 🤦  `watch` methods are the key, here.

PR #1847, opened 2023-02-20
Alternative compose

willmcgugan:
I've added an alternative (and complimentary) way of composing widgets.

Building compose methods gets a little awkward at a certain level. Consider this method:

```python
    yield Foo()
    yield Bar()
    yield Baz()
```

If I want to wrap that in a `Horizontal`, the syntax changes:

```python
    yield Horizontal(    
        Foo(),
        Bar(),
        Baz(),
        id="wrapper"
    )
```

This change allows widgets to be used as a context manager, to introduce a new *level*.

```python
    with Horizontal(id="wrapper"):
        yield Foo()
        yield Bar()
        yield Baz()
```

Both methods can coexist, but I find there are the following benefits with the context managers:

- It's easy to wrap something, add a line and indent the contents.
- The `id` and other parameters remain at the top.
- No need to add commas and braces.
- Adding many widgets doesn't require clumsy looking spat syntax `Vertical(*[MyWidget() for _ in range(10)])`.

This isn't documented yet. I've only changed internal apps for now. I want to see how we find this.


davep:
Just from the description alone this gets an enthusiastic "oh hell yes!" from me.

src/textual/widget.py
```diff
@@ -363,6 +365,22 @@ def offset(self) -> Offset:
     def offset(self, offset: Offset) -> None:
         self.styles.offset = ScalarOffset.from_offset(offset)
 
+    def __enter__(self) -> None:
```
davep, line 368: Is there a reason why we can't return `self` from this? I'm thinking folk may want to sometimes hang on to a reference to the main wrapper widget for use elsewhere in the code:

```python
    with Horizontal(id="wrapper") as frame:
        self.frame = frame
        yield Foo()
        yield Bar()
        yield Baz()
```

willmcgugan: Good idea.

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## Unreleased
+
+### Changed
+
+- Added alternative method of composing Widgets
```
rodrigogiraoserrao, line 12: Link to PR?

src/textual/_compose.py
```diff
@@ -0,0 +1,39 @@
...
+def compose(node: App | Widget) -> list[Widget]:
+    """Compose child widgets.
+
+    Args:
+        node: The parent node.
+
+    Returns:
+        A list of widgets.
+    """
+    app = node.app
```
rodrigogiraoserrao, line 19: You typed `node` as `App | Node`. If `node` is the app, can you really do `node.app`?

src/textual/_compose.py
```diff
@@ -0,0 +1,39 @@
...
+    nodes: list[Widget] = []
+    compose_stack: list[Widget] = []
+    composed: list[Widget] = []
+    app._compose_stacks.append(compose_stack)
+    app._composed.append(composed)
```
rodrigogiraoserrao, line 24: I'm a bit lost. The lists `compose_stack` and `composed` do not seem to be modified directly in this function.
Are they being modified as a side effect of having been appended to `app._compose_stacks`/`app._composed` and the later call to `node.compose()`?

rodrigogiraoserrao: Now, I see they are.
I don't _love_ side-effects here and they make the code harder to follow but I can't come up with a better suggestion off the top of my head.
Can you see any other possible strategies here or is this the only feasible route?

rodrigogiraoserrao (commented):
I like this use of context managers!

src/textual/_compose.py
willmcgugan: That is an oddity, but App inherits from MessagePump and MessagePump has an `app` property. Mypy would inform me if that wasn't the case!

src/textual/_compose.py
willmcgugan: I don't think there is any other way to make this magic work.

ofek:
Very nice!!!

davep (approved):

Issue #1816, opened 2023-02-16
MarkdownViewer - moving focus to the tree causes markdown content to scroll to top


darrenburns:
When focus moves to the tree widget within MarkdownViewer, the content on the right hand side of the viewer scrolls to the top.

PR #1851, opened 2023-02-21
don't scroll to docked widgets

willmcgugan:
Fixed an issue with the Markdown example. When you focused the tree it scrolled the screen (and the markdown it contained). The problem was the docked things can't be scrolled in to view, since they are fixed.

Fixes https://github.com/Textualize/textual/issues/1816


darrenburns:
Does this fix https://github.com/Textualize/textual/issues/1816?

willmcgugan:
It does! I erroneously said the dictionary example, but it was the Markdown examples I was thinking of.

rodrigogiraoserrao (approved):

PR #1866, opened 2023-02-23
Add a "prevent" context manager

willmcgugan:
Context manager to temporarily prevent a message from being sent.

### DIFF
--- a/docs/examples/events/prevent.py
+++ b/docs/examples/events/prevent.py
@@ -18,7 +18,7 @@ class PreventApp(App):
 
     def on_input_changed(self) -> None:
         """Called as the user types."""
-        self.bell()  # (2)
+        self.bell()  # (2)!
 
 
 if __name__ == "__main__":
--- a/docs/examples/events/prevent.py
+++ b/docs/examples/events/prevent.py
@@ -1,0 +1,26 @@
+from textual.app import App, ComposeResult
+from textual.containers import Horizontal
+from textual.widgets import Button, Input
+
+
+class PreventApp(App):
+    """Demonstrates `prevent` context manager."""
+
+    def compose(self) -> ComposeResult:
+        yield Input()
+        yield Button("Clear", id="clear")
+
+    def on_button_pressed(self) -> None:
+        """Clear the text input."""
+        input = self.query_one(Input)
+        with input.prevent(Input.Changed):  # (1)
+            input.value = ""
+
+    def on_input_changed(self) -> None:
+        """Called as the user types."""
+        self.bell()  # (2)
+
+
+if __name__ == "__main__":
+    app = PreventApp()
+    app.run()
--- a/docs/examples/events/prevent.py
+++ b/docs/examples/events/prevent.py
@@ -13,7 +13,7 @@ class PreventApp(App):
     def on_button_pressed(self) -> None:
         """Clear the text input."""
         input = self.query_one(Input)
-        with input.prevent(Input.Changed):  # (1)
+        with input.prevent(Input.Changed):  # (1)!
             input.value = ""
 
     def on_input_changed(self) -> None:
--- a/docs/guide/events.md
+++ b/docs/guide/events.md
@@ -120,9 +120,9 @@ There are other ways of sending (posting) messages, which you may need to use le
 
 ## Preventing messages
 
-You can *temporarily* prevent a Widget or App from posting messages or events of a particular type by calling [prevent][textual.message_pump.MessagePump.prevent], which returns a context manager. This is typically used when updating data in a child widget and you don't want to receive notifications that something has changed.
+You can *temporarily* disable posting of messages of a particular type by calling [prevent][textual.message_pump.MessagePump.prevent], which returns a context manager (used with Python's `with` keyword). This is typically used when updating data in a child widget and you don't want to receive notifications that something has changed.
 
-The following example will play the terminal bell as you type, by handling the [Input.Changed][textual.widgets.Input.Changed] event. There's also a button to clear the input by setting `input.value` to empty string. We don't want to play the bell when the button is clicked so we wrap setting the attribute with a call to [prevent][textual.message_pump.MessagePump.prevent].
+The following example will play the terminal bell as you type. It does this by handling [Input.Changed][textual.widgets.Input.Changed] and calling [bell()][textual.app.App.bell]. There is a Clear button which sets the input's value to an empty string. This would normally also result in a `Input.Changed` event being sent (and the bell playing). Since don't want the button to make a sound, the assignment to `value` is wrapped in with a [prevent][textual.message_pump.MessagePump.prevent] context manager.
 
 !!! tip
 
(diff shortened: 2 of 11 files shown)


darrenburns:
Not particularly bothered but I'll just mention that I think I prefer `suppress` as it makes me think of reducing/preventing the _output_ of something from travelling further. `prevent` sounds like I'd be prevented from using the button.

e.g. `suppress` in the standard library doesn't _prevent_ an exception, it stops it from travelling further; a suppressor on a gun stops the sound from travelling further; when you suppress speech you don't stop it from occurring but instead limit its reach.

darrenburns:
Oh, here we are preventing the message from being posted. Never mind then.

darrenburns:
Although, I do think `prevent_message` is a better name personally 😄 otherwise it's kind of like "prevent what?"

docs/guide/events.md
```diff
@@ -118,6 +118,32 @@ There are other ways of sending (posting) messages, which you may need to use le
 - [post_message_no_wait][textual.message_pump.MessagePump.post_message_no_wait] The non-async version of `post_message`.
 
 
+## Preventing messages
+
+You can *temporarily* disable posting of messages of a particular type by calling [prevent][textual.message_pump.MessagePump.prevent], which returns a context manager (used with Python's `with` keyword). This is typically used when updating data in a child widget and you don't want to receive notifications that something has changed.
+
+The following example will play the terminal bell as you type. It does this by handling [Input.Changed][textual.widgets.Input.Changed] and calling [bell()][textual.app.App.bell]. There is a Clear button which sets the input's value to an empty string. This would normally also result in a `Input.Changed` event being sent (and the bell playing). Since don't want the button to make a sound, the assignment to `value` is wrapped in with a [prevent][textual.message_pump.MessagePump.prevent] context manager.
```
rodrigogiraoserrao, line 125:
```suggestion
The following example will play the terminal bell as you type. It does this by handling [Input.Changed][textual.widgets.Input.Changed] and calling [bell()][textual.app.App.bell]. There is a Clear button which sets the input's value to an empty string. This would normally also result in a `Input.Changed` event being sent (and the bell playing). Since we don't want the button to make a sound, the assignment to `value` is wrapped within a [prevent][textual.message_pump.MessagePump.prevent] context manager.
```

docs/examples/events/prevent.py
```diff
@@ -0,0 +1,26 @@
...
+    """Demonstrates `prevent` context manager."""
+
+    def compose(self) -> ComposeResult:
+        yield Input()
+        yield Button("Clear", id="clear")
+
+    def on_button_pressed(self) -> None:
+        """Clear the text input."""
+        input = self.query_one(Input)
+        with input.prevent(Input.Changed):  # (1)
```
rodrigogiraoserrao, line 16:
```suggestion
        with input.prevent(Input.Changed):  # (1)!
```

Consider adding the `!` to make sure that that comment disappears when the code is pasted into the documentation.

docs/examples/events/prevent.py
```diff
@@ -0,0 +1,26 @@
...
+            input.value = ""
+
+    def on_input_changed(self) -> None:
+        """Called as the user types."""
+        self.bell()  # (2)
```
rodrigogiraoserrao, line 21:
```suggestion
        self.bell()  # (2)!
```

Consider adding the `!` to make sure that that comment disappears when the code is pasted into the documentation.

src/textual/dom.py
```diff
@@ -1,11 +1,13 @@
 from __future__ import annotations
 
 import re
-from functools import lru_cache
+from contextlib import contextmanager
+from functools import lru_cache, reduce
```
rodrigogiraoserrao, line 5:
```suggestion
from functools import lru_cache
```

src/textual/dom.py
```diff
@@ -1,11 +1,13 @@
...
 from inspect import getfile
 from typing import (
     TYPE_CHECKING,
     ClassVar,
+    Generator,
```
rodrigogiraoserrao, line 10:
```suggestion
```

src/textual/dom.py
```diff
@@ -39,6 +41,7 @@
 
 if TYPE_CHECKING:
     from .app import App
+    from .messages import Message
```
rodrigogiraoserrao, line 44:
```suggestion
```

src/textual/message_pump.py
```diff
@@ -78,6 +85,58 @@ def __init__(self, parent: MessagePump | None = None) -> None:
         self._mounted_event = asyncio.Event()
         self._next_callbacks: list[CallbackType] = []
 
+    @property
+    def _prevent_message_types_stack(self) -> list[set[type[Message]]]:
+        """A stack that manages prevented messages.
+
+        Returns:
+            A list of sets of Message Types.
+        """
```
rodrigogiraoserrao, line 94:
```suggestion
        """The stack that manages prevented messages."""
```

https://github.com/Textualize/textual/wiki/Documentation-standards#property

src/textual/message_pump.py
```diff
@@ -149,6 +208,14 @@ def _detach(self) -> None:
         self._parent = None
 
     def check_message_enabled(self, message: Message) -> bool:
+        """Check if a given message is enabled (allowed to be sent).
+
+        Args:
+            message: A message object
```
rodrigogiraoserrao, line 214:
```suggestion
            message: A message object.
```

src/textual/message_pump.py
```diff
@@ -542,6 +610,7 @@ async def post_message(self, message: Message) -> bool:
             return False
         if not self.check_message_enabled(message):
             return True
+        message._prevent.update(self._get_prevented_messages())
```
rodrigogiraoserrao, line 613: This line confuses me a bit.
Is this the line that had to do with the extra thing you mentioned you ended up figuring out you had to do, in that as the message bubbles up, you need to continuously update the prevent stack with these things?

If so, do you want to write something with regard to that? If this only came to you as an afterthought or when some tests broke, it may be a good idea to comment above (or in the docstring) why we do this kind of update.

src/textual/message_pump.py
```diff
@@ -580,6 +649,7 @@ def post_message_no_wait(self, message: Message) -> bool:
             return False
         if not self.check_message_enabled(message):
             return False
+        message._prevent.update(self._get_prevented_messages())
```
rodrigogiraoserrao, line 652: Same comment as in `post_message`.

src/textual/message.py
```diff
@@ -31,6 +31,7 @@ class Message:
         "_no_default_action",
         "_stop_propagation",
         "_handler_name",
+        "_prevent",
```
rodrigogiraoserrao, line 34: Might be worth documenting this (and the other private attributes).

rodrigogiraoserrao (changes requested):
I like the idea and the implementation! I just left a couple of comments regarding nitpicks that are up for discussion and some tidying up.

docs/guide/events.md
rodrigogiraoserrao: (Paragraphs and diffs like this are the reason why I personally write one sentence per line in markdown comments.
Might be worth considering adding that suggestion to the documentation standards.)

docs/examples/events/prevent.py
willmcgugan: Didn't know that, thanks!

src/textual/message.py
willmcgugan: At some point, once things settle.

PR #1867, opened 2023-02-23
Correct the wording of the description of `Switch.value`

davep:
It's not just the default value, it's the ongoing value too and can be used to change the switch.


davep:
> Of course! Sorry for that... 😓

And there was me silently judging @willmcgugan 🤷‍♀️

willmcgugan (approved):

Issue #1846, opened 2023-02-20
Bugs related to multi-screen approach


mzebrak:
Version: 0.11.0

Consider a following code:
```python
from textual.app import App
from textual.binding import Binding
from textual.screen import Screen
from textual.widget import Widget
from textual.widgets import Footer, Header, Placeholder


class SomeWidget(Widget):
    def compose(self):
        yield Placeholder('SomeWidget')


class BaseScreen(Screen):

    def compose(self):
        # looks like heres also a problem - adding this line breaks the app `NoMatches: No nodes match <DOMQuery Header() filter='HeaderTitle'>`
        # yield Header()

        yield SomeWidget()
        yield Footer()


class FirstScreen(BaseScreen):
    BINDINGS = [
        Binding('n', 'switch_screen("second")', 'Second screen - via name'),
        Binding('m', 'switch', 'Second screen - via object'),
    ]

    def action_switch(self):
        self.app.switch_screen(SecondScreen())


class SecondScreen(BaseScreen):
    BINDINGS = [
        Binding('n', 'switch_screen("first")', 'First screen - via name'),
        Binding('m', 'switch', 'First screen - via objecct'),

    ]

    def action_switch(self):
        self.app.switch_screen(FirstScreen())


class MyApp(App):
    SCREENS = {
        'first': FirstScreen,
        'second': SecondScreen,
    }

    BINDINGS = [
        Binding('q', 'query', 'Query for SomeWidget'),
    ]

    def on_mount(self):
        # with 'first', we got 2 of SomeWidget's after first keypress (m). When there is `First()`, we got a 1 SomeWidget
        self.push_screen(FirstScreen())

    def action_query(self):
        self.log(f"Current stack: {self.screen_stack}")

        # Should be always equal to 1, but sometimes it's 2 or even 3 (while switching via both 'n' and 'm' keys)
        self.log(f"Number of SomeWidget's: {len(self.query(SomeWidget))}")


if __name__ == '__main__':
    MyApp().run()

```

The example provided above shows some problems. Try launching it and switching views via "n" and "m" keys.
I think it's a descriptive example so there's not much to add, but it seems that there are 2 bugs in this piece of code - one is a problem with displaying the header, because after uncommenting it throws an error, and second problem occurs when switching screens.

willmcgugan:
[...]

PR #1873, opened 2023-02-24
fix walk children

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/1846

The `walk_children` method was looking at `node._nodes` rather then `node.children` which mean that all screens were walked, rather than the active screen.


rodrigogiraoserrao (approved):
Merge away! 🚀

PR #1876, opened 2023-02-24
new blog post

willmcgugan:


docs/blog/posts/release0-12-0.md
```diff
@@ -0,0 +1,103 @@
...
+with self.app.batch_update():
+    await self.query("MarkdownBlock").remove()
+    await self.mount_all(output)
+```
+
+Without the batch update there are a few frames where the old markdown blocks are removed and the new blocks are added, which the user will perceive as a brief flicker.
+
+## Disabled widgets
+
+A few widgets (such as [Button](./../../widgets/button.md) had a `disabled` attribute which would fade the widget a little and make it unselectable. We've extended this to all widgets. Although it is particularly applicable to input controls, anything may be disabled. Disabling a container makes it children disabled, so you could use this for disabling a form, for example.
```
davep, line 91:
```suggestion
A few widgets (such as [Button](./../../widgets/button.md) had a `disabled` attribute which would fade the widget a little and make it unselectable. We've extended this to all widgets. Although it is particularly applicable to input controls, anything may be disabled. Disabling a container makes its children disabled, so you could use this for disabling a form, for example.
```

docs/blog/posts/release0-12-0.md
```diff
@@ -0,0 +1,103 @@
...
+It's been just 9 days since the previous release, but we have a few interesting enhancements to the Textual API to talk about.
+
+<!-- more -->
+
+## Better compose
+
+We've added a little *syntactical sugar* to Textual's `compose` methods, which aids both
+readability and *editability* (that might not be a word).
+
+First, lets look at the old way of building compose methods. This snippet is taken from the `textual colors` command.
```
davep, line 22:
```suggestion
First, let's look at the old way of building compose methods. This snippet is taken from the `textual colors` command.
```

rodrigogiraoserrao:
> Merging because it is deployed. Will only accept the most egregious of typos now. @rodrigogiraoserrao that means you!

Maybe it's best if I don't read the blog post, at all! 🤣

PR #1880, opened 2023-02-25
batch update fix, and optimization

willmcgugan:
- Fix for glitch (see screenshot) caused by the batch update, I noticed in markdown example. Essentially removed a race condition.
- Cached link ids on Strip

Screenshot 2023-02-25 at 08 15 45


willmcgugan:
#executivedecision Bit of a hot fix

Issue #1364, opened 2022-12-14
Review what causes some classes to get linked in our docs, and not others


davep:
As of the time of writing, something like [the docs for `App.mount_all`](https://textual.textualize.io/api/app/#textual.app.App.mount_all) have (auto-?)links to `Widget`, but not for `AwaitMount` or `MountError`. The issue is: why one but not the other two? What causes the links to happen, should we encourage linking to all public-facing and published types and classes, and if so how do we encourage the docs to do this?

First thought about in a comment in #1363.

vitrioil:
Isn't this because of no reference to that class which is somewhat determined from [here](https://github.com/Textualize/textual/blob/main/docs/api/widget.md)?

Example if I change it to say:
```markdown
::: textual.widget
```
Then I do see dedicated sections to `/api/widget/#textual.widget.AwaitMount` (and `/api/widget/#textual.widget.MountError`).

Although what's still weird is that there is a hyperlink to `AwaitMount` but not to `MountError` from `mount_all`'s documentation.

![mount_all_doc](https://user-images.githubusercontent.com/41074533/210633268-ac6ffb1c-9242-4bfe-8766-3138b335b318.png)

davep:
Nothing leapt out at me relating to this specific issue, although in fairness I wasn't concentrating on it. Now I've found the relevant docs I'll dip back in with this in mind at some point soon.

willmcgugan:
I wonder if @rodrigogiraoserrao figured this out in his tour of the docs?

rodrigogiraoserrao:
> I wonder if @rodrigogiraoserrao figured this out in his tour of the docs?

Not yet, didn't have this present in my mind. But you can assign this to me and I'll have a look during my next batch of dealing with documentation tickets.

davep:
Looking into this some more this morning and @vitrioil was spot on. Changing `docs/api/widget.md` to read like this:

```markdown
::: textual.widget
```

ensures all the public interfaces get documented and linkable. Moreover, doing this I'm finding that in the `mount_all` example mentioned above, everything that should be linkable is:

<img width="852" alt="Screenshot 2023-02-28 at 11 40 21" src="https://user-images.githubusercontent.com/28237/221843438-1a0d98b5-a083-4c20-95b1-8fa54abd58d1.png">

This seems like the correct way to go, at least for the API documentation: don't pull in the specific widget class, instead pull in the module and let everything inside it get documented (and so become linkable).

[...]

PR #1898, opened 2023-02-28
Increase the coverage of API documentation

davep:
See #1364 -- rather than pull in specific classes, this pulls in every publicly-documented item (function, class, type, "constant", etc) that can be found in each of the modules.

### DIFF
--- a/src/textual/await_remove.py
+++ b/src/textual/await_remove.py
@@ -5,7 +5,11 @@ from typing import Generator
 
 
 class AwaitRemove:
-    """An awaitable returned by App.remove and DOMQuery.remove."""
+    """An awaitable returned by a method that removes DOM nodes.
+
+    Returned by [Widget.remove][textual.widget.Widget.remove] and
+    [DOMQuery.remove][textual.css.query.DOMQuery.remove].
+    """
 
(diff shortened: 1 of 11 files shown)


src/textual/await_remove.py
```diff
@@ -5,7 +5,11 @@
 
 
 class AwaitRemove:
-    """An awaitable returned by App.remove and DOMQuery.remove."""
+    """An awaitable returned by a method that removes DOM nodes.
+
+    Returned by [Widget.remove][textual.widget.Widget.remove] and
```
rodrigogiraoserrao, line 10: Instead of listing categorically the places where this thing can be returned from, might I suggest we rephrase it as "returned, for example, from ..." or "[...] by ..."?
The rationale is to make it less likely that the docstring will become wrong in the future.

rodrigogiraoserrao (approved):
I left a comment for your consideration but looks good to me.

src/textual/await_remove.py
davep: Funnily enough I had more or less that to start with and then decided against it as I felt this was one time where I'd want this to be categorically correct (in that we shouldn't really want more than those instances of `remove` in Textual).

PR #1900, opened 2023-02-28
Added DEBUG switch, scrollbar_size property, fixed text_log line length

willmcgugan:
Fixed a few issues while looking in to something else.

- Added a `TEXTUAL_DEBUG` constant set from an env var, that checks for a common logic error.
- Added a `Strip.apply_style` which caches the moderately expensive operation.
- Fixes a miscalculation in TextLog

### DIFF
--- a/src/textual/constants.py
+++ b/src/textual/constants.py
@@ -8,7 +8,8 @@ A number of app level constants where the value is pulled from environment varia
 """
 
 import os
-from typing import Final
+
+from typing_extensions import Final
 
 get_environ = os.environ.get
 
--- a/src/textual/constants.py
+++ b/src/textual/constants.py
@@ -3,9 +3,34 @@ Constants that we might want to expose via the public API.
 
 """
 
+"""
+A number of app level constants where the value is pulled from environment variables.
+"""
+
+import os
+from typing import Final
+
+get_environ = os.environ.get
+
+
+def get_environ_bool(name: str) -> bool:
+    """Check an environment variable switch.
+
+    Args:
+        name: Name of environment variable.
+
+    Returns:
+        True if the env var is "1", otherwise False.
+    """
+    has_environ = os.environ.get(name) == "1"
+    return has_environ
+
+
 from ._border import BORDER_CHARS
 
 __all__ = ["BORDERS"]
 
 
 BORDERS = list(BORDER_CHARS)
+
+DEBUG: Final[bool] = get_environ_bool("TEXTUAL_DEBUG")
--- a/src/textual/constants.py
+++ b/src/textual/constants.py
@@ -11,6 +11,11 @@ import os
 
 from typing_extensions import Final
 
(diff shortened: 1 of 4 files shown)


src/textual/strip.py
```diff
@@ -39,8 +54,13 @@ def __init__(
         self._cell_length = cell_length
         self._divide_cache: FIFOCache[tuple[int, ...], list[Strip]] = FIFOCache(4)
         self._crop_cache: FIFOCache[tuple[int, int], Strip] = FIFOCache(4)
+        self._style_cache: FIFOCache[Style, Strip] = FIFOCache(4)
```
davep, line 57: Is there something in the context that makes it clear what `4` is for?

willmcgugan: It's fairly arbitrary, predicated on observation that there are relatively few combinations of inputs.

src/textual/widget.py
```diff
@@ -1029,6 +1029,11 @@ def scrollbars_enabled(self) -> tuple[bool, bool]:
         enabled = self.show_vertical_scrollbar, self.show_horizontal_scrollbar
         return enabled
 
+    @property
+    def scrollbar_size(self) -> Size:
+        """The dimensions of the scrollbar stored in a Size object."""
```
davep, line 1034: I'm finding this docstring a wee bit confusing; it makes a `Size` from `scrollbar_size_vertical` and `scrollbar_size_horizontal`, but they're two *different* scrollbars, as I understand it. Perhaps the docstring could do with being expanded to make it clear that it's about two different scrollbars and also make it clear which is in which `Size` attribute?

davep (approved):
Small docstring thing but otherwise all good with me!

src/textual/constants.py
```diff
@@ -3,9 +3,32 @@
 
 """
 
+"""
+A number of app level constants where the value is pulled from environment variables.
+"""
```
rodrigogiraoserrao, line 8: I don't understand why we have two module-level docstrings here! 😕

src/textual/strip.py
```diff
@@ -10,9 +10,23 @@
...
+def get_line_length(segments: Iterable[Segment]) -> int:
+    """Get the line length (total length of all segments).
+
+    Args:
+        segments: Iterable of segments.
+
+    Returns:
+        Length of line in cells.
+    """
+    _cell_len = cell_len
```
rodrigogiraoserrao, line 26: Is this a micro-optimisation?

rodrigogiraoserrao (approved):

src/textual/widget.py
willmcgugan: It's poorly named, and could just be a tuple. I will clarify.

src/textual/strip.py
willmcgugan: It is. Globals are slower than locals.

Issue #1897, opened 2023-02-28
The widget single cell scrolling methods no longer work as expected from the keyboard


davep:
Consider the following (slightly over-complicated, but designed to show all the related information) code:

```python
from rich.segment        import Segment
from textual.app         import App, ComposeResult
from textual.widgets     import Header, Footer
from textual.scroll_view import ScrollView
from textual.geometry    import Size
from textual.strip       import Strip

class ScrollViewTest( ScrollView, can_focus=True ):

    def __init__( self, height: int ) -> None:
        super().__init__()
        self.virtual_size = Size( 0, height )

    def render_line( self, y: int ) -> Strip:
        return Strip( [
            Segment( f"{self.scroll_offset.y + y:10}" ),
            Segment( f"{y:10}" ),
            *(
                [ Segment( f"{' ' * 10}scroll_offset.y == {self.scroll_offset.y}") ]
                if y == 0 else []
            )
        ] )

class ScrollTestApp( App[ None ] ):

    def compose( self ) -> ComposeResult:
        yield Header()
        yield ScrollViewTest( 1000 )
        yield Footer()

    def on_mount( self ) -> None:
        self.query_one( ScrollViewTest ).focus()

if __name__ == "__main__":
    ScrollTestApp().run()
```

If you run it and, using just the keyboard, try and scroll down just one line, or up one line, you'll find it's not possible. It will always scroll two lines.

It seems that this was introduced as part of #1689 ("Scroll sensitivity"). Two `App`-level properties were introduced, called `scroll_sensitivity_x` and `scroll_sensitivity_y`. Both properties say they dictate the number of cells to scroll in the given direction *"with wheel or trackpad"*. However, they're used in the related `Widget.scroll_*` methods regardless of the input method.

At the very least `scroll_up`, `scroll_down`, `scroll_left` and `scroll_right` should be modified to only apply the sensitivity setting if the input device is a pointing device rather than a key binding.

davep:
Bumping this to TODO and claiming it as it intersects with what'll need doing with #1774. This should be merged first I feel, then #1774 can follow (or it may be done as part of the same change; I've yet to decide -- either way I'm promoting it and claiming it).

willmcgugan:
I think we need to differentiate between mouse scroll (which *should* respect sensitivity and keyboard scroll which should *not* respect sensitivity). Suspect we will need to add a "mouse_initiated" bool to the scroll up / down / left / right methods.

Issue #1774, opened 2023-02-13
`scroll_*` doesn't work immediately after loading a `Syntax` into a `Static`


barisione:
I'm trying to load a source file into a `Static` and then jump to a line but it looks like `scroll_*` methods are not reliable.

I tried using https://github.com/Textualize/textual/blob/main/examples/code_browser.py and changed `scroll_home` to other functions like `scroll_end` and `scroll_to_region`. If used immediately after creating the widgets and loading a `Syntax` they never seem to work.
After switching files they sometimes work and sometimes not.

This is a minimal reproducer:
```
import os
import tempfile

from rich.syntax import Syntax
from textual import events
from textual.app import App, ComposeResult
from textual.containers import Vertical
from textual.geometry import Region
from textual.widgets import Static


class ScrollBugApp(App):
    def compose(self) -> ComposeResult:
        yield Vertical(Static(id="code", expand=True), id="code-view")

    def on_mount(self, event: events.Mount) -> None:
        tmp = tempfile.NamedTemporaryFile("w", delete=False)
        try:
            for i in range(1, 500):
                print(f"Line {i}", file=tmp)
            tmp.close()

            self.query_one("#code", Static).update(
                Syntax.from_path(
                    tmp.name,
                    line_numbers=True,
                    word_wrap=False,
                    indent_guides=True,
                    theme="github-dark",
                )
            )
            self.query_one("#code-view").scroll_to_region(
                Region(y=250, height=20),
                animate=False,
            )
        finally:
            os.unlink(tmp.name)


if __name__ == "__main__":
    ScrollBugApp().run()
```

# Textual Diagnostics

## Versions

| Name    | Value  |
|---------|--------|
| Textual | 0.10.1 |
| Rich    | 13.3.1 |

## Python

| Name           | Value                                                                     |
|----------------|---------------------------------------------------------------------------|
| Version        | 3.10.1                                                                    |
| Implementation | CPython                                                                   |
| Compiler       | Clang 13.0.0 (clang-1300.0.29.30)                                         |
| Executable     | /Users/bari/.pyenv/versions/3.10.1/envs/undo-core-venv3.10/bin/python3.10 |

## Operating System

| Name    | Value                                                                                             |
|---------|---------------------------------------------------------------------------------------------------|
| System  | Darwin                                                                                            |
| Release | 21.6.0                                                                                            |
[...]

Issue #1811, opened 2023-02-16
Add missing docstrings to a number of widgets


davep:
Some widgets seem to be missing docstrings, either on the class, on their `__init__`, or both. Review the following widgets and correct this; also pay attention to any attributes that need docstrings too:

- [x] `Header`
- [x] `ListItem`
- [x] `Placeholder`
- [x] `Pretty`
- [x] `Tree`
- [x] `TextLog`
- [x] `Welcome`

Also noticed while working elsewhere:

- [x] `MessagePump.check_message_enabled` and implementations in child classes.

Also also:

- [x] Remove a lot of the "default" info from the various `scroll_*` methods in `Widget` (see comments on #1902).

PR #1902, opened 2023-02-28
Fixes and improvements relating to scrolling

davep:
The changes here roll two issues into one change. With this commit:

- Scrolling up/down/etc using the keyboard now moves just one cell, rather than moving the number of cells specified by the scroll sensitivity that's intended for pointing devices. #1897
- Where appropriate the scrolling is done lazily; that is it is done after the next refresh, helping to ensure that the scroll will take into account any updates in the same parent call. #1774


src/textual/widget.py
```diff
@@ -1493,6 +1493,43 @@ def scroll_to(
...
+
+        Args:
+            x: X coordinate (column) to scroll to, or None for no change. Defaults to None.
+            y: Y coordinate (row) to scroll to, or None for no change. Defaults to None.
+            animate: Animate to new scroll position. Defaults to True.
+            speed: Speed of scroll if animate is True. Or None to use duration.
+            duration: Duration of animation, if animate is True and speed is None.
+            easing: An easing method for the scrolling animation. Defaults to "None",
+                which will result in Textual choosing the default scrolling easing function.
+            force: Force scrolling even when prohibited by overflow styling. Defaults to `False`.
```
rodrigogiraoserrao, line 1517: Please remove all the defaults from the docstring :)

rodrigogiraoserrao: I see this was likely a copy & paste from other methods, like `scroll_relative`.
If you want, you can also delete the defaults from there 😆

src/textual/widget.py
```diff
@@ -1581,24 +1612,58 @@ def scroll_end(
...
+    ) -> None:
+        """Scroll one cell left.
+
+        Args:
+            animate: Animate scroll. Defaults to True.
+            speed: Speed of scroll if animate is True. Or None to use duration.
+            duration: Duration of animation, if animate is True and speed is None.
+            easing: An easing method for the scrolling animation. Defaults to "None",
+                which will result in Textual choosing the configured default scrolling easing function.
+            force: Force scrolling even when prohibited by overflow styling. Defaults to `False`.
```
rodrigogiraoserrao, line 1655: Please remove defaults.

src/textual/widget.py
```diff
@@ -1638,8 +1706,36 @@ def scroll_right(
...
+    ) -> None:
+        """Scroll one cell right.
+
+        Args:
+            animate: Animate scroll. Defaults to True.
+            speed: Speed of scroll if animate is True. Or None to use duration.
+            duration: Duration of animation, if animate is True and speed is None.
+            easing: An easing method for the scrolling animation. Defaults to "None",
+                which will result in Textual choosing the configured default scrolling easing function.
+            force: Force scrolling even when prohibited by overflow styling. Defaults to `False`.
```
rodrigogiraoserrao, line 1718: 😁

src/textual/widget.py
```diff
@@ -1670,9 +1769,37 @@ def scroll_down(
...
+    ) -> None:
         """Scroll one line down.
 
+        Args:
+            animate: Animate scroll. Defaults to True.
+            speed: Speed of scroll if animate is True. Or None to use duration.
+            duration: Duration of animation, if animate is True and speed is None.
+            easing: An easing method for the scrolling animation. Defaults to "None",
+                which will result in Textual choosing the configured default scrolling easing function.
+            force: Force scrolling even when prohibited by overflow styling. Defaults to `False`.
```
rodrigogiraoserrao, line 1781: 😬

src/textual/widget.py
```diff
@@ -1702,9 +1832,37 @@ def scroll_up(
...
+    ) -> None:
         """Scroll one line up.
 
+        Args:
+            animate: Animate scroll. Defaults to True.
+            speed: Speed of scroll if animate is True. Or None to use duration.
+            duration: Duration of animation, if animate is True and speed is None.
+            easing: An easing method for the scrolling animation. Defaults to "None",
+                which will result in Textual choosing the configured default scrolling easing function.
+            force: Force scrolling even when prohibited by overflow styling. Defaults to `False`.
```
rodrigogiraoserrao, line 1844: 🧐

rodrigogiraoserrao (approved):
Docstrings just need a little bit of trimming! Otherwise, good to go :)

src/textual/widget.py
davep: Let's make this a followup "clean widget.py docs" PR in that case (although I do tend to like to see something in the docstring that makes it clear it's optional and what the default may be and why).

src/textual/widget.py
davep: Ditto followup `widget.py` docstring cleanup.

src/textual/widget.py
davep: Ditto followup `widget.py` docstring cleanup.

src/textual/widget.py
davep: Ditto followup `widget.py` docstring cleanup.

src/textual/widget.py
rodrigogiraoserrao: Right. Explaining what the default does or why it is a specific value makes more sense than just writing "Defaults to x".
I don't like "Defaults to x" because it ages badly: in a couple of iterations it will become a lie and it is easy to forget updating that.

A more complete sentence saying something like "Defaults to `None`, which will result in Textual [...]" is _more_ sensible but it can also be rewritten in a way that doesn't even specify the default, e.g., "An easing method for the scrolling animation to replace the one Textual chooses by default".

What do you think?

rodrigogiraoserrao: But given the follow-up issue, I think this can be merged.

davep: > What do you think?

I think if we explain that something is optional, but don't say what the default is and means, that makes more work for the reader. So *"An easing method for the scrolling animation to replace the one Textual chooses by default"* means either a) we've got to somehow link to somewhere that says what that default is (and so the reader ends up being taken out of their current context and then has to remember they need to go back rather than disappear down a documentation rabbit hole) or b) the reader has to make a note to go find how and where that default behaviour is documented.

I'm totally on board with the whole thing of not repeating facts of the code in the docs, for fear if it going stale, but I do worry that we end up leaving the docs lacking in useful information.

(and now I've seen I've written a fair bit here perhaps this is a consideration for us to spin out into its own issue)

src/textual/widget.py
```diff
@@ -1581,24 +1612,58 @@ def scroll_end(
...
-            0,
-            self.max_scroll_y,
+
+        # In most cases we'd call self.scroll_to and let it handle the call
+        # to do things after a refresh, but here we need the refresh to
+        # happen first so that we can get the new self.max_scroll_y (that
+        # is, we need the layout to work out and then figure out how big
+        # things are). Because of this we'll create a closure over the call
+        # here and make our own call to call_after_refresh.
+        def _lazily_scroll_end():
```
willmcgugan, line 1625: Typing and docstring please.

willmcgugan (approved):
Just one request

src/textual/widget.py
davep: I wouldn't have thought we'd docstring hidden closures, but sure thing.

davep: Done.

davep:
PS: Currently working on #1811 so will roll the `scroll_*` docstring trimming into that.

Issue #1905, opened 2023-02-28
[Docs] `question02.py` in the guide could do wth an extra highlight


davep:
Identified by @driscollis over on Discord, in [this section of the guide](https://textual.textualize.io/guide/app/#css) the `Label` acquires an `id` but that change isn't highlighted, so anyone playing along in their editor might miss out on it.

We should highlight that change.

PR #1906, opened 2023-03-01
Highlight that the label acquires an id in the example code

davep:
While the code here isn't exactly written as a play-along tutorial, it's reasonable to expect that some one can and will. This commit has been written because exactly that situation happened.

The problem here is without the ID on the label the layout will end up looking very different from that shown in the docs; which will halt the reader's flow as they try and figure out what they did wrong.

Counterpoint: this is a guide, not a tutorial, so it seemed reasonable to only highlight the important thing that's changing. Personally I think I'd argue in favour of highlighting all the changes from question01 to question02.

Addresses #1905.

### DIFF
--- a/docs/guide/app.md
+++ b/docs/guide/app.md
@@ -157,9 +157,13 @@ The chapter on [Textual CSS](CSS.md) describes how to use CSS in detail. For now
 
 The following example enables loading of CSS by adding a `CSS_PATH` class variable:
 
-```python title="question02.py" hl_lines="6"
+```python title="question02.py" hl_lines="6 9"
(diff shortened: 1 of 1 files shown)


rodrigogiraoserrao (approved):
I'm in favour of highlighting small changes that would be hard to miss, or adding a note in a callout, or providing a diff, or just a Python comment.

Agreeing somewhat with your counterpoint as well, one can always say that it is only worth adding the "tooltips" to those important changes.

willmcgugan (approved):

davep:
> one can always say that it is only worth adding the "tooltips" to those important changes

Good call! I always forget about the tooltips.

PR #1907, opened 2023-03-01
A warning from the future

davep:
Add a warning to my first blog post, letting the attentive reader know that `emit_no_wait` doesn't exist any more, and linking them to what they should be using if they want to do their own custom messages.


rodrigogiraoserrao (approved):
Time travelling for the win!

willmcgugan (approved):

PR #1909, opened 2023-03-01
Widget documentation sweep

davep:
A wander through various widgets, adding missing docstrings, trying to make others better fit in with our standards, and with the occasional type hint tweak where necessary.

See #1811.

Also delivers on the `scroll_` docstring promise made in #1902.

---
*NOTE:* Part of this PR also experiments with a slight change to the layout of the API docs; currently only with the `Tree`. Rather than document `Tree` and `TreeNode` as two very different things, it documents them both under `Tree`. This means that it's then possible to document some of the other public-facing types without the reader having to try and guess which might relate to `Tree`, which to `TreeNode`, and which to both (and then guess which will be living where).


davep:
Managed to typo an issue reference in the last set of changes relating to the `scroll_` docstring updates, resulting in a future issues/PR/something (hello everyone reading #1920) possibly getting a bunch of unexpected references, when I meant to reference #1902.

(╯°□°）╯︵ ┻━┻

src/textual/widgets/_header.py
```diff
@@ -112,13 +137,21 @@ def __init__(
         id: str | None = None,
         classes: str | None = None,
     ):
+        """Initialise the header widget.
+
+        Args:
+            show_clock: ``True`` if the clock should be shown on the right of the header.
```
rodrigogiraoserrao, line 143: Is there a difference between surrounding `True` with a pair of backticks versus two pairs of backticks? I have seen both and haven't understood the difference yet.

src/textual/widgets/_placeholder.py
```diff
@@ -8,9 +10,11 @@
 from .. import events
 from ..css._error_tools import friendly_list
 from ..reactive import Reactive, reactive
-from ..widget import RenderResult, Widget
+from ..widget import Widget
 
 PlaceholderVariant = Literal["default", "size", "text"]
+"""The different variants of placeholder."""
```
rodrigogiraoserrao, line 16: What do you think of moving the variant table from `Placeholder` up to here?

src/textual/widgets/_placeholder.py
```diff
@@ -115,12 +117,19 @@ def __init__(
...
         # Set a cycle through the variants with the correct starting point.
         self._variants_cycle = cycle(_VALID_PLACEHOLDER_VARIANTS_ORDERED)
         while next(self._variants_cycle) != self.variant:
             pass
 
     def render(self) -> RenderableType:
+        """Render the placeholder.
+
+        Returns:
+            The value to render.
```
rodrigogiraoserrao, line 131:
```suggestion
            The rendered placeholder.
```

src/textual/widgets/_pretty.py
```diff
@@ -30,8 +45,18 @@ def __init__(
         self._renderable = PrettyRenderable(object)
 
     def render(self) -> PrettyRenderable:
+        """Render the pretty-printed object.
+
+        Returns:
+            The value to render.
```
rodrigogiraoserrao, line 51:
```suggestion
            The rendered pretty-print.
```

src/textual/widgets/_text_log.py
```diff
@@ -27,7 +30,7 @@ class TextLog(ScrollView, can_focus=True):
     }
     """
 
-    max_lines: var[int | None] = var(None)
+    max_lines: var[int | None] = var[Optional[int]](None)
```
rodrigogiraoserrao, line 33: As far as my understanding goes, `var` is just an "alias" for a quick way to create a reactive attribute with some parameters in pre-defined values.
Would it be worth typing `max_lines` as `max_lines: reactive[int | None] = var[Optional[int]](None)`?
(This was not a rhetoric question.)

src/textual/widgets/_text_log.py
```diff
@@ -54,10 +57,10 @@ def __init__(
...
             highlight: Automatically highlight content.
             markup: Apply Rich console markup.
-            name: The name of the button.
-            id: The ID of the button in the DOM.
-            classes: The CSS classes of the button.
-            disabled: Whether the button is disabled or not.
+            name: The name of the text log.
+            id: The ID of the text log in the DOM.
+            classes: The CSS classes of the text log.
+            disabled: Whether the text log is disabled or not.
```
rodrigogiraoserrao, line 63: Good catch.

src/textual/widgets/_tree.py
```diff
@@ -498,8 +534,10 @@ def last_line(self) -> int:
         """The index of the last line."""
         return len(self._tree_lines) - 1
 
-    def process_label(self, label: TextType):
-        """Process a str or Text in to a label. Maybe overridden in a subclass to change modify how labels are rendered.
+    def process_label(self, label: TextType) -> Text:
+        """Process a `str` or `Text` value in to a label.
```
rodrigogiraoserrao, line 538:
```suggestion
        """Process a `str` or `Text` value into a label.
```

src/textual/widgets/_tree.py
```diff
@@ -498,8 +534,10 @@ def last_line(self) -> int:
...
+
+        Maybe overridden in a subclass to change modify how labels are rendered.
```
rodrigogiraoserrao, line 540:
```suggestion
        Maybe overridden in a subclass to change how labels are rendered.
```

rodrigogiraoserrao (approved):

src/textual/widget.py
```diff
@@ -1436,17 +1436,16 @@ def _scroll_to(
...
-            y: Y coordinate (row) to scroll to, or None for no change. Defaults to None.
-            animate: Animate to new scroll position. Defaults to True.
-            speed: Speed of scroll if animate is True. Or None to use duration.
-            duration: Duration of animation, if animate is True and speed is None.
-            easing: An easing method for the scrolling animation. Defaults to "None",
-                which will result in Textual choosing the default scrolling easing function.
-            force: Force scrolling even when prohibited by overflow styling. Defaults to `False`.
+            x: X coordinate (column) to scroll to, or `None` for no change.
+            y: Y coordinate (row) to scroll to, or `None` for no change.
+            animate: Animate to new scroll position.
```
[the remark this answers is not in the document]
davep: Happy to have improvements made to it; I was just cleaning up prior docs really.

src/textual/widgets/_header.py
davep: Good question; I was asking myself the same thing as I was going through the docstrings and seeing a difference. I tested it and from what I can tell there's zero difference; both have the same effect (my uninformed guess here would be it's mkdocstrings or something supporting a couple of other doc systems' styles).

src/textual/widgets/_text_log.py
davep: What benefit do we get from that? A downside I can see is that it might be confusing to a reader of the code, seeing two "different" types like that.

src/textual/widgets/_placeholder.py
davep: Missed this. Yeah, I was debating that too.

Issue #1904, opened 2023-02-28
radio_set.py example glitches when scrollbars appear


willmcgugan:
When scrollbars are added to the radio set widget, the display glitches until it is refreshed.

This is unlikely to be an issue with the RadioSet widget itself.

![Image](https://user-images.githubusercontent.com/554369/221926785-8540b3e5-6e2c-4496-b5ad-8baf45972f7e.png)

PR #1910, opened 2023-03-01
Fix content width

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/1904
Fixes https://github.com/Textualize/textual/issues/1920


src/textual/scroll_view.py
```diff
@@ -82,7 +82,7 @@ def _size_updated(
         Returns:
             True if anything changed, or False if nothing changed.
         """
-        if self._size != size or container_size != container_size:
+        if self._size != size or self._container_size != container_size:
```
rodrigogiraoserrao, line 85: Good catch.

src/textual/widget.py
```diff
@@ -978,34 +977,36 @@ def _refresh_scrollbars(self) -> None:
         styles = self.styles
         overflow_x = styles.overflow_x
         overflow_y = styles.overflow_y
-        width, height = self.container_size
+        width, height = self._container_size
 
-        show_horizontal = self.show_horizontal_scrollbar
+        show_horizontal = False
```
rodrigogiraoserrao, line 982: Can you help me understand why the default behaviour is to turn off `show_horizontal` instead of leaving the previous value?

src/textual/widget.py
```diff
@@ -515,6 +515,8 @@ def _arrange(self, size: Size) -> DockArrangeResult:
     def _clear_arrangement_cache(self) -> None:
         """Clear arrangement cache, forcing a new arrange operation."""
         self._arrangement_cache.clear()
+        self._stabilized_scrollbar_size = None
+        self._scrollbar_stabilizer = 0
```
rodrigogiraoserrao, line 519: Consider adding docstrings to these two attributes.
`self._stabilized_scrollbar_size` is already typed in the body of `Widget` but `_scrollbar_stabilizer` is brand new.

tests/snapshot_tests/test_snapshots.py
```diff
@@ -256,3 +256,7 @@ def test_disabled_widgets(snap_compare):
 
 def test_focus_component_class(snap_compare):
     assert snap_compare(SNAPSHOT_APPS_DIR / "focus_component_class.py", press=["tab"])
+
+
+def test_line_api_scrollbars(snap_compare):
+    assert snap_compare(SNAPSHOT_APPS_DIR / "test_line_api_scrollbars.py", press=["_"])
```
rodrigogiraoserrao, line 262: The app examples in `snapshot_apps` don't have the leading `test_` in their names.
I don't think this will break anything, but it may create the expectation (for humans and for `pytest`) that there are tests written in this file.

rodrigogiraoserrao (approved):

src/textual/widget.py
willmcgugan: The previous value shouldn't come in to consideration.  It should make the same calculation, no matter the current state of the scrollbars.

PR #1913, opened 2023-03-01
Add a Makefile target for installing pre-commit hooks

davep:
Any time I cause my venv to be rebuilt, I need to remember to do this, and I always have to search my shell history to remind myself what the command is. That feels like it deserves to go into the Makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,3 +76,7 @@ setup:
 .PHONY: update
 update:
 	poetry update
+
+.PHONY: install-pre-commit
+install-pre-commit:
+	$(run) pre-commit install


rodrigogiraoserrao (approved):

rodrigogiraoserrao:
> Any time I cause my venv to be rebuilt, I need to remember to do this, and I always have to search my shell history to remind myself what the command is. That feels like it deserves to go into the Makefile.

Ironically, both the make command and the original one use the exact same words! 😆 
But I think this is good, because the makefile can be used as a reference if it is complete.

davep:
> But I think this is good, because the makefile can be used as a reference if it is complete.

Yup, and any reasonable shell will auto-complete targets from `Makefile` when typing `make` and hitting <kbd>Tab</tab>. Also at some point I should probably drag this:

```make
.PHONY: help
help:				# Display this help
	@grep -Eh "^[a-z]+:.+# " $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.+# "}; {printf "%-20s %s\n", $$1, $$2}'
```

into our `Makefile`, add "docstrings" to each of the targets and then `make help` will document all the targets (see [this Makefile in one of my own projects](https://github.com/davep/oidia/blob/main/Makefile) as an example).

rodrigogiraoserrao:
> [...] then `make help` will document all the targets (see [this Makefile in one of my own projects](https://github.com/davep/oidia/blob/main/Makefile) as an example).

Looks like an excellent idea!

rodrigogiraoserrao:
> The only downside I see here is that it's unlikely to play nicely out of the box on Windows; but then I'm not sure if the `Makefile` as was, and as is, plays nicely there anyway.

I've used makefile with some degree of success on Windows... But just because it doesn't work perfectly on Windows, it doesn't mean we can't make the lives of Mac/Linux users even better :D

davep:
True. It's more about the availability of awk, and sort (if Windows still has it? Pretty sure DOS did), really. 

But yeah, if taken as a Un*x convenience...

PR #1915, opened 2023-03-01
Interlink the docs with the Python and Rich manuals

davep:
All through our docs there are mentions of Python types and also Rich types. This change will make the vast majority of mentions of them into actual links that people can follow. So now, for example, when someone sees a method that returns a `Style`, they can click on it and see what a `Style` actually is.

### DIFF
--- a/mkdocs-common.yml
+++ b/mkdocs-common.yml
@@ -67,6 +67,9 @@ plugins:
     default_handler: python
     handlers:
       python:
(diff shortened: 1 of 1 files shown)


willmcgugan (approved):
Nice

rodrigogiraoserrao (approved):
It's in alphabetical order and all!

PR #1916, opened 2023-03-01
Button docs update

davep:
Quick end-of-day tweak to the `Button` docs; getting some more information about the types in the module into the docs so there's more inter-linking going on.


davep:
Ugh, actually, I think some sort of merge issue happened here too. Looks like I might have branched off the big earlier widget docs branch. If it looks like this won't merge into `main` well I can quickly recreate.

This is what you get for doing a thing just before heading for the bus.

davep:
Actually, on second thoughts, scratch it. I'll nuke this and do again.

PR #1922, opened 2023-03-02
Explicitly type attribute `sender` in messages.

### LINKED ISSUE #1813: [Typing] Add type hint to `sender` in widget messages?
The class `Message` defines the attribute `sender` as being of the type `MessageTarget`, which is a fairly generic protocol. Fine.

When writing custom messages, wouldn't it make sense to add an explicit type hint to mark the class variable `sender` as being of the type of the widget within which the message is being defined?

E.g., won't this make sense:

```py
class MyWidget(Widget):
   class MyMessage(Message):

        sender: MyWidget  # Isn't this something sensible to do?

        def __init__(self, sender: MyWidget, ...):
            ...
```

By adding `sender: MyWidget` in the body of the class, IDEs and static checkers will let me make use of `.sender` as of being the type I already _know_ it is.

rodrigogiraoserrao:
This will fix #1813.

This PR also removes some superfluous event attributes, like `Button.Pressed.button` or `Input.Changed.input`.
These are noted in the CHANGELOG.

Finally, this PR makes `DataTable` events generics over the same data type as the `DataTable` itself, akin to how `Tree` does it.

I typechecked all my changes and I did not introduce new type errors.
(Or rather, I did but I fixed them before opening the PR! 🤪 )

### DIFF
--- a/src/textual/widgets/_data_table.py
+++ b/src/textual/widgets/_data_table.py
@@ -3,7 +3,9 @@ from __future__ import annotations
 import functools
 from dataclasses import dataclass
 from itertools import chain, zip_longest
+from multiprocessing import Event
 from operator import itemgetter
+from tkinter import E
 from typing import Any, ClassVar, Generic, Iterable, NamedTuple, TypeVar, cast
 
 import rich.repr
@@ -38,6 +40,9 @@ RowCacheKey: TypeAlias = (
 )
 CursorType = Literal["cell", "row", "column", "none"]
 CellType = TypeVar("CellType")
+"""Used to type a `DataTable` as a container of cells."""
+EventCellType = TypeVar("EventCellType")
+"""Used to type the sender attribute of `DataTable`-related events."""
 
 CELL_X_PADDING = 2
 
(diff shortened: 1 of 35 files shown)


src/textual/widgets/_data_table.py
```diff
@@ -3,7 +3,9 @@
 import functools
 from dataclasses import dataclass
 from itertools import chain, zip_longest
+from multiprocessing import Event
```
willmcgugan, line 6: If you have some kind of auto-import adder in your IDE. I think you should turn it off. Pretty sure we don't use multiprocessing here.

willmcgugan (commented):
After some thought, I don't think we can merge this.

More accurate typing is generally a win, but `sender` is an internal detail. It just so happens that `sender` is the thing you interacted with for most widgets, but we don't guarantee that. In the future, maybe something else sends the message. Or we have different child widgets send the event, but we want them to be treated in the same way by a parent. In that case `sender` may be of different types. 

There's also the issue that `event.sender.id` could be anything, but `event.input.id` is self documenting.

src/textual/widgets/_data_table.py
rodrigogiraoserrao: I _also_ don't know how this ended up here! 😅 
I _might_ have some auto-import thing turned on.

davep: > I might have some auto-import thing turned on.

I *think* it's part of pylance/pyright; it ha a habit of doing the same thing via eglot in Emacs. Generally it only does the import if you pick an auto-imported completion (in Emacs it's generally easy to see where it's comping from so you can avoid it), and normally I noticed if it's managed to sneak one in when it comes to reviewing the commit in magit (atomic commits for the win here, of course).

What I've yet to find is how to tell pylance/pyright to turn it off. There is apparently a setting (`python.analysis.autoImportCompletions`) but I've yet to figure out a non-project-specific way of configuring it.

PR #1926, opened 2023-03-02
Minor `keys.py` tidy up.

### LINKED ISSUE #1815: Pilot cannot press key minus
The way `App._press_keys` is implemented is flawed in here:

https://github.com/Textualize/textual/blob/ba6c8afac6ffb78dffeab9ebdc7c115fd28cd40b/src/textual/app.py#L830-L842

There is a clear bug because the transformation done on `key` with two consecutive replacements cannot be undone in line 840 with a single replacement.
In particular, the keys HYPHEN-MINUS and REVERSE SOLIDUS show that the underscores that we get after the two consecutive replacements cannot be unambiguously reverted.

On top of that, I have to understand what is the issue with a key that has length 1 and is not "alnum".
For example, `"-"` has length 1, is not "alnum", and could very well be pressed as is. Maybe we want to check if `key.isprintable` instead?

### LINKED ISSUE #1830: Duplicated code to handle (sequences of) key presses
There is definitely some duplicated code in the codebase with regards to how keys are handled and converted from codepoints to events.
Consider these four instances, which are not necessarily the only ones:

https://github.com/Textualize/textual/blob/be850635c2d764dfa4d45707a0cb92bfd63f355d/src/textual/app.py#L830-L837

https://github.com/Textualize/textual/blob/main/src/textual/_xterm_parser.py#L253-L264

https://github.com/Textualize/textual/blob/be850635c2d764dfa4d45707a0cb92bfd63f355d/src/textual/keys.py#L250

https://github.com/Textualize/textual/blob/be850635c2d764dfa4d45707a0cb92bfd63f355d/src/textual/keys.py#L266-269

I see lots of very similar bits and bobs, and things like multiple attempts at normalising key/codepoint names, etc.
These should probably be refactored into methods that we can use that will ensure consistency.

Related to: #1815

rodrigogiraoserrao:
This will fix #1815 and will fix #1830.

 - This PR adds a helper function to `keys.py` (`_get_unicode_name_from_key`) that tries its best to undo the transformation done by `_character_to_key`.
 - We use `_character_to_key` more consistently throughout the code, as it was essentially reimplemented in a couple of places.
 - We use `_get_unicode_name_from_key` when we need to revert the transformation done by `_character_to_key`.


rodrigogiraoserrao:
For reference, the helper code that was used to generate `KEY_TO_UNICODE_NAME` was:

```py
import unicodedata as u
import string

for char in string.punctuation:
    name = u.name(char)
    if " " in name or "-" in name:
        friendly = name.lower().replace("-", "_").replace(" ", "_")
        print(f'"{friendly}": "{name}",')
```

src/textual/app.py
```diff
@@ -864,19 +869,15 @@ async def _press_keys(self, keys: Iterable[str]) -> None:
...
                 if len(key) == 1 and not key.isalnum():
-                    key = (
-                        unicodedata.name(key)
-                        .lower()
-                        .replace("-", "_")
-                        .replace(" ", "_")
-                    )
+                    key = _character_to_key(key)
                 original_key = REPLACED_KEYS.get(key, key)
+                print(f"original key is {original_key}")
```
willmcgugan, line 874: Can we combine these two print statements? The intent is just to show a little information re the keys pressed in the docs build output.

willmcgugan (approved):

Issue #1805, opened 2023-02-15
Wrong type hint of DOMNode.watch method


mzebrak:
It looks like after the latest changes `0.11.0` (because in 0.10.1 there was no issue), the wrong type hint was used for DOMNode.watch method.

After upgrading textual in my project I see mypy errorrs like below:

```bash
Argument 3 to "watch" of "DOMNode" has incompatible type "Callable[[Any], None]"; expected "Union[Callable[[], Awaitable[None]], Callable[[], None]]"  [arg-type]
```

After some investigation, argument no.3 in the `0.11.0` is:
```python
callback: CallbackType,
```

which is:
```python
CallbackType = Union[Callable[[], Awaitable[None]], Callable[[], None]]
```

This is not true because Callabe can take additional arguments - current_value and previous_value (both of the Any type).

Previously, in the `reactive.watch` method, it was: 
```python
 callback: Callable[[Any], object]
 ```

**Comment by mzebrak** [comment]
@willmcgugan  This one is not fixed in 0.13.0 so shouldn't be closed.

**Comment by mzebrak** [comment]
0.14.0 fixed most of the errors, but there is still one minor issue:
```bash
error: Argument "callback" to "watch" of "DOMNode" has incompatible type "Callable[[], None]"; expected "Union[Callable[[Any], Awaitable[None]], Callable[[Any], None]]"  [arg-type]
```

It shows up when I attach callack method without any params:

```python
self.watch(self.app, "something", callback=self.__sync_cart_items)

def __sync_cart_items(self) -> None:
    pass
```

I have to apply workaround such as:

```python
self.watch(self.app, "something", callback=self.__sync_cart_items)

def __sync_cart_items(self, _: Something) -> None:
    pass
```

PR #1929, opened 2023-03-03
Added typing for watch callbacks

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/1805

I think we might adopt a policy of requiring a comment explaining why `type: ignore` was necessary.

### DIFF
--- a/src/textual/_types.py
+++ b/src/textual/_types.py
@@ -1,4 +1,4 @@
-from typing import TYPE_CHECKING, Awaitable, Callable, List, Union
+from typing import TYPE_CHECKING, Any, Awaitable, Callable, List, Union
 
 from rich.segment import Segment
 from typing_extensions import Protocol
@@ -28,3 +28,4 @@ class EventTarget(Protocol):
 
 SegmentLines = List[List["Segment"]]
 CallbackType = Union[Callable[[], Awaitable[None]], Callable[[], None]]
+WatchCallbackType = Union[Callable[[Any], Awaitable[None]], Callable[[Any], None]]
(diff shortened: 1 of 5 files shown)


src/textual/_types.py
```diff
@@ -28,3 +28,4 @@ def post_message_no_wait(self, message: "Message") -> bool:
 
 SegmentLines = List[List["Segment"]]
 CallbackType = Union[Callable[[], Awaitable[None]], Callable[[], None]]
+WatchCallbackType = Union[Callable[[Any], Awaitable[None]], Callable[[Any], None]]
```
rodrigogiraoserrao, line 31: I'm still not familiarised with how `ParamSpec` works, but wouldn't this be the type of place where it makes sense, to somehow tie the signature of the function that is the callback to the place where it gets called?

rodrigogiraoserrao (approved):
Looks good to me.
I did leave a question about `ParamSpec`, but that is because I don't know how it works.

rodrigogiraoserrao:
> I think we might adopt a policy of requiring a comment explaining why `type: ignore` was necessary.

I like this.

src/textual/_types.py
willmcgugan: I don't think it would work here. At the point where you add watchers, it is too dynamic to type. You do something like `self.watch(app, "dark", on_dark_change)`, but the typing information isn't there at that point. You can only know the type of "dark" at runtime.

rodrigogiraoserrao: I see where you're going. Thanks!

PR #1930, opened 2023-03-04
Codebase Refactor

HighnessAtharva:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=LKXDgEH)
Several improvements to make the code more Pythonic. Few of these include:-

- Use of f-strings everywhere instead of string concatenation
- Merging several if conditionals and usage of if expressions instead of complicated nested structures
- Replacing == with is wherever applicable Inline returns from a function
- Overall, better readable and maintainable code structure.

- [ ] Updated documentation

### DIFF
--- a/examples/dictionary.py
+++ b/examples/dictionary.py
@@ -56,17 +56,16 @@ class DictionaryApp(App):
         """Convert the results in to markdown."""
         lines = []
         if isinstance(results, dict):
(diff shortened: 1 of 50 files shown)


willmcgugan:
I'm assuming this is some automated refactoring, because in almost all cases it has made the code worse or broken.

Please do not submit PRs like this.

PR #1935, opened 2023-03-04
Add an example of using a RadioSet.Changed message

davep:
Unlike a few other widgets, the RadioSet is pretty much all about reacting to the selection result; the question of how you go about it has already come up and while the message is documented, complete with all properties, it can't hurt to have an illustrative example of code that uses it.

Here I add an extra RadioSet example that sits with the message in the reference. This should help the reader better follow how to use it, and also gives something to link to if someone hasn't got that far into the documentation yet but is attempting to use the RadioSet.

### DIFF
--- a/docs/examples/widgets/radio_set_changed.py
+++ b/docs/examples/widgets/radio_set_changed.py
@@ -1,0 +1,43 @@
+from textual.app import App, ComposeResult
+from textual.containers import Horizontal, Vertical
+from textual.widgets import Label, RadioButton, RadioSet
+
+
+class RadioSetChangedApp(App[None]):
+    CSS_PATH = "radio_set_changed.css"
+
+    def compose(self) -> ComposeResult:
+        with Vertical():
+            with Horizontal():
+                with RadioSet():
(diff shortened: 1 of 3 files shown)


docs/examples/widgets/radio_set_changed.py
```diff
@@ -0,0 +1,43 @@
...
+            with Horizontal():
+                yield Label(id="pressed")
+            with Horizontal():
+                yield Label(id="index")
+
+    def on_mount(self) -> None:
+        self.query_one("#focus_me", RadioButton).focus()
+
+    def on_radio_set_changed(self, event: RadioSet.Changed) -> None:
+        self.query_one("#pressed", Label).update(
```
rodrigogiraoserrao, line 34: Oh, I didn't know I could write queries like this! Cool 👍

rodrigogiraoserrao (approved):
Great idea.

Do you think it is worth highlighting the line(s) that implement the handler and/or add a written note in the example? (Those that are marked with `# (n)!` in the code.)

willmcgugan (approved):

PR #1940, opened 2023-03-05
Drop explicit sender attribute from messages

willmcgugan:
Rectifies two API issues.

The `sender` attribute create boiler plate when sending messages. It was also a source of confusion as to its purpose.

Having two messages `post_message` and `post_message_no_wait` is rather ugly, and difficult to explain. Awaiting posting something on a message queue is only necessary if the queue is bounded, but they aren't any more.

- Dropped `sender` from the Message base class.
- Dropped the async `post_message`. There is now just `post_message` which is syncronous. 
- Added a `control` alias to messages from controls.

See CHANGELOG for details

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [0.14.0] - Unreleased
+
+### Changes
+
+- There is now only `post_message` to post events, which is non-async https://github.com/Textualize/textual/pull/1940
+- The Timer class now has just one method to stop it, `Timer.stop` which is non sync https://github.com/Textualize/textual/pull/1940
+- Messages don't require a `sender` in their constructor https://github.com/Textualize/textual/pull/1940
+
 ## [0.13.0] - 2023-03-02
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,15 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Messages don't require a `sender` in their constructor https://github.com/Textualize/textual/pull/1940
 - Many messages have grown a `control` property which returns the control they relate to. https://github.com/Textualize/textual/pull/1940
 
+
+### Added
+
+- Added `data_table` attribute to DataTable events https://github.com/Textualize/textual/pull/1940
+- Added `list_view` attribute to `ListView` events https://github.com/Textualize/textual/pull/1940
+- Added `radio_set` attribute to `RadioSet` events https://github.com/Textualize/textual/pull/1940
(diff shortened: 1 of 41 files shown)


davep:
I'll review properly tomorrow morning, if this doesn't go in beforehand, but much like with the alternate compose method I'm throwing a very hearty OH HELL YES at this.

CHANGELOG.md
```diff
@@ -5,6 +5,24 @@ All notable changes to this project will be documented in this file.
...
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [0.14.0] - Unreleased
+
+### Changes
+
+- There is now only `post_message` to post events, which is non-async https://github.com/Textualize/textual/pull/1940
+- The Timer class now has just one method to stop it, `Timer.stop` which is non sync https://github.com/Textualize/textual/pull/1940
+- Messages don't require a `sender` in their constructor https://github.com/Textualize/textual/pull/1940
+- Many messages have grown a `control` property which returns the control they relate to. https://github.com/Textualize/textual/pull/1940
```
davep, line 15: Elsewhere in the CHANGELOG we've adopted the convention of prefixing breaking changes with `Breaking change:` -- perhaps a good idea to do this with this changes too?

willmcgugan: ok

src/textual/widgets/_radio_set.py
```diff
@@ -114,16 +113,14 @@ def on_radio_button_changed(self, event: RadioButton.Changed) -> None:
...
         """
         # If the button is changing to be the pressed button...
-        if event.input.value:
+        if event.radio_button.value:
             # ...send off a message to say that the pressed state has
             # changed.
-            self.post_message_no_wait(
-                self.Changed(self, cast(RadioButton, event.input))
-            )
+            self.post_message(self.Changed(self, cast(RadioButton, event.radio_button)))
```
davep, line 119: I suspect we don't need this cast now.

src/textual/widgets/_toggle_button.py
```diff
@@ -218,19 +218,24 @@ def on_click(self) -> None:
...
             Args:
-                sender: The toggle button sending the message.
+                toggle_button: The toggle button sending the message.
                 value: The value of the toggle button.
             """
-            super().__init__(sender)
-            self.input = sender
+            super().__init__()
+            self._toggle_button = toggle_button
             """A reference to the toggle button that was changed."""
```
davep, line 230: This docstring can go now.

src/textual/widgets/_toggle_button.py
```diff
@@ -218,19 +218,24 @@ def on_click(self) -> None:
...
             self.value = value
             """The value of the toggle button after the change."""
 
+        @property
+        def control(self) -> ToggleButton:
+            """Alias for self.toggle_button."""
```
davep, line 236: This will need a reword given that `toggle_button` was moved to being "private". Actually, I'm wondering if `control` should move to the two individual child message classes.

davep (approved):
Couple of tweaks needed around `ToggleButton` and friends, but otherwise looks good. This is a really nice spring-clean!

Issue #1863, opened 2023-02-23
Add a percentage to the border definition to apply an alpha


willmcgugan:
It's often a requirement to set a border that is muted a little, but there is no easy way of applying an alpha to a border if it is not already set.

I think we should add an optional percentage to the border definition, to multiple the borders alpha. Something like the following:

```
border: solid $foreground 50%;
```

rodrigogiraoserrao:
Just to make sure, you _can_ mute the border a little if you bake the alpha value into the colour itself, right?

willmcgugan:
Yeah, you can set a color with alpha. But you apply new alpha to that color.

PR #1954, opened 2023-03-06
Border colour percentage

davep:
Adds the ability to say something like:

```scss
Widget {
    border: solid red 50%;
}
```

See #1863.


docs/styles/border.md
```diff
@@ -126,7 +127,7 @@ This example also shows that a widget cannot contain both a `border` and an `out
 /* Set a heavy white border */
 border: heavy white;
 
-/* set a red border on the left */
+/* Set a red border on the left */
 border-left: outer red;
```
darrenburns, line 131: Worth adding an example of using a percentage here?

src/textual/css/_styles_builder.py
```diff
@@ -441,6 +441,7 @@ def _process_space_partial(self, name: str, tokens: list[Token]) -> None:
     def _parse_border(self, name: str, tokens: list[Token]) -> BorderValue:
         border_type: EdgeType = "solid"
         border_color = Color(0, 255, 0)
+        border_alpha: float | None = None
```
darrenburns, line 444: Is the None here instead of 100.0 to differentiate between someone who has explicitly set 100% and someone who hasn't gave a percentage?

Would it make sense to default it to 100.0?

Have you checked the interactions when we default to None here and a user tries to animate the border percentage to e.g. 0% using code?

darrenburns (commented):
Looks good - left a few questions and comments.

Does this work with online styles? Can I update the % using widget.styles.border_left for example?

davep:
> Does this work with online styles? Can I update the % using widget.styles.border_left for example?

No, this is just for CSS (by design -- checked with @willmcgugan on that this morning).

docs/styles/border.md
davep: Aye, can't hurt to.

src/textual/css/_styles_builder.py
davep: > Is the None here instead of 100.0 to differentiate between someone who has explicitly set 100% and someone who hasn't gave a percentage?
>
> Would it make sense to default it to 100.0?

Aye, it started out that way (following the example set by the `color` style) but now you mention it it can likely be reduced to that.

davep: Actually, thinking about it some more, I think it's better as it is in that if we go with `None` it saves the need to create a fresh instance of a `Color` just to end up with the same `Color`. Would save a wee bit of time I think?

davep: > vHave you checked the interactions when we default to None here and a user tries to animate the border percentage to e.g. 0% using code?

This part I don't think I follow.

darrenburns: The part about animation is less relevant if this can't be set inline.

src/textual/css/_styles_builder.py
```diff
@@ -462,9 +463,18 @@ def border_value_error():
...
+                alpha_scalar = Scalar.parse(token.value)
+                if alpha_scalar.unit != Unit.PERCENT:
+                    self.error(name, token, "alpha must be given as a percentage.")
+                border_alpha = alpha_scalar.value / 100.0
+
             else:
                 border_value_error()
 
+        if border_alpha is not None:
+            border_color = border_color.with_alpha(border_alpha)
```
willmcgugan, line 476: I don't think you should set the alpha, but rather multiply it.

If a color has 80% alpha, and you set the alpha value to 50%, then the border color should have 40% alpha.

willmcgugan (changes requested):

src/textual/css/_styles_builder.py
davep: As discussed in person, this was following what the `color` property does in CSS right now, so that it worked the same.

As a followup from that conversation, we're going to change how `color` works so that it does the multiplication too, and then this will be modified to work the same (it's likely that we'll have `Color` grow a `multiply_alpha` method, or perhaps even implement `__mul__` (I think I prefer the former over the latter).

willmcgugan (approved):

PR #1959, opened 2023-03-07
Extend the centering Q in the FAQ

davep:
Add an example of how to center multiple widgets, rather than center the bounding box of multiple widgets.

### DIFF
--- a/questions/align-center-middle.question.md
+++ b/questions/align-center-middle.question.md
@@ -2,8 +2,11 @@
 title: "How do I center a widget in a screen?"
 alt_titles:
   - "centre a widget"
+  - "centre a widgets"
   - "center a control"
   - "centre a control"
+  - "center controls"
+  - "centre controls"
 ---
 
 To center a widget within a container use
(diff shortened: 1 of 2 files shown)


questions/align-center-middle.question.md
```diff
@@ -32,3 +35,71 @@ class ButtonApp(App):
 if __name__ == "__main__":
     ButtonApp().run()
 ```
+
+If you use the above on multiple widgets, you'll find they appear to
+"left-align" in the center of the screen, like this:
```
darrenburns, line 40: "Left align in the center of the screen" sounds strange.

darrenburns (changes requested):

questions/align-center-middle.question.md
davep: This is how folk have termed it on a couple of occasions when asking about it, so it's as much faqtory-bait as anything.

darrenburns (approved):

Issue #1944, opened 2023-03-06
Warn users about macOS Terminal


willmcgugan:
I'm thinking we might detect macOS terminal when running `textual run` and direct them to a better terminal, like iTerm.

Otherwise, I suspect many people wouldn't realise it could look a lot better.

On exit from `textual run`, we should print out a short message which a link to our site somewhere. Suggest adding a FAQ entry.

PR #1964, opened 2023-03-07
Add a post-run warning hook to `textual run`

davep:
Implements #1944 and in doing so adds a place to add more checks and warnings to a list. The idea here being that if a user is using `textual run`, on exit, problematic environments can be tested for and warnings displayed.

Right now the only warning added is for using `terminal.app` under macOS.

This also adds a new entry to the FAQ explaining how `terminal.app` isn't good for Textual.

### DIFF
--- a/src/textual/cli/cli.py
+++ b/src/textual/cli/cli.py
@@ -49,6 +49,7 @@ def _post_run_warnings() -> None:
     import platform
 
     from rich.console import Console
+    from rich.panel import Panel
 
     console = Console()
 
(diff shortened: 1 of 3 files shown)


src/textual/cli/cli.py
```diff
@@ -38,6 +38,38 @@ def console(verbose: bool, exclude: list[str]) -> None:
...
+    warnings = [
+        (
+            platform.system() == "Darwin"
+            and os.environ.get("TERM_PROGRAM") == "Apple_Terminal",
+            "The default terminal app on macOS is limited to 256 colors. See our FAQ for more details:\n\n"
+            "https://github.com/Textualize/textual/blob/main/FAQ.md#why-doesn't-textual-look-good-on-macos",
+        )
+    ]
+
+    for concering, concern in warnings:
```
willmcgugan, line 68: typo

willmcgugan (approved):
Just a typo

src/textual/cli/cli.py
davep: Excellent spot!

PR #1968, opened 2023-03-07
Generate MouseUp events

darrenburns:

- [ ] Updated documentation


rodrigogiraoserrao (commented):
Why don't we need to also modify the `windows_driver.py` implementation?

davep:
> Why don't we need to also modify the windows_driver.py implementation?

@rodrigogiraoserrao Not speaking for Darren, of course, but as I look at the changes they're in the core driver, not any platform-specific driver. The only change I see to the Linux driver is to remove an unused import.

willmcgugan:
Yeah, that looks to be the case. Could use a test on Windows just to be sure.

willmcgugan (approved):

PR #1971, opened 2023-03-07
Be more assertive when shutting down

### LINKED ISSUE #1949: CSS errors printed twice
Sometimes CSS errors are duplicated.

```python
from textual.widgets import Label, Input


class InputWithLabel(Widget):
    """An input with a label."""

    def __init__(self, input_label: str) -> None:
        self.input_label = input_label
        super().__init__()

    DEFAULT_CSS = """
    InputWithLabel {
        layout: horizontal;        
        height: 3;
        
    }
    InputWithLabel > * {
        width: 1fr;        
    }
    InputWithLabel Label {
        text-align: right
        widget: 1fr;
    }
    
    """

        yield Label(self.input_label)
        yield Input()


class CompoundApp(App):
        yield InputWithLabel("Fist Name")
        yield InputWithLabel("Last Name")
        yield InputWithLabel("Email")


if __name__ == "__main__":
    app = CompoundApp()
    app.run()
```

Image

rodrigogiraoserrao:
Possibly fixes #1949 

This feels more like a patch and less like a proper fix.
Why?

When running the example app (below) you would eventually try to shutdown the app via `_shutdown`, which would in turn try to close all the message pumps.
Even if you say you do NOT want to wait for the message pumps to process previous messages, we close the message pump for the app screen and then, later on, the code starts running `Screen._process_messages` either way, which will then trigger the composition of the screen, which would raise the second instance of the CSS error.
That is why I added the check

```py
if self.app._closing or self.app._closed: return
```

to the method `MessagePump._process_messages`.

What is funny is that I _cannot_ tell why `Screen._process_messages` is running.
I tried printing the call stack and I get this:

```py
File "myapp.py", line 31, in <module>
    app.run()
File "/Users/rodrigogs/Documents/textualize/textual/src/textual/app.py", line 1044, in run
    event_loop.run_until_complete(run_app())
File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/asyncio/base_events.py", line 574, in run_until_complete
    self.run_forever()
File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/asyncio/base_events.py", line 541, in run_forever
    self._run_once()
File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/asyncio/base_events.py", line 1786, in _run_once
    handle._run()
File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/asyncio/events.py", line 88, in _run
    self._context.run(self._callback, *self._args)
File "/Users/rodrigogs/Documents/textualize/textual/src/textual/message_pump.py", line 420, in _process_messages
    for line in traceback.format_stack():
```

It is just a bunch of `asyncio` calls and our very own `run_app` call...

<details>
<summary>Example app</summary>

```py


class InputWithLabel(Widget):
    """An input with a label."""

    DEFAULT_CSS = """
    InputWithLabel {
        text-align: right
        width: 1fr;
    }
    """

        print("Composing widget")
        yield Label("oi")
        print("Done composing widget")


class CompoundApp(App):
        print("Composing custom app")
        yield InputWithLabel()
        print("Custom app composed.")


if __name__ == "__main__":
    app = CompoundApp()
    app.run()
```

</details>

### DIFF
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -985,7 +985,7 @@ class App(Generic[ReturnType], DOMNode):
             )
         finally:
             try:
-                if auto_pilot_task is not None:
+                if auto_pilot_task is not None and not self._exit_renderables:
                     await auto_pilot_task
             finally:
                 await app._shutdown()
@@ -1809,11 +1809,11 @@ class App(Generic[ReturnType], DOMNode):
             if isinstance(screen, Screen) and screen._running:
                 await self._prune_node(screen)
 
-        # Close any remaining nodes
-        # Should be empty by now
+        # Close any remaining nodes. This should be empty by now, unless we couldn't
+        # even compose the app cleanly when starting up.
         remaining_nodes = list(self._registry)
         for child in remaining_nodes:
-            await child._close_messages()
+            await child._close_messages(wait=False)
 
     async def _shutdown(self) -> None:
         self._begin_batch()  # Prevents any layout / repaint while shutting down
--- a/src/textual/message_pump.py
+++ b/src/textual/message_pump.py
@@ -372,7 +372,7 @@ class MessagePump(metaclass=MessagePumpMeta):
         """Request the message queue to immediately exit."""
         self._message_queue.put_nowait(messages.CloseMessages())
 
-    async def _on_close_messages(self, message: messages.CloseMessages) -> None:
+    async def _on_close_messages(self) -> None:
         await self._close_messages()
 
     async def _close_messages(self, wait: bool = True) -> None:
@@ -408,6 +408,10 @@ class MessagePump(metaclass=MessagePumpMeta):
             self._closed = True
 
     async def _process_messages(self) -> None:
+        # Don't even bother if the app is already shutting down.
+        if self.app._closing or self.app._closed:
+            return
+
         self._running = True
 
         await self._pre_process()


src/textual/app.py
```diff
@@ -985,7 +985,7 @@ async def run_auto_pilot(
             )
         finally:
             try:
-                if auto_pilot_task is not None:
+                if auto_pilot_task is not None and not self._exit_renderables:
```
rodrigogiraoserrao, line 988: I don't find this a very elegant way of checking if we are shutting down already, but it works...

src/textual/app.py
```diff
@@ -1809,11 +1809,11 @@ async def _close_all(self) -> None:
...
                 await self._prune_node(screen)
 
-        # Close any remaining nodes
-        # Should be empty by now
+        # Close any remaining nodes. This should be empty by now, unless we couldn't
+        # even compose the app cleanly when starting up.
         remaining_nodes = list(self._registry)
         for child in remaining_nodes:
-            await child._close_messages()
+            await child._close_messages(wait=False)
```
willmcgugan, line 1816: You may be misunderstanding this line. This method will always complete processing the message queue. The `wait` argument is so we can wait until it is complete.

Setting `wait=False` feels like it may have consequences.

willmcgugan (commented):
I think you may be on to something. But I think we need to dig a little deeper.

Could the issue be that if `pre_process` fails, it carries on to process further messages?

I'm wondering if there is an error in `pre_process` that the `_process_messages` should return abruptly.

src/textual/app.py
rodrigogiraoserrao: It definitely has consequences: if I remove it, the error gets printed twice again.
I think the point is that I do not wait for the messages to process neither now nor never.

rodrigogiraoserrao: > I think you may be on to something. But I think we need to dig a little deeper.

I think the same...
I'll take your thoughts into consideration.

rodrigogiraoserrao:
> I think you may be on to something. But I think we need to dig a little deeper.

I think the same...
I'll take your thoughts into consideration.

src/textual/app.py
willmcgugan: Right, but I fear that it will prevent things from shutting down cleanly. It's important to wait for those tasks to finish before exiting.

rodrigogiraoserrao:
Superseded by #1998.

Issue #1945, opened 2023-03-06
Switcher Widget


willmcgugan:
We need a widget to switch between child widgets. Not sure what we would call it yet.

Essentially a single child would be visible at one time. Setting a reactive attribute to an query selector would make a new pane visible and hide the others.

rodrigogiraoserrao:
Carousel? Gallery?

willmcgugan:
Kind of, but more fundamental than that. Think of it as the foundation of a tabbed widget.

PR #1983, opened 2023-03-08
Add `ContentSwitcher`

davep:
Implements the requirement outlined in #1945. Simply put: this PR adds a simple container widget (that inherits from `Container` and can be styled as desired) that hides all its immediate children, and expects them all to have a widget id. A `current` property is then set to switch between the different children. Essentially providing the foundation for a tabbed view or other similar widgets.


docs/widgets/content_switcher.md
```diff
@@ -0,0 +1,54 @@
...
+    ```{.textual path="docs/examples/widgets/content_switcher.py"}
+    ```
+
+=== "content_switcher.py"
+
+    ~~~python
+    --8<-- "docs/examples/widgets/content_switcher.py"
+    ~~~
+
+    1. A `Horizontal` to hold the buttons, which with a unique ID.
```
rodrigogiraoserrao, line 28:
```suggestion
    1. A `Horizontal` to hold the buttons, each with a unique ID.
```

docs/widgets/content_switcher.md
```diff
@@ -0,0 +1,54 @@
...
+    2. This button will select the `DataTable` in the `ContentSwitcher`.
+    3. This button will select the `Markdown` in the `ContentSwitcher`
```
rodrigogiraoserrao, line 30:
```suggestion
    3. This button will select the `Markdown` in the `ContentSwitcher`.
```

docs/widgets/content_switcher.md
```diff
@@ -0,0 +1,54 @@
...
+=== "content_switcher.css"
+
+    ~~~sass
+    --8<-- "docs/examples/widgets/content_switcher.css"
+    ~~~
+
+When the user presses the "Markdown" button the view is switched:
+
+```{.textual path="docs/examples/widgets/content_switcher.py" lines="40" press="tab,tab,enter"}
+```
```
rodrigogiraoserrao, line 43: Do you think it would look better if this was yet another tab?

Something like:

```suggestion
=== "After pressing 'Markdown'"

    ```{.textual path="docs/examples/widgets/content_switcher.py" lines="40" press="tab,tab,enter"}
    ```
```

docs/widgets/content_switcher.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Reactive Attributes
+
+| Name      | Type            | Default | Description                                                          |
+|-----------|-----------------|---------|----------------------------------------------------------------------|
+| `current` | `str` \| `None` | `None`  | The ID of the currently-visible child. `None` means nothing visible. |
```
rodrigogiraoserrao, line 49:
```suggestion
| `current` | `str` \| `None` | `None`  | The ID of the currently-visible child. `None` means nothing is visible. |
```

rodrigogiraoserrao (approved):
Good job. I like how the implementation turned out to be quite simple.

docs/widgets/content_switcher.md
davep: No. I much prefer it unembellished as it makes it very clear it's an illustration of what's above with zero code changes, just with user input, and with text beforehand making it clear what's about to follow.

rodrigogiraoserrao: Ok!

davep:
> Good job. I like how the implementation turned out to be quite simple.

Thanks. Was 100% one of those where it took *way* longer to think up the example and do the docs than it did to actually write the code for the widget itself.

willmcgugan:
> Thanks. Was 100% one of those where it took _way_ longer to think up the example and do the docs than it did to actually write the code for the widget itself.

That's a very good sign!

davep:
Really struggling to get the tests to pass with this, and it seems to be nothing at all to do with the code but with the timing of the snapshot tests. Looking at the failures, it seems to wander around operating system and Python version, which suggests it's down to timing. On the other hand it's *always* on the new snapshot test included in here. It works fine locally.

What's really odd though is I've downloaded the results a couple of times now and the cause of the difference was different both times.

- First time was the vertical scrollbar wasn't quite the same length.
- Second time the second button making the "tab bar" wasn't depressed.

These were with different pauses at the end.

At this point I'm stumped on how to make this pass in CI.

willmcgugan:
Suspect there may be something else at play here, beyond the occasional flakiness. Chat about it tomorrow.

PR #1985, opened 2023-03-08
Don't watch the generated mkdocs configs

davep:
I got a bit carried away with what to watch, and it looks like the mkdocs watch plugin doesn't much care for watching for files that don't exist but may.

@willmcgugan This should correct the error you were seeing when deploying.


willmcgugan (approved):

willmcgugan:
Thanks, Dave.

PR #1998, opened 2023-03-09
prevent double errors

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/1949

Here's what I think was happening:

When screen is composed, it registers children and starts their message pumps. Each of which will get a compose message.

When an app is composed it will parse the CSS (all of the CSS) and trigger the error. This is because parsing the CSS is a lazy operation, done on demand.

The solution is for the parser to store the fact that parsing failed, and not attempt to reparse it (because it can only fail again).

### DIFF
--- a/src/textual/css/stylesheet.py
+++ b/src/textual/css/stylesheet.py
@@ -146,7 +146,12 @@ class Stylesheet:
             List of rules sets for this stylesheet.
         """
         if self._require_parse:
-            self.parse()
+            try:
+                self.parse()
+            except Exception:
+                self._require_parse = False
+                self._rules = []
+                raise
             self._require_parse = False
         assert self._rules is not None
         return self._rules
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -1460,7 +1460,6 @@ class App(Generic[ReturnType], DOMNode):
         Args:
             error: An exception instance.
         """
-
         if hasattr(error, "__rich__"):
             # Exception has a rich method, so we can defer to that for the rendering
             self.panic(error)


rodrigogiraoserrao:
Ok... 🤷.

Ignoring the context of the issue, this makes sense: if parsing failed, don't bother trying again.
So, I think this change makes sense.

In the context of the issue, is this a better fix than trying to prevent the second parse from being triggered by exiting early inside `_process_messages`?

src/textual/css/stylesheet.py
```diff
@@ -304,10 +306,20 @@ def parse(self) -> None:
...
+            try:
+                css_rules = self._parse_rules(
+                    css,
+                    path,
+                    is_default_rules=is_default_rules,
+                    tie_breaker=tie_breaker,
+                )
+            except Exception:
+                self._invalid_css.add(css)
+                raise
```
rodrigogiraoserrao, line 322: Is this equivalent to the alternative below?

```py
except Exception as error:
    ...
    raise error from None
```

rodrigogiraoserrao:
I'm also trusting that this change will not interfere with the CSS watcher?
E.g., if I make an error while watching the CSS and then fix it, will Textual parse the file the second time?

willmcgugan:
> In the context of the issue, is this a better fix than trying to prevent the second parse from being triggered by exiting early inside `_process_messages`?

I don't think that is desirable. We might actually want further errors to be displayed. Just not the exact same error.

But I've not entirely abandoned the idea of bailing out early.

willmcgugan:
> I'm also trusting that this change will not interfere with the CSS watcher? E.g., if I make an error while watching the CSS and then fix it, will Textual parse the file the second time?

Shouldn't do. That attempts to create a new stylesheet, and parse it all over again.

src/textual/css/stylesheet.py
willmcgugan: I'm not entirely sure. Doesn't impact anything here since we never display the traceback.

PR #2001, opened 2023-03-09
More documentation tweaks

davep:
- Cleans up the titles of widgets in the reference section (makes them all their class name)
- Adds `ContentSwitcher` (sans screenshot for obvious reasons) to the widget gallery
- Tells mkdocs that it should watch the main nav file for changes


docs/widget_gallery.md
```diff
@@ -35,6 +35,14 @@ A classic checkbox control.
...
 
 
+## ContentSwitcher
+
+A widget for containing and switching display between multiple child
+widgets.
+
+[ContentSwitcher reference](./widgets/content_switcher.md){ .md-button .md-button--primary }
+
+
```
rodrigogiraoserrao, line 45: You didn't add an example because you don't think there is a canonical example to show here or because you forgot?

davep: Someone didn't read every single commit 😛

https://github.com/Textualize/textual/commit/91a33a45aa31393ae24df8abaaffb7716b0e730b

willmcgugan (approved):

docs/widget_gallery.md
rodrigogiraoserrao: Woops! Is it standard practise to go through all the commit messages when reviewing a PR?
I have to be honest: I never read them...

davep: > I never read them...

My neither; but I'm gonna be smug about this anyway. :-P

PR #2003, opened 2023-03-09
prevent stuck scrollbar

willmcgugan:
Fixes the issue when you grab a scrollbar, move outside the terminal, and release the mouse button. The scrollbar would get attached to the mouse until you clicked the button again.

There was a previous attempt at this in https://github.com/Textualize/textual/pull/1968 but it didn't quite do the job.

### DIFF
--- a/src/textual/driver.py
+++ b/src/textual/driver.py
@@ -27,8 +27,8 @@ class Driver(ABC):
         self._size = size
         self._loop = asyncio.get_running_loop()
         self._mouse_down_time = _clock.get_time_no_wait()
-        self._dragging = False
-        self._dragging_button = None
+        self._down_buttons: list[int] = []
+        self._last_move_event: events.MouseMove | None = None
 
     @property
     def is_headless(self) -> bool:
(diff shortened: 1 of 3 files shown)


src/textual/driver.py
```diff
@@ -44,29 +44,37 @@ def process_event(self, event: events.Event) -> None:
...
-                        ctrl=event.ctrl,
-                        screen_x=event.screen_x,
-                        screen_y=event.screen_y,
-                        style=event.style,
+            if (
+                self._down_buttons
+                and not event.button
+                and self._last_move_event is not None
+            ):
+                buttons = list(dict.fromkeys(self._down_buttons).keys())
```
rodrigogiraoserrao, line 58: Isn't this the same as `list(self._down_buttons)`?

rodrigogiraoserrao (approved):

src/textual/driver.py
willmcgugan: Nope, it will additionally unique-ify them. Wether that is strictly necessary, I'm not certain.

rodrigogiraoserrao: Ah, right. In that case, why not just `buttons = set(self._down_buttons)` and then iterate over the `set`?
Or do we _need_ to preserve order?

It makes a bit more sense, it is probably faster, and doesn't require a comment explaining the hoops you are going over.
Otherwise, I think this warrants a comment.

Depending on what you reply to this comment, I don't mind PR'ing the change.

willmcgugan: It is intended to preserve order. So that any button ups, arrive in the same order as they went down. Feel free to add a comment.

Issue #2007, opened 2023-03-09
If you `remove` a child from a `Vertical` that is `height: auto` the `Vertical` doesn't shrink in heigh


davep:
Given this example code:

```python
from textual.app        import App, ComposeResult
from textual.containers import Vertical
from textual.widgets    import Header, Footer, Label

class VerticalRemoveApp( App[ None ] ):

    CSS = """
    Vertical {
        border: round green;
        height: auto;
    }

    Label {
        border: round yellow;
        background: red;
        color: yellow;
    }
    """
    BINDINGS = [
        ( "a", "add", "Add" ),
        ( "d", "del", "Delete" ),
    ]

    def compose( self ) -> ComposeResult:
        yield Header()
        yield Vertical()
        yield Footer()

    def action_add( self ) -> None:
        self.query_one( Vertical ).mount( Label( "This is a test label" ) )

    def action_del( self ) -> None:
        if self.query_one( Vertical ).children:
            self.query_one( Vertical ).children[ -1 ].remove()

if __name__ == "__main__":
    VerticalRemoveApp().run()
```

You can see that the `Vertical` will grow in hight fine as you add items, but as you remove them it doesn't shrink again. Doing a `refresh( layout=True )` on the `Vertical` after the `remove` on the child makes it work as expected. For example:

```diff
diff --git a/vertical_remove.py b/vertical_remove.py
index f252456..4195abe 100644
--- a/vertical_remove.py
+++ b/vertical_remove.py
@@ -34,6 +34,7 @@ class VerticalRemoveApp( App[ None ] ):
     def action_del( self ) -> None:
         if self.query_one( Vertical ).children:
             self.query_one( Vertical ).children[ -1 ].remove()
+            self.query_one( Vertical ).refresh( layout=True )

 if __name__ == "__main__":
     VerticalRemoveApp().run()
```

PR #2008, opened 2023-03-09
fix refresh on remove

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/2007

### DIFF
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -2198,7 +2198,8 @@ class App(Generic[ReturnType], DOMNode):
                 await self._prune_nodes(widgets)
             finally:
                 finished_event.set()
-                parent.refresh(layout=True)
+                if parent.styles.auto_dimensions:
+                    parent.refresh(layout=True)
 
         removed_widgets = self._detach_from_dom(widgets)
 
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -2175,7 +2175,7 @@ class App(Generic[ReturnType], DOMNode):
             for child in widget._nodes:
                 push(child)
 
-    def _remove_nodes(self, widgets: list[Widget]) -> AwaitRemove:
+    def _remove_nodes(self, widgets: list[Widget], parent: DOMNode) -> AwaitRemove:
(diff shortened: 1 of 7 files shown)


src/textual/app.py
```diff
@@ -2198,7 +2198,8 @@ async def prune_widgets_task(
                 await self._prune_nodes(widgets)
             finally:
                 finished_event.set()
-                self.refresh(layout=True)
+                if parent.styles.auto_dimensions:
+                    parent.refresh(layout=True)
```
rodrigogiraoserrao, line 2202: Don't we need to refresh the layout of the parent, regardless of whether they have auto dimensions or not?
Imagine the parent has fixed dimensions but the children have relative dimensions (e.g., `fr` units).
Don't we need the refresh to make sure the children make use of the space that was just made available?

willmcgugan: auto dimensions and relative units kind of cancel each other out.

Consider a container with auto height and a single child with height `1fr`. The widget needs to know the height of the container, but the container needs to know the height of the child. So auto height considers all relative dimensions to be zero. It will still respect minimums, so essentially what it calculates is the minimal value which satisfies all constraints.

davep (approved):

davep:
Not a blocker to approval, but I wonder if we should have a snapshot test for this?

src/textual/app.py
rodrigogiraoserrao: That's interesting.
Up until now, I interpreted `1fr` to mean that it would expand as much as possible while also respecting relative ratios to other widgets with `fr` dimensions.
Then, if the parent has `auto` and the child has `1fr`, the child should look at the grandparent to find the maximum size it can occupy and the parent would expand to that size.
This is the behaviour I find more consistent with the interaction between absolute sizes and `1fr`.
If a parent has a fixed size and the child has `1fr`, the child will _expand as much as possible_. Hence, I have been taking the "expand as much as possible" as the interpretation of `1fr`, which I don't think should fall apart when the parent is `auto`.

Issue #1994, opened 2023-03-09
Remove underscore from press keys


willmcgugan:
The snapshot tests and screenshots used to accept an underscore to insert a pause. This isn't supported anymore. We should remove the underscores from all tests, and the condition in press_keys which catches the underscores.

davep:
Do we want to just simply remove `_`, or do we want to replace most instances with an appropriate `wait:<n>`?

willmcgugan:
The underscore is literally doing nothing now, so they can all go.

PR #2011, opened 2023-03-09
Remove underscore pauses from (hopefully) all snapshots

davep:
This seeks to remove the `"_"` from the key presses of snapshots and inline runs in the documentation. The `"_"` no longer does anything, but got legacy reasons at the moment we can't actually allow `"_"` to be an input to a "faked" keypress in the docs and snapshots.

Removing these clears the way to letting `"_"` have the same status as any other character.

See #1994.

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -44,7 +44,7 @@ python -m textual
 
 If Textual is installed you should see the following:
 
-```{.textual path="src/textual/demo.py" columns="127" lines="53" press="enter,_,_,_,_,_,_,tab,_,w,i,l,l"}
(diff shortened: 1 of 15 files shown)


willmcgugan (approved):

rodrigogiraoserrao:
> [...] but for legacy reasons at the moment we can't actually allow `"_"` to be an input to a "faked" keypress in the docs and snapshots.
> 
> Removing these clears the way to letting `"_"` have the same status as any other character.

What needs to be done to allow `_` to become a regular keypress?

davep:
Just a couple of lines need removing in (I think) the pilot.

PR #2013, opened 2023-03-09
fix(markdown): add formatter to render ordered lists correctly

### LINKED ISSUE #2002: The `Markdown` widget isn't handling ordered lists correctly
In Markdown the number prefix for an item in an ordered list [normally isn't important](https://www.markdownguide.org/basic-syntax/#ordered-lists). Every item can start with any valid numeric value; rendering will have them as an ordered sequence of numbers starting with the number of the first item.

For example:

```markdown
1. One
1. Two
1. Three
1. Four
1. Five
```

comes out as:

1. One
1. Two
1. Three
1. Four
1. Five

However, in our `Markdown` widget at the moment, it appears like this:

Screenshot 2023-03-09 at 13 04 05

This may be down to [the underlying markdown parser we're using](https://github.com/executablebooks/markdown-it-py). It says it conforms to [the CommonMark specification](https://spec.commonmark.org/), and that specification does state:

> The start number of an ordered list is determined by the list number of its initial list item. The numbers of subsequent list items are disregarded.
> -- <cite>[Section 5.3](https://spec.commonmark.org/0.30/#lists) of [v0.30](https://spec.commonmark.org/0.30/#lists)</cite>

so either we need to do something at our end to do that or enable that, or we may need to raise an issue upstream.

TomJGooding:
### Description

Parse a `mdformat` formatted version of the markdown. This ensures that ordered lists will be rendered correctly by applying consecutive numbering using the `"number": True` option.

This adds the dependency `mdformat-gfm` for [GitHub Flavored Markdown (GFM)](https://github.github.com/gfm/) support.

### Related Issue

Fixes #2002 

### Types of Changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Other

### Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] All new and existing tests passed.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,6 +40,7 @@ textual = "textual.cli.cli:run"
 python = "^3.7"
 rich = ">12.6.0"
 markdown-it-py = {extras = ["plugins", "linkify"], version = "^2.1.0"}
+mdformat-gfm = "^0.3.5"
 #rich = {path="../rich", develop=true}
 importlib-metadata = "^4.11.3"
 typing-extensions = "^4.4.0"
--- a/src/textual/widgets/_markdown.py
+++ b/src/textual/widgets/_markdown.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from pathlib import Path, PurePath
 from typing import Iterable
 
+import mdformat
 from markdown_it import MarkdownIt
 from rich.style import Style
 from rich.syntax import Syntax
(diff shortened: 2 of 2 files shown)


TomJGooding:
This would probably need some tests adding, but honestly I have no idea how to do that!

I'm not really too familiar with poetry so after several failed attempts I just edited `pyproject.toml` by hand. This means that `poetry.lock` is not consisent, but whatever I tried with poetry it created enormous diffs relating to files and hashes...

willmcgugan:
Appreciate the effort, but I’m sure we can solve the ordered list issue without an additional dependency.

TomJGooding:
No worries Will. I just thought the numering logic could get a bit complex when dealing with nested lists and the like, so this seemed like a quick fix.

PR #2015, opened 2023-03-10
docs: fix Markdown headers

xakon:
Fixed the headers of "Getting Started" guide. Still, there is an "INFO" box, probably coming from reStructuredText, which I don't know how to display it in Markdown.

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -6,17 +6,17 @@ Textual requires Python 3.7 or later (if you have a choice, pick the most recent
 
 !!! info inline end "Your platform"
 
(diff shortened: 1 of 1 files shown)


xakon:
Fix the headers of "Getting Started" guide.  Still, there is an "INFO" box, probably coming from reStructuredText, which I don't know how to display it in Markdown.

willmcgugan:
The docs use mkdocs-material which has some extensions to Markdown. You can see how it is rendered here: https://textual.textualize.io/getting_started/

xakon:
Oh, then the commit is completely irrelevant.

Thank you for the immediate feedback!

PR #2018, opened 2023-03-10
Loading indicator

willmcgugan:
- Added a LoadingIndicator widget
- Added color.Gradient

https://user-images.githubusercontent.com/554369/224347096-5e4ac8c4-c4c0-4b6d-9929-6d437610bac9.mov


src/textual/widgets/_loading_indicator.py
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    LoadingIndicator > .loading-indicator--dot {
+        color: $accent;
+    }
+
+    """
+
+    def on_mount(self) -> None:
+        self.start_time = time()
```
davep, line 32: Is this intended to be public? If so it would be a good idea to add a docstring.

src/textual/widgets/_loading_indicator.py
```diff
@@ -0,0 +1,63 @@
...
+        self.auto_refresh = 1 / 16
```
davep, line 33: Is this intended to be public? If so it would be a good idea to add a docstring.

src/textual/widgets/_loading_indicator.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def render(self) -> RenderableType:
+        elapsed = time() - self.start_time
+        speed = 0.8
+        dot = "\u25CF"
```
davep, line 38: I think I can guarantee that folk will start asking how they change the character. I think ideally it should be possible for the dev to set it when constructing the widget. If not that it should be possible for someone to modify it in a subclass with minimal effort.

davep (commented):
Couple of observations at first glance.

src/textual/widgets/_loading_indicator.py
willmcgugan: I'm going to leave it for the time being. The problem with offering everything as an option is that you accumulate a lot of arguments, and you have to handle the edge cases. I'll leave in until we know for certain the demand is there.

davep (approved):

PR #2020, opened 2023-03-11
tabs widget

willmcgugan:
Adds a Tabs widget.

This will form part of a larger `TabbedContent` widget, but could be used independently.

Screenshot 2023-03-13 at 11 11 36

https://user-images.githubusercontent.com/554369/224687824-67f1a436-872a-495b-8d46-aaaf33ee4a07.mov


docs/api/tabs.md
```diff
@@ -0,0 +1,2 @@
+::: textual.widgets.Tabs
+::: textual.widgets.Tab
```
rodrigogiraoserrao, line 2: Swap order..? :D

docs/guide/actions.md
```diff
@@ -50,7 +50,7 @@ Action strings have the following format:
 
 ### Parameters
 
-If the action string contains parameters, these must be valid Python literals. Which means you can include numbers, strings, dicts, lists etc. but you can't include variables or references to any other python symbol.
+If the action string contains parameters, these must be valid Python literals. Which means you can include numbers, strings, dicts, lists etc. but you can't include variables or references to any other python symbols.
```
rodrigogiraoserrao, line 53:
```suggestion
If the action string contains parameters, these must be valid Python literals. Which means you can include numbers, strings, dicts, lists etc. but you can't include variables or references to any other Python symbols.
```

docs/widgets/tabs.md
```diff
@@ -0,0 +1,67 @@
+# Tabs
+
+Displays a number of tab headers which may be activated with click or navigated with cursor keys.
```
rodrigogiraoserrao, line 3:
```suggestion
Displays a number of tab headers which may be activated with a click or navigated with cursor keys.
```
```suggestion
Displays a number of tab headers which may be activated with clicks or navigated with cursor keys.
```

docs/widgets/tabs.md
```diff
@@ -0,0 +1,67 @@
...
+
+- [x] Focusable
+- [ ] Container
+
+Construct a `Tabs` widget with strings or [Text][rich.text.Text] objects as positional arguments, which will set the label in the tabs. Here's an example with three tabs:
```
rodrigogiraoserrao, line 8:
```suggestion
Construct a `Tabs` widget with strings or [Text][rich.text.Text] objects as positional arguments, which will set the labels in the tabs. Here's an example with three tabs:
```

docs/widgets/tabs.md
```diff
@@ -0,0 +1,67 @@
...
+
+```python
+def compose(self) -> ComposeResult:
+    yield Tabs("First tab", "Second tab", Text.from_markup("[u]Third[/u] tab"))
+```
+
+This will create [Tab][textual.widgets.Tab] widgets internally, with an auto-incrementing `id` attribute (`"tab-1"`, `"tab-2"` etc).
```
rodrigogiraoserrao, line 15:
```suggestion
This will create [Tab][textual.widgets.Tab] widget internally, with an auto-incrementing `id` attribute (`"tab-1"`, `"tab-2"` etc).
```

docs/widgets/tabs.md
```diff
@@ -0,0 +1,67 @@
...
+You can also supply `Tab` objects directly in the constructor, which will allow you to explicitly set an `id`. Here's an example:
+
+```python
+def compose(self) -> ComposeResult:
+    yield Tabs(
+        Tab("First tab", id="one"),
+        Tab("Second tab", id="two")
```
rodrigogiraoserrao, line 22:
```suggestion
        Tab("Second tab", id="two"),
```

docs/widgets/tabs.md
```diff
@@ -0,0 +1,67 @@
...
+    )
+```
+
+When the user switches to a tab by clicking or pressing keys, then `Tabs` will send a [Tabs.TabActivated][textual.widgets.Tabs.TabActivated] message which contains the `tab` that was activated.
+You can then use `event.tab.id` attribute to perform any related actions.
+
+## Clearing tabs
+
+Clear tabs by calling the [clear][textual.widgets.Tabs.clear] method. Clearing the tabs will send a [Tabs.TabActivated][textual.widgets.Tabs.TabActivated] message with the `tab` attribute set to `None`.
```
rodrigogiraoserrao, line 31: Shouldn't clearing the tabs send a `TabsCleared` message?

docs/widgets/tabs.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Adding tabs
+
+Tabs may be added dynamically with the [add_tab][textual.widgets.Tabs.add_tab] method, which accepts strings, [Text][rich.text.Text], or [Tab][textual.widgets.Tab] objects.
+
+## Example
+
+The following example adds a Tabs widget above a text label. Press ++a++ to add a tab, ++c++ to clear the tabs.
```
rodrigogiraoserrao, line 39:
```suggestion
The following example adds a `Tabs` widget above a text label. Press ++a++ to add a tab, ++c++ to clear the tabs.
```

src/textual/widgets/_tabs.py
```diff
@@ -0,0 +1,440 @@
...
+        text-style: bold;
+        color: $text;
+    }
+    Tab:hover {
+        text-style: bold;
+    }
+    Tab.-active:hover {
+        color: $text;
+    }
+
```
rodrigogiraoserrao, line 88:
```suggestion
```

src/textual/widgets/_tabs.py
```diff
@@ -0,0 +1,440 @@
...
+from ..css.query import NoMatches
+from ..geometry import Offset
+from ..message import Message
+from ..reactive import reactive
+from ..renderables.underline_bar import UnderlineBar
+from ..widget import Widget
+from ..widgets import Static
+
+
+class Underline(Widget):
```
rodrigogiraoserrao, line 22: I find this to be a really cool widget. Is it worth documenting it so that others can use it for other things?
Or is this really a `Tabs` thing, in which case we may want to change the name to `_Underline`, or `TabUnderline`, or a similar thing?

src/textual/widgets/_tabs.py
```diff
@@ -0,0 +1,440 @@
...
+    Underline > .underline--bar {
+        background: $foreground 10%;
+        color: $accent;
+    }
+    """
+
+    COMPONENT_CLASSES = {"underline--bar"}
+
+    highlight_start = reactive(0)
+    highlight_end = reactive(0)
```
rodrigogiraoserrao, line 39: Docstrings, please!

src/textual/widgets/_tabs.py
```diff
@@ -0,0 +1,440 @@
...
+        width: 1fr;
+        height: 1;
+    }
+    Underline > .underline--bar {
+        background: $foreground 10%;
+        color: $accent;
+    }
+    """
+
+    COMPONENT_CLASSES = {"underline--bar"}
```
rodrigogiraoserrao, line 36: Docstring, please.

src/textual/widgets/_tabs.py
```diff
@@ -0,0 +1,440 @@
...
+        bar_style = self.get_component_rich_style("underline--bar")
+        return UnderlineBar(
+            highlight_range=self._highlight_range,
+            highlight_style=Style.from_color(bar_style.color),
+            background_style=Style.from_color(bar_style.bgcolor),
+        )
+
+    def on_click(self, event: events.Click):
+        """Catch clicks, so that the underline can activate the tabs."""
+        self.post_message(self.Clicked(event.screen_offset))
```
rodrigogiraoserrao, line 64: Did you forget to stop `event` or do you think it is beneficial to let the original click bubble up? If you think so, can you help me understand why, please?

src/textual/widgets/_tabs.py
```diff
@@ -0,0 +1,440 @@
...
+    def __init__(
+        self,
+        label: TextType,
+        *,
+        id: str | None = None,
+    ) -> None:
+        """Initialise a Tab.
+
+        Args:
+            label: The label to use in the tab. May be a str or a Text object.
```
rodrigogiraoserrao, line 107: I would remove this as the info comes from the type, no?

```suggestion
            label: The label to use in the tab.
```

rodrigogiraoserrao: Otherwise, consider styling as code and linking the `Text`.

src/textual/widgets/_tabs.py
```diff
@@ -0,0 +1,440 @@
...
+            id: Optional ID for the widget.
+
+        """
+        self.label = Text.from_markup(label) if isinstance(label, str) else label
+        if id is None:
+            id = f"tab-{self.label_text.lower().replace(' ', '-')}"
```
rodrigogiraoserrao, line 113: I find it a bit confusing that `Tab` creates an id out of its label by default, but `Tabs` counts IDs from one.

In particular, the two pieces of code below are very similar and yet return tabs with different IDs:

```py
Tabs("One", "Two", "Three")
```

```py
Tabs(
    Tab("One"),
    Tab("Two"),
    Tab("Three"),
)
```

src/textual/widgets/_tabs.py
```diff
@@ -0,0 +1,440 @@
...
+    """
+
+    BINDINGS: ClassVar[list[BindingType]] = [
+        Binding("left", "previous_tab", "Previous tab", show=False),
+        Binding("right", "next_tab", "Next tab", show=False),
+    ]
+
+    class TabActivated(Message):
+        """Sent when a new tab is activated."""
+
```
rodrigogiraoserrao, line 160:
```suggestion

        tab: Tab
        """The tab that was activated."""

```

src/textual/widgets/_tabs.py
```diff
@@ -0,0 +1,440 @@
...
+        def __init__(self, tab: Tab | None) -> None:
+            self.tab = tab
+            super().__init__()
+
+        def __rich_repr__(self) -> rich.repr.Result:
+            yield self.tab
+
+    active: reactive[str] = reactive(str, init=False)
```
rodrigogiraoserrao, line 168: Docstring, please.

rodrigogiraoserrao: Also, passing the built-in `str` as the initial value looks quite odd.

src/textual/widgets/_tabs.py
```diff
@@ -0,0 +1,440 @@
...
+        name: str | None = None,
+        id: str | None = None,
+        classes: str | None = None,
+        disabled: bool = False,
+    ):
+        """Construct a Tabs widget.
+
+        Args:
+            *tabs: Positional argument should be explicit Tab objects, or a str or Text.
+            active (_type_, optional): ID of the tab which should be active on start.
```
rodrigogiraoserrao, line 183:
```suggestion
            active: ID of the tab which should be active on start.
```

src/textual/widgets/_tabs.py
```diff
@@ -0,0 +1,440 @@
...
+            id=id,
+            classes=classes,
+            disabled=disabled,
+        )
+        self._tabs = add_tabs
+        self._first_active = active
+
+    @property
+    def _new_tab_id(self) -> int:
+        """Get the next tab id in a sequence."""
```
rodrigogiraoserrao, line 209:
```suggestion
        """Next tab ID in a sequence."""
```

Also, what's "a sequence"? "The next available tab ID"?

src/textual/widgets/_tabs.py
```diff
@@ -0,0 +1,440 @@
...
+        self._tabs_counter += 1
+        return self._tabs_counter
+
+    @property
+    def tab_count(self) -> int:
+        """The total number of tabs."""
```
rodrigogiraoserrao, line 215:
```suggestion
        """Total number of tabs."""
```

src/textual/widgets/_tabs.py
```diff
@@ -0,0 +1,440 @@
...
+        return len(self.query("#tabs-list > Tab"))
+
+    @property
+    def _next_active(self) -> Tab | None:
+        """Get the next tab to make active, if the active tab is removed."""
```
rodrigogiraoserrao, line 220:
```suggestion
        """Next tab to make active if the active tab is removed."""
```

src/textual/widgets/_tabs.py
```diff
@@ -0,0 +1,440 @@
...
+            return None
+        del tabs[active_index]
+        try:
+            return tabs[active_index]
+        except IndexError:
+            try:
+                return tabs[active_index - 1]
+            except IndexError:
+                pass
+        return None
```
rodrigogiraoserrao, line 236: There is something I don't get. Can't we just do this by inspecting the length of `tabs` with respect to the `active_index`?
Why do we have to `del tabs[active_index]` and whatnot?

Does `del tabs[active_index]` do anything other than deleting that specific entry from the list `tabs`?

src/textual/widgets/_tabs.py
```diff
@@ -0,0 +1,440 @@
...
+            remove_tab = tab_or_id
+        else:
+            try:
+                remove_tab = self.query_one(f"#tabs-list > #{tab_or_id}", Tab)
+            except NoMatches:
+                return
+        removing_active_tab = remove_tab.has_class("-active")
+
+        next_tab = self._next_active
+        self.post_message(self.TabActivated(next_tab))
```
rodrigogiraoserrao, line 289: Why do we do this outside of `do_remove`? Shouldn't all these actions be done together?

src/textual/widgets/_tabs.py
```diff
@@ -0,0 +1,440 @@
...
+            self._highlight_active(animate=previously_active != "")
+            self.post_message(self.TabActivated(active_tab))
+        else:
+            underline = self.query_one(Underline)
+            underline.highlight_start = 0
+            underline.highlight_end = 0
+            self.post_message(self.TabActivated(None))
+
+    def _highlight_active(self, animate: bool = True) -> None:
+        """Move the underline bar to under the active tab,
```
rodrigogiraoserrao, line 348:
```suggestion
        """Move the underline bar to under the active tab.
```

src/textual/widgets/_tabs.py
```diff
@@ -0,0 +1,440 @@
...
+        event.stop()
+        offset = event.offset + (0, -1)
+        self.focus()
+        for tab in self.query(Tab):
+            if offset in tab.region:
+                self._activate_tab(tab)
+                break
+
+    def _scroll_active_tab(self) -> None:
+        """Scroll the active tab in to view."""
```
rodrigogiraoserrao, line 403:
```suggestion
        """Scroll the active tab into view."""
```

docs/widgets/tabs.md
willmcgugan: Don't know if this warrants adding a new message. I will consider it for a later revision.

src/textual/widgets/_tabs.py
willmcgugan: It is tightly coupled with the tabs widget ATM. It is essentially private since its in `_tabs.py` and no exported anywhere.

Maybe at some point it can be its own thing.

src/textual/widgets/_tabs.py
willmcgugan: I was thinking that click events don't bubble, but of course they do.

src/textual/widgets/_tabs.py
willmcgugan: Exactly.

src/textual/widgets/_tabs.py
willmcgugan: I'm sure there is a more elegant solution. But we want to get the next active tab, which is normally the one to the right. Unless we're at the end, then its the one to the left.

davep (approved):

src/textual/widgets/_tabs.py
willmcgugan: The `do_remove` is mainly for the underline bar. So that it can move the start and end post-layout. This message doesn't need to wait, it can go straight away.

PR #2025, opened 2023-03-12
Properly indicate dependency is only required for development

ofek:
I discovered this when trying to build Textual and its dependencies in such a way that required every package to have a wheel and it stood out to me that this was unlikely to be an actual dependency.

I also found https://github.com/apenwarr/mkdocs-exclude/issues/16, it might be unmaintained


willmcgugan (approved):

willmcgugan:
Thanks

PR #2030, opened 2023-03-13
Add containers

rodrigogiraoserrao:

- [x] Updated documentation


src/textual/containers.py
```diff
@@ -14,45 +14,93 @@ class Container(Widget):
 
 
 class Vertical(Widget):
-    """A container widget which aligns children vertically."""
+    """A container which lays children vertically."""
```
davep, line 17: I don't think "lays" is quite the right way to phrase this here; `aligns` seemed to make more says, or "lays out" would work.

src/textual/containers.py
```diff
@@ -14,45 +14,93 @@ class Container(Widget):
...
+        width: 1fr;
         layout: vertical;
         overflow-y: auto;
     }
     """
 
 
 class Horizontal(Widget):
-    """A container widget which aligns children horizontally."""
+    """A container which lays children horizontally."""
```
davep, line 41: Similar comment to earlier for "lays" here.

davep (approved):
Just some docstring phrasing issues from me, but I don't see anything that's a blocker.

src/textual/containers.py
rodrigogiraoserrao: I don't like "align" because that can be confused with the `align-*` styles. The children are not aligned vertically (e.g., one might be to the left and the other might be to the right).

What if I write "A container which composes children vertically."?
I don't love this wording either, but I think it is better than using "align".

docs/examples/widgets/button.css
```diff
@@ -2,7 +2,7 @@ Button {
     margin: 1 2;
 }
 
-Horizontal > Vertical {
+Horizontal>VerticalScroll {
```
willmcgugan, line 5: Can we keep the spaces around the `>`

src/textual/containers.py
willmcgugan: I don't think `composes` works, as it might confuse with our compose method.

How about `arranges`? "A container which arranges children vertically."

willmcgugan (commented):
Code is good. Just a few quibbles.

src/textual/containers.py
rodrigogiraoserrao: I like "arranges".

src/textual/containers.py
```diff
@@ -14,45 +14,93 @@ class Container(Widget):
...
-        height: 1fr;
+        width: 1fr;
+        layout: vertical;
+        overflow: hidden hidden;
+    }
+    """
+
+
+class VerticalScroll(Widget):
+    """A container which aligns children vertically and overflows automatically."""
```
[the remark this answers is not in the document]
rodrigogiraoserrao: Sure. I was just trying to stick to the line length limit for the one-line, short description.

willmcgugan (approved):
Looks good!

Would you mind looking through all the examples, `python -m textual`, `textual colors`, etc.

Just to confirm scrolling behaviour hasn't broken.

If everything is working, feel free to merge.

rodrigogiraoserrao:
> Looks good!
> 
> Would you mind looking through all the examples, `python -m textual`, `textual colors`, etc.
> 
> Just to confirm scrolling behaviour hasn't broken.
> 
> If everything is working, feel free to merge.

Everything looking good. 🚀

PR #2036, opened 2023-03-13
Driver class environ

willmcgugan:
Adds a `TEXTUAL_DRIVER` environment variable. This allows you to change the driver class dynamically, and will be used by Ganglion.


willmcgugan:
#executivedecision

PR #2037, opened 2023-03-13
Pilot can now press _

rodrigogiraoserrao:
Follow-up to #2011.

I did this in a jiffy because I had been lurking around this code when I worked on a bug with keys recently, so I figured I'd just make this clean PR while I still knew what needed to be done.


willmcgugan (approved):

davep:
Good work on spotting the last few lurking _s.

PR #2042, opened 2023-03-13
Get rid of `_Clock` and move utility time-related functions to `_time.py`.

### LINKED ISSUE #1961: Remove clock concept
clock.py was indented to be a virtual clock to driver Textual apps. The ideas was that in testing we could create a mocked clock that would make the event system predictable.

That never happened, and I suspect the abstraction in clock.py is superfluous. Suggest we ditch it in favour of simple module to expose time related functions.

May require an investigation to see if I am correct about this.

rodrigogiraoserrao:
This will close #1961.


src/textual/_time.py
```diff
@@ -42,3 +42,26 @@ async def sleep(secs: float) -> None:
...
+    """Get the current wall clock (monotonic) time.
+
+    Returns:
+        The value (in fractional seconds) of a monotonic clock,
+        i.e. a clock that cannot go backwards.
+    """
+    return time()
+
+
+async def get_time() -> float:
```
willmcgugan, line 57: I don't think we will need an `async` time. It was an artifact of the mockable clock idea I had.

rodrigogiraoserrao: The tests seem to agree with you.
I saw it being used in `timer.py` and I thought it was needed, but it works just fine with the synchronous version as well.

davep (approved):

src/textual/_time.py
```diff
@@ -42,3 +42,13 @@ async def sleep(secs: float) -> None:
         sleep_for = secs - 0.0005
         if sleep_for > 0:
             await asyncio_sleep(sleep_for)
+
+
+def get_time() -> float:
```
willmcgugan, line 47: Could we replace this with:

```python
get_time = time
```

willmcgugan (approved):

Issue #1911, opened 2023-03-01
Horizontal scrollbar for Markdown Viewer table is overriding the last row


learnbyexample:
While using the Markdown Viewer, tables requiring horizontal scrollbar seems to be overriding the last row. I'd expect the scrollbar to be added as a new row instead. Used the https://textual.textualize.io/widgets/markdown_viewer/#example (with non-table sections removed) for the demo below:

https://user-images.githubusercontent.com/17766317/222116203-3e02c3f9-ee57-4eb9-b18e-4cc99523dc6a.mp4

learnbyexample:
I just tried adding an empty row at the end, and that's a good enough workaround for me.

willmcgugan:
There's some incoming changes which should fix this.

PR #2055, opened 2023-03-14
Replace markdown table

willmcgugan:
Replaces  DataTables in Markdown with Rich tables. It makes more sense to use Rich tables in a document, so that it they will wrap nicely.

Fixes https://github.com/Textualize/textual/issues/1911

## Before

Screenshot 2023-03-14 at 13 30 15

## After

Screenshot 2023-03-14 at 13 30 00


src/textual/widgets/_markdown.py
```diff
@@ -316,17 +318,59 @@ def compose(self) -> ComposeResult:
...
+        height: auto;
+
+    }
+    MarkdownTableContent > .markdown-table--header {
+        text-style: bold;
+
+    }
+    MarkdownTableContent > .markdown-table--lines {
+
+    }
```
davep, line 334: Are empty `DEFAULT_CSS` rules needed to make component styles work?

src/textual/widgets/_markdown.py
```diff
@@ -316,17 +318,59 @@ def compose(self) -> ComposeResult:
         self._blocks.clear()
 
 
+class MarkdownTableContent(Widget):
```
davep, line 321: Class could use a docstring.

src/textual/widgets/_markdown.py
```diff
@@ -316,17 +318,59 @@ def compose(self) -> ComposeResult:
...
+
+    }
+    MarkdownTableContent > .markdown-table--lines {
+
+    }
+    """
+
+    COMPONENT_CLASSES = {"markdown-table--header", "markdown-table--lines"}
+
+    def __init__(self, headers: list[Text], rows: list[list[Text]]):
```
davep, line 339: The public properties could use docstrings.

davep (approved):
A question about the CSS and a couple of docstring reminders but looks good otherwise!

mrossinek (commented):
Looking at that screenshot in the PR description, I think the text in `example.md` should be updated to not refer to `DataTable` any longer.
That text there also refers to clickable column headers to allow sorting. Not sure if you deem this out of scope now (since this could have been possible based on `DataTable` but not based on `rich.Table`). Or you could update it to say that this is totally if one wants to implement such a feature in a document viewer but with the caveats of wrapping which this PR is solving.

Just some thoughts... :slightly_smiling_face:

src/textual/widgets/_markdown.py
willmcgugan: No, just a side-effect of editing. Will remove.

willmcgugan:
@mrossinek Sorting by columns should still be possible. It's straightforward to turn those headers in to links.

Issue #1972, opened 2023-03-07
Pilot tests incorrectly pass when exception thrown in action method


darrenburns:
They pass, but an exception traceback is printed. Textual appears to be catching the exception and printing it, but it doesn't bubble up to the test framework. The test framework therefore passes the test because nothing went wrong from its POV.

I added a `1/0` into a handler for a binding and used `pilot.press` to press the corresponding key. The test passed unexpectedly.

I think we need to make it so that when we’re using run_test, Textual will print the exceptions as normal, but also allow them to bubble up.

willmcgugan:
I'm not sure they can bubble up here. Given that everything is running on independent tasks.

Perhaps what we need to do is reraise those errors from `run_test` when it exits.

Just a suggestion. This needs a bit of an investigation.

Issue #1838, opened 2023-02-19
Add a mega-stylesheet test


davep:
Following on from #1837 and #1836: add a mega-stylesheet-unit-test. This will be one huge stylesheet that contains everything that should be valid, in as many fun combinations as possible, and the test is it should just work.

Issue #1836, opened 2023-02-19
[CSS] Descendant type selectors can't have a numeric in their name


davep:
Consider the following code:

```python
from textual.app        import App, ComposeResult
from textual.containers import Vertical
from textual.widgets    import Header, Footer, Label

class LabelH1( Label ):
    ...

class CSSOddnessApp( App[ None ] ):

    CSS = """
    Vertical LabelH1 {
        background: red;
    }
    """

    def compose( self ) -> ComposeResult:
        yield Header()
        yield Vertical(
            Label( "Label" ),
            LabelH1( "LabelH1" ),
        )
        yield Footer()

if __name__ == "__main__":
    CSSOddnessApp().run()
```

When run we get the following error:

```
 Error in stylesheet:
 /Users/davep/develop/python/textual-sandbox/css_oddness.py:CSSOddnessApp:1:19
╭────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ ❱ 1 │                                                                                                                                                      │
│   2 │   Vertical LabelH1 {                                                                                                                                 │
│   3 │   │   background: red;                                                                                                                               │
╰────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
 • Expected one of 'combinator child', 'comment start', 'declaration set start', 'new selector', 'pseudo class', 'selector', 'selector class', 'selector id',
 'selector universal', or 'whitespace'.
 • Did you forget a semicolon at the end of a line?
```

The same thing happens with `Vertical LabelH1`. On the other hand, if I remove the number from the inherited label widget:

```python
from textual.app        import App, ComposeResult
from textual.containers import Vertical
from textual.widgets    import Header, Footer, Label

class LabelHOne( Label ):
    ...

class CSSOddnessApp( App[ None ] ):

    CSS = """
    Vertical LabelHOne {
        background: red;
    }
    """

    def compose( self ) -> ComposeResult:
        yield Header()
        yield Vertical(
            Label( "Label" ),
            LabelHOne( "LabelHOne" ),
        )
        yield Footer()

if __name__ == "__main__":
    CSSOddnessApp().run()
```

this works fine. Likewise, if I retain the name but *don't* use combination:

```python
from textual.app        import App, ComposeResult
from textual.containers import Vertical
from textual.widgets    import Header, Footer, Label

class LabelH1( Label ):
    ...

class CSSOddnessApp( App[ None ] ):

    CSS = """
[...]

Issue #1253, opened 2022-11-23
Type selectors can't contain a number


davep:
It seems that type selectors can't handle types that have a number in their name. For example this:

```python
from textual.app import App, ComposeResult
from textual.widgets import Label

class H1( Label ):
    pass

class NumberError( App[ None ] ):

    CSS = """
    H1 {
        text-style: bold;
        color: yellow;
        background: red;
    }
    """
    def compose( self ) -> ComposeResult:
        yield H1( "This is a very important heading" )

if __name__ == "__main__":
    NumberError().run()
```

throws this error:

```
Error in stylesheet:
 /Users/davep/develop/python/textual-sandbox/not_all_types.py:NumberError:1:5
╭───────────────────────────────────────────────────────────────────────────────────────────╮
│ ❱ 1 │                                                                                     │
│   2 │   H1 {                                                                              │
│   3 │   │   text-style: bold;                                                             │
╰───────────────────────────────────────────────────────────────────────────────────────────╯
 • Expected one of 'combinator child', 'comment start', 'declaration set start', 'new
 selector', 'pseudo class', 'selector', 'selector class', 'selector id', 'selector
 universal', or 'whitespace'.
 • Did you forget a semicolon at the end of a line?
```

whereas this works fine:

```python
from textual.app import App, ComposeResult
from textual.widgets import Label

class HOne( Label ):
    pass

class NumberError( App[ None ] ):

    CSS = """
    HOne {
        text-style: bold;
        color: yellow;
        background: red;
    }
    """
    def compose( self ) -> ComposeResult:
        yield HOne( "This is a very important heading" )

if __name__ == "__main__":
    NumberError().run()
```

as does, of course, this:

```python
from textual.app import App, ComposeResult
from textual.widgets import Label

class H1( Label ):
    pass

class NumberError( App[ None ] ):

    CSS = """
    Label {
        text-style: bold;
        color: yellow;
        background: red;
    }
    """
    def compose( self ) -> ComposeResult:
        yield H1( "This is a very important heading" )

if __name__ == "__main__":
    NumberError().run()
```

willmcgugan:
Probably requires a tweak to the regex. It should at least allow any valid Python identifier.

PR #2057, opened 2023-03-14
Add some initial rules to a "mega stylesheet"

davep:
The start of what will hopefully become a one-stop comprehensive test of all the fun corners of stylesheet parsing. While not intended to replace all the other unit tests for CSS, this test should help to quickly and easily be sure that nothing breaks any valid CSS.

Influenced by #1253 and #1836 this code starts the process of implementing the requirement outlined in #1838.

Adding more rules -- simple and/or weird -- that are expected and known to parse without a problem is encouraged.

---
PS: Initially the intention with this test was to load up the stylesheet as part of an actual `App`, using the pilot. At the moment though it looks like the pilot eats any CSS parsing errors (possibly a variation on #1972?) so the test instead does this in a lower-level way.


tests/css/test_mega_stylesheet.css
```diff
@@ -0,0 +1,227 @@
...
+* > * {}
+* > * > * {}
+
+/**********************************************************************/
+
+A {}
+A A {}
+A A A {}
+A > A {}
+A > A > A {}
```
rodrigogiraoserrao, line 19: I see you have the pattern `X Y Z` and `X > Y > Z` but you skipped both `X Y > Z` and `X > Y Z`

rodrigogiraoserrao (approved):

tests/css/test_mega_stylesheet.css
davep: Yup, given how the rules seem to be constructed, either of those implies the other is fine. I just wanted to mostly kick off testing the one, two, many scenario.

We can add as many as we like from now on though. Adding stuff is encouraged!

PR #2059, opened 2023-03-14
tabbed content widget

willmcgugan:
- Adds `TabbedContent` widget
- Adds `TabPane` widget
- New Tabs.TabCleared message
- Adds `get_child_by_type` method

https://user-images.githubusercontent.com/554369/225863769-0837f5ec-2a74-4be5-98df-47dbfe9be323.mov


src/textual/app.py
```diff
@@ -2202,7 +2216,7 @@ async def prune_widgets_task(
                 await self._prune_nodes(widgets)
             finally:
                 finished_event.set()
-                if parent is not None and parent.styles.auto_dimensions:
+                if parent is not None:
```
willmcgugan, line 2219: Caused things not to update. A bad optimization on my part.

src/textual/widgets/_tabbed_content.py
```diff
@@ -0,0 +1,148 @@
...
+
+    def compose_add_child(self, widget: Widget) -> None:
+        """When using the context manager compose syntax, we want to attach nodes to the switcher."""
+        self._tab_content.append(widget)
+
+    def _on_tabs_tab_activated(self, event: Tabs.TabActivated) -> None:
+        """User clicked a tab."""
+        self.log("on_tabs_tab_activated", event)
+        event.stop()
+        switcher = self.query_one(ContentSwitcher)
```
davep, line 138: This is going to cause an error if one of the tabs contains a `ContentSwitcher` of its own. Either retaining a reference to the `ContentSwitcher` in `compose`, or use of `get_child_by_type`, might be a better approach here.

docs/widgets/tabbed_content.md
```diff
@@ -0,0 +1,106 @@
...
+You can then switch tabs by setting the `active` reactive attribute:
+
+```python
+# Switch to Jessica tab
+self.query_one(TabbedContent).active = "jessica"
+```
+
+!!! note
+
+    If you don't provide `id` attributes to the tab panes, they will be assigned sequentially starting at `tab-1` (then `tab-2` etc).
```
davep, line 63: Out of curiosity, why start at 1?

docs/widgets/tabbed_content.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Initial tab
+
+The first child of `TabbedContent` will be the initial active tab by default. You can pick a different initial tab by setting the `initial` argument to the `id` of the tab:
```
davep, line 67: While testing this, it seems that `initial` is required. Or rather, it's optional, but then that results in this:

```python
NoMatches: No child found with id=''
```

src/textual/widgets/_tabbed_content.py
```diff
@@ -0,0 +1,148 @@
...
+
+    active: reactive[str] = reactive("", init=False)
+    """The ID of the active tab, or empty string if none are active."""
+
+    def __init__(self, *titles: TextType, initial: str = "") -> None:
+        """Initialize a TabbedContent widgets.
+
+        Args:
+            *titles: Positional argument will be used as title.
+            initial: The id of the initial tab, or empty string to select the first tab.
```
davep, line 91: As mentioned in a comment on the docs, but I felt it worth a mention any place this gets said, leaving this empty results in an error that `NoMatches: No child found with id=''`.

src/textual/widgets/_tabbed_content.py
```diff
@@ -0,0 +1,148 @@
...
+            )
+        ]
+        # Get a tab for each pane
+        tabs = [
+            ContentTab(content._title, content.id or "") for content in pane_content
+        ]
+        # Yield the tabs
+        yield Tabs(*tabs, active=self._initial)
+        # Yield the content switcher and panes
+        with ContentSwitcher(initial=self._initial):
```
davep, line 127: I think an extra step is required here, where `self._initial` needs to be tested if it's empty and, if so, it's set to the `id` of the first tab.

davep (changes requested):
Couple or so small documentation things, but also a couple of errors that seem like they're easily fixed.

I feel this also deserves some dedicated unit tests too given that this is a fairly involved container, testing the messages and the like, and also testing things like `TabbedContent` inside `TabbedContent`, etc (I did try that "by hand" and seems fine once the issue around pulling back the `ContentSwitcher` was fixed).

Overall though this is really nice! Looks and feels nice and smooth.

docs/widgets/tabbed_content.md
willmcgugan: It's not used as an index per se. And regular people like to start counting from one.

docs/widgets/tabbed_content.md
willmcgugan: Pushed a fix.

src/textual/widget.py
```diff
@@ -496,6 +515,24 @@ def get_component_rich_style(self, name: str, *, partial: bool = False) -> Style
 
         return partial_style if partial else style
 
+    def render_str(self, text_content: str | Text) -> Text:
```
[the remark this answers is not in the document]
willmcgugan: At the moment it doesn't do anything with widget, but I am expecting it to refer to styles for things like text-wrap.

src/textual/widgets/_tabbed_content.py
willmcgugan: The first tab is set via the TabActivated event.

src/textual/widgets/_tabbed_content.py
```diff
@@ -0,0 +1,148 @@
...
+
+    def _on_tabs_tab_activated(self, event: Tabs.TabActivated) -> None:
+        """User clicked a tab."""
+        self.log("on_tabs_tab_activated", event)
+        event.stop()
+        switcher = self.query_one(ContentSwitcher)
+        assert isinstance(event.tab, ContentTab)
+        switcher.current = event.tab.id
+
+    def _on_tabs_tabs_cleared(self, event: Tabs.TabsCleared) -> None:
```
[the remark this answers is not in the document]
willmcgugan: We've never really documented the event  handlers. Perhaps because it feels redundant.

If we are going to start, I think we should do it one PR. So I'm inclined to leave it for then.

src/textual/widgets/_tabbed_content.py
willmcgugan: True. I suppose nested tabs are possible!

src/textual/widgets/_tabbed_content.py
```diff
@@ -0,0 +1,148 @@
...
+        # Yield the content switcher and panes
+        with ContentSwitcher(initial=self._initial):
+            yield from pane_content
+
+    def compose_add_child(self, widget: Widget) -> None:
+        """When using the context manager compose syntax, we want to attach nodes to the switcher."""
+        self._tab_content.append(widget)
+
+    def _on_tabs_tab_activated(self, event: Tabs.TabActivated) -> None:
+        """User clicked a tab."""
```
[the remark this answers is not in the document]
willmcgugan: We've never documented the event parameter in handlers. If we're going to start now I'd be inclined to do it in a single pass. So I think I'll leave it till then...

src/textual/widgets/_tabs.py
```diff
@@ -175,16 +175,33 @@ class Tabs(Widget, can_focus=True):
...
+        def __init__(self, tabs: Tabs, tab: Tab) -> None:
+            self.tabs = tabs
             self.tab = tab
             super().__init__()
 
         def __rich_repr__(self) -> rich.repr.Result:
+            yield self.tabs
             yield self.tab
 
+    class TabsCleared(Message):
```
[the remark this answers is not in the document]
willmcgugan: Yeah. Makes sense.

docs/widgets/tabbed_content.md
davep: Works for me. I'd hope that "regular people" working with this would be used to their language of choice doing the zero-based thing, but this makes sense as a choice.

src/textual/widget.py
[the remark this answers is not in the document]
davep: Perhaps make it a static method until then?

src/textual/widgets/_tabbed_content.py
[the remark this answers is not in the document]
davep: Works for me!

src/textual/widgets/_tabbed_content.py
davep: Imagine what I tested first. 😏

davep:
Random thought while playing with this some more: I feel like it might be a good idea to have `home` and `end` bindings on tabs to support easily going to the first/last tab (I realise this isn't an issue for this particular PR, but playing with this fully had my hitting them more than once out of habit).

src/textual/widget.py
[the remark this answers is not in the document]
willmcgugan: I don't see any benefit in that. It would just lead to a future breaking change.

willmcgugan:
@davep Added some tests. The new pilot.click is awesome.

docs/widgets/tabbed_content.md
```diff
@@ -0,0 +1,106 @@
...
+
+    ```python
+    --8<-- "docs/examples/widgets/tabbed_content.py"
+    ```
+
+## Reactive attributes
+
+| Name     | Type  | Default | Description                                                    |
+| -------- | ----- | ------- | -------------------------------------------------------------- |
+| `active` | `str` | `""`    | The `id` attribute of the active tab. Set this to switch tabs. |
```
davep, line 99: One last nit... as of 2ede38f653fb92299ab77d5a8e8a788df1692d1c this needs to show that it can be `None`.

davep:
> Added some tests. The new pilot.click is awesome.

@willmcgugan Looks good! I suspect I'll be giving pilot.click a workout with the menu stuff too; looking forward to it.

For the sake of completeness I think I'd be temped to add a tabs-within-tabs test (given that cause an error initially and could possibly be a source of regression at some point), but that's your call.

Looking good! Looks very mergeable to me now!

davep (approved):

PR #2064, opened 2023-03-14
Implement border (sub)title.

### LINKED ISSUE #1864: Add title to CSS borders
We should add the ability to render a title (displayed in the top) and a subtitle (displayed in the bottom) to CSS borders. Something akin to Rich panels.

Widget should grow `border_title` and `border_subtitle` attributes, which should accept a str or Text in the setter, and return a Text in the getter.

Additionally, there should be new CSS declaration s:

`border-title-align: <left, center, right>`
`border-subtitle-align: <left, center, right>`

Note that it only makes sense to render a single line in the titles. So we need a little logic to get the first line.

If the line doesn't fit, it should be truncated with ellipsis. The Text object can do this (set no_wrap=True, and overflow="ellipsis")

See panel.py in Rich for inspiration.

rodrigogiraoserrao:
This will close #1864.


willmcgugan:
@rodrigogiraoserrao would you mind adding a screenshot or two of a widget with title / subtitle.

src/textual/_border.py
```diff
@@ -261,8 +274,70 @@ def get_box(
...
+@lru_cache(maxsize=1024)
+def render_border_label(
+    label: str,
+    is_title: bool,
+    name: EdgeType,
+    width: int,
+    inner_style: Style,
+    outer_style: Style,
+    style: Style,
+) -> Segment:
```
willmcgugan, line 286: This can't be right. If the title contains markup,  like "[bold]Hello [magenta]World[/][blink]![/]", then the it will consist of several segments. But you are returning a single segment.

src/textual/_styles_cache.py
```diff
@@ -66,6 +66,11 @@ class StylesCache:
 
     """
 
+    _border_title: str = ""
```
willmcgugan, line 69: I don't think we will need these, since we can inspect the widget object when we come to render it.

willmcgugan (changes requested):
Good first pass. A few requests. Might be worth going over this tomorrow.

rodrigogiraoserrao:
_borders_example

src/textual/widget.py
```diff
@@ -2353,6 +2357,28 @@ def watch_disabled(self) -> None:
...
         self._update_styles()
 
+    def validate_border_title(self, title: str) -> str:
+        """Ensure we only use a single line for the border title."""
+        if not title:
+            return title
+        first, *_ = title.splitlines()
+        return first
+
+    def watch_border_title(self, title: str) -> None:
```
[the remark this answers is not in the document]
rodrigogiraoserrao: Done.

src/textual/_border.py
rodrigogiraoserrao: You were right, I misunderstood how text rendering works.

src/textual/_border.py
```diff
@@ -261,8 +274,70 @@ def get_box(
...
+            This accounts for the 2 cells made available for the corner, which means
+            that the space available for the label is effectively `width - 2`.
+        inner_style: The inner style (widget background).
+        outer_style: The outer style (parent background).
+        style: Widget style.
+    """
+    if not label:
+        return Segment("", Style())
+
+    console = Console(width=width - 2)
```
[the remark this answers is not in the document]
rodrigogiraoserrao: Thanks for the tips.

rodrigogiraoserrao:
@willmcgugan because I was catenating lists left, right, and centre, I reworked the two methods in `_border.py` to return iterables (to yield pieces one by one).
Let me know if this is _not_ ok.

rodrigogiraoserrao:
Untitled

Run this demo with `textual run docs/examples/styles/border_sub_title_align_all.py`

docs/snippets/border_sub_title_align_all_example.md
```diff
@@ -0,0 +1,38 @@
...
+Open the code tabs to see the details of the code examples.
+
+=== "Output"
+
+    ```{.textual path="docs/examples/styles/border_sub_title_align_all.py"}
+    ```
+
+=== "border_sub_title_align_all.py"
+
+    ```py hl_lines=""
```
willmcgugan: An empty `hl_lines` breaks the markup.

docs/examples/styles/border_subtitle_align.py
```diff
@@ -0,0 +1,20 @@
+from textual.app import App
+from textual.widgets import Label
+
+
+class BorderSubtitleAlignApp(App):
+    def compose(self):
+        lbl = Label("My border is on the left.", id="label1")
```
willmcgugan, line 7: Suggest "My **subtitle** ..."

docs/examples/styles/border_title_align.py
```diff
@@ -0,0 +1,20 @@
+from textual.app import App
+from textual.widgets import Label
+
+
+class BorderTitleAlignApp(App):
+    def compose(self):
+        lbl = Label("My border is on the left.", id="label1")
```
willmcgugan, line 7: Ditto, "My **title** ..."

tests/test_styles_cache.py
```diff
@@ -9,6 +10,8 @@
 from textual.geometry import Region, Size
 from textual.strip import Strip
 
+_CONSOLE = Console()
```
willmcgugan, line 13: Probably not a good idea to share this across tests. Don't see any particular issues, but we don't want to share state accross tests.

Suggest just using `Console()` in the tests.

src/textual/_styles_cache.py
```diff
@@ -275,20 +295,42 @@ def post(segments: Iterable[Segment]) -> Iterable[Segment]:
         line: Iterable[Segment]
         # Draw top or bottom borders (A)
         if (border_top and y == 0) or (border_bottom and y == height - 1):
+            is_top = y == 0
             border_color = base_background + (
-                border_top_color if y == 0 else border_bottom_color
+                border_top_color if is_top else border_bottom_color
```
willmcgugan, line 300: There is a significant amount of work done here which may not be required if there is no label.

Can we short-circuit this for the most common cases where there is no border or no title?

willmcgugan (changes requested):
Looking good. A few requests.

docs/examples/styles/border_subtitle_align.py
rodrigogiraoserrao: Right. I don't know why the heck I thought "border" was a good word, there.

src/textual/_styles_cache.py
rodrigogiraoserrao: I don't follow!
`border_color` creates `border_color_as_style` which is needed by `get_box` regardless of whether there is a label or not.
(I just renamed a couple of things.)

So, I don't see what can be short-circuited here.

tests/test_styles_cache.py
rodrigogiraoserrao: That was me trying to not waste time, since you recently told me `Console` does a fair bit of work that ends up being wasted in these tests.

docs/guide/styles.md
```diff
@@ -257,6 +257,22 @@ There are many other border types. Run the following from the command prompt to
...
+Similarly, a widget with a border can have a subtitle in its bottom edge.
+To change it, assign to the attribute [`border_subtitle`][textual.widget.Widget.border_subtitle].
+
+These labels must be one-line strings, but they can contain rich markup.
+Additionally, the labels can be aligned horizontally with the respective styles:
+
+ - [`border-title-align`](../styles/border_title_align.md) for the title; and
+ - [`border-subtitle-align`](../styles/border_subtitle_align.md) for the subtitle.
+
+--8<-- "docs/snippets/border_sub_title_align_all_example.md"
```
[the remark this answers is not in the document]
rodrigogiraoserrao: I added a title and a subtitle to the previous border example (for continuity) instead of making a brand new example.

src/textual/_styles_cache.py
willmcgugan: By "here" I meant the code that follows, not necessarily just this line. If there was no label, could you avoid calling that `render_border_label` entirely (and possibly avoid calculating some of its parameters)?

rodrigogiraoserrao: Commit e05e50b should do it.

PR #2074, opened 2023-03-15
fix border issue

willmcgugan:
Fixes an issue where a border was partially black. I'd seen this before but it was a mystery why until recently...

You can see the issue in this snapshot:

Screenshot 2023-03-15 at 21 36 10


tests/snapshot_tests/snapshot_apps/border_alpha.py
```diff
@@ -1,43 +1,40 @@
 from textual.app import App, ComposeResult
-from textual.containers import Grid
-from textual.widget import Widget
+from textual.containers import Vertical
```
willmcgugan, line 2: Rewrote this snapshot test, because it wasn't clear to me what it should look like.

davep: The problem I see with the rewrite is that it doesn't exercise the CSS parser in quite the same way. The original test was designed to try out a good combination of styles, colours and alpha values to ensure it all worked together.

That looks to have been lost.

rodrigogiraoserrao (approved):
Barring Dave's comment on the snapshot test, looks good to me.

tests/snapshot_tests/snapshot_apps/border_alpha.py
willmcgugan: True. But those are better captured with a unit test. Snapshots should be used to validate that things look correct only, where you can't tell from the data alone.

PR #2075, opened 2023-03-15
Allow customizing the markdown parser

jepler:
For instance, code using Markdown might wish to create a markdown parser that does not parse embedded HTML:
```py
def parser_factory():
    parser = MarkdownIt()
    parser.options['html'] = False
    return parser
```

This is needed to properly render chatgpt's markdown; there are many "knobs" of MarkdownIt and users of the textual library may need to tune them to their needs.


src/textual/widgets/_markdown.py
```diff
@@ -574,7 +577,7 @@ async def update(self, markdown: str) -> None:
         """
         output: list[MarkdownBlock] = []
         stack: list[MarkdownBlock] = []
-        parser = MarkdownIt("gfm-like")
+        parser = self._parser_factory()
```
willmcgugan, line 580: Does this still create a "gfm-like" parser?

jepler: oh, no, it's totally wrong now :)

jepler: It should be fixed now. thank you for the feedback!

src/textual/widgets/_markdown.py
```diff
@@ -509,6 +514,7 @@ def __init__(
         name: str | None = None,
         id: str | None = None,
         classes: str | None = None,
+        parser_factory: Callable[[], MarkdownIt] = default_parser_factory,
```
willmcgugan, line 517: Can we make this None-able. When you need to construct the parser, check for None and create `MarkdownIt("gfm-like")`.

This means we could drop the `default_parser_factory` function.

src/textual/widgets/_markdown.py
```diff
@@ -799,6 +807,7 @@ def __init__(
...
```
willmcgugan, line 810: Some comment as above. Let's make this None able.

willmcgugan (changes requested):
Like it. Just a request re the API.

src/textual/widgets/_markdown.py
jepler: I can rewrite it this way, sure.

Sorry for prematurely requesting your re-review.

src/textual/widgets/_markdown.py
```diff
@@ -509,6 +508,7 @@ def __init__(
         name: str | None = None,
         id: str | None = None,
         classes: str | None = None,
+        parser_factory: Optional[Callable[[], MarkdownIt]] = None
```
willmcgugan, line 511: Can you use `| None` here, just for consistency sake...

jepler:
Thanks again for the feedback. I've updated the PR once more.

willmcgugan (approved):
Looks good. Thanks!

willmcgugan:
Test fail is unrelated to this change. Merging.

jepler:
Thank you!

PR #2087, opened 2023-03-18
Use cleaner API and allow custom file in json_tree example

royatt:

- [x] Updated documentation


willmcgugan:
Since this is an example, we don't need to over-engineer it.

If you could make a PR that just replaced the private methods, I would merge that.

royatt:
> 

Fair enough, https://github.com/Textualize/textual/pull/2138

Issue #1908, opened 2023-03-01
Returning `self` instead of `None` for `Widget` "verb" methods


davep:
### Discussed in https://github.com/Textualize/textual/discussions/1817

<div type='discussions-op-text'>

<sup>Originally posted by **liancheng** February 16, 2023</sup>
My app has a form-like interface, and bound one key to each widget to set focus and optionally adjust the input, e.g., toggle a switch, move the cursor to the next `ListItem` in a `ListView`, etc.

For `Switch`es, I wrote the following method to achieve this:

```python
def action_toggle_switch(self, selector: str) -> None:
    switch = self.query_one(selector, Switch)
    switch.focus()
    switch.toggle()
```

The fact that both `focus()` and `toggle()` return `None` prevented me from chaining the calls. If they returned `self` instead, the above method can be simplified to:

```python
def action_toggle_switch(self, selector: str) -> None:
    self.query_one(selector, Switch).focus().toggle()
```

I'm not quite sure whether this violates certain well-established Python best practices or conventions, though.</div>

Chatted with @willmcgugan about this and the general feeling is this would be a good thing to do (if done carefully).

rodrigogiraoserrao:
This was implemented in #2102.

PR #2102, opened 2023-03-22
Return 'self' in some widget verb methods.

rodrigogiraoserrao:
I compiled a list of all widget methods that return `None` and for which it _could_ make sense to make this change. (I filtered out some methods, like watch and action methods.)

I tried choosing a subset of those methods, trying to only pick methods for which there weren't two things that could be returned (e.g., `Widget.move_child` _could_ return either the widget or the child that was moved) and I also tried to only pick methods that have little or no parameters (e.g., `Widget.animate` has many parameters and is typically called with quite a few.)

These are all the `Widget` methods for which this could make sense:
- `move_child` (either return `self` or the actual `child` that was moved…)
- `animate`
- `scroll_to` / `scroll_relate` / `scroll_home` / `scroll_end` / `scroll_left` / `scroll_right` / `scroll_down` / `scroll_up` / `scroll_page_up` / `scroll_page_down` / `scroll_page_left` / `scroll_page_right` / `scroll_visible`
- `refresh`
- `focus` / `reset_focus`
- `capture_mouse` / `release_mouse`

Additionally, I looked at each widget, and found these methods:
- `Tree`
  - `TreeNode`
     - `expand` / `expand_all` / `collapse` / `collapse_all` / `toggle` / `toggle_all`
     - `set_label`
   - `clear` / `reset`
   - `select_node` (either return `self` or the actual `node` that was selected)
   - `scroll_to_line` / `scroll_to_node`
   - `refresh_line`
- `ToggleButton`
  - `toggle` (and `action_toggle`?)
- `TextLog`
  - `write`
  - `clear`
- `Tabs`
  - `add_tab` / `remove_tab`
  - `clear`
- `Switch`
  - `toggle` (and `action_toggle`?)
- `Static`
  - `update`
- `Pretty`
  - `update`
- `Placeholder`
  - `cycle_variant`
- `_markdown.py`
  - `MarkdownBlock`
    - `set_content`
  - `MarkdownTableOfContents`
    - `set_table_of_contents`
- `Input`
  - `insert_text_at_cursor`
- `DirectoryTree`
  - `load_directory`
- `DataTable`
  - `update_cell` / `update_cell_at`
  - `clear`
  - `refresh_coordinate` / `refresh_row` / `refresh_column`
  - `sort`
- `Button`
  - `press`

Related issues: #1908
Related discussions: #1817

### DIFF
--- a/src/textual/widgets/_data_table.py
+++ b/src/textual/widgets/_data_table.py
@@ -13,7 +13,7 @@ from rich.protocol import is_renderable
 from rich.segment import Segment
 from rich.style import Style
 from rich.text import Text, TextType
-from typing_extensions import Literal, TypeAlias
+from typing_extensions import Literal, Self, TypeAlias
 
 from .. import events
 from .._cache import LRUCache
@@ -1156,7 +1156,7 @@ class DataTable(ScrollView, Generic[CellType], can_focus=True):
         full_column_region = Region(x, 0, width, height)
         return full_column_region
 
-    def clear(self, columns: bool = False) -> None:
+    def clear(self, columns: bool = False) -> Self:
         """Clear the table.
 
         Args:
@@ -1175,7 +1175,7 @@ class DataTable(ScrollView, Generic[CellType], can_focus=True):
         self.hover_coordinate = Coordinate(0, 0)
         self._label_column = Column(self._label_column_key, Text(), auto_width=True)
         self._labelled_row_exists = False
-        self.refresh()
+        return self.refresh()
 
     def add_column(
         self, label: TextType, *, width: int | None = None, key: str | None = None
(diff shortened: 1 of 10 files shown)


src/textual/widgets/_tree.py
```diff
@@ -89,7 +89,7 @@ def __init__(
         *,
         expanded: bool = True,
         allow_expand: bool = True,
-    ) -> None:
+    ):
```
davep, line 92: Accidental removal of return type hint here?

rodrigogiraoserrao: 😭  why are my fingers so fat?
Thanks for staying on top of things.

src/textual/widgets/_data_table.py
```diff
@@ -1175,7 +1178,7 @@ def clear(self, columns: bool = False) -> None:
         self.hover_coordinate = Coordinate(0, 0)
         self._label_column = Column(self._label_column_key, Text(), auto_width=True)
         self._labelled_row_exists = False
-        self.refresh()
+        return self.refresh()
```
willmcgugan, line 1181: I think it is better to explicitly return self, rather than return the result of `refresh`.

Otherwise we create a tight coupling between the two methods. i.e. if `refresh` changes, we break `clear`

src/textual/widgets/_data_table.py
```diff
@@ -1333,49 +1336,62 @@ def on_idle(self) -> None:
...
+
+        Returns:
+            The `DataTable` instance.
         """
         if not self.is_valid_coordinate(coordinate):
-            return
+            return self
         region = self._get_cell_region(coordinate)
-        self._refresh_region(region)
+        return self._refresh_region(region)
```
willmcgugan, line 1351: explicit self would be getter.

willmcgugan (changes requested):
Looks good. But I think it would be best to always explicitly return `self`, rather than return the result of another method.

rodrigogiraoserrao:
> [...] always explicitly return `self` [...]

That makes sense.

willmcgugan (approved):

PR #2112, opened 2023-03-22
Add tests for action parsing.

### LINKED ISSUE #2088: `)` cannot appear as part of the parameter passed to an action
Reproduction:

```python
from textual.app import App


class ActionBugApp(App):
    BINDINGS = [("a", "test(')')", "Test")]

    def action_test(self, _: str) -> None:
        pass


if __name__ == '__main__':
    app = ActionBugApp()
    app.run()
```

Omitting the full stack trace (since it's fairly easy to reproduce), the key error message is:

```
ActionError: unable to parse "(')" in action "test(')')"
```

Seems that [this regex](https://github.com/Textualize/textual/blob/2a6368754a8b3a11f1772b52298b5d3b50ceebaa/src/textual/actions.py#L20) is not general enough.

rodrigogiraoserrao:
This will close #2088.

This also uncovered a couple of more nuanced bugs that I have fixed in this PR and that are covered by tests.
These have to do with the previous way in which we decided if we should wrap the parsed parameters into another tuple or not.

It is easier to see the issue than to explain it.
The REPL session below uses  the "current" tip of `main`, e15805b93bbd2a4b2e321b0c0accba23bf198b86, _**after**_ changing the regex to `r"([\w\.]+)(\(.*\))"` as suggested by Dave in the linked issue to allow using `)` in the arguments.

```pycon
>>>  _, args = parse("f((1, 2, 3))")
>>> args
(1, 2, 3)
```

This is wrong! The result should be `((1, 2, 3),)`, a 1-item tuple with the tuple `(1, 2, 3)` inside it.

The returned action arguments must be "a tuple with all arguments".
Hence, if the only argument is a tuple, then we must return a 1-item tuple with that argument tuple inside it.
Previously, we would return the tuple itself, as if it were a collection of arguments.

To tackle this, we tweak the regex (so that the action arguments group doesn't capture the outer () from the function call) and add an auxiliary function that determines if a given string that parses as a tuple is a single tuple argument or not to cover all edge cases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+
+## Unreleased
+
+### Fixed
+
+- Issue with parsing action strings whose arguments contained quoted closing parenthesis https://github.com/Textualize/textual/pull/2112
+- Issues with parsing action strings with tuple arguments https://github.com/Textualize/textual/pull/2112
+
 ## [0.16.0] - 2023-03-22
 
 ### Added
--- a/src/textual/actions.py
+++ b/src/textual/actions.py
@@ -6,7 +6,7 @@ import re
 from typing_extensions import Any, TypeAlias
 
 ActionParseResult: TypeAlias = "tuple[str, tuple[Any, ...]]"
-"""An action is its name and the arbitrary tuple of its parameters."""
+"""An action is its name and the arbitrary tuple of its arguments."""
 
 
 class SkipAction(Exception):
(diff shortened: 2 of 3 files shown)


rodrigogiraoserrao:
The bugs fixed, which are discussed with some detail in the relevant parts of the code, have to do with the previous way in which we decided if we should wrap the parsed parameters into another tuple or not.

The returned action arguments must be "a tuple with all arguments".
Hence, if the _only_ argument is a tuple, then we must return a 1-item tuple with that argument tuple inside it.
Previously, we would return the tuple itself, as if it were a collection of arguments.

To tackle this, we tweak the regex (so that the action arguments group doesn't capture the outer `()` from the function call) and add an auxiliary function that determines if a given string that parses as a tuple is a single tuple argument or not.

@willmcgugan let me know if this suffices ^

willmcgugan:
Can you edit the description please. So we have a record of what was fixed.

And can you add some explicit examples that were broken prior to this PR. I'm struggling to understand the problem you fixed...

willmcgugan:
Ok, I see what you are trying to do. The `literal_eval` doesn't quite parse the function call syntax that we are trying to emulate.

The solution of stripping braces and attempting to feels risky though. We should be a bit more confident than just attempting and catching errors.

Ideally there would be some kind of `parse_literal_call_syntax`, which would be possible to implement, but would require significant effort.

What would happen if we were to take the string in braces, and wrap that in tuple syntax? Something like this:

```python
ast.literal_eval(f"({action},)")
```

That should always return a tuple.

What do you think?

rodrigogiraoserrao:
> What do you think?

I think I'm sad I didn't come up with this myself, but it works! Good job 🚀

willmcgugan (approved):

Issue #2105, opened 2023-03-22
`TextLog.write`: optional auto scroll


notEvil:
Hi,

`TextLog.write` always scrolls to the end. I use

```python
class TextLog(textual.widgets.TextLog):
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)

        self._auto_scroll = True
        self._writing = False

    def watch_scroll_y(self, *args, **kwargs):
        if len(args) < 2:
            return super().watch_scroll_y(*args, **kwargs)
        self._auto_scroll = args[1] == self.max_scroll_y
        return super().watch_scroll_y(*args, **kwargs)

    def write(self, *args, **kwargs):
        self._writing = True
        try:
            object = super().write(*args, **kwargs)
        finally:
            self._writing = False
        return object

    def scroll_end(self, *args, **kwargs):
        if self._writing and not self._auto_scroll:
            return
        return super().scroll_end(*args, **kwargs)
```

to disable auto scroll when scrolling off the last line (`max_scroll_y`), similar to most terminals, but would prefer an official solution.

PR #2127, opened 2023-03-23
Text log scroll end

willmcgugan:
Adds options to influence scrolling behaviour in TextLog

Fixes https://github.com/Textualize/textual/issues/2105

### DIFF
--- a/src/textual/widgets/_text_log.py
+++ b/src/textual/widgets/_text_log.py
@@ -38,6 +38,7 @@ class TextLog(ScrollView, can_focus=True):
     wrap: var[bool] = var(False)
     highlight: var[bool] = var(False)
     markup: var[bool] = var(False)
+    auto_scroll: var[bool] = var(True)
 
     def __init__(
         self,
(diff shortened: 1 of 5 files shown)


davep (approved):

src/textual/widgets/_text_log.py
```diff
@@ -92,19 +97,26 @@ def write(
...
+            scroll_end: Enable automatic scroll to end, or `None` to use `self.auto_scroll`.
 
         Returns:
             The `TextLog` instance.
         """
 
+        if scroll_end is None:
+            auto_scroll = self.auto_scroll
+        else:
+            auto_scroll = scroll_end
```
rodrigogiraoserrao, line 118: I thought we liked conditional expressions?

How do you typically determine whether you'll use a conditional expression or an `if: ... else: ...`?

```suggestion
        auto_scroll = self.auto_scroll if scroll_end is None else scroll_end
```

tests/snapshot_tests/snapshot_apps/textlog_scroll.py
rodrigogiraoserrao, line 1: Kudos on remembering the snapshot test file, this time! 😜

rodrigogiraoserrao (approved):

src/textual/widgets/_text_log.py
willmcgugan: I think it is just a stylistic thing for the most part.

PR #2128, opened 2023-03-23
Fix issues with watching CSS reloading

### LINKED ISSUE #2063: Watching CSS for border changes renders misplaced content
When (un)setting the border while watching for live CSS changes, the content of the label gets misplaced.

https://user-images.githubusercontent.com/5621605/225134232-c680b9dc-c157-4725-946f-e550425c02fc.mov

The app and CSS follow.

(This is copy & pasted from an example for the border title and subtitle, but the issue also occurs in the currently released version, and this app will run just fine despite the assignment to two meaningless attributes)

```py
from textual.app import App
from textual.containers import Grid, Container


def make_label_container(
    text: str, id: str, border_title: str, border_subtitle: str
) -> Container:
    lbl = Label(text, id=id)
    lbl.border_title = border_title
    lbl.border_subtitle = border_subtitle
    return Container(lbl)


class BorderSubTitleAlignAll(App[None]):
    def compose(self):
        with Grid():
            yield make_label_container(
                "Just some label text here...", "lbl1", "[b]Title.[/b]", "[u]Subtitle."
            )
            yield make_label_container(
                "Just some label text here...", "lbl2", "[b]Title.[/b]", "[u]Subtitle."
            )
            yield make_label_container(
                "Just some label text here...", "lbl3", "[b]Title.[/b]", "[u]Subtitle."
            )
            yield make_label_container(
                "Just some label text here...", "lbl4", "[b]Title.[/b]", "[u]Subtitle."
            )
            yield make_label_container(
                "Just some label text here...", "lbl5", "[b]Title.[/b]", "[u]Subtitle."
            )
            yield make_label_container(
                "Just some label text here...", "lbl6", "[b]Title.[/b]", "[u]Subtitle."
            )
            yield make_label_container(
                "Just some label text here...", "lbl7", "[b]Title.[/b]", "[u]Subtitle."
            )
            yield make_label_container(
                "Just some label text here...", "lbl8", "[b]Title.[/b]", "[u]Subtitle."
            )
            yield make_label_container(
                "Just some label text here...", "lbl9", "[b]Title.[/b]", "[u]Subtitle."
            )


app = BorderSubTitleAlignAll(css_path="border_sub_title_align_all.css")
```

```sass
Grid {
    grid-size: 3 3;
    align: center middle;
}

Container {
    width: 100%;
    height: 100%;
    align: center middle;
}

#lbl2 {
    border: round $secondary;
}

```

rodrigogiraoserrao:
This will close #2063.

The only reason I am a bit confident in this fix is because all the tests pass.
I am not 100% **sure** why this works, but I have a conjecture: the call to `reset_styles` was flagging the need to repaint a widget too soon and it was either preventing the latter refresh from happening or it was keeping some outdated information around.

I got to this conclusion because setting the border property programmatically works, and the code that runs in that case is _just_ the `__set__` in the descriptor and the layout refresh, as opposed to what happens when the CSS is reloaded, which runs _many more_ lines of code, so I figured that the CSS reloading was “doing too much work”.

**Note**: Playing with the code and running the tests suggests we can actually get rid of the method `reset_styles` altogether.


willmcgugan:
That may be right. I'm assuming you've pushed the hot reloading quite hard and can't break it.

If that method isn't called anywhere, then I'd go ahead and remove it.

willmcgugan:
@rodrigogiraoserrao Still working on those tests?

rodrigogiraoserrao:
> @rodrigogiraoserrao Still working on those tests?

Yeah, there wasn't a good way to test the hot reloading and take a snapshot test.
Wrapping up now.

rodrigogiraoserrao:
@willmcgugan now I'm happy.

willmcgugan (approved):

PR #2134, opened 2023-03-24
Fix scrollbar thumb

willmcgugan:
The scrollbar thumb would sometimes not extend to the bottom of the window.

There was a clever "fix" for this I implemented a while ago. It was so clever it was broken.


willmcgugan:
@davep You may want to try this branch against your options list work.

davep:
Noted. Will do.

willmcgugan:
#executivedecision feel free to do a retroactive review

PR #2139, opened 2023-03-25
compositor refactor and transparent screens

willmcgugan:
- Implements transparent screens via background opacity
- Refactor of compositor to break up monolithic function
- Bumps Rich to 13.3.0 for new method
- Fixed a bug where `fr` units didn't work within Grid layout

Screenshot 2023-03-27 at 12 18 30


src/textual/app.py
```diff
@@ -1321,6 +1337,14 @@ def push_screen(self, screen: Screen | str) -> AwaitMount:
...
 
         """
+        if not isinstance(screen, (Screen, str)):
+            raise TypeError(
+                f"push_screen requires a Screen instance or str; not {screen!r}"
+            )
+
+        if self._screen_stack:
+            self.screen.post_message(events.ScreenSuspend())
+            self.screen.refresh()
```
rodrigogiraoserrao, line 1347: I don't understand why these two method calls on `self.screen` are guarded by the `if`.
Is this to check if the currently active screen is _not_ the app default screen?

src/textual/app.py
```diff
@@ -1321,6 +1337,14 @@ def push_screen(self, screen: Screen | str) -> AwaitMount:
...
         next_screen, await_mount = self._get_screen(screen)
         self._screen_stack.append(next_screen)
         self.screen.post_message(events.ScreenResume())
```
rodrigogiraoserrao, line 1350: By appending to `_screen_stack`, then `self.screen` will already point to `next_screen`?
I think it would be clearer if this was
```suggestion
        self._screen_stack.append(next_screen)
        next_screen.post_message(events.ScreenResume())
```

Make it easier to follow to what screen the message is being posted.

src/textual/_compositor.py
```diff
@@ -800,41 +806,83 @@ def _get_renders(
...
-            A renderable
+            A renderable for the update, or `None` if no update was required.
         """
 
-        width, height = self.size
-        screen_region = Region(0, 0, width, height)
-
-        if full:
-            update_regions: set[Region] = set()
+        visible_screen_stack.set(screen_stack or [])
```
rodrigogiraoserrao, line 821: If `screen_stack` is `[]`, then this evaluates to the right-hand empty list `[]`.
So, if later someone mutates the `screen_stack` that was `[]`, the change will not be reflected in `visible_screen_stack` because we have a reference to a different list.

Can this be an issue in any way?

src/textual/_compositor.py
```diff
@@ -800,41 +806,83 @@ def _get_renders(
...
-            if screen_region in update_regions:
-                # If one of the updates is the entire screen, then we only need one update
-                full = True
+            return self.render_partial_update()
+
+    def render_full_update(self) -> LayoutUpdate:
+        """Render a full update.
+
+        Returns:
+            A LayoutUpdate renderable
```
rodrigogiraoserrao, line 832:
```suggestion
            A LayoutUpdate renderable.
```

rodrigogiraoserrao (commented):
This was very hard to review, for me.
So, I left plenty of comments, but most are questions for you to help me understand what is going on.

Depending on your answers, I might feel more confident in hitting the "Approve" button.

docs/examples/guide/screens/modal02.css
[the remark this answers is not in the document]
willmcgugan: Turned out I didn't need it.

docs/examples/guide/screens/modal02.py
```diff
@@ -0,0 +1,50 @@
...
+        if event.button.id == "quit":
+            self.app.exit()
+        else:
+            self.app.pop_screen()
+
+
+class ModalApp(App):
+    """A app with a modal dialog."""
+
+    CSS_PATH = "modal01.css"
```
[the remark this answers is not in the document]
willmcgugan: It does indicate that we're using the same CSS file. And it follows a convention...

src/textual/_compositor.py
willmcgugan: I don't think it will be an issue, but it can't hurt to be explicit with `None`...

CHANGELOG.md
```diff
@@ -27,12 +27,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   - `TextLog`: `write`, `clear`
   - `TreeNode`: `expand`, `expand_all`, `collapse`, `collapse_all`, `toggle`, `toggle_all`
   - `Tree`: `clear`, `reset`
-- Replaced some private attributes with public ones in the json tree example. https://github.com/Textualize/textual/pull/2138
+- Screens with alpha in their background color will now blend with the background. https://github.com/Textualize/textual/pull/2139
```
davep, line 30: Was it intentional that this nuked the CHANGELOG entry about the JSON tree example?

src/textual/_compositor.py
```diff
@@ -800,41 +806,83 @@ def _get_renders(
...
+        """
+        chops = self._render_chops(self.size.region, lambda y: True)
+        render_strips = [Strip.join(chop.values()) for chop in chops]
+        return render_strips
+
+    def _render_chops(
+        self,
+        crop: Region,
+        is_rendered_line: Callable[[int], bool],
+    ) -> list[dict[int, Strip | None]]:
```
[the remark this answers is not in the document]
willmcgugan: Not yet, it's a bit lower level.

src/textual/renderables/background_screen.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def __init__(
+        self,
+        screen: Screen,
+        color: Color,
+    ) -> None:
+        """Tints a renderable and removes links / meta.
+
+        Args:
+            renderable: A renderable.
```
davep, line 26: Looks like a copy/paste-o here? This looks like it should be `screen`.

src/textual/renderables/background_screen.py
```diff
@@ -0,0 +1,82 @@
...
+            color: A color (presumably with alpha).
+        """
+        self.screen = screen
+        self.color = color
```
davep, line 30: This should likely get a docstring.

src/textual/app.py
willmcgugan: This may be called when there are no screens, and hence `self.screen` raises an exception.

davep (approved):
Pretty much a handful of docstring things; looks like the European Office got to the other stuff first. ;-)

src/textual/app.py
```diff
@@ -1938,22 +1966,37 @@ def bell(self) -> None:
...
-                (node, node._bindings) for node in focused.ancestors_with_self
-            ]
+            if screen.is_modal:
+                namespace_bindings = [
+                    (node, node._bindings) for node in focused.ancestors
+                ]
+            else:
+                namespace_bindings = [
+                    (node, node._bindings) for node in focused.ancestors_with_self
+                ]
```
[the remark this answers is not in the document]
willmcgugan: Personally I would find that less easy to follow.  You could go one step further and combine the condition with another ternary expression

CHANGELOG.md
willmcgugan: Yeah, examples aren't relevant. Folk only care about the API.

src/textual/renderables/background_screen.py
```diff
@@ -0,0 +1,82 @@
...
+            segments: Incoming segments.
+            color: Color of tint.
+
+        Returns:
+            Segments with applied tint.
+
+        """
+        from_rich_color = Color.from_rich_color
+        style_from_color = Style.from_color
+        _Segment = Segment
```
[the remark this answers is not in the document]
willmcgugan: Yeah, locals are faster than globals. If you are doing many lookups this can make a difference.

The difference is much less with more recent Pythons. but I think it is still worth it.

src/textual/screen.py
```diff
@@ -43,19 +45,27 @@ class Screen(Widget):
     """
 
     focused: Reactive[Widget | None] = Reactive(None)
+    """The focused widget or `None` for no focus."""
+    stack_updates: Reactive[int] = Reactive(0, repaint=False)
```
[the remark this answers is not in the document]
willmcgugan: It is to allow the footer to watch for changes. When the screen stack is updated, the footer can update itself with the new bindings.

src/textual/strip.py
```diff
@@ -27,6 +28,21 @@ def get_line_length(segments: Iterable[Segment]) -> int:
     return sum(_cell_len(text) for text, _, control in segments if not control)
 
 
+class StripRenderable:
+    """A renderable which renders a list of strips."""
```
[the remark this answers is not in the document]
willmcgugan: Strips are used to render lines. But it couldn't hurt to mention that in the docstring.

src/textual/widget.py
```diff
@@ -291,14 +291,20 @@ def __init__(
...
         self._add_children(*children)
         self.disabled = disabled
 
     virtual_size = Reactive(Size(0, 0), layout=True)
     auto_width = Reactive(True)
     auto_height = Reactive(True)
-    has_focus = Reactive(False)
-    mouse_over = Reactive(False)
+    has_focus = Reactive(False, repaint=False)
+    mouse_over = Reactive(False, repaint=False)
```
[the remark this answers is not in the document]
willmcgugan: Yeah. The update is conditional, and done in `watch_` methods. Without this change, Textual was doing some superfluous updates.

Issue #2053, opened 2023-03-14
Menu Line API widget


willmcgugan:
We need a `Menu` widget which displays a list of items that may be navigated and selected.

Similar to ListView but using the Line API.

The menu should accept a list of renderables (not just str and Text), which will be presented in a vertically stacked list. It should display a cursor that may be moved up and down.

- We would need to identify the menu items with an optional id. Perhaps also by index.
- Renderables may be an arbitrary size. The `add_item` method should accept a fixed height, but a height of None should render the item with an automatic height.
- Menu items should allow an automatic checkbox control which updates a set on the parent.

PR #2154, opened 2023-03-28
Add `OptionList` widget

davep:
Adds an `OptionList` widget, which can form the basis of traditional bounce-bar menus, and also can provide a low-cost item list, etc.

Initially started to implement the requirement outlined in #2053 but turned into something more foundational.

Key features include:

- Just a single widget, so ideally allowing *many* options within the list with little cost to the application.
- Each option's prompt can be any sort of Rich renderable, with multi-line renderables supported.
- Messages for highlight and selection.
- Can be built up from a list of strings, or `Option` classes.
- Provides support for non-option content -- currently supporting a separator.

Screenshot 2023-03-29 at 09 36 20

Screenshot 2023-03-29 at 09 52 40

Screenshot 2023-03-29 at 09 52 51

Screenshot 2023-03-29 at 09 53 03


src/textual/widgets/_option_list.py
```diff
@@ -0,0 +1,904 @@
...
+        color: $text-disabled;
+        background: $accent 30%;
+    }
+
+    OptionList:focus > .option-list--option-highlighted-disabled {
+        background: $accent 40%;
+    }
+
+    OptionList > .option-list--option-hover {
+        background: $panel-lighten-2;
```
willmcgugan, line 203: Consider $boost for these states, which brightens or darkens depending on the start of dark mode.

src/textual/widgets/_option_list.py
```diff
@@ -0,0 +1,904 @@
...
+        Args:
+            index: The ID of the option to get.
+
+        Returns:
+            The option at with the ID.
+
+        Raises:
+            KeyError: If no option has the given ID.
+        """
+        return self.get_option_at_index(self._option_ids[option_id])
```
willmcgugan, line 689: I think we might want to have a dedicated exception here. Probably the same exception as the previous method.

willmcgugan (commented):
Great work. A few suggestions.

src/textual/widgets/_option_list.py
```diff
@@ -0,0 +1,903 @@
...
+        self, prompt: RenderableType, id: str | None = None, disabled: bool = False
+    ) -> None:
+        """Initialise the option.
+
+        Args:
+            prompt: The prompt for the option.
+            id: An optional ID for the option.
+            disabled: The initial enabled/disabled state. Enabled by default.
+        """
+        self.__prompt = prompt
```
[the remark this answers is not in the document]
davep: Folk will be encouraged to inherit from `Option` if they want to add extra data as cargo for options, but I want to discourage messing with the prompt under the hood in any child class as much as possible (because there is no `Option` -> `OptionList` coupling that would have the prompt update.

src/textual/widgets/_option_list.py
```diff
@@ -0,0 +1,904 @@
...
+This type represnets all of the types that will be accepted when adding new
+content to the option list. This is a superset of `OptionListContent`.
+"""
+
+
+class OptionList(ScrollView, can_focus=True):
+    """A vertical option list with bounce-bar highlighting."""
+
+    BINDINGS: ClassVar[list[BindingType]] = [
+        Binding("down", "down", "Down", show=False),
```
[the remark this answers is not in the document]
davep: Good call.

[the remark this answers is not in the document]
davep: Done.

src/textual/widgets/_option_list.py
```diff
@@ -0,0 +1,904 @@
...
+        self._spans: list[OptionLineSpan] = []
+        """A list of the locations and sizes of all options in the option list.
+
+        This will be the same size as the number of prompts; each entry in
+        the list contains the line offset of the start of the prompt, and
+        the count of the lines in the prompt.
+        """
+
+        # Initial calculation of the content tracking.
+        self._request_content_tracking_refresh()
```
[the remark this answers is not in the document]
davep: As per offline conversation about this; kept it as is but tweaked `check_idle` to not post a `Prompt` if the widget isn't running yet.

src/textual/widgets/_option_list.py
```diff
@@ -0,0 +1,904 @@
...
+        Args:
+            old: The old status.
+            new: The new status.
+
+        `show_vertical_scrollbar` is watched because it has an impact on the
+        available width in which to render the renderables that make up the
+        options in the list. If a vertical scrollbar appears or disappears
+        we need to recalculate all the lines that make up the list.
+        """
+        if old != new:
```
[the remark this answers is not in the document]
davep: D'oh! Good call; always seem to forget about that.

src/textual/widgets/_option_list.py
```diff
@@ -0,0 +1,904 @@
...
+        """
+        return self.disable_option_at_index(self._option_ids[option_id])
+
+    @property
+    def option_count(self) -> int:
+        """The count of options."""
+        return len(self._options)
+
+    @property
+    def options(self) -> Iterator[Option]:
```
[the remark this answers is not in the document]
davep: It's only provided for public consumption; it isn't used in any capacity within the code itself. My thinking being that someone may wish to walk back over the options for some reason. Shall I just remove it?

src/textual/widgets/_option_list.py
davep: Nice idea, I always seem to forget about `$boost`. Done.

src/textual/widgets/_option_list.py
```diff
@@ -0,0 +1,904 @@
...
+        content that isn't an option.
+        """
+
+        self._option_ids: dict[str, int] = {}
+        """A dictionary of option IDs and the option indexes they relate to."""
+
+        self._lines: list[Line] = []
+        """A list of all of the individual lines that make up the option list.
+
+        Note that the size of this list will at least the same as the number
```
rodrigogiraoserrao, line 328:
```suggestion
        Note that the size of this list will be at least the same as the number
```

src/textual/widgets/_option_list.py
```diff
@@ -0,0 +1,904 @@
...
+        """Refresh the layout of the renderables in the list when resized."""
+        self._request_content_tracking_refresh(rescroll_to_highlight=True)
+
+    def on_mouse_move(self, event: MouseMove) -> None:
+        """React to the mouse moving.
+
+        Args:
+            event: The mouse movement event.
+        """
+        self._mouse_hovering_over = event.style.meta.get("option")
```
rodrigogiraoserrao, line 394: I still don't get what this `style.meta` is all about.
How does it always have all the information you need..?

src/textual/widgets/_option_list.py
```diff
@@ -0,0 +1,904 @@
...
+        Returns:
+            The `OptionList` instance.
+        """
+        # Turn any incoming value into valid content for the list.
+        content = self._make_content(item)
+        self._contents.append(content)
+        # If the content is a genuine option, add it to the list of options.
+        if isinstance(content, Option):
+            self._options.append(content)
+        self._refresh_content_tracking(force=True)
```
rodrigogiraoserrao, line 521: At this point, couldn't we do _less_ work because we know we are just adding a new option to the _end_ of the list?
Or do we always have to go through all of the `_refresh_content_tracking`?
(Or is it irrelevant that we go through that whole method again..?)

src/textual/widgets/_option_list.py
```diff
@@ -0,0 +1,904 @@
...
+        Args:
+            index: The index of the item to remove.
+
+        Raises:
+            IndexError: If there is no option of the given index.
+        """
+        option = self._options[index]
+        del self._options[index]
+        del self._contents[self._contents.index(option)]
+        self._refresh_content_tracking(force=True)
```
rodrigogiraoserrao, line 537: Similar question here: couldn't we "just" go over the options that show up after this one and offset their starting positions?

src/textual/widgets/_option_list.py
```diff
@@ -0,0 +1,904 @@
...
+        # Force a re-validation of the highlight.
+        self.highlighted = self.highlighted
+        self.refresh()
+
+    def remove_option(self, option_id: str) -> Self:
+        """Remove the open with the given ID.
```
rodrigogiraoserrao, line 543:
```suggestion
        """Remove the option with the given ID.
```

src/textual/widgets/_option_list.py
```diff
@@ -0,0 +1,943 @@
...
+        # content tracking isn't necessary. Can possibly improve this a wee
+        # bit.
+        self._option_ids.clear()
+
+    def _refresh_content_tracking(self, force: bool = False) -> None:
+        """Refresh the various forms of option list content tracking.
+
+        Args:
+            force: Optionally force the refresh.
+
```
rodrigogiraoserrao, line 440: Raises: DuplicateID missing.

src/textual/widgets/_option_list.py
```diff
@@ -0,0 +1,943 @@
...
+            OptionDoesNotExist: If there is no option with the index.
+        """
+        try:
+            return self._options[index]
+        except IndexError:
+            raise OptionDoesNotExist(
+                f"There is no option with an index of {index}"
+            ) from None
+
+    def get_option(self, option_id: str) -> Option:
```
rodrigogiraoserrao, line 711: I see there are a lot of `do_something_option(_at_index)`.
Would it be messy if all those pairs were collapsed into a single method that just received an index or an ID, and then retrieved the appropriate option?
@davep @willmcgugan

src/textual/widgets/_option_list.py
```diff
@@ -0,0 +1,943 @@
...
+        """Ensure that the highlighted option is in view."""
+        highlighted = self.highlighted
+        if highlighted is None:
+            return
+        try:
+            span = self._spans[highlighted]
+        except IndexError:
+            # Index error means we're being asked to scroll to a highlight
+            # before all the tracking information has been worked out.
+            # That's fine; let's just NoP that.
```
rodrigogiraoserrao, line 833: What's "NoP"?

src/textual/widgets/_option_list.py
```diff
@@ -0,0 +1,943 @@
...
+            return
+        self.scroll_to_region(
+            Region(
+                0, span.first, self.scrollable_content_region.width, span.line_count
+            ),
+            force=True,
+            animate=self.show_vertical_scrollbar,  # https://github.com/Textualize/textual/issues/2077
+        )
```
rodrigogiraoserrao, line 841: Thanks for this! This gave me an idea for another issue I'm tackling! 😆

src/textual/widgets/_option_list.py
```diff
@@ -0,0 +1,943 @@
...
+
+    def validate_highlighted(self, highlighted: int | None) -> int | None:
+        """Validate the `highlighted` property value on access."""
+        if not self._options:
+            return None
+        if highlighted is None or highlighted < 0:
+            return 0
```
rodrigogiraoserrao, line 848: I actually expected this to be
```py
if not self._options or highlighted is None:
    return None
if highlighted < 0:
    return None
```

It is a bit weird that `highlighted` _can_ be `None` but _I_ can't set it to `None` to remove highlighting.

src/textual/widgets/_option_list.py
```diff
@@ -0,0 +1,943 @@
...
+            if not self._options[highlighted].disabled:
+                self.post_message(self.OptionHighlighted(self, highlighted))
+
+    def action_cursor_up(self) -> None:
+        """Move the highlight up by one option."""
+        if self.highlighted is not None:
+            if self.highlighted > 0:
+                self.highlighted -= 1
+            else:
+                self.highlighted = len(self._options) - 1
```
rodrigogiraoserrao, line 864: A suggestion:
```suggestion
            self.highlighted = (self.highlighted - 1) % len(self._options)
```

src/textual/widgets/_option_list.py
```diff
@@ -0,0 +1,943 @@
...
+        elif self._options:
+            self.action_first()
+
+    def action_cursor_down(self) -> None:
+        """Move the highlight down by one option."""
+        if self.highlighted is not None:
+            if self.highlighted < len(self._options) - 1:
+                self.highlighted += 1
+            else:
+                self.highlighted = 0
```
rodrigogiraoserrao, line 874: For your consideration:
```suggestion
            self.highlighted = (self.highlighted + 1) % len(self._options)
```

src/textual/widgets/_option_list.py
```diff
@@ -0,0 +1,943 @@
...
+        elif self._options:
+            self.action_first()
+
+    def action_first(self) -> None:
+        """Move the highlight to the first option."""
+        if self._options:
+            self.highlighted = 0
```
rodrigogiraoserrao, line 881: Explicitly set to `None` otherwise?
I'm not sure if my suggestion makes sense here because I am not entirely aware of all the bookkeeping that is done with the property `highlighted`.
I'd say "better safe than sorry", but if you say everything is in order, then I can live without the addition of the `else`.

src/textual/widgets/_option_list.py
```diff
@@ -0,0 +1,943 @@
...
+
+    def action_last(self) -> None:
+        """Move the highlight to the last option."""
+        if self._options:
+            self.highlighted = len(self._options) - 1
```
rodrigogiraoserrao, line 886: See my comment in `action_first`.

src/textual/widgets/_option_list.py
```diff
@@ -0,0 +1,943 @@
...
+        fallback = self.action_first if direction == -1 else self.action_last
+
+        highlighted = self.highlighted
+        if highlighted is None:
+            # There is no highlight yet so let's go to the default position.
+            fallback()
+        else:
+            # We want to page roughly by lines, but we're dealing with
+            # options that can be a varying number of lines in height. So
+            # let's start with the a target line alone.
```
rodrigogiraoserrao, line 907: "with the a"?

rodrigogiraoserrao (changes requested):

src/textual/widgets/_option_list.py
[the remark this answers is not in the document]
willmcgugan: I have a feeling that we should remove it if there isn't an immediate need.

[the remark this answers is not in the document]
davep: Cool. Removed.

src/textual/widgets/_option_list.py
davep: Done.

src/textual/widgets/_option_list.py
```diff
@@ -0,0 +1,904 @@
...
+        Args:
+            index: The index of the option to get.
+
+        Returns:
+            The option at that index.
+
+        Raises:
+            IndexError: If there is no option with the index.
+        """
+        return self._options[index]
```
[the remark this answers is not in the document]
davep: Done. Also modified a bunch of other similar methods to use the same.

src/textual/widgets/_option_list.py
```diff
@@ -0,0 +1,904 @@
...
+
+
+class Separator:
+    """Class that denotes that a particular option is really a separator."""
+
+
+class Line(NamedTuple):
+    """Class that holds a list of segments for the line of a option."""
+
+    segments: list[Segment]
```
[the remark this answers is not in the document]
davep: Excellent call! Done.

src/textual/widgets/_option_list.py
```diff
@@ -0,0 +1,904 @@
...
+        # Now that we know which line we're on, pull out the option index so
+        # we have a "local" copy to refer to rather than needing to do a
+        # property access multiple times.
+        option_index = line.option_index
+
+        # Knowing which line we're going to be drawing, we can now go pull
+        # the relevant segments for the line of that particular prompt.
+        # While we're here, let's also attach which option index the line is
+        # with.
+        strip = Strip(line.segments).apply_style(Style(meta={"option": option_index}))
```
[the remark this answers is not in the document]
davep: Yup, makes sense now. Moved the application of the meta to when the `Line` is created.

src/textual/widgets/_option_list.py
davep: I feel this leaves the code just a little less readable to the casual reader without improving anything.

src/textual/widgets/_option_list.py
davep: I feel this leaves the code just a little less readable to the casual reader without improving anything.

src/textual/widgets/_option_list.py
davep: Aye, typo.

src/textual/widgets/_option_list.py
davep: `validate_highlighted` ensures `highlighted` is `None` if there are no options, so this code only deals with the action itself.

src/textual/widgets/_option_list.py
```diff
@@ -0,0 +1,904 @@
...
+        self.disabled = disabled
+
+    @property
+    def prompt(self) -> RenderableType:
+        """The prompt for the option."""
+        return self.__prompt
+
+    @property
+    def id(self) -> str | None:
+        """An optional ID for the option."""
```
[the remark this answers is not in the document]
davep: "The" is better here but I don't think optional needs to be parenthetical. ;-)

docs/widgets/option_list.md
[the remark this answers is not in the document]
davep: It is. Bindings too!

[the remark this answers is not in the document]
davep: And component classes. 🤷‍♀️

docs/widgets/option_list.md
```diff
@@ -0,0 +1,84 @@
+# OptionList
+
+!!! tip "Added in version 0.17.0"
+
+A widget for showing a vertical list of Rich renderable options.
+
+- [x] Focusable
+- [ ] Container
+
+## Examples
```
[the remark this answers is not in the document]
davep: Good call, done.

src/textual/widgets/_option_list.py
davep: We can look at this as a tweak later on if becomes necessary. Mostly it feels like the overhead of tracking this (in terms of following the code) seems to outweigh the benefit it'll give. There's a few places in the code where we can squeeze a bit more performance out. Adding/removing solitary items feels like one of the lesser wins.

src/textual/widgets/_option_list.py
davep: Same answer here.

src/textual/widgets/_option_list.py
davep: And more importantly on `add_option` too, where the user will actually see it.

src/textual/widgets/_option_list.py
```diff
@@ -0,0 +1,943 @@
...
+
+        Args:
+            index: The index of the option to set the disabled state of.
+            disabled: The disabled state to set.
+
+        Returns:
+            The `OptionList` instance.
+        """
+        self._options[index].disabled = disabled
+        # TODO: Refresh only if the affected option is visible.
```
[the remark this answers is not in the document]
davep: More a note to future-self if we want to squeeze a wee bit more speed out at some point.

src/textual/widgets/_option_list.py
davep: Amusingly this reminded me that I needed to just turn it off (at Will's request! WILL WANTED NO ANIMATION!)

src/textual/widgets/_option_list.py
davep: I don't think the meta stuff is very well documented yet (or at all). Might be worth having a chat with Will.

willmcgugan:
Is this good to go?

src/textual/widgets/_option_list.py
```diff
@@ -0,0 +1,904 @@
...
+        return self.__id
+
+    def __rich_repr__(self) -> Result:
+        yield "prompt", self.prompt
+        yield "id", self.id, None
+        yield "disabled", self.disabled, False
+
+
+class Separator:
+    """Class that denotes that a particular option is really a separator."""
```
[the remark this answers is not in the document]
rodrigogiraoserrao: Thanks! I'm quite a dumb reader myself 😆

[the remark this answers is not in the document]
davep: That's okay, I'm a terrible writer. 🤪

Issue #1081, opened 2022-10-31
Add a method of either setting classes to a specific collection, or clearing all classes


davep:
I've run into one situation where I wanted to be able to clear all of the classes on a `Widget` and set them to a very specific selection -- no toggle or the like. Perhaps it's worth considering adding a mechanism for doing this? Either "set to this very specific collection", or "clear all classes".

willmcgugan:
Suggest `replace_classes` which takes class names as positional args. Without any args it is equievelent to clearing the classes. With args, it will set the classes to just those specified.

PR #2160, opened 2023-03-28
Set classes

willmcgugan:
Adds the ability to change / clear classes.

- Adds a setter for classes, which may be intuitive to some.
- Also adds explicit `set_classes`

Fixes https://github.com/Textualize/textual/issues/1081

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,7 +40,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Added Screen.ModalScreen which prevents App from handling bindings. https://github.com/Textualize/textual/pull/2139
 - Added TEXTUAL_LOG env var which should be a path that Textual will write verbose logs to (textual devtools is generally preferred) https://github.com/Textualize/textual/pull/2148
 - Added textual.logging.TextualHandler logging handler
-- Added Query.set_classes DOMNode.set_classes and `classes` setter for Widget https://github.com/Textualize/textual/issues/1081
+- Added Query.set_classes, DOMNode.set_classes, and `classes` setter for Widget https://github.com/Textualize/textual/issues/1081
 
 ## [0.16.0] - 2023-03-22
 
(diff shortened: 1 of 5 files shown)


src/textual/dom.py
```diff
@@ -83,6 +83,26 @@ class NoScreen(DOMError):
     pass
 
 
+class _ClassesDescriptor:
```
willmcgugan, line 86: Mypy doesn't like it if you have a property where the setter and getter have different types. However, it is happy with a descriptor.

CHANGELOG.md
```diff
@@ -40,6 +40,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Added Screen.ModalScreen which prevents App from handling bindings. https://github.com/Textualize/textual/pull/2139
 - Added TEXTUAL_LOG env var which should be a path that Textual will write verbose logs to (textual devtools is generally preferred) https://github.com/Textualize/textual/pull/2148
 - Added textual.logging.TextualHandler logging handler
+- Added Query.set_classes DOMNode.set_classes and `classes` setter for Widget https://github.com/Textualize/textual/issues/1081
```
davep, line 43:
```suggestion
- Added Query.set_classes, DOMNode.set_classes, and `classes` setter for Widget https://github.com/Textualize/textual/issues/1081
```

src/textual/dom.py
```diff
@@ -888,6 +905,15 @@ def set_class(self, add: bool, *class_names: str) -> None:
         else:
             self.remove_class(*class_names)
 
+    def set_classes(self, classes: str | Iterable[str]) -> None:
```
davep, line 908: Perhaps this should return `self` so folk can chain calls?

davep (approved):
Text tweak and an idle question; but nothing of consequence form me. This looks super handy!

src/textual/dom.py
willmcgugan: It probably should. We don't have that on the other classes either.

rodrigogiraoserrao (approved):

PR #2172, opened 2023-03-30
Some docstrings for actions

darrenburns:
Some docstrings I had on a local branch on this machine.

### DIFF
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -2382,11 +2382,20 @@ class App(Generic[ReturnType], DOMNode):
             pass
 
     async def action_add_class_(self, selector: str, class_name: str) -> None:
-        """Add a CSS class on the selected widget."""
+        """Add a CSS class on the selected widget.
+
+        Args:
+            selector: Selects the widget to add the class to.
+            class_name: The class to add to the selected widget.
+        """
         self.screen.query(selector).add_class(class_name)
 
     async def action_remove_class_(self, selector: str, class_name: str) -> None:
-        """Remove a CSS class on the selected widget."""
+        """Remove a CSS class on the selected widget.
+
+        Args:
+            selector: Selects the widget to remove the class from.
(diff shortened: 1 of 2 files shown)


src/textual/app.py
```diff
@@ -2375,18 +2375,22 @@ async def action_pop_screen(self) -> None:
...
 
     async def action_back(self) -> None:
+        """Go back to the previous screen (pop the current screen)."""
         try:
             self.pop_screen()
         except ScreenStackError:
             pass
 
     async def action_add_class_(self, selector: str, class_name: str) -> None:
+        """Add a CSS class on the selected widget."""
```
davep: Should probably add the `Args` while here.

src/textual/app.py
```diff
@@ -2375,18 +2375,22 @@ async def action_pop_screen(self) -> None:
...
         self.screen.query(selector).add_class(class_name)
 
     async def action_remove_class_(self, selector: str, class_name: str) -> None:
+        """Remove a CSS class on the selected widget."""
```
davep: Should probably add the `Args` while here.

src/textual/app.py
```diff
@@ -2375,18 +2375,22 @@ async def action_pop_screen(self) -> None:
...
         self.screen.query(selector).remove_class(class_name)
 
     async def action_toggle_class(self, selector: str, class_name: str) -> None:
+        """Toggle a CSS class on the selected widget."""
```
davep, line 2402: Should probably add the `Args` while here.

davep (approved):
Suggest adding `Args` where there are arguments, but otherwise good.

src/textual/app.py
darrenburns: Sorry, I meant to make this PR a draft.

davep: But I'm sure this will be a perfect PR and can be pre-approved. :-P

PR #2174, opened 2023-03-30
Add FAQ about some keys not making it to a Textual app

davep:
Mostly this is about things like Cmd and Alt key combinations not being available in some environments; I've attempted to word it as a catch-all for the issue, trying to drive people to using `textual keys` and ensuring they understand their environments, their users' environments, and finding the most portable combinations for their applications.

Also added `make faq` suppport.

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -181,14 +181,28 @@ Textual can only ever support key combinations that are passed on by your
 terminal application. Which keys get passed on can differ from terminal to
 terminal, and from operating system to operating system.
 
+Because of this it's best to stick to key combinations that are known to be
+universally-supported; these include the likes of:
+
+- Letters
+- Numbers
+- Numbered function keys (especially F1 through F10)
+- Space
+- Return
+- Arrow, home, end and page keys
+- Control
+- Shift
+
 When [creating bindings for your
 application](https://textual.textualize.io/guide/input/#bindings) we
-recommend picking keys and key combinations that are supported on as many
-platforms as possible.
+recommend picking keys and key combinations from the above.
 
-The easiest way to test different environments to see which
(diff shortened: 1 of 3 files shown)


davep:
> I find this reply unhelpful. It basically says we don't know which keys work, and the dev would need to have access to all platforms to know which they can use.

My experience is that it does differ from terminal-to-terminal, and platform-to-platform. Can you suggest a better wording that doesn't mislead?

willmcgugan:
> My experience is that it does differ from terminal-to-terminal, and platform-to-platform.

I don't doubt that, but there is a huge overlap with what works.

I think its best to say what we know is universal: letters, numbers, function keys, space , return, digits etc. And ctrl, shift modifiers.

We know that meta and option doesn't work on macOS.

Anything else recommend trying `textual keys`

davep:
Take 2 is here...

FAQ.md
```diff
@@ -173,6 +174,36 @@ Greetings(to_greet="davep").run()
...
 
+<a name="why-do-some-key-combinations-never-make-it-to-my-app"></a>
+## Why do some key combinations never make it to my app?
+
+Textual can only ever support key combinations that are passed on by your
+terminal application. Which keys get passed on can differ from terminal to
+terminal, and from operating system to operating system.
+
+Because of this it's best to stick to key combinations that are known to be
+universally-supported; these include the likes of:
```
rodrigogiraoserrao, line 185:
```suggestion
universally-supported; these include:
```

rodrigogiraoserrao (approved):
Still looks good to me.
It also looks _better_ now.

rodrigogiraoserrao (approved):

willmcgugan (approved):

PR #2182, opened 2023-03-30
Worker API

willmcgugan:
- Adds a Worker API
- Adds Worker API docs
- Makes Markdown.update non-async
- Adds release blog post
- Makes post_message thread safe


src/textual/_work_decorator.py
```diff
@@ -0,0 +1,102 @@
...
+
+if TYPE_CHECKING:
+    from .worker import Worker
+
+
+FactoryParamSpec = ParamSpec("FactoryParamSpec")
+DecoratorParamSpec = ParamSpec("DecoratorParamSpec")
+ReturnType = TypeVar("ReturnType")
+
+Decorator: TypeAlias = Callable[
```
rodrigogiraoserrao, line 16: "Decorator" sounds like a very generic name that encompasses Python decorators.
Maybe something like "WorkerDecorator" would be more appropriate?

src/textual/_worker_manager.py
```diff
@@ -0,0 +1,164 @@
...
+            worker
+            for worker in self._workers
+            if (worker.group == group and worker.node == node)
+        ]
+        for worker in workers:
+            worker.cancel()
+        return workers
+
+    def cancel_node(self, node: DOMNode) -> list[Worker]:
+        """Cancel all workers associated with a given node."""
```
davep, line 151: Docstring is missing args and return.

src/textual/_work_decorator.py
```diff
@@ -0,0 +1,102 @@
...
+    method: Callable[FactoryParamSpec, ReturnType]
+    | Callable[FactoryParamSpec, Coroutine[None, None, ReturnType]]
+    | None = None,
+    *,
+    name: str = "",
+    group: str = "default",
+    exclusive: bool = False,
+    exit_on_error: bool = True,
+) -> Callable[FactoryParamSpec, Worker[ReturnType]] | Decorator:
+    """Worker decorator factory."""
```
rodrigogiraoserrao: Docstring, please!
What's `name`, `group`, `exclusive`, and `exit_on_error`?

src/textual/_work_decorator.py
```diff
@@ -0,0 +1,102 @@
...
+        def decorated(
+            *args: DecoratorParamSpec.args, **kwargs: DecoratorParamSpec.kwargs
+        ) -> Worker[ReturnType]:
+            """The replaced callable."""
+            from .dom import DOMNode
+
+            self = args[0]
+            assert isinstance(self, DOMNode)
+
+            positional_arguments = ", ".join(repr(arg) for arg in args[1:])
```
rodrigogiraoserrao, line 85: Faster alternatives:

```suggestion
            positional_arguments = ", ".join([repr(arg) for arg in args[1:]])
```
```suggestion
            positional_arguments = ", ".join(map(repr, args[1:]))
```

Sources: [1](https://mathspp.com/blog/til/008), [2](https://docs.python.org/3/library/timeit.html#basic-examples)

src/textual/_work_decorator.py
```diff
@@ -0,0 +1,102 @@
...
+            keyword_arguments = ", ".join(
+                f"{name}={value!r}" for name, value in kwargs.items()
+            )
```
rodrigogiraoserrao, line 88: Faster alternative:

```suggestion
            keyword_arguments = ", ".join(
                [f"{name}={value!r}" for name, value in kwargs.items()]
            )
```

src/textual/_work_decorator.py
```diff
@@ -0,0 +1,102 @@
...
+            tokens = [positional_arguments, keyword_arguments]
+            worker_description = (
+                f"{method.__name__}({', '.join(token for token in tokens if token)})"
```
rodrigogiraoserrao, line 91: Faster:

```suggestion
                f"{method.__name__}({', '.join([token for token in tokens if token])})"
```

_Probably_ even faster:

```suggestion
                f"{method.__name__}({', '.join(filter(None, tokens))})"
```

davep (approved):
Typo-squishing suggestions and stuff, and a question or two; looks pretty damn powerful!

src/textual/_worker_manager.py
```diff
@@ -0,0 +1,164 @@
...
+            start: Automatically start the worker.
+            exclusive: Cancel all workers in the same group.
+
+        Returns:
+            A Worker instance.
+        """
+        worker: Worker[Any] = Worker(
+            node,
+            work,
+            name=name or getattr(work, "__name__", "") or "",
```
rodrigogiraoserrao, line 105: Don't we already get an empty string from `getattr`?

```suggestion
            name=name or getattr(work, "__name__", ""),
```

src/textual/_worker_manager.py
```diff
@@ -0,0 +1,164 @@
...
+        """Start all the workers."""
+        for worker in self._workers:
+            worker._start(self._app, self._remove_worker)
+
+    def cancel_all(self) -> None:
+        """Cancel all workers."""
+        for worker in self._workers:
+            worker.cancel()
+
+    def cancel_group(self, node: DOMNode, group: str) -> list[Worker]:
```
rodrigogiraoserrao, line 131: It confused me seeing `node` up here, especially after finding the `cancel_node` method below.
Why don't we just accept the group here? (Maybe with an _optional_ node?)

src/textual/app.py
```diff
@@ -427,6 +429,11 @@ def __init__(
         self.set_class(self.dark, "-dark-mode")
         self.set_class(not self.dark, "-light-mode")
 
+    @property
+    def workers(self) -> WorkerManager:
+        """A worker manager."""
+        return self._workers
```
rodrigogiraoserrao, line 435: I think "workers" makes us think this will give us the workers being manager, not the worker manager itself.
I suggest `workers -> worker_manager` and `_workers -> _worker_manager`.

src/textual/dom.py
```diff
@@ -208,6 +210,49 @@ def auto_refresh(self, interval: float | None) -> None:
             )
         self._auto_refresh = interval
 
+    @property
+    def workers(self) -> WorkerManager:
+        """A worker manager."""
+        return self.app.workers
```
rodrigogiraoserrao, line 216: See similar comment under `App.workers`

src/textual/_animator.py
```diff
@@ -27,12 +27,14 @@ class AnimationError(Exception):
     """An issue prevented animation from starting."""
 
 
-T = TypeVar("T")
+ReturnType = TypeVar("T")
```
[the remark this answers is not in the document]
willmcgugan: No, that's a mistake. will fix.

src/textual/widgets/_markdown.py
```diff
@@ -589,7 +589,7 @@ def __init__(self, href: str) -> None:
 
     async def on_mount(self) -> None:
```
rodrigogiraoserrao, line 590: Make method non-async?

```suggestion
    def on_mount(self) -> None:
```

src/textual/widgets/_markdown.py
```diff
@@ -589,7 +589,7 @@ def __init__(self, href: str) -> None:
...
         if self._markdown is not None:
-            await self.update(self._markdown)
+            self.update(self._markdown)
 
     async def load(self, path: Path) -> bool:
```
rodrigogiraoserrao, line 594: Make method non-async?

```suggestion
    def load(self, path: Path) -> bool:
```

src/textual/widgets/_markdown.py
```diff
@@ -874,7 +874,7 @@ def table_of_contents(self) -> MarkdownTableOfContents:
 
     async def on_mount(self) -> None:
```
rodrigogiraoserrao, line 875: Make method non-async?

```suggestion
    def on_mount(self) -> None:
```

src/textual/worker.py
```diff
@@ -0,0 +1,388 @@
...
+active_worker: ContextVar[Worker] = ContextVar("active_worker")
+"""Currently active worker context var."""
+
+
+class NoActiveWorker(Exception):
+    """There is no active worker."""
+
+
+class WorkerError(Exception):
+    """A worker related error."""
```
rodrigogiraoserrao, line 39:
```suggestion
    """A worker-related error."""
```

src/textual/worker.py
```diff
@@ -0,0 +1,388 @@
...
+
+
+class WorkerFailed(WorkerError):
+    """The worker raised an exception and did not complete."""
+
+    def __init__(self, error: BaseException) -> None:
+        self.error = error
+        super().__init__(f"Worker raise exception: {error!r}")
```
rodrigogiraoserrao, line 47:
```suggestion
        super().__init__(f"Worker raised exception: {error!r}")
```

src/textual/worker.py
```diff
@@ -0,0 +1,388 @@
...
+    Raises:
+        NoActiveWorker: If there is no active worker.
+
+    Returns:
+        A Worker instance.
+    """
+    try:
+        return active_worker.get()
+    except LookupError:
+        raise NoActiveWorker("There is no active worker in this task or thread.")
```
rodrigogiraoserrao, line 70: Wouldn't we want to suppress the part of the error that is related to the `LookupError`?

```suggestion
        raise NoActiveWorker("There is no active worker in this task or thread.") from None
```

docs/guide/workers.md
```diff
@@ -0,0 +1,171 @@
...
+
+!!! info "What are threads?"
+
+    Threads are a form of concurrency supplied by your Operating System. Threads allow your code to run more than a single function simultaneously.
+
+You can create threads by applying `run_worker` or the `work` decorator to a plain (non async) method or function.
+The API for thread workers is identical to async workers, but there are a few differences you need to be aware of when writing threaded code.
+
+The first difference is that you should avoid calling methods on your UI directly, or setting reactive variables.
+You can work around this with the [App.call_from_thread][textual.app.App.call_from_thread] method which schedules a call in the main thread.
```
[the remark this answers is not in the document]
willmcgugan: Let me do that in a future revision.

src/textual/worker.py
```diff
@@ -0,0 +1,388 @@
...
+
+WorkType: TypeAlias = Union[
+    Callable[[], Coroutine[None, None, ResultType]],
+    Callable[[], ResultType],
+    Awaitable[ResultType],
+]
+
+
+class _ReprText:
+    """Shim to insert a word in to the Worker's repr."""
```
rodrigogiraoserrao, line 99:
```suggestion
    """Shim to insert a word into the Worker's repr."""
```

src/textual/worker.py
```diff
@@ -0,0 +1,388 @@
...
+        *,
+        name: str = "",
+        group: str = "default",
+        description: str = "",
+        exit_on_error: bool = True,
+    ) -> None:
+        """Initialize a Worker.
+
+        Args:
+            node: THe widget, screen, or App that initiated the work.
```
rodrigogiraoserrao, line 146:
```suggestion
            node: The widget, screen, or App that initiated the work.
```

src/textual/worker.py
```diff
@@ -0,0 +1,388 @@
...
+            work: A callable, coroutine, or other awaitable.
```
rodrigogiraoserrao, line 147: Instead of the types, wouldn't it make more sense that `work` is supposed to be an object that holds the work / does the work / makes things happen?
I.e. the thing that will actually "run" and update the app?

src/textual/worker.py
```diff
@@ -0,0 +1,388 @@
...
+        """Cancel the task."""
+        self._cancelled = True
+        if self._task is not None:
+            self._task.cancel()
+
+    async def wait(self) -> ResultType:
+        """Wait for the work to complete.
+
+        Raises:
+            WorkerFailed: If the Worker raised an exception
```
rodrigogiraoserrao, line 360:
```suggestion
            WorkerFailed: If the Worker raised an exception.
```

src/textual/worker.py
```diff
@@ -0,0 +1,388 @@
...
+                raise DeadlockError(
+                    "Can't call worker.wait from within the worker function!"
+                )
+        except LookupError:
+            # Not in a worker
+            pass
+
+        if self.state == WorkerState.PENDING:
+            raise WorkerError("Worker must be started before calling this method.")
+        if self._task is not None:
```
rodrigogiraoserrao, line 377: Shouldn't we also check that `self.state` is `RUNNING` or if I await a worker that has been completed, it won't run again?

tests/test_work_decorator.py
```diff
@@ -0,0 +1,33 @@
...
+            self.worker = self.foo()
+
+        def on_worker_state_changed(self, event: Worker.StateChanged) -> None:
+            states.append(event.state)
+
+    app = WorkApp()
+
+    async with app.run_test() as pilot:
+        await app.workers.wait_for_complete()
+        result = await app.worker.wait()
```
rodrigogiraoserrao, line 30: Is this how we extract the result out of a worker? I thought the `await` above meant that the worker _already_ finished?

src/textual/_work_decorator.py
willmcgugan: The original is more idiomatic, even if it is not the faster. In this instance, speed is not much of an issue,

src/textual/_work_decorator.py
willmcgugan: True, but not necessarily clearer.

src/textual/_worker_manager.py
willmcgugan: The original fixed a typing issue IIRC. the `getattr` could in theory return anything.

src/textual/_worker_manager.py
willmcgugan: There may be unrelated nodes that share the `group`.

src/textual/app.py
willmcgugan: The manager does contain the workers, so it is in a sense the workers being managed.

src/textual/worker.py
willmcgugan: I don't follow what you are suggesting. Can you elaborate?

tests/test_work_decorator.py
willmcgugan: It will have already finished, so this will give us the return value. We could also do `worker.result` IIRC>

src/textual/_worker_manager.py
rodrigogiraoserrao: Adding `or ""` won't fix typing because `getattr` can _still_ return anything in theory.
The type of `name` is `Any | str` with or without `or ""` at the end.

src/textual/_worker_manager.py
rodrigogiraoserrao: Exactly. What if I want to cancel all workers in a single group across multiple nodes?

willmcgugan: It doesn’t make sense to share groups. I can’t set a use case. Shared groups could also potentially introduce clashes if a third party library happens to have used the same name.

rodrigogiraoserrao: Understood.

src/textual/worker.py
rodrigogiraoserrao: A word was eaten out of my comment.
"Wouldn't it make more sense to _write_ that `work` is ..."
The current docstring lists the valid types for `work` but it doesn't say what that callable/coroutine/awaitable is used for.

rodrigogiraoserrao (approved):

Issue #2220, opened 2023-04-05
Margin and `fr` don't mix well


rodrigogiraoserrao:
**Edit**: original MRE was wrong.
[See the real issue below](https://github.com/Textualize/textual/issues/2220#issuecomment-1497849880)

willmcgugan:
This seems to work as I would expect. A blank line, followed by the lines with text.

What do you get?

rodrigogiraoserrao:
Don't mind me, I got the MRE mixed up.

Real issue: setting margins in widgets inside a container doesn't mix well with also using `fr` in those widgets.

The labels in the MRE below have 3 lines each. Each sets a margin to 1.
We set the outer container to `height: 8`.
Instead of splitting the 8 lines 4-4, the `fr` gives 4 lines to each label and then we need two _extra_ lines for the margins.

What we get:

![gh](https://user-images.githubusercontent.com/5621605/230155062-8eb46416-30f3-4cf6-af41-dd674dcb6049.png)


Expected output:

![gh](https://user-images.githubusercontent.com/5621605/230154950-5526a261-a623-452b-b674-88d119670b9e.png)


 - `margin.css`:

```sass
Container {
    height: 8;
    background: pink;
}

#top {
    background: red;
    margin-top: 1;
    height: 1fr;
}

#bot {
    background: blue;
    margin-bottom: 1;
    height: 1fr;
}
```

 - `margin.py`:

```py
from textual.app import App, ComposeResult
from textual.containers import Container
from textual.widgets import Label


class MyApp(App[None]):
    CSS_PATH = "margin.css"

    def compose(self) -> ComposeResult:
        with Container():
            yield Label("one\ntwo\nthree", id="top")
            yield Label("four\nfive\nsix", id="bot")
```

rodrigogiraoserrao:
I have had all sorts of weird interactions with `margin` in the past. May be related with the above.

willmcgugan:
Awesome, looks like my branch fixes that.

<img width="899" alt="Screenshot 2023-04-05 at 18 38 33" src="https://user-images.githubusercontent.com/554369/230160002-93692840-2acb-4454-8df3-e00cdb5b1da4.png">

PR #2221, opened 2023-04-05
Expanding fr

willmcgugan:
Expand "auto" container if any child has a relative  dimension (`fr` or `%`).

Fix margins not being respected when containers have auto dimensions.

Fixes https://github.com/Textualize/textual/issues/2220


src/textual/css/styles.py
```diff
@@ -384,6 +384,21 @@ def auto_dimensions(self) -> bool:
...
+        return width is not None and width.unit in (
+            Unit.FRACTION,
+            Unit.PERCENT,
+        )
+
+    @property
+    def is_relative_height(self) -> bool:
+        """Does the node have a relative width?"""
+        height = self.height
+        return height is not None and height.unit in (Unit.FRACTION, Unit.PERCENT)
```
rodrigogiraoserrao, line 400: Aren't `w`, `h`, `vh`, and `vw`, relative units too?

src/textual/layouts/horizontal.py
```diff
@@ -23,10 +23,23 @@ def arrange(
         x = max_height = Fraction(0)
         parent_size = parent.outer_size
 
+        child_styles = [child.styles for child in children]
+        box_margins = [styles.margin for styles in child_styles]
```
rodrigogiraoserrao, line 27: Is there any reason you didn't do
`box_margins = [child.styles.margin for child in children]`?

src/textual/layouts/horizontal.py
```diff
@@ -23,10 +23,23 @@ def arrange(
...
+                max(margin1.right, margin2.left)
+                for margin1, margin2 in zip(box_margins, box_margins[1:])
+            )
+            if len(box_margins) > 1:
+                total_margin += box_margins[0].left + box_margins[-1].right
+        else:
+            total_margin = 0
+
         box_models = resolve_box_models(
             [child.styles.width for child in children],
```
rodrigogiraoserrao, line 39: Did you mean to reuse `child_styles` here?

tests/snapshot_tests/snapshot_apps/auto_fr.py
```diff
@@ -0,0 +1,40 @@
...
+        margin: 1 2;    
+        background: $primary;
+    }
+    """
+
+    def compose(self) -> ComposeResult:
+        with Widget(id="container"):
+            yield Label("Hello one line", id="top")
+            yield Widget(id="child")
+            yield Label("Two\nLines with 2x2 margin", id="bottom")
```
rodrigogiraoserrao, line 35: May not be super relevant but the margin set in the CSS is actually 1x2!

rodrigogiraoserrao (commented):
Looking forward to using the new mechanism to handle fr units!

davep (approved):

src/textual/css/styles.py
willmcgugan: `w` and `h` become percent units.

`vh` and `vw` aren't relative to the container, unless the container happens to be screen. Maybe the name could be clarified?

tests/snapshot_tests/snapshot_apps/auto_fr.py
willmcgugan: We spotted

src/textual/layouts/horizontal.py
willmcgugan: Yes, an optimization I forgot to port from horizontal.

src/textual/css/styles.py
rodrigogiraoserrao: A clarification is always welcome, in my opinion.

PR #2234, opened 2023-04-06
optimize horizontal and vertical

willmcgugan:
Applying some micro-optimizations to the new code in horizontal and vertical

### DIFF
--- a/src/textual/_resolve.py
+++ b/src/textual/_resolve.py
@@ -118,14 +118,10 @@ def resolve_box_models(
     ]
 
     if dimension == "width":
-        total_remaining = sum(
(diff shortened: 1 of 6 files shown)


willmcgugan:
Haven't profile all of this. Just going on what @rodrigogiraoserrao tells me is faster...

willmcgugan:
#executivedecision

rodrigogiraoserrao:
😬 the list comp is better than the generator specifically inside `join`, and I even linked specifically [this reference article](https://mathspp.com/blog/til/two-pass-join) that only talks about `join`.
I don't know if that is the case with `sum`!

PR #2239, opened 2023-04-07
docs update

willmcgugan:
Some customization of the API docs templates, which I always found difficult to visually parse.

It's live now. https://textual.textualize.io/api/app/


willmcgugan:
#executivedecision (its only docs)

PR #2240, opened 2023-04-07
api docs

willmcgugan:
Refactor of docs.

- Additional template and CSS tweaks
- Moved Widget API docs from the API reference, in to the widget pages. This keeps widget related things in one place, and also reduces noise in the API reference.
- Messages section in widgets changed to a simple list.
- Added "Returns:" section to properties. Without it, you don't get links to the return type. Suggest we do this for anything other than builtin types.
- Bunch of updated docstrings.
- Added docstring to modules. 
- Notes are now open by default (totally baffling why this isn't the case).
- Remove "Defaults to..." text


willmcgugan:
#executivedecision

PR #2243, opened 2023-04-08
Use string representation of the item to order

msempere:
**Please review the following checklist.** [PENDING]

- [ ] Updated documentation

Following PR modifies `sort` method as part  of `DataTable` so it uses the string representation of the item when ordering.
I have a project that uses a `DataTable` and has `rich.Text` in its cells, to make use of different styles depending of the cell content, but I cannot order the table because `rich.Text` entities cannot be compared, raising a `TypeError: '<' not supported between instances of 'Text' and 'Text'`.
With this patch, `rich.Text` entities use  their plain content, the string value, so they can be compared and ordered via `DataTable.sort`.

### DIFF
--- a/src/textual/widgets/_data_table.py
+++ b/src/textual/widgets/_data_table.py
@@ -1900,7 +1900,7 @@ class DataTable(ScrollView, Generic[CellType], can_focus=True):
             row: tuple[RowKey, dict[ColumnKey | str, CellType]]
         ) -> Any:
             _, row_data = row
-            result = itemgetter(*columns)(row_data)
+            result = str(itemgetter(*columns)(row_data))
             return result
 
         ordered_rows = sorted(


rodrigogiraoserrao:
This can't be accepted as-is because it will break sorting columns with numbers, e.g., `"20" < "100"` returns `False`.

Maybe we could add a `key` parameter to the `sort` methods, akin to how `sorted(..., key=...)` works.

If you are interested in moving this forward, I suggest you open an issue with a feature request where the feature can be discussed before it is implemented.

msempere:
> This can't be accepted as-is because it will break sorting columns with numbers, e.g., `"20" < "100"` returns `False`.
> 
> Maybe we could add a `key` parameter to the `sort` methods, akin to how `sorted(..., key=...)` works.
> 
> If you are interested in moving this forward, I suggest you open an issue with a feature request where the feature can be discussed before it is implemented.

Thanks @rodrigogiraoserrao for your comment. 
My intention was initially to trigger the discussion around that issue, so I guess it'd be a better place to start with an issue instead of PR.

PR #2249, opened 2023-04-10
fix arrangement cache

willmcgugan:
Fix an issue with arrangement cache not being cleared.

Bump to 0.19.1


willmcgugan:
#executivedecision

@darrenburns This fixes the issue with the DT not refreshing.

PR #2252, opened 2023-04-11
Simplify driver

willmcgugan:
A refactor of driver for the benefit of textual-serve.

### DIFF
--- a/src/textual/driver.py
+++ b/src/textual/driver.py
@@ -18,7 +18,6 @@ class Driver(ABC):
         self,
         app: App,
         *,
-        debug: bool = False,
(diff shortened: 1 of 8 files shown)


src/textual/driver.py
```diff
@@ -117,6 +114,9 @@ def write(self, data: str) -> None:
             data: Raw data.
         """
 
+    def flush(self) -> None:
```
darrenburns, line 117: Missing decorator?

willmcgugan: No, this one doesn't need to be abstract. Default is to to nothing...

PR #2253, opened 2023-04-11
Datatable remove row

darrenburns:

- [x] Updated documentation


src/textual/widgets/_data_table.py
```diff
@@ -1322,6 +1324,40 @@ def add_rows(self, rows: Iterable[Iterable[CellType]]) -> list[RowKey]:
...
+        Raises:
+            RowDoesNotExist: If the row key does not exist.
+        """
+        if row_key not in self._row_locations:
+            raise RowDoesNotExist(f"Row key {row_key!r} is not valid.")
+
+        self._require_update_dimensions = True
+
+        index_to_delete = self._row_locations.get(row_key)
+        new_row_locations = TwoWayDict({})
```
willmcgugan, line 1342: Could `TwoWayDict` grow a `clear`?

src/textual/widgets/_data_table.py
```diff
@@ -1322,6 +1324,40 @@ def add_rows(self, rows: Iterable[Iterable[CellType]]) -> list[RowKey]:
...
+        Args:
+            row_key: The key identifying the row to remove.
+
+        Raises:
+            RowDoesNotExist: If the row key does not exist.
+        """
+        if row_key not in self._row_locations:
+            raise RowDoesNotExist(f"Row key {row_key!r} is not valid.")
+
+        self._require_update_dimensions = True
```
willmcgugan, line 1339: Will you not need `check_idle` after this? There is no guarantee an `on_idle` will follow, if this method is called in the context of another widget.

src/textual/widgets/_data_table.py
```diff
@@ -1322,6 +1324,40 @@ def add_rows(self, rows: Iterable[Iterable[CellType]]) -> list[RowKey]:
...
+            row_index = self._row_locations.get(row_location_key)
+            if row_index > index_to_delete:
+                new_row_locations[row_location_key] = row_index - 1
+            elif row_index < index_to_delete:
+                new_row_locations[row_location_key] = row_index
+
+        self._row_locations = new_row_locations
+
+        del self.rows[row_key]
+        del self._data[row_key]
```
rodrigogiraoserrao, line 1353: I don't get this code.
Wouldn't it be simpler to do something like

```py
del self.rows[row_key]
del self._data[row_key]

row_locations = self._row_locations
for row_key, index in row_locations.items():
    row_locations[row_key] -= index > index_to_delete
```

If subtracting a Boolean is not something we'd like to do,

```py
del self.rows[row_key]
del self._data[row_key]

row_locations = self._row_locations
for row_key, index in row_locations.items():
    row_locations[row_key] -= 1 if index > index_to_delete else 0
```

There's no point in building a new dictionary from scratch.

willmcgugan (approved):

PR #2255, opened 2023-04-11
Add scroll_to_center method.

### LINKED ISSUE #2031: Implement scroll_to_center method
We have methods to scroll widgets in to view, but these tend to do as little scrolling as possible.

We may have a need to scroll something to the center. This would benefit the Tabs control.

rodrigogiraoserrao:
This will close #2031.

### DIFF
--- a/src/textual/widget.py
+++ b/src/textual/widget.py
@@ -2464,15 +2464,11 @@ class Widget(DOMNode):
             # We add that to `container_virtual_region.y` to find the total vertical
             # offset of the central point with respect to the container of `container`.
             # A similar calculation is made for the horizontal update.
-            central_point = Offset(
-                container_virtual_region.x
-                + central_point.x
-                - container.scroll_offset.x
-                - scroll.x,
-                container_virtual_region.y
-                + central_point.y
-                - container.scroll_offset.y
-                - scroll.y,
+            central_point = (
+                container_virtual_region.offset
+                + central_point
+                - container.scroll_offset
+                - scroll
             )
             widget = container
             container = widget.parent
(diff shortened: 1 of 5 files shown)


src/textual/widget.py
```diff
@@ -2408,6 +2408,104 @@ def scroll_visible(
...
+            # We scroll `widget` within `container` with the central point written in
+            # the frame of reference of `container`. However, we need to update it so
+            # that we are ready to scroll `container` within _its_ container.
+            # To do this, notice that
+            # (central_point.y - container.scroll_offset.y - scroll.y) is the number
+            # of rows of `widget` that are visible within `container`.
+            # We add that to `container_virtual_region.y` to find the total vertical
+            # offset of the central point with respect to the container of `container`.
+            # A similar calculation is made for the horizontal update.
+            central_point = Offset(
```
willmcgugan, line 2467: Could this not be expressed using offset operators?

```python
central_point = (
    container_virtual_region.offset
    + central_point
    - container.scroll_offset
    - scroll
 )
```

tests/snapshot_tests/test_snapshots.py
```diff
@@ -406,3 +406,13 @@ def test_fr_margins(snap_compare):
 def test_scroll_visible(snap_compare):
     # https://github.com/Textualize/textual/issues/2181
     assert snap_compare(SNAPSHOT_APPS_DIR / "scroll_visible.py", press=["t"])
+
+
+def test_scroll_to_center(snap_compare):
```
willmcgugan, line 411: If this test is expected to fail, can you mark is as an `xfail` ?

rodrigogiraoserrao: This test is not expected to fail _now_. It will fail eventually, when the bug is fixed.

PR #2258, opened 2023-04-11
Add --port option to textual console.

### LINKED ISSUE #2014: Allow command-line option to change development port
I couldn't use `textual run --dev` because the server I was on had AV software listening on port 8081.

Luckily a quick `git grep` was able to fix it, but it would be nice to have that as a command-line option.

rodrigogiraoserrao:

- [x] Updated documentation

This will close #2014

### DIFF
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -95,7 +95,7 @@ from .screen import Screen
 from .widget import AwaitMount, Widget
 
 if TYPE_CHECKING:
-    from typing_extensions import Coroutine, Final, TypeAlias
+    from typing_extensions import Coroutine, TypeAlias
 
     from .devtools.client import DevtoolsClient
     from .pilot import Pilot
@@ -162,6 +162,15 @@ class CssPathError(Exception):
 ReturnType = TypeVar("ReturnType")
 
 
+CSSPathType = Union[
+    str,
+    PurePath,
+    List[Union[str, PurePath]],
+]
+
+CallThreadReturnType = TypeVar("CallThreadReturnType")
+
+
 class _NullFile:
     """A file-like where writes go nowhere."""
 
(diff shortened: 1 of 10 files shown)


src/textual/app.py
```diff
@@ -58,6 +58,8 @@
 from rich.segment import Segment, Segments
 from rich.traceback import Traceback
 
+from textual.devtools.client import get_port_for_devtools
```
willmcgugan, line 61: Can you use relative imports please.

src/textual/cli/cli.py
```diff
@@ -21,14 +23,27 @@ def run():
 
 
 @run.command(help="Run the Textual Devtools console.")
+@click.option(
+    "--port",
+    "port",
+    type=int,
+    default=None,
```
willmcgugan, line 30: Can the default not be an integer? So it shows in the help.

src/textual/devtools/client.py
```diff
@@ -17,11 +18,20 @@
...
 from .._log import LogGroup, LogVerbosity
 
-DEVTOOLS_PORT = 8081
+DEVTOOLS_PORT_ENVIRON_VARIABLE = "TEXTUAL_CONSOLE_PORT"
+DEFAULT_DEVTOOLS_PORT = 8081
 WEBSOCKET_CONNECT_TIMEOUT = 3
 LOG_QUEUE_MAXSIZE = 512
 
 
+def get_port_for_devtools() -> int:
```
willmcgugan, line 27: Can you move this to `constants.py` where we read environment variables.

src/textual/cli/cli.py
rodrigogiraoserrao: I don't understand. The default is _not_ an integer (it is `None`).
If I change the default to an integer, it still doesn't show when I run `textual console --help`.

willmcgugan: I thought it would show the default. That's disappointing. In that case, would you mind updating the help text with the default?

rodrigogiraoserrao: 👍

src/textual/constants.py
```diff
@@ -25,10 +28,18 @@ def get_environ_bool(name: str) -> bool:
     Returns:
         `True` if the env var is "1", otherwise `False`.
     """
-    has_environ = os.environ.get(name) == "1"
+    has_environ = get_environ(name) == "1"
     return has_environ
 
 
+def get_port_for_devtools() -> int:
```
willmcgugan, line 35: This doesn't need to be a callable, as it is never going to change.

Can you create a constant, in the same way as the other symbols in this module.

willmcgugan (changes requested):

src/textual/constants.py
rodrigogiraoserrao: I thought a callable made sense because we had to check if we'd use the default, or not.
Let me try again.

src/textual/constants.py
```diff
@@ -37,3 +37,12 @@ def get_environ_bool(name: str) -> bool:
 
 LOG_FILE: Final[str | None] = get_environ("TEXTUAL_LOG", None)
 """A last resort log file that appends all logs, when devtools isn't working."""
+
+
+DEVTOOLS_PORT_ENVIRON_VARIABLE = "TEXTUAL_CONSOLE_PORT"
+DEFAULT_DEVTOOLS_PORT = 8081
+try:
```
willmcgugan, line 44: Let's factor this out in to a function. Something like `get_environment_int(env_var, default)`

willmcgugan (changes requested):

src/textual/constants.py
rodrigogiraoserrao: I pondered whether the `default` could be set to `None` (to return `None`) if the env var wasn't set but then decided against it because it would take an `overload` to ensure consistent typing and that seemed like overkill for this situation.

willmcgugan (approved):

src/textual/constants.py
willmcgugan: I agree. These constants should have simple typing.

PR #2266, opened 2023-04-12
Add docstring to action_check_bindings

darrenburns:
There was only one action without a docstring.

Also, proposing we remove `action_bang` as it's unused.

### DIFF
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -2448,7 +2448,7 @@ class App(Generic[ReturnType], DOMNode):
         self._unregister(root)
 
     async def action_check_bindings(self, key: str) -> None:
-        """Handle a key press using the bindings system."""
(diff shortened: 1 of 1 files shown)


willmcgugan:
Feel free to remove action_bang

src/textual/app.py
```diff
@@ -2448,16 +2448,14 @@ async def _prune_node(self, root: Widget) -> None:
         self._unregister(root)
 
     async def action_check_bindings(self, key: str) -> None:
+        """Handle a key press using the bindings system."""
```
willmcgugan, line 2451: Could you look at the other actions in app. I've linked `actions` to the docs. I think we should follow this convention.

willmcgugan (approved):

PR #2274, opened 2023-04-12
Fix issue with tabbed content underline

### LINKED ISSUE #2229: It's possible to somehow break the working of `TabbedContent`
This one isn't exactly easy to explain, but hopefully easy to see and recreate with the following code (which is a distilled version of what I'm doing in a bigger application, where I found it).

Worth noting: the motivation here is that (in the application I'm working on) the `TabbedContent` is acting as a sidebar, where each pane has content that can be focused, and I want folk to be people to switch tabs without needing to navigate to the tab bar and back into the pane again. As such there's some bindings in place that call into the `Tabs` and uses their prev/next tab actions.

```python
from textual.widgets    import Header, Footer, TabbedContent, TabPane, Tabs, DirectoryTree

class SelfFocusPane(TabPane):

    DEFAULT_CSS = """
    SelfFocusPane {
        height: 100% !important;
    }
    DirectoryTree {
        width: 100%;
        height: 100% !important;
    }
    """

        """Compose the child widgets."""
        yield DirectoryTree(".")

    def on_show( self ) -> None:
        self.query_one( DirectoryTree ).focus()

class TabbedContentIssueApp( App[ None ] ):

    CSS = """
    Screen {
        align: center middle;
    }

    Screen > Vertical {
        width: 42;
    }

    TabbedContent {
        border: round red;
        max-width: 40;
        height: 100%;
    }

    ContentSwitcher {
        height: 1fr !important;
    }
    """

    BINDINGS = [
        Binding( "shift+left", "previous", "Previous" ),
        Binding( "shift+right", "next", "Next" ),
    ]

        yield Header()
        with Vertical():
            with TabbedContent():
                for n in range( 6 ):
                    yield SelfFocusPane( f"Tab {n}")
        yield Footer()

    def on_mount(self) -> None:
        self.query_one(Tabs).focus()

    def action_previous(self) -> None:
        self.query_one(Tabs).action_previous_tab()

    def action_next(self) -> None:
        self.query_one(Tabs).action_next_tab()

if __name__ == "__main__":
    TabbedContentIssueApp().run()
```

In experimenting, it looks like the `SelfFocusPane.on_show` setting focus to the child of the pane is key here; remove that and I can't recreate the issue.

rodrigogiraoserrao:
I did some investigation around #2229 and I figured out the issue is related to the fact that it takes a lot of time to swap in and out the contents of the tabbed content.
Commenting out this line makes the issue go away:

https://github.com/Textualize/textual/blob/c249548c43f6831db76d59aba5cc10c600335afa/src/textual/widgets/_tabbed_content.py#L188

That's called when the message `Tabs.TabActivated` is handled in `TabbedContent`, and that message is coming from `Tabs.watch_active`.

A simple fix would be to throttle `Tabs.watch_active` so that it doesn't get called _too_ often.
That is what this PR does, and merging it will close #2229.

Another possible solution, that is more involved, is to add a task-based mechanism around swapping the visible sub-widget of the `ContentSwitcher` in `ContentSwitch.watch_current`, such that when we schedule a new switch, old ones get thrown away.

### DIFF
--- a/src/textual/widgets/_tabs.py
+++ b/src/textual/widgets/_tabs.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from time import monotonic
 from typing import TYPE_CHECKING, ClassVar
 
 import rich.repr
@@ -20,6 +21,10 @@ from ..widgets import Static
 
 if TYPE_CHECKING:
     from typing_extensions import Self
+
+
+THROTTLE_DELTA = 0.005
+"""Don't call `watch_active` more than once every THROTTLE_DELTA seconds."""
 
 
 class Underline(Widget):
@@ -174,6 +179,9 @@ class Tabs(Widget, can_focus=True):
     | left | Move to the previous tab. |
     | right | Move to the next tab. |
     """
+
+    _last_update: float = -1
+    """Time of last update to throttle updates."""
 
     class TabActivated(Message):
         """Sent when a new tab is activated."""
@@ -399,7 +407,16 @@ class Tabs(Widget, can_focus=True):
                 yield Underline()
 
     def watch_active(self, previously_active: str, active: str) -> None:
-        """Handle a change to the active tab."""
+        """Handle a change to the active tab.
+
+        This function is throttled to only run once every THROTTLE_DELTA seconds.
+        """
+
+        now = monotonic()
+        if now < self._last_update + THROTTLE_DELTA:
+            return
+        self._last_update = now
+
         if active:
             active_tab = self.query_one(f"#tabs-list > #{active}", Tab)
             self.query("#tabs-list > Tab.-active").remove_class("-active")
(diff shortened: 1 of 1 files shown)


src/textual/widgets/_tabs.py
```diff
@@ -504,5 +521,6 @@ def _move_tab(self, direction: int) -> None:
             return
         tab_count = len(tabs)
         new_tab_index = (tabs.index(active_tab) + direction) % tab_count
+        print(tabs[new_tab_index].id or "")
```
willmcgugan, line 524: Debug print?

willmcgugan (changes requested):
This feels like it might be masking the error, rather than fixing it. I don't think you guarantee that time mechanism won't fail on a slower machine or with faster key repeat.

Have you established why this occurs?

rodrigogiraoserrao:
> This feels like it might be masking the error, rather than fixing it. [...]

Sadly, I share this sentiment.

> Have you established why this occurs?

Not really... I will dig some more.

rodrigogiraoserrao:
Closing this because the actual fix (coming soon) is unrelated to this throttling attempt.

PR #2275, opened 2023-04-12
fix(app): explicitly check if action_target is none

### LINKED ISSUE #2265: Bindings are not fired on an empty `ListView`
Stemming from #2262 asked by @TomJGooding:

Given the following code:

```python
from textual.widgets    import Header, Footer, ListView, Input

class MyListView(ListView):

    BINDINGS = [
        Binding( "i", "input", "Focus Input" ),
    ]

    def action_input(self) -> None:
        self.app.bell()
        self.app.query_one( Input ).focus()

class LVFocusApp( App[ None ] ):

    CSS = """
    MyListView {
        height: 1fr;
        border: solid blue;
    }

    MyListView:focus {
        border: double green;
    }
    """

        yield Header()
        with Vertical():
            yield Input()
            yield MyListView()
        yield Footer()

if __name__ == "__main__":
    LVFocusApp().run()
```

When the `ListView` is focused pressing the <kbd>i</kbd> key doesn't fire the action on `MyListView`. On the other hand, if one or more items are added to the `ListView`, the action *does* fire.

Another way to make the action fire is to add a `__len__` to `MyListView` that always returns a value that's non-zero. That is, a non-zero return value on `MyListView.__len__`, even when there are no list items, results in the action being fired when the bound key is pressed.

This feels very similar to something that @darrenburns and I found in Textual some time back (late last year I think), where there was some code that made a decision akin to:

```python
if self.screen.focused:
   ...
```

when it should really have been (and was fixed by):

```python
if self.screen.focused is not None:
   ...
```

I suspect something similar here when it comes to handling keys, events or bindings (I've not dived in to really look yet).

I sense this raises two related meta-issues:

1. We should probably do an audit of the code and ensure that every location that tests if a widget reference is not `None` actually does test that it is not `None`, as opposed to test if it's falsey.
2. We should probably have a careful thing about using various dunder methods on widgets in general. Is this something we should be doing; is it perhaps something we should discourage and also discourage developers from doing, due to the chance that it may have unintended consequences in Textual.

TomJGooding:
Fixes #2265 by explicitly checking if `action_target` is None.

Thanks again to @davep for his help with this issue.

### DIFF
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -2149,7 +2149,9 @@ class App(Generic[ReturnType], DOMNode):
             action_target = getattr(self, destination)
             implicit_destination = True
         else:
-            action_target = default_namespace or self
+            action_target = default_namespace
+            if action_target is None:
+                action_target = self
             action_name = target
 
         handled = await self._dispatch_action(action_target, action_name, params)


rodrigogiraoserrao (commented):
Looks good to me!
Can I trouble you for a regression test?

willmcgugan:
Thanks. We've been tripped up by this before! Regression test would be appreciated.

willmcgugan:
Looks like Darren got there first! https://github.com/Textualize/textual/pull/2281

darrenburns:
Sorry @TomJGooding, I went ahead and implemented this/merged it without realising this PR was here!

TomJGooding:
No problem @darrenburns! I wouldn't be too confident tbh about adding a regresssion test anyway (and I see your solution is a couple lines shorter!)

PR #2283, opened 2023-04-13
Fix retrieval mechanism for devtools port.

rodrigogiraoserrao:
Defined as a static constant, the value of the devtools port is assigned before getting a chance to modify the relevant environment variable.
Make it a function, and everything works just fine.

Related PRs: https://github.com/Textualize/textual/pull/2258

### DIFF
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -249,6 +249,8 @@ class App(Generic[ReturnType], DOMNode):
         ```
     """
 
+    _devtools: DevtoolsClient | None = None
+    """The devtools client associated with this app, if any."""
     _devtools_port: int = constants.DEFAULT_DEVTOOLS_PORT
     """Port to use if/when connecting to the devtools server.
 
@@ -371,16 +373,6 @@ class App(Generic[ReturnType], DOMNode):
         self._compose_stacks: list[list[Widget]] = []
         self._composed: list[list[Widget]] = []
 
-        self.devtools: DevtoolsClient | None = None
-        if "devtools" in self.features:
-            try:
-                from .devtools.client import DevtoolsClient
-            except ImportError:
-                # Dev dependencies not installed
-                pass
-            else:
-                self.devtools = DevtoolsClient(port=self._devtools_port)
-
         self._loop: asyncio.AbstractEventLoop | None = None
         self._return_value: ReturnType | None = None
         self._exit = False
@@ -396,6 +388,19 @@ class App(Generic[ReturnType], DOMNode):
         self._batch_count = 0
         self.set_class(self.dark, "-dark-mode")
         self.set_class(not self.dark, "-light-mode")
+
+    @property
+    def devtools(self) -> DevtoolsClient | None:
+        """The devtools associated with this app."""
+        if self._devtools is None and "devtools" in self.features:
+            try:
+                from .devtools.client import DevtoolsClient
+            except ImportError:
+                # Dev dependencies not installed
+                pass
+            else:
+                self._devtools = DevtoolsClient(port=self._devtools_port)
+        return self._devtools
 
     @property
     def workers(self) -> WorkerManager:
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -29,7 +29,6 @@ from contextlib import (
 from datetime import datetime
 from functools import partial
 from pathlib import Path, PurePath
-from queue import Queue
 from time import perf_counter
 from typing import (
     TYPE_CHECKING,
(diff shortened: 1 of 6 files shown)


rodrigogiraoserrao:
@willmcgugan I ended up having to make it a function again but I decided to keep `get_environ_int` because it looks useful.
Let me know what you think.

willmcgugan:
I'm struggling to understand this one. I don't think we should be modifying environment variables. They should really be read-only for the most part. Can we not get the port from the environment, then pass it to `_run_devtools` ?

rodrigogiraoserrao:
The port to be used is configurable via the `--port` option in both commands, right?
We only write to the environment variable at that time.

Later, we just read from it.

willmcgugan:
But I don't think we should be writing to env vars at all. Writing to `os.environ` is much like using globals, and bad for the same reasons.

rodrigogiraoserrao:
@willmcgugan I know we chatted about passing the value for the port down the call stack.
However, to change the port that the app uses in the `DevtoolsClient`, I decided to use a class variable for the reason that I outline below.

I implemented the change where the port was passed down the call stack and it wasn't very deep.
What I needed to do was to make sure the port got to the import app mechanism and then I passed it to the app as a keyword argument.

However, by putting that keyword argument in the import mechanism, that means that all apps that run with `textual run` will receive such a keyword argument (even if the value is the default one).
What does that mean?
It means that if you implement an app and you need to override `__init__` to do some initialisation, you _will_ have to add the keyword argument `devtools_port` and be sure to remember to pass it to `super().__init__()`.
I thought that didn't look very friendly and a bit unexpected, so I decided against that.

E.g., the app in `tests/snapshot_tests/snapshot_apps/label_widths.py` wouldn't run if you used the command `textual run label_widths.py` because its `__init__` doesn't expect the keyword argument for the port:

https://github.com/Textualize/textual/blob/898de5a69c03a6323b872d8775fa673d8d7d6cdf/tests/snapshot_tests/snapshot_apps/label_widths.py#L26-L32

willmcgugan:
Just been looking at this, and I see the problem now. Just as you described, we don't necessarily have much control over the app instantiation when we import the app.

I think the flaw is that we're doing too much in the constructor, which violates the "keep constructors simple" guideline. We construct the features set there, which is probably the reason I did the nasty hack for setting env variables.

The solution is probably to make certain objects lazy in the constructor. So `devtools` should be a property that instantiates the devtools object on demand. We then add properties or a methods to update devtools port, features etc, which we would do prior to running the app. What do you think?

There's obviously a little more to this than the original issue you were working on. You could file an new issue, or just go ahead and make these changes if you're confident enough in the solution. Up to you!

willmcgugan:
Actually, an even better solution may be to not load the app in to the same process, but run it is a subprocess. That way we could run it with a fresh environment. Hmm.

rodrigogiraoserrao:
> There's obviously a little more to this than the original issue you were working on. You could file an new issue, or just go ahead and make these changes if you're confident enough in the solution. Up to you!

I'm happy to file a new issue to keep track of things and assign it to myself.
I expect to get to this on Monday.

rodrigogiraoserrao:
I'll also gladly take #2285 down with me as I do so.
Feels like all of these changes could be done at the same time.

rodrigogiraoserrao:
I've reworked this PR to take into account #2285 and #2288.

What I _haven't_ done is address [this comment](https://github.com/Textualize/textual/pull/2283#issuecomment-1508257247), which I didn't understand:

 > Actually, an even better solution may be to not load the app in to the same process, but run it is a subprocess. That way we could run it with a fresh environment. Hmm.

src/textual/app.py
```diff
@@ -248,17 +247,37 @@ class App(Generic[ReturnType], DOMNode):
         ```python
         self.app.dark = not self.app.dark  # Toggle dark mode
         ```
+    """
+
+    _devtools_port: int = constants.DEFAULT_DEVTOOLS_PORT
```
willmcgugan, line 252: I don't think any of these need to be classvars.

Any reason why they shouldn't be instance variables?

willmcgugan (commented):
I'm going to park this, because I think we need to revise how we pass such variables in to apps.

I think we need to exclusively go via environment variables rather than importing and monkey patching. 

Thanks for doing the groundwork.

src/textual/app.py
rodrigogiraoserrao: Not really, no.

rodrigogiraoserrao:
> [...] I think we need to exclusively go via environment variables rather than importing and monkey patching. [...]

Ok. Please ping me when you attempt this again because I'm having a hard time seeing how we're going to have a CLI that has options that the app can then use, which the app reads from environment variables, while also not _writing_ into those env variables!
My brain is all twisted 🤪

rodrigogiraoserrao:
Superseded by #2338

Issue #2263, opened 2023-04-12
Update the FAQ entry about centring widgets


davep:
Remove the custom container and use the newer standard ones.

PR #2286, opened 2023-04-14
Update the centring FAQ to reference `Center`

davep:
See #2263.

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -105,22 +105,14 @@ If you want them more like this:
 +---------------+
 ```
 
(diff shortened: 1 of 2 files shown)


rodrigogiraoserrao (commented):
`Center` only centres one axis. Is this what we want?

rodrigogiraoserrao (approved):

rodrigogiraoserrao:
> That's the FAQ, this just removes the example and uses the recently-added container in the answer.

Right. I was just confused because the CSS deleted contained `align: center middle`.

PR #2304, opened 2023-04-17
Updating styles on demand instead of on_idle

darrenburns:
The `update_styles` call was being done in `on_idle`.

This meant that when you hover over a widget, the `:hover` pseudoclass was not immediately applied.

As a result, there was a short window where calls to `get_component_styles` would return incorrect values. 

We close this window by ensuring `update_styles` is called immediately rather than `on_idle`.

- [ ] Updated documentation


rodrigogiraoserrao (approved):
Looks good to me. 🚀

PR #2305, opened 2023-04-17
Prevent reactive-watcher loop in Tabs / TabbedContent.

### LINKED ISSUE #2229: It's possible to somehow break the working of `TabbedContent`

rodrigogiraoserrao:
This will close #2229.
See [this comment](https://github.com/Textualize/textual/issues/2229#issuecomment-1511636895) wherein for an explanation of the issue.

We use a descriptor to proxy the attribute `active` from `Tabs` to `TabbedContent` without having to keep it in sync in two different places.

### DIFF
--- a/src/textual/widgets/_tabbed_content.py
+++ b/src/textual/widgets/_tabbed_content.py
@@ -120,6 +120,8 @@ class TabbedContent(Widget):
 
     @active.setter
     def active(self, active: str) -> None:
+        if not active:
+            raise ValueError("'active' tab must not be empty string.")
         self.get_child_by_type(Tabs).active = active
 
     def compose(self) -> ComposeResult:
--- a/src/textual/widgets/_tabbed_content.py
+++ b/src/textual/widgets/_tabbed_content.py
@@ -71,6 +71,27 @@ class TabPane(Widget):
         )
 
 
+class _ActiveTabProxyDescriptor:
+    """Descriptor to proxy the attribute `active` from `Tabs` to the `TabbedContent`.
+
+    For the rationale that prompted the implementation of this descriptor, refer to
+    https://github.com/Textualize/textual/issues/2229 and, in particular, this comment:
+    https://github.com/Textualize/textual/issues/2229#issuecomment-1511636895.
+    """
+
+    def __get__(
+        self, obj: TabbedContent, objtype: type[TabbedContent] | None = None
+    ) -> str:
+        """Get the active tab by proxy."""
+        return obj.get_child_by_type(Tabs).active
+
+    def __set__(self, obj: TabbedContent, active: str) -> None:
+        """Set the active tab by proxy."""
+        if not active:
+            raise ValueError("'active' tab must not be empty string.")
+        obj.get_child_by_type(Tabs).active = active
+
+
 class TabbedContent(Widget):
     """A container with associated tabs to toggle content visibility."""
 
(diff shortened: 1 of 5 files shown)


src/textual/widgets/_tabbed_content.py
```diff
@@ -116,21 +113,16 @@ def __init__(self, *titles: TextType, initial: str = "") -> None:
...
+        """The ID of the active tab, or empty string if none are active."""
+        return self.get_child_by_type(Tabs).active
 
-        Raises:
-            ValueError: If the active attribute is set to empty string.
-        """
+    @active.setter
+    def active(self, active: str) -> None:
         if not active:
             raise ValueError("'active' tab must not be empty string.")
```
rodrigogiraoserrao, line 124: Originally I just let the setter set the attribute `active` on the child `Tabs` but the tests failed because they expected a `ValueError` when setting the tab to an empty string, so I added it back in 6eb141c.

However, to me it doesn't make much sense that I can't set `TabbedContent.active = ""` but _**I can**_ set `Tabs.active = ""`.
Assuming you agree, should I change the validation in `Tabs` so that it checks for an empty active tab ID at that point?

willmcgugan (approved):

src/textual/widgets/_tabbed_content.py
rodrigogiraoserrao: @willmcgugan pinging you because you already merged but this was left unresolved.

PR #2318, opened 2023-04-18
Fix frame rate limiter

willmcgugan:
Fixes the frame rate limiter, which wasn't working correctly.

The default is to limit the frame rate to 60fps. There doesn't seem any point in increasing beyond that given we only have the granularity of cells to work with.


src/textual/_compositor.py
```diff
@@ -245,19 +245,15 @@ def render_segments(self, console: Console) -> str:
...
 
-                if x < x1:
-                    strip = strip.crop(x1 - x, x2 - x)
-                    append(move_to(x1, y).segment.text)
-                else:
-                    strip = strip.crop(0, x2 - x)
-                    append(move_to(x, y).segment.text)
-
+                strip = strip.crop(0, min(end, x2) - x)
+                append(move_to(x, y).segment.text)
```
rodrigogiraoserrao, line 249: Without any more context, the two lines are not equivalent to the old ones.
These two seem to be more aligned with what was there before:

```suggestion
                strip = strip.crop(max(x1 - x, 0), min(end, x2) - x)
                append(move_to(max(x, x1), y).segment.text)
```

So, the further differences come from extra knowledge/context?

rodrigogiraoserrao (approved):

src/textual/_compositor.py
willmcgugan: No, they're not equivalent. But now the `crop` method handles the case of the start crop being negative correctly.

Issue #2267, opened 2023-04-12
We need a way to report results from modal Screens


willmcgugan:
We need a way for modal screens to report their result.

Up for discussion, but I think a callback makes sense here:

```python
async def on_result(screen:MyModal) -> None:
     ...

self.push_screen(MyModal(), on_result)
```

The screen would have some method to report its result, like `def dismiss(result)`, which pops the screen and calls the callback.

willmcgugan:
@davep Feel free to comment

PR #2321, opened 2023-04-18
Add support for "returning" a result value from screens.

davep:
This is still in the "let's see if we like this approach" stage, although I think @willmcgugan and myself are liking this approach. Testing and review of these changes is highly encouraged. Beyond docstrings there are no documentation changes at the moment. This will need a slight reworking of the `Screen` guide so it's better that we're happy with the approach and the code before we write this up fully.

Please review without worrying about documentation for now.

This PR seeks to satisfy #2267 and does take the suggested callback approach. The alternative was to go with a "use messages" story that, while working well, places more work on the developer using Textual and potentially results in more boilerplate.

Changes of note here are:

- Much like with `App`, a `Screen` can now be typed. The type signifies the type of the result that it will "return".
- `App.push_screen` has acquired a `callback` parameter. If provided the callback function should be a function that takes a single argument, that is the type of the type given to the screen being pushed.
- `Screen` has acquired a new method called `dismiss`. This is a thin wrapper around `App.pop_screen`. Called with no parameter it performs the same as if `App.pop_screen` were called. If called with a parameter the callback provided with `App.push_screen` is called and then the screen is popped.
- `Screen` has also acquired an `action_dismiss` method to allow for easy dismissal of a screen from within an action.

Other things to keep in mind:

- A callback can optionally be awaitable.
- Because any given instance of a screen can end up on the screen stack multiple times, the result callback hook is done as a stack.

As an illustration of how this might be used, here's a small program that uses a `ModalScreen` to ask the user a question and then act on their answer:

```python
from textual.widgets import Header, Footer, OptionList, Button, Label


class AyeNaw(ModalScreen[bool]):
    DEFAULT_CSS = """
    AyeNaw {
        align: center middle;
    }

    AyeNaw > Vertical {
        background: $secondary;
        width: auto;
        height: auto;
        border: thick $primary;
        padding: 2 4;
    }

    AyeNaw > Vertical > * {
        width: auto;
        height: auto;
    }

    AyeNaw > Vertical > Label {
        padding-bottom: 2;
    }

    AyeNaw > Vertical > Horizontal {
        align: right middle;
    }

    AyeNaw Button {
        margin-left: 2;
    }
    """

        with Vertical():
            yield Label("Well, what do you think?")
            with Horizontal():
                yield Button("Aye!", id="aye")
                yield Button("Naw!", id="naw")

    def on_button_pressed(self, event: Button.Pressed) -> None:
        self.dismiss(result=event.button.id == "aye")


class ScreenResultExampleApp(App[None]):
    BINDINGS = [
        ("question_mark", "ask", "Ask a question"),
    ]

        yield Header()
        yield OptionList()
        yield Footer()

    def accept_answer(self, answer: bool) -> None:
            f"Computer said {'[green]Aye' if answer else '[red]Naw'}[/]"
        )

    def action_ask(self) -> None:
        self.push_screen(AyeNaw(), callback=self.accept_answer)


if __name__ == "__main__":
    ScreenResultExampleApp().run()
```

https://user-images.githubusercontent.com/28237/232797119-2b711c08-a2ab-43be-817d-a36db64f995a.mov

### DIFF
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -1402,7 +1402,9 @@ class App(Generic[ReturnType], DOMNode):
             self.screen.post_message(events.ScreenSuspend())
             self.screen.refresh()
         next_screen, await_mount = self._get_screen(screen)
-        next_screen._push_result_callback(self.screen, callback)
+        next_screen._push_result_callback(
+            self.screen if self._screen_stack else None, callback
+        )
         self._screen_stack.append(next_screen)
         next_screen.post_message(events.ScreenResume())
         self.log.system(f"{self.screen} is current (PUSHED)")
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -92,7 +92,7 @@ from .keys import (
 from .messages import CallbackType
 from .reactive import Reactive
 from .renderables.blank import Blank
-from .screen import Screen, ScreenResultCallbackType
+from .screen import Screen, ScreenResultCallbackType, ScreenResultType
 from .widget import AwaitMount, Widget
 
 if TYPE_CHECKING:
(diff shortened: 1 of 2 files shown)


src/textual/app.py
```diff
@@ -1380,12 +1380,19 @@ def _replace_screen(self, screen: Screen) -> Screen:
...
+        callback: ScreenResultCallbackType[ScreenResultType] | None = None,
+    ) -> AwaitMount:
         """Push a new [screen](/guide/screens) on the screen stack, making it the current screen.
 
         Args:
             screen: A Screen instance or the name of an installed screen.
+            callback: An optional callback function that is called if the screen is dismissed with a result.
 
+        Returns:
+            An awaitable that awaits the mounting of the screen and its children.
```
willmcgugan, line 1395:
```suggestion
            An optional awaitable that awaits the mounting of the screen and its children.
```

willmcgugan (changes requested):
I think we will need to discard callbacks for switch screens.

davep:
> I think we will need to discard callbacks for switch screens.

@willmcgugan Good call. Done: 9123a80aa54c452042c128c1dd43ced5b1ebf7b3

willmcgugan (approved):

PR #2323, opened 2023-04-18
dim filter

willmcgugan:
Adds a `DimFilter` which pre-applies dim styles by blending the foreground with the background, and reseting the dim attribute.

This will be used to workaround an issue in xterm.js https://github.com/xtermjs/xterm.js/issues/4161 which doesn't render the dim attribute correctly.

### DIFF
--- a/src/textual/filter.py
+++ b/src/textual/filter.py
@@ -57,6 +57,46 @@ NO_DIM = Style(dim=False)
 """A Style to set dim to False."""
 
 
+@lru_cache(1024)
+def dim_color(background: RichColor, color: RichColor, factor: float) -> RichColor:
+    """Dim a color by blending towards the background
+
+    Args:
+        background: background color.
+        color: Foreground color.
+        factor: Blend factor
+
+    Returns:
+        _type_: _description_
+    """
+    red1, green1, blue1 = background.triplet
+    red2, green2, blue2 = color.triplet
+
+    return RichColor.from_rgb(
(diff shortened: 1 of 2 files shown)


src/textual/filter.py
```diff
@@ -17,36 +18,115 @@ def apply(self, segments: list[Segment]) -> list[Segment]:
...
+    )
+    return style + Style.from_color(color, background)
+
+
 class Monochrome(LineFilter):
     """Convert all colors to monochrome."""
 
     def apply(self, segments: list[Segment]) -> list[Segment]:
-        to_monochrome = self.to_monochrome
+        """Apply monochrome style to all segments."""
```
davep, line 50: Missing `Args` and `Returns` in the docstring.

src/textual/filter.py
```diff
@@ -17,36 +18,115 @@ def apply(self, segments: list[Segment]) -> list[Segment]:
...
-        background = (
-            None
-            if style_background is None
-            else Color.from_rich_color(style_background).monochrome.rich_color
-        )
-        return style + Style.from_color(color, background)
+        self.dim_factor = dim_factor
+
+    def apply(self, segments: list[Segment]) -> list[Segment]:
+        """Modify color of segments with dim style."""
```
davep, line 116: Missing `Args` and `Returns` in the docstring.

davep (approved):
Just a couple of docstrings that likely could do with args and returns adding for the sake of completeness.

PR #2329, opened 2023-04-19
Export types & doc improvements

### LINKED ISSUE #2168: Document custom types used and actually show their definition
I think we should document, perhaps in the API page, the custom types we use for things that are already documented.

For example, `App.CSS_PATH` shows up in the docs and is typed as `CSSPathType`, and the type itself is pretty much all the documentation needed for `App.CSS_PATH`.
However, `CSSPathType` isn't documented.

In my opinion, we chould/should document these custom types in the module they are defined in, or they should all be collected in a “Types” API page, for example.

**Note**: I spent some time trying to do this for a single type (`CSSPathType`) and I managed to put it in a separate API page.
However, I wasn't getting two important things:
 1. The signature of `App.CSS_PATH` wasn't linking to `CSSPathType`.
 2. I didn't manage to have `CSSPathType` show its signature, which is what I really wanted to have documented.

rodrigogiraoserrao:
This will close #2168 when it's done.

I did a pass on the App API page and looked for types that are used in signatures and that aren't documented.
This is the type of change I suggest.

If everyone is ok with this type of changes, I will go over the remainder of the API pages.

We still need to figure out how to document type aliases without duplicating their definition.


willmcgugan:
Looks good. Please add screenshots when its ready!

rodrigogiraoserrao:
As of bbde665, this PR exports _almost everything_ that the docs link to via argument types, return types, and exceptions raised.

To accommodate this, the API section grew 4 pages:

 - Errors (`errors.py`)
 - Filter (`filter.py`)
 - Scrollbar (`scrollbar.py`)
 - Types (new file `types.py`)

The first three pages contained 2+ objects that were referenced in other API pages so I thought it was best to add the whole API page.

Questions for @willmcgugan:

 - I have no idea what rule(s) we've been following to determine what API pages to include, so I'm not sure it's ok to add the API pages errors, filter, and scrollbar.
 - The page "Types" was named like that because I thought I'd end up only documenting type aliases there. As it turns out, there are a couple of odd exceptions, literals, type variables, and classes, that come from private modules but that can be used/returned by/raised by public methods/attributes. I feel tempted to rename this API page to "Miscellaneous" but still keep it at the end of the API menu.
 - I exported `RenderStyles` via the page "Types"; should I add the new API page "Styles" for `styles.py`?

CC @davep because you seem to have sensible ideas and suggestions when it comes to the docs :P

src/textual/app.py
```diff
@@ -399,33 +409,24 @@ def __init__(
 
     @property
     def workers(self) -> WorkerManager:
-        """The [worker](guide/workers/) manager.
-
-        Returns:
-            An object to manage workers.
-
-        """
+        """The [worker](guide/workers/) manager."""
```
willmcgugan, line 412: Without the `Returns:` these is nothing in the docs to link to the `WorkerManager` type.

src/textual/app.py
```diff
@@ -399,33 +409,24 @@ def __init__(
...
 
         The return value is set when calling [exit][textual.app.App.exit].
-
         """
         return self._return_value
 
     @property
     def children(self) -> Sequence["Widget"]:
-        """A view on to the App's children.
+        """A view onto the app's direct children.
```
willmcgugan, line 425: I think "immediate" would be more apt than "direct"

src/textual/app.py
```diff
@@ -500,18 +500,12 @@ def is_headless(self) -> bool:
...
 
     @property
     def screen_stack(self) -> Sequence[Screen]:
-        """The current screen stack.
-
-        Returns:
-            A snapshot of the current state of the screen stack.
-
-        """
+        """A snapshot of the current screen stack."""
```
willmcgugan, line 508: Ditto with the `Return:`. I couldn't make the docs link to `Screen` without it. Unless you have found a way of doing that?

src/textual/message_pump.py
```diff
@@ -340,7 +340,7 @@ def set_interval(
         self._timers.add(timer)
         return timer
 
-    def call_after_refresh(self, callback: Callable, *args, **kwargs) -> None:
+    def call_after_refresh(self, callback: Callable, *args: Any, **kwargs: Any) -> None:
```
willmcgugan, line 343: `object` may be a better choice if you are going to type these.

willmcgugan (changes requested):
You've removed the `Returns:` sections I intentionally put it.

Without them I couldn't get the docs to actually link to the return values of properties. Unless you have found a way of making that work automatically, can you restore the `Returns:`

davep:
All looking good to me; although it does raise one thing that overlaps with this quite a bit. This might be best as another issue but perhaps some of this can be rolled in here as it is about types and linking.

When I see something like this:

```markdown
[VideoPlayer][textual.widgets._video_player.VideoPlayer]
```

I feel it should really be like this:

```markdown
[`VideoPlayer`][textual.widgets._video_player.VideoPlayer]
```

as the result in the docs does give the type name a slightly different, and I think beneficial, styling.

Strong feelings either way, and scope for this or a followup PR?

davep (approved):
Approving within the bounds of Will's requested changes.

rodrigogiraoserrao:
> Strong feelings either way, and scope for this or a followup PR?

I feel strongly that you are right and that mentions to code objects/entities should be styled with backticks. I'll add those in.
Feel free to open the issue if you think it's a good idea, but I'll tackle that issue here.

src/textual/message_pump.py
rodrigogiraoserrao: You think so?

Typing them as `object` means that you can't do _anything_ with them without having the type checkers complain.
I'll wait for you to comment further before determining whether I'll leave them as `Any` or whether to type them as `object`.

rodrigogiraoserrao:
Attributes now get their source, after I reverted a change from a couple of months ago:

Untitled

The issue is that ALL attributes now show their first assignment and this can look a bit odd in some places, e.g., here are the two reactive attributes for `Button`:

Untitled

willmcgugan (approved):

PR #2333, opened 2023-04-19
Implement a Progress Bar widget.

rodrigogiraoserrao:
@willmcgugan here is a first prototype that leans heavily on `rich.progress.Progress`.
Does this feel like the way forward?

Or should we implement things more from scratch so that we can do things like have component styles to style parts of the progress bar?

https://user-images.githubusercontent.com/5621605/233153493-ac790e1d-6c0e-4c36-8a22-7cfed273d1d6.mov


willmcgugan:
I think we should implement things from scratch, yes.

Functionality would be quite similar to Rich's Progress. Although simpler. It only needs to render a single progress, with percentage and optional ETA.

rodrigogiraoserrao:
@willmcgugan I think things are going in a pretty good direction as of 8f2fae5.

But I have an issue I don't understand.
I don't know why I need this line:

https://github.com/Textualize/textual/blob/8f2fae5d1e0524b68d100b7202bfa45b3718dad7/src/textual/widgets/_progress_bar.py#L248

<details>
<summary>App to reproduce the issue</summary>

```py
from textual.widgets import ProgressBar, Footer


class MyApp(App[None]):
    BINDINGS = [
        ("a", "advance", "Advance by 1"),
        ("j", "jump", "Jump to 25"),
        ("u", "update", "Update to start"),
        ("d", "duplicate", "Duplicate total"),
    ]

        self.pb = ProgressBar(total=None)
        yield self.pb
        yield Footer()

    def action_update(self):
        self.pb.update(total=100)

    def action_advance(self):
        self.pb.advance()

    def action_jump(self):
        self.pb.progress = 25

    def action_duplicate(self):
        self.pb.total *= 2


if __name__ == "__main__":
    MyApp().run()
```

</details>

src/textual/renderables/bar.py
```diff
@@ -5,7 +5,7 @@
 from rich.text import Text
 
 
-class UnderlineBar:
+class RenderableBar:
```
willmcgugan, line 8: I'd rather not have "Renderable" in the name, since they are all renderables. I guess you renamed it, because you repurposed it as a progress bar. How about just "Bar" then?

src/textual/widgets/_progress_bar.py
```diff
@@ -0,0 +1,432 @@
...
+
+        The percentage is a value between 0 and 1 and the returned value is only
+        `None` if the total progress of the bar hasn't been set yet.
+        In other words, after the progress bar emits the message
+        [`ProgressBar.Started`][textual.widgets.ProgressBar.Started],
+        the value of `percentage` is always not `None`.
+
+        Example:
+            ```py
+            pb = ProgressBar()
```
davep, line 380: In the other examples a variable called `progress_bar` tends to be used, presumably avoiding the "no abbreviations" rule; perhaps do the same here too?

src/textual/widgets/_progress_bar.py
```diff
@@ -0,0 +1,432 @@
...
+    }
+    """
+
+    _completion_percentage: reactive[float | None] = reactive[Optional[float]](None)
+    """The percentage of the bar that should be filled or `None` if indeterminate.
+
+    If the completion percentage is `None`, then the bar will render as
+    an indeterminate progress bar.
+    """
+    _refresh_timer: Timer | None
```
willmcgugan, line 59: This should really be initialized in the constructor, but there doesn't seem to be one.

I think we need a `__init__` for this reason, and that we want to prevent the dev from passing in child widgets.

src/textual/widgets/_progress_bar.py
```diff
@@ -0,0 +1,432 @@
...
+
+    def watch__completion_percentage(self, percentage: float | None) -> None:
+        """Manage the timer that enables the indeterminate bar animation."""
+        if percentage is not None:
+            if self._refresh_timer:
+                self._refresh_timer.stop()
+                self._refresh_timer = None
+
+        else:
+            self._refresh_timer = self.set_interval(1 / 60, self.refresh)
```
willmcgugan, line 70: Whoah, 60FPS.  Way to high: suggest 15 FPS max for a passable animation. 

Be careful about creating things like timers. It looks like you could be creating multiple timers here.

Do you need a custom timer? Would `self.auto_refresh = 15` suffice?

src/textual/widgets/_progress_bar.py
```diff
@@ -0,0 +1,432 @@
...
+    def render_indeterminate(self) -> RenderResult:
+        """Render a frame of the indeterminate progress bar animation."""
+        width = self.size.width
+        highlighted_bar_width = 0.25 * width
+        # Width used to enable the visual effect of the bar going into the corners.
+        total_imaginary_width = width + highlighted_bar_width
+
+        speed = 30  # Cells per second.
+        # Compute the position of the bar.
+        start = (speed * monotonic()) % (2 * total_imaginary_width)
```
willmcgugan, line 97: I have feeling this may cause test flakiness in CI.

Suggest you have a method called `_get_time`. In tests you can monkey patch this.

src/textual/widgets/_progress_bar.py
```diff
@@ -0,0 +1,432 @@
...
+    def watch__completion_percentage(self, percentage: float | None) -> None:
+        if percentage is None:
+            self.renderable = "--:--:--"
+        else:
+            # Keep track of the starting time.
+            if self._start_time is None:
+                self._start_time = monotonic()
+            # If we are not notified of progress for some time, periodically refresh
+            # the countdown so that the ETA is updated either way.
+            if self._refresh_timer is None:
```
willmcgugan, line 159: It makes me nervous creating objects in watch functions. Looks fine, but the code implies you will be creating new objects here.

Suggest we move setting the interval outside of the watch function. How about create it `on_mount`? That way it is clear there will only be a single instance.

src/textual/widgets/_progress_bar.py
```diff
@@ -0,0 +1,432 @@
...
+    """The percentage of progress that has been completed or `None` if indeterminate."""
+
+    DEFAULT_CSS = """
+    ETAStatus {
+        width: 9;
+        content-align-horizontal: right;
+    }
+    """
+
+    _refresh_timer: Timer | None = None
```
willmcgugan, line 145: Not keen on setting initial values as classvars. Initial values should be the domain of `__init__`.

src/textual/widgets/_progress_bar.py
```diff
@@ -0,0 +1,432 @@
...
+    _percentage: reactive[float | None] = reactive[Optional[float]](None)
+    """The percentage of progress that has been completed.
+
+    We keep this private to prevent the user from having multiple reactive attributes
+    that can be used to advance the progress of the bar.
+    """
+    _started = False
+    """Whether the total was set to a non-`None` value before."""
+
+    class Completed(Message):
```
willmcgugan, line 227: Struggling to think why we would need a message for this (ditto `Started`).

The dev tells this widget about progress, which means the dev already knows if it is complete. Don't see any value in informing them of something the probably know. Suggest we remove these messages unless you have a compelling reason to keep them...

As an aside, messages usually use present tense. It's `on_click`, not `on_clicked`. This this would be `Complete` rather than `Completed`.

src/textual/widgets/_progress_bar.py
```diff
@@ -0,0 +1,432 @@
...
+            """
+            self.progress_bar = progress_bar
+            self.total = total
+            super().__init__()
+
+    def __init__(
+        self,
+        total: float | None = None,
+        *,
+        hide_bar: bool = False,
```
willmcgugan, line 269: I think these parameter may be better as `show_*`, otherwise it feels like a double negetive.

src/textual/widgets/_progress_bar.py
```diff
@@ -0,0 +1,432 @@
...
+            hide_bar: Whether to hide the bar portion from the progress bar.
+            hide_percentage: Whether to hide the percentage status from the bar.
+            hide_eta: Whether to hide the ETA countdown from the progress bar.
+            name: The name of the widget.
+            id: The ID of the widget in the DOM.
+            classes: The CSS classes for the widget.
+            disabled: Whether the widget is disabled or not.
+        """
+        super().__init__(name=name, id=id, classes=classes, disabled=disabled)
+        self._bar = Bar()
```
willmcgugan, line 302: You are constructing these here, but also creating the same objects in `compose`. These don't appear to be referenced.

src/textual/widgets/_progress_bar.py
```diff
@@ -0,0 +1,432 @@
...
+
+        self._percentage = None
+        self.total = total
+
+    def compose(self) -> ComposeResult:
+        with Horizontal():
+            if not self.hide_bar:
+                bar = Bar()
+                # Notify the bar when the percentage is updated.
+                self.watch(
```
willmcgugan, line 317: If everything is dependant on `percentage`, can we not have a `watch_percentage` method here that updates its children?

I think that would be easier to follow than the watches and partials.

src/textual/widgets/_progress_bar.py
```diff
@@ -0,0 +1,432 @@
...
+
+    def watch_total(self, total: float | None) -> None:
+        """Post the message [`ProgressBar.Started`][textual.widgets.ProgressBar.Started]."""
+        if total is None:
+            return
+        if not self._started:
+            self.post_message(ProgressBar.Started(self, total))
+            self._started = True
+
+    def compute__percentage(self) -> float | None:
```
willmcgugan, line 354: Confused about `_percentage` and `percentage` which appear to calculate the same thing?

src/textual/widgets/_progress_bar.py
```diff
@@ -0,0 +1,432 @@
...
+            left = ceil((delta / percentage) * (1 - percentage))
+            minutes, seconds = divmod(left, 60)
+            hours, minutes = divmod(minutes, 60)
+            if hours > 999999:
+                self.renderable = "+999999h"
+            elif hours > 99:
+                self.renderable = f"{hours}h"
+            else:
+                self.renderable = f"{hours:02}:{minutes:02}:{seconds:02}"
+        self.refresh()
```
willmcgugan, line 193: Can we avoid doing a refresh if the eta doesn't change?

src/textual/widgets/_progress_bar.py
```diff
@@ -0,0 +1,432 @@
...
+                if self._completion_percentage < 1
+                else self.get_component_rich_style("bar--complete")
+            )
+            return RenderableBar(
+                highlight_range=(0, self.size.width * self._completion_percentage),
+                highlight_style=Style.from_color(bar_style.color),
+                background_style=Style.from_color(bar_style.bgcolor),
+            )
+
+    def render_indeterminate(self) -> RenderResult:
```
willmcgugan, line 88: Can I see a video of the intermediate effect?

src/textual/widgets/_progress_bar.py
```diff
@@ -0,0 +1,432 @@
...
+    """The percentage of progress that has been completed or `None` if indeterminate."""
+
+    DEFAULT_CSS = """
+    PercentageStatus {
+        width: 5;
+        content-align-horizontal: right;
+    }
+    """
+
+    def watch__completion_percentage(self, percentage: float | None) -> None:
```
willmcgugan, line 125: Setting the renderable won't prompt a refresh. There is an `update` method which will set the renderable and prompt a refresh.

We want to avoid unnecessary updates. If the rounded percentage hasn't changed, we don't want to refresh.

Suggest we have a second reactive for the status text, which refresh when it changes.

willmcgugan (changes requested):
Functionality is about right I think. Looks good. But needs another pass at the implementation.

src/textual/widgets/_progress_bar.py
rodrigogiraoserrao: Sure thing.

src/textual/widgets/_progress_bar.py
willmcgugan: You can set `auto_refresh` to None when you don't need it to refresh. Which seems similar to what you are doing here.

rodrigogiraoserrao: That's exactly it. On it.

src/textual/widgets/_progress_bar.py
rodrigogiraoserrao: [Here you go](https://github.com/Textualize/textual/pull/2333#issuecomment-1522142784)

docs/examples/widgets/progress_bar_isolated.py
```diff
@@ -0,0 +1,26 @@
...
+                yield ProgressBar()
+
+    def key_f(self) -> None:
+        # Freeze time for the indeterminate progress bar.
+        self.query_one(ProgressBar).query_one("#bar")._elapsed_time = lambda: 5
+
+    def key_t(self) -> None:
+        # Freeze time to show always the same ETA.
+        self.query_one(ProgressBar).query_one("#eta")._elapsed_time = lambda: 4.2
+        self.query_one(ProgressBar).update(total=100, progress=39)
```
rodrigogiraoserrao, line 19: We need this so that the screenshots in the tests and in the docs always look the same.
I don't like this assignment to `_elapsed_time` to be so publicly exposed but I couldn't think of anything better.

willmcgugan: In a test, this would be fine. But I'm afraid that we can't do this in an example. Call me if you want to brainstorm an alternative solution.

willmcgugan (commented):
Looks good! But that private attribute in the examples may be an issue.

We want examples to be show canonical usage, so we shouldn't be using private attributes.

I think we could fake it, show a screenshot with the hack to set the time, but link to the real example.

src/textual/widgets/_progress_bar.py
```diff
@@ -0,0 +1,434 @@
...
+        end = start + highlighted_bar_width
+
+        bar_style = self.get_component_rich_style("bar--indeterminate")
+        return BarRenderable(
+            highlight_range=(max(0, start), min(end, width)),
+            highlight_style=Style.from_color(bar_style.color),
+            background_style=Style.from_color(bar_style.bgcolor),
+        )
+
+    def _elapsed_time(self) -> float:
```
willmcgugan, line 119: If this isn't a property. It should begin with a verb, i.e. `_get_elapsed_time`.

src/textual/widgets/_progress_bar.py
```diff
@@ -0,0 +1,434 @@
...
+        name: str | None = None,
+        id: str | None = None,
+        classes: str | None = None,
+        disabled: bool = False,
+    ):
+        super().__init__(name=name, id=id, classes=classes, disabled=disabled)
+        self._percentage = None
+        self._label_text = "--%"
+
+    def watch__percentage(self, percentage: float | None) -> None:
```
willmcgugan, line 161: Docstring por favor.

src/textual/widgets/_progress_bar.py
```diff
@@ -0,0 +1,434 @@
...
+        self.show_eta = show_eta
+
+        self.percentage = None
+        self.total = total
+
+    def compose(self) -> ComposeResult:
+        # We create a closure so that we can determine what are the sub-widgets
+        # that are present and, therefore, will need to be notified about changes
+        # to the percentage.
+        def update_percentage(widget: Widget):
```
willmcgugan, line 343: Return types please

src/textual/widgets/_progress_bar.py
```diff
@@ -0,0 +1,434 @@
...
+            if self.show_percentage:
+                percentage_status = PercentageStatus(id="percentage")
+                self.watch(self, "percentage", update_percentage(percentage_status))
+                yield percentage_status
+            if self.show_eta:
+                eta_status = ETAStatus(id="eta")
+                self.watch(self, "percentage", update_percentage(eta_status))
+                yield eta_status
+
+    def validate_progress(self, progress: float) -> float:
```
willmcgugan, line 366: You could use  `clamp` here.

willmcgugan (approved):
LGTM. Just some nitpicks to consider.

Issue #2331, opened 2023-04-19
Bindings on non-scroll containers should have `inherit_bindings=False`


willmcgugan:
Containers which don't scroll, shouldn't inherit bindings from Widget.

Issue #2100, opened 2023-03-21
Some keys are no longer showing in the `Footer`


davep:
Given this code in Textual v0.15.1:

```python
from textual.app     import App, ComposeResult
from textual.widgets import Header, Footer
from textual.binding import Binding

class MissingFooterStuffApp( App[ None ] ):

    BINDINGS = [
        Binding( "up", "oops(0)", "Up" ),
        Binding( "down", "oops(1)", "Down" ),
        Binding( "left", "oops(2)", "Left" ),
        Binding( "right", "oops(3)", "Right" ),
        Binding( "ctrl+up", "oops(4)", "Up" ),
        Binding( "ctrl+down", "oops(5)", "Down" ),
        Binding( "ctrl+left", "oops(6)", "Left" ),
        Binding( "ctrl+right", "oops(7)", "Right" ),
        Binding( "space", "oops(8)", "Space" ),
        Binding( "enter", "oops(9)", "Space" ),
        Binding( "q", "oops(10)", "Q" ),
    ]

    def compose( self ) -> ComposeResult:
        yield Header()
        yield Footer()

    def action_oops( self, _ ):
        self.app.exit()

if __name__ == "__main__":
    MissingFooterStuffApp().run()
```

while all the bindings work, not all of them are showing in the `Footer` any more. The above isn't an exhaustive list (I'm told that <kbd>Home</kbd> and <kbd>End</kbd> aren't showing up either).

Fairly sure these were working okay until recently.

Thanks to [DigitalSeraphim on Discord for alerting us to that](https://discord.com/channels/1026214085173461072/1033754296224841768/1087847328813174824).

digitalseraphim:
This was due to the swapping of the order of "precedence" in the "binding chain" applied in https://github.com/Textualize/textual/commit/32d18a

As this was obviously an intentional change, I'll leave it to the main team to fix 😄

willmcgugan:
Probably a dupe of https://github.com/Textualize/textual/issues/2219

davep:
Looking at this a little further, the issue here seems to come down to doing things on the application vs, say, doing them on screen or a widget. To narrow it down to a simple example, here's an app where we bind <kbd>Up</kbd> on the app:

```python
from textual.app      import App, ComposeResult
from textual.binding  import Binding
from textual.reactive import reactive
from textual.widgets  import Header, Footer, Label

class FooterBindingOnApp( App[ None ] ):

    BINDINGS = [
[...]

PR #2337, opened 2023-04-20
Move default scroll navigation keys out of `Widget` and into `ScrollableContainer`

davep:
This PR removes the default bindings for navigation keys -- binding them to various scroll actions -- from `Widget` and introduces them in a new `ScrollableContainer` container. Primarily this is to address #2100; the problem here being that not all widgets want to scroll, but many non-scrolling widgets will want to do other things with navigation keys and the bindings can end up clashing when it comes to sorting out descriptions and entries (or lack of them) in the `Footer` of an application.

This is a breaking change. Any widget that scrolls, and expects keyboard-driven scrolling "out of the box", will cease to be navigable via the keyboard. This is easily solved by switching to `ScrollableContainer`, `VerticalScroll` or `HorizontalScroll` as the parent (so far any examples I've seen where this is an issue have inherited from `Widget` or `Container` so changing the parent class won't be much of a change).

This PR also goes on to change the demo, plus `ScrollView`, to ensure scrolling works as before there.

I've run through the main examples and they all look fine.

While the main thrust of this PR sort of renders the need to do it moot, I've also rolled #2331 in here too; mostly now this will work as a guard against any default bindings that may get added to `Widget` in the future, such that we should think about the consequences for containers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Breaking change: standard keyboard scrollable navigation bindings have been moved off `Widget` and onto a new base class for scrollable containers (see also below addition) https://github.com/Textualize/textual/issues/2332
 - `ScrollView` now inherits from `Scrollablecontainer` rather than `Widget` https://github.com/Textualize/textual/issues/2332
+- Containers no longer inherit any bindings from `Widget` https://github.com/Textualize/textual/issues/2331
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
-## [0.20.1] - 2023-04-18
+## Unreleased
 
 ### Changed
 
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,18 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## Unreleased
+
+### Changed
+
+- Breaking change: standard keyboard scrollable navigation bindings have been moved off `Widget` and onto a new base class for scrollable containers (see also below addition) https://github.com/Textualize/textual/issues/2332
+- `ScrollView` now inherits from `Scrollablecontainer` rather than `Widget` https://github.com/Textualize/textual/issues/2332
```
rodrigogiraoserrao, line 13:
```suggestion
- `ScrollView` now inherits from `ScrollableContainer` rather than `Widget` https://github.com/Textualize/textual/issues/2332
```

src/textual/scroll_view.py
```diff
@@ -7,11 +7,11 @@
...
 
+from .containers import ScrollableContainer
 from .geometry import Size
-from .widget import Widget
 
 
-class ScrollView(Widget):
+class ScrollView(ScrollableContainer):
     """
     A base class for a Widget that handles it's own scrolling (i.e. doesn't rely
```
rodrigogiraoserrao, line 16:
```suggestion
    A base class for a Widget that handles its own scrolling (i.e. doesn't rely
```

rodrigogiraoserrao (approved):
Nice! 🚀 

For what is worth, to me it makes a LOT of sense to have a base class for scrollable containers, irregardless of the fuss I made yesterday about bindings and `show=False` and etc.

src/textual/scroll_view.py
davep: I'd normally discourage fixing unrelated things in a PR, but seeing as it's one character I'll let it slide. ;-)

willmcgugan (approved):

PR #2338, opened 2023-04-20
New CLI runner

willmcgugan:
Alternative way of running Textual apps via `textual run`

- `console run` Now execs apps in a new process, with updated environment
- Added `-c` switch to `console.run` which runs commands in a dev context


src/textual/cli/cli.py
```diff
@@ -132,53 +158,39 @@ def run_app(
     in quotes:
 
         textual run "foo.py arg --option"
+
+    If your Textual app isn't easily accessible via a script or import, you can add the -c switch
```
rodrigogiraoserrao, line 162:
```suggestion
    If your Textual app isn't easily accessible via a script or import, you can add the -c switch.
```

rodrigogiraoserrao, line 162: I genuinely don't understand this sentence.

src/textual/app.py
```diff
@@ -1762,23 +1772,15 @@ async def _ready(self) -> None:
...
-            svg = self.export_screenshot(title=screenshot_title)
-            self._screenshot = svg  # type: ignore
+        async def take_screenshot() -> None:
+            """Take a screenshot and exit."""
+            self.save_screenshot()
             self.exit()
 
-        self.set_timer(screenshot_timer, on_screenshot, name="screenshot timer")
+        self.set_timer(
+            constants.SCREENSHOT_DELAY, take_screenshot, name="screenshot timer"
```
rodrigogiraoserrao, line 1782: The default value is `-1`. The timer just ignores the callback in that case?

src/textual/app.py
```diff
@@ -1966,6 +1968,14 @@ async def _shutdown(self) -> None:
 
         self._print_error_renderables()
 
+        if constants.SHOW_RETURN and self._exit and self._return_value is not None:
```
rodrigogiraoserrao, line 1971: Even if the return value is `None`, shouldn't it be printed?

E.g., I'm debugging an app and the bug makes it exit at the wrong time (but I don't know yet).
I'm expecting to see a return value but I don't see the final panel.
I'll think the issue lies in the `textual run -r` option.
However, if the `None` is shown, I'll be able to tell that my app returned the wrong thing.

src/textual/cli/_run.py
```diff
@@ -0,0 +1,134 @@
...
+        True if the path references Python code, False it it doesn't.
+    """
+    if path.endswith(".py"):
+        return True
+    try:
+        with open(path, "r") as source:
+            first_line = source.readline()
+    except IOError:
+        return False
+    return first_line.startswith("#!") and "py" in first_line
```
davep, line 82: I'll be honest and say I'm struggling to think of one off the top of my head right now, although of course "#!" can point to *anything*,  but I feel that looking for just `py` is a little too permissive here and could result in a false positive.

Perhaps consider being just a wee bit more guarded here?

src/textual/cli/_run.py
```diff
@@ -0,0 +1,134 @@
...
+    app.run()
+else:
+    # Otherwise it is assumed to be a class
+    app().run()
+"""
+)
+"""A template script to import and run an app."""
+
+
+def run_app(command_args: str, environment: dict[str, str] | None = None) -> None:
```
davep, line 37: The args in the docstring don't match function here.

davep, line 37: `NoReturn` here too?

davep (approved):
Couple of notes rather than must-changes, and aside from the thing mentioned in person it's all looking good to me.

src/textual/cli/cli.py
willmcgugan: It's when your app runs via a CLI app. Like `textual keys`.

Suggestions for a rewording?

src/textual/cli/_run.py
willmcgugan: Will go with this for now, and handle any clashes if they occur.

src/textual/cli/_run.py
willmcgugan: THis one may actually return.

src/textual/cli/cli.py
rodrigogiraoserrao: I finally got it!

How about something like "Use `-c` if you are specifying a custom command to run your app."?

Issue #1999, opened 2023-03-09
Dark mode toggle in a stacked screen doesn't cause mode switch in other screens


davep:
Initially reported by @learnbyexample [over in discussions](https://github.com/Textualize/textual/discussions/1992), if dark mode is toggled inside a screen, other screens on the stack don't refresh as you'd expect.

Consider this code:

```python
from textual.app        import App, ComposeResult
from textual.containers import Vertical
from textual.widgets    import Header, Footer, Label, Button
from textual.screen     import Screen
from textual.binding    import Binding

class Child( Screen ):

    def compose( self ) -> ComposeResult:
        yield Header()
        with Vertical():
            yield Label( "This is the child screen" )
            yield Button( "Close Me" )
        yield Footer()

    def on_button_pressed( self, _: Button.Pressed ) -> None:
        self.app.pop_screen()

class Main( Screen ):

    def compose( self ) -> ComposeResult:
        yield Header()
        with Vertical():
            yield Label( "This is the main screen" )
            yield Button( "Show Child" )
        yield Footer()

    def on_button_pressed( self, _: Button.Pressed ) -> None:
        self.app.push_screen( Child() )

class DarkToggleTest( App[ None ] ):

    BINDINGS = [
        Binding( "d", "app.toggle_dark", "Toggle Light/Dark" ),
    ]


    CSS = """
    Vertical {
        align: center middle;
    }
    """

    def on_mount( self ) -> None:
        """"""
        self.push_screen( Main() )

if __name__ == "__main__":
    DarkToggleTest().run()
```

If you run it up and press the "Show Child" button, then press <kbd>d</kbd> to toggle dark mode, then press the "Close Me" button, you'll notice that the "parent" screen is still in dark mode. Further, if you then press <kbd>d</kbd> you'll notice that nothing happens (because then that screen is switching *to* dark mode).

When watching a change to dark mode we should refresh the current screen but also alert other screens in the stack that they need to refresh when they become shown.

willmcgugan:
Suspect this is fixed. Needs confirmation.

learnbyexample:
Checked with Textual version 0.20.1 - the test app mentioned in the discussion seemed stable with theme switches. Then checked with my original app where I had the issue, and seems there are still issues. Here's a sample video:

https://user-images.githubusercontent.com/17766317/233251755-28778d77-8d7b-409b-84b7-86a6c5e074d1.mp4

[...]

PR #2339, opened 2023-04-20
Ensure an app's CSS refresh updates all screens in the stack

davep:
Mainly to address #1999 (dark mode toggles in a "child" screen not updating the "parent" screen). In turn also fixes what happens when making CSS changes in a situation where the frontmost screen has a transparent background.


rodrigogiraoserrao (approved):
Makes sense to me.

willmcgugan (approved):

Issue #2352, opened 2023-04-22
`TabbedContent.active` no longer working as of 0.20.1


davep:
It seems that the fix for #2229 made in v0.20.1 (#2322) has broken the ability to use `TabbedContent.active` to change the active tab from code. This can be see by running [`docs/examples/widgets/tabbed_content.py`](https://textual.textualize.io/widgets/tabbed_content/#example) with v0.20.0 and then with v0.20.1 and pressing any of the bound keys. With v0.20.0 the tab *and* the content changes; with v0.20.1 the tab changes but the content doesn't.

PR #2353, opened 2023-04-22
Fix `TabbedContent.active` not changing actual content

davep:
Ensures that the actual content gets changed when TabbedContent.active is updated via code. Also adds more testing for TabbedContent.

See #2352 for details.


willmcgugan (approved):

willmcgugan:
Thanks

PR #2358, opened 2023-04-22
Fix scroll flicker

willmcgugan:
Fixes flicker when the cursor on DataTable and Tree caused the widget to scroll.

### DIFF
--- a/src/textual/devtools/client.py
+++ b/src/textual/devtools/client.py
@@ -211,9 +211,9 @@ class DevtoolsClient:
             log: The log to write to devtools
         """
         if isinstance(log.objects_or_string, str):
-            self.console.print(log.objects_or_string)
+            self.console.print(log.objects_or_string, markup=False)
         else:
-            self.console.print(*log.objects_or_string)
+            self.console.print(*log.objects_or_string, markup=False)
 
         segments = self.console.export_segments()
 
--- a/src/textual/message_pump.py
+++ b/src/textual/message_pump.py
@@ -374,6 +374,7 @@ class MessagePump(metaclass=_MessagePumpMeta):
             **kwargs: Keyword arguments to pass to the callable.
         """
         self._next_callbacks.append(partial(callback, *args, **kwargs))
+        self.check_idle()
 
     def _on_invoke_later(self, message: messages.InvokeLater) -> None:
         # Forward InvokeLater message to the Screen
@@ -506,6 +507,7 @@ class MessagePump(metaclass=_MessagePumpMeta):
                             except Exception as error:
                                 self.app._handle_exception(error)
                                 break
+                    await self._flush_next_callbacks()
 
     async def _flush_next_callbacks(self) -> None:
         """Invoke pending callbacks in next callbacks queue."""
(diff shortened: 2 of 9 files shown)


src/textual/devtools/client.py
```diff
@@ -211,9 +211,9 @@ def log(
             log: The log to write to devtools
         """
         if isinstance(log.objects_or_string, str):
-            self.console.print(log.objects_or_string)
+            self.console.print(log.objects_or_string, markup=False)
```
willmcgugan, line 214: Disabling markup on log print. It risks accidentally hiding content.

src/textual/message_pump.py
```diff
@@ -374,6 +374,7 @@ def call_next(self, callback: Callable, *args, **kwargs) -> None:
             **kwargs: Keyword arguments to pass to the callable.
         """
         self._next_callbacks.append(partial(callback, *args, **kwargs))
+        self.check_idle()
```
willmcgugan, line 377: A bug! Without this, the next callbacks wouldn't run outside of a different message handler.

src/textual/message_pump.py
```diff
@@ -506,6 +507,7 @@ async def _process_messages_loop(self) -> None:
                             except Exception as error:
                                 self.app._handle_exception(error)
                                 break
+                    await self._flush_next_callbacks()
```
willmcgugan, line 510: Another bug! the call next callbacks wouldn't run if created via on_idle

src/textual/pilot.py
```diff
@@ -100,11 +100,11 @@ async def click(
             target_widget, offset, button=1, shift=shift, meta=meta, control=control
         )
         app.post_message(MouseDown(**message_arguments))
-        await self.pause()
+        await self.pause(0.1)
```
willmcgugan, line 103: Testing failing on Windows. Not ideal.

src/textual/screen.py
```diff
@@ -479,20 +475,12 @@ def _on_timer_update(self) -> None:
                 self._dirty_widgets.clear()
 
         if self._callbacks:
-            self.post_message(events.InvokeCallbacks())
-
-    async def _on_invoke_callbacks(self, event: events.InvokeCallbacks) -> None:
```
willmcgugan, line 484: This was written before adding `cell_next`, and is no longer needed.

src/textual/screen.py
```diff
@@ -479,20 +475,12 @@ def _on_timer_update(self) -> None:
...
-        """Handle PostScreenUpdate events, which are sent after the screen is updated"""
-        await self._invoke_and_clear_callbacks()
+            self.call_next(self._invoke_and_clear_callbacks)
 
     async def _invoke_and_clear_callbacks(self) -> None:
         """If there are scheduled callbacks to run, call them and clear
         the callback queue."""
         if self._callbacks:
-            display_update = self._compositor.render_update(
```
willmcgugan, line 492: Suspect this was a bit of a hail Mary to ensure that the callbacks happened after a refresh. I'm 99% certain this isn't required now.

davep (approved):

PR #2374, opened 2023-04-25
pass through unprocessed args

willmcgugan:
Uses a feature of Click to pass through "unprocessed arguments".

Prevuously if you want to run a command you would have to quote everything

```
textual run -c "python -m textual"
```

Now, you can omit the quotes:

```
textual run -c python -m textual
```


willmcgugan:
#executivedecision

rodrigogiraoserrao (commented):
LGTM.

PR #2377, opened 2023-04-26
container refactor

willmcgugan:
A modest refactor of containers, which defines what a container *is*.

Other than Center, and Middle, containers will by default console all remaining space in the window. There are variants for containers that with automatic scrolling on one or both axis.

### DIFF
--- a/docs/examples/styles/visibility_containers.css
+++ b/docs/examples/styles/visibility_containers.css
@@ -1,6 +1,7 @@
 Horizontal {
     padding: 1 2;     /* (1)! */
     background: white;
+    height: 1fr;
 }
 
 #top {}               /* (2)! */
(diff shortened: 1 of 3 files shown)


rodrigogiraoserrao:
I am confused. If you removed the `1fr` from the dimensions of the containers, how is their default behaviour to take up the _remaining_ space available?

willmcgugan:
That is the default behaviour if there is no value for width or height.

rodrigogiraoserrao:
> That is the default behaviour if there is no value for width or height.

I don't think it is. If you run the app below, the label isn't visible because it is pushed off by the inner `Vertical`, probably with a default height of 100%:

```py


class MyApp(App[None]):
        with Vertical():
            yield Vertical()
            yield Label("asdfsf\nasdfasdfa\nasdfasdfa\n")


if __name__ == "__main__":
    MyApp().run()
```

willmcgugan:
If you add a container, there will be zero remaining space left. So it has used up all the remaining space.

Perhaps a better way to put it would be that it will consume all the space of its parent?

Issue #2385, opened 2023-04-26
Superfluous vertical scrollbar for Tabs widget since 0.21.0


ankraft:
It seems that since 0.21.0 a "Tabs" widget always displays a vertical scrollbar, even when no content is present in one of the tabs.
The extra size of the scrolling area is 3 lines, the same as the size of the tab area (2 lines + 1 underline).

![screenshot tabs extra vertical scrollbar](https://user-images.githubusercontent.com/8394617/234603510-4051616b-32ad-4c84-bc9e-2cdfeb4f4863.png)

Running on MacOS 13.3 and iTerm 3.4.19.

Thanks for the continuous developments and support!

rodrigogiraoserrao:
@willmcgugan following up on my first comment in the PR, this is more evidence in favour of setting the default width and height of containers to `1fr`.

davep:
Thanks, this looks like an expected while also unintended consequence of some of the changes we're making to the default styling of various containers. But... it feels unlikely that this would affect `Tabs`; could it be that you're actually using `TabbedContent`?

Either way, having more of a chat about it here now, expect a 0.21.1 fairly soon; we're likely going to roll this particular aspect of the change back.

ankraft:
Thanks a lot for the quick feedback. Looking forward to 0.21.1 ;)

> could it be that you're actually using TabbedContent?

No, for the screenshot I removed all the actual containers (except the header, footer and the Tabs) from the app. The Tabs widget I added like this in `compose()`:

``` python
yield Tabs(Tab('Resources', id = ...),
           Tab('Requests', id = ...),
           Tab('Registrations', id = ...),
           Tab('Configurations', id = ...),
           Tab('Infos', id = ...),
           Tab('About', id = ...))
```

davep:
> No, for the screenshot I removed all the actual containers (except the header, footer and the Tabs) from the app.

@ankraft Hmm, that's sort of unexpected. Do you have a stand-alone minimal example you could post that shows the problem you're seeing, just so we can be sure that the incoming fix will address this?

ankraft:
@davep Ah, perhaps it is not the `Tab` widget, it seems to be the `ContentSwitcher`:

```python
from textual.app import App, ComposeResult
from textual.widgets import Tabs, Tab, ContentSwitcher

class TabsApp(App):

    def compose(self) -> ComposeResult:
        yield Tabs(Tab('1'), Tab('2'))
        yield ContentSwitcher()

[...]

PR #2386, opened 2023-04-26
Make `textual.containers` `1fr` in meaningful dimensions by default

davep:
An unintended consequence of changes made to containers in v0.21.0 (#2377) is something like #2385; so this PR sort of rolls that change back, and solidifies how containers are styled by default in respect to their width and height. Where appropriate the dimensions will be `1fr`.

Also includes a wee tweak to `ProgressBar` to ensure its `DEFAULT_CSS` styling of `Vertical` isn't quite so greedy (noticed because without this the `ProgressBar` snapshot tests all failed when the main purpose of this PR was introduced).

Note that, as a followup to this, as identified by @darrenburns, we will need to think about revamping the layout section of the manual as it's a little out of date in respect to how containers work now (and what's available).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unrelased
 
+### Added
+
+- Added `ProgressBar` widget https://github.com/Textualize/textual/pull/2333
+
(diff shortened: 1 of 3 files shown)


rodrigogiraoserrao (approved):
🚀

PR #2393, opened 2023-04-26
Fix demo example CSS (fixes #2389)

### LINKED ISSUE #2389: Issues with in-demo CSS code display — pull request welcome?
Hi, thanks for making textual and sharing it as Open Source! :pray: :+1: 

When I played with `python -m textual` today, I noticed that the CSS section — after scrolling to the bottom — …

textual_Screenshot_20230426_175034

…makes it _appears_ as if `OptionGroup {` was the last of 50 lines of the [file `demo.css`](https://github.com/Textualize/textual/blob/4148b1d4501f405eac9c74485ff876ffc8201ca7/src/textual/demo.css#L50). From a user point of view it felt broken and unintended so I had a closer look to find that:
- commit e419a0821cda7c5860c171a20c09034daaf82dea added explicit `[:50]`
- the same commit resolves use of `EXAMPLE_CSS` and that variable is now unused.

I would like to make one or two pull requests dropping the `[:50]` and dropping unused variable `EXAMPLE_CSS`.
Would that be welcome, should it be one or two, or would you rather quickfix this yourself?

Thanks and best, Sebastian

hartwork:
Fixes #2389

- [x] Updated documentation

### DIFF
--- a/src/textual/demo.py
+++ b/src/textual/demo.py
@@ -294,7 +294,7 @@ class DemoApp(App[None]):
         self.query_one(TextLog).write(renderable)
 
     def compose(self) -> ComposeResult:
-        example_css = "\n".join(Path(self.css_path[0]).read_text().splitlines())
+        example_css = Path(self.css_path[0]).read_text()
         yield Container(
             Sidebar(classes="-hidden"),
             Header(show_clock=False),
--- a/src/textual/demo.py
+++ b/src/textual/demo.py
@@ -94,25 +94,6 @@ Textual uses Cascading Stylesheets (CSS) to create Rich interactive User Interfa
 Here's an example of some CSS used in this app:
 """
 
-
-EXAMPLE_CSS = """\
-Screen {
-    layers: base overlay notes;
-    overflow: hidden;
-}
-
-Sidebar {
-    width: 40;
-    background: $panel;
-    transition: offset 500ms in_out_cubic;
-    layer: overlay;
-
-}
-
-Sidebar.-hidden {
-    offset-x: -100%;
-}"""
-
 DATA = {
     "foo": [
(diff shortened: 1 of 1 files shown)


rodrigogiraoserrao (approved):
Looks good to me!

src/textual/demo.py
```diff
@@ -313,7 +294,7 @@ def add_note(self, renderable: RenderableType) -> None:
         self.query_one(TextLog).write(renderable)
 
     def compose(self) -> ComposeResult:
-        example_css = "\n".join(Path(self.css_path[0]).read_text().splitlines()[:50])
+        example_css = "\n".join(Path(self.css_path[0]).read_text().splitlines())
```
rodrigogiraoserrao, line 297: The `"\n".join(...splitlines())` was only used because we wanted to be able to display full lines (otherwise slicing would be difficult).

If we are not slicing, and instead we display the full CSS, no need to split and join.

rodrigogiraoserrao (changes requested):

src/textual/demo.py
hartwork: @rodrigogiraoserrao I see, done.

rodrigogiraoserrao (approved):
Thanks.

src/textual/demo.py
rodrigogiraoserrao: Thanks for the quick turnaround.

hartwork:
@rodrigogiraoserrao rebased onto latest `main` for https://github.com/Textualize/textual/pull/2405/files#diff-50e0e891ca17c6f08e6e8ba0ed6b6b5240ede0b72a64c633a070e663f3fa9c0bL322 .

Issue #2413, opened 2023-04-28
Update the various `Tree` node events to add a reference to the `Tree`


davep:
I'm currently updating an application that was last built with Textual 0.11.0, and the application has multiple `Tree` widgets in the DOM. The `TreeNode` event handlers could see which tree was sending the messages thanks to checking the `sender` of the event.

Since then Textual has removed `sender` from events and so there's now no way to easily know which `Tree` is sending a `TreeNode`-oriented event. This can be fixed by doing one (or both) of the following:

- Add a `tree` property to `NodeCollapsed`, `NodeExpanded`, `NodeHighlighted` and `NodeSelected`
- Expose `TreeNode._tree` as a read-only property `TreeNode.tree`

davep:
Thinking some more, I think I prefer the latter option; the event is about a node, once you've got the node you can ask it what tree it's from.

PR #2415, opened 2023-04-28
Add `TreeNode.tree`

davep:
Currently, in the various TreeNode messages, and the handlers you'd write to handle them, there's no way to easily know *which* tree sent the message and so which tree the node belongs to.

This commit adds public access to the tree reference to the nodes, so that in an event handler the developer can check the tree involved in the event.

See #2413.


rodrigogiraoserrao (approved):
Let's go 🚀

PR #2423, opened 2023-04-28
faster focus within

willmcgugan:
The mechanism to refresh styles when the `focus-within` state changes was doing way more work than required. The root cause of slow tabs in Frogmouth.


willmcgugan:
#executivedecision

Issue #2382, opened 2023-04-26
Allow watch methods to be private


willmcgugan:
It should be possible for watch methods to be private.

If there is a `_watch_*` reactive, it should take precedence over the `watch_*` version.

This will allow us to hide them from the docs.

PR #2442, opened 2023-05-01
Add support for private watch methods

davep:
This change allows for private watch methods. By convention they start with an underscore. If a reactive or var has a private watch method, it will be used in preference to a public watch method.

With this change it becomes easier to have a private reactive/var whose watcher is also private. For example:

````python
    _counter = var(0)
    """This is a private counter, it won't appear in the docs."""

    ...

    def _watch__counter(self) -> None:
        """Watch _counter, but don't appear in the docs either."""
        ...
````

See #2382.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## Unreleased
+
+### Added
+
+- Watch methods can now optionally be private https://github.com/Textualize/textual/issues/2382
 
 ## [0.22.3] - 2023-04-29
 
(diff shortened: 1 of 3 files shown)


davep:
Note to self: Once merged `RadioSet.watch__selected` will benefit from a rename.

willmcgugan:
Just had a thought. If I subclass a widget with a private watcher, and implement my own watcher for an existing reactive var. I am almost certainly going to break the widget's functionality.

This is not true of method handlers which call the handler on the base class regardless.

I'm thinking that we should call both the public and the private handler.

davep:
@willmcgugan So when you specified:

> it will be used in preference to a public watch method

you meant "called first", not "called instead of"?

willmcgugan:
I meant instead of, but now I think it should call both.

davep:
Okay, cool, thanks for clarifying. Testing it out now and, yeah, it makes more sense.

willmcgugan (approved):

Issue #2420, opened 2023-04-28
!important should apply to border


willmcgugan:
It looks like !important works for border-left, border-bottom etc. But *not* an undecorated `border`.

I suspect this will be the same for other compound rules.

`!important` should make all "sub" rules important.

davep:
To help with testing this:

```python
from textual.app        import App, ComposeResult
from textual.containers import Grid
from textual.widgets    import Header, Footer, Static

class ImportantBorderApp( App[ None ] ):

    CSS = """
    Grid {
        grid-size: 3;
    }

    Static {
        content-align: center middle;
        border: round green !important;
        height: 1fr;
    }

    .subsequent {
        border: round red;
    }
    """

    def compose( self ) -> ComposeResult:
        yield Header()
        with Grid():
            for n in range( 9 ):
                yield Static(str(n), classes="subsequent")
        yield Footer()

if __name__ == "__main__":
    ImportantBorderApp().run()
```

All the borders should come out as green, but right now they come out as red. Sure enough though, with this version:

```python
from textual.app        import App, ComposeResult
from textual.containers import Grid
from textual.widgets    import Header, Footer, Static

class ImportantBorderApp( App[ None ] ):

    CSS = """
    Grid {
        grid-size: 3;
    }

    Static {
        content-align: center middle;
        border-top: round green !important;
        border-left: round green !important;
        border-bottom: round green !important;
        border-right: round green !important;
        height: 1fr;
    }

    .subsequent {
        border: round red;
    }
    """

    def compose( self ) -> ComposeResult:
        yield Header()
        with Grid():
            for n in range( 9 ):
                yield Static(str(n), classes="subsequent")
        yield Footer()

if __name__ == "__main__":
    ImportantBorderApp().run()
```

they come out as green.

PR #2444, opened 2023-05-01
Make `!important` apply to rules that have "sub-rules"

davep:
This PR fixes things so that `!important` applies to all sides/directions/etc, when it is applied to the whole rule. In other words this:

```python
border: solid red !important;
```

never came through as important, but this did:

```python
border-top: solid red !important;
```

This fixes that issue. The issue was found to exist for the following styles:

- `border`
- `outline`
- `align`
- `content-align`
- `overflow`
- `scrollbar-size`

See #2420.

Also includes a wee fix for #2446 too, which I noticed in the code as I was researching the fix for the above.

### DIFF
--- a/src/textual/css/_styles_builder.py
+++ b/src/textual/css/_styles_builder.py
@@ -486,6 +486,16 @@ class StylesBuilder:
         rules = self.styles._rules
         rules["border_top"] = rules["border_right"] = border
         rules["border_bottom"] = rules["border_left"] = border
+        # If border is marked as important...
+        if "border" in self.styles.important:
+            # ...border alone at this depth isn't really a thing, only
+            # border edges (that's to say, border gets expanded out to all 4
+            # edges). So we remove "border" as important and mark all the
+            # edges as important.
+            self.styles.important.remove("border")
+            self.styles.important.update(
+                f"border_{edge}" for edge in ("top", "left", "bottom", "right")
+            )
 
     def process_border_top(self, name: str, tokens: list[Token]) -> None:
         self._process_border_edge("top", name, tokens)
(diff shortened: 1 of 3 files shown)


davep:
In conversation during standup we've realised that this might actually affect more than just styles that have "sides". First job in the morning is to wander through the docs and try and compose one or more tests that cover `!important` and all the styles that have subsets of some description, and then apply a similar fix.

davep:
Having written tests for all of the above, the following styles also have the same bug:

- [ ] `align`
- [ ] `content-align`
- [ ] `overflow`
- [ ] `scrollbar-size`

src/textual/css/_styles_builder.py
```diff
@@ -252,6 +252,23 @@ def scalar_error():
...
+
+        A number of styles can be set with the 'prefix' of the style,
+        providing the values as a series of parameters; or they can be set
+        with specific suffixes. Think `border` vs `border-left`, etc. This
+        method is used to ensure that if the former is set, `!important` is
+        distributed amongst all the suffixes.
+        """
+        if prefix in self.styles.important:
+            self.styles.important.remove(prefix)
+            self.styles.important.update(f"{prefix}_{suffix}" for suffix in suffixes)
```
willmcgugan, line 270:
```suggestion
        self.styles.important.discard(prefix)
        self.styles.important.update(f"{prefix}_{suffix}" for suffix in suffixes)
```

The discard will save a lookup. I think `prefix` is always going to be in there anyway?

willmcgugan (approved):
Just one suggestion

src/textual/css/_styles_builder.py
davep: The check is the check that the style is `!important`, otherwise we don't want all the suffixes marked as important.

PR #2445, opened 2023-05-01
Invalid pseudo selectors

willmcgugan:
Error on bad pseudo selectors.

It catches bad pseudo selectors at the tokenizer level. I could had changed the regex, but this way we can generate a helpful error.


davep (approved):

davep:
Forgot to say: I think this deserves a mention in the CHANGELOG.

willmcgugan:
I'll add it to main

Issue #2398, opened 2023-04-27
Decorator approach to event handling


willmcgugan:
Add a decorator approach to handling messages, in addition to the naming convention.

Implement a function called `on`, importable from the root. This would take the message type, and an optional selector.

The message system would check for decorated handlers before the naming convention.

Note: there could be more than one handler associated with a method.

```python
from textual import on

class MyApp:

    @on(Button.Pressed, "#quit")
    def quit_app(self, event:Button.Pressed):
        ...
```

I think Textual should look for the decorated handler first, then the naming convention.

Note that we have a `sender` attribute, which for the most part could be used to match against the selector. Consequently I think we need a method on the message itself, so that any given message hand override the matching behaviour.

willmcgugan:
Updated to use `@on` which compliments the current naming convention, and reads better.

willmcgugan:
https://github.com/Textualize/textual/pull/2453

PR #2453, opened 2023-05-02
On decorator

willmcgugan:
Implement the @on decorator

Fixes https://github.com/Textualize/textual/issues/2398

### DIFF
--- a/docs/examples/events/on_decorator.css
+++ b/docs/examples/events/on_decorator.css
@@ -1,0 +1,8 @@
+Screen {
+    align: center middle;
+    layout: horizontal;
+}
+
+Button {
+    margin:2  4;
+}
--- a/docs/examples/events/on_decorator.css
+++ b/docs/examples/events/on_decorator.css
@@ -4,5 +4,5 @@ Screen {
 }
 
 Button {
-    margin:2  4;
+    margin: 2 4;
 }
--- a/docs/guide/events.md
+++ b/docs/guide/events.md
@@ -155,6 +155,65 @@ Textual uses the following scheme to map messages classes on to a Python method.
 --8<-- "docs/images/events/naming.excalidraw.svg"
 </div>
 
+### On decorator
+
+In addition to the naming convention, message handlers may be created with the [on][textual.on] decorator, which turns a method in to a handler for the given message or event.
+
+For instance, the two methods declared below are equivalent:
+
+```python
+@on(Button.Pressed)
+def handle_button_pressed(self):
+    ...
+
+def on_button_pressed(self):
+    ...
+```
+
+While this allows you to name your method handlers anything you want, the main advantage of the decorator approach over the naming convention is that you can specify *which* widget(s) you want to handle messages for.
+
+Let's first explore where this can be useful.
+In the following example we have three buttons, each of which does something different; one plays the bell, one toggles dark mode, and the other quits the app.
+
+=== "on_decorator01.py"
+
+    ```python title="on_decorator01.py"
+    --8<-- "docs/examples/events/on_decorator01.py"
+    ```
+
+=== "Output"
+
+    ```{.textual path="docs/examples/events/on_decorator01.py"}
+    ```
+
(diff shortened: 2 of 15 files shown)


docs/guide/events.md
```diff
@@ -155,6 +155,66 @@ Textual uses the following scheme to map messages classes on to a Python method.
 --8<-- "docs/images/events/naming.excalidraw.svg"
 </div>
 
+### On decorator
+
+In addition to the naming convention, message handlers may be created with the [on][textual.on] decorator, which turns a method in to a handler for the given message or event.
```
davep, line 160:
```suggestion
In addition to the naming convention, message handlers may be created with the [`on`][textual.on] decorator, which turns a method into a handler for the given message or event.
```

docs/guide/events.md
```diff
@@ -155,6 +155,66 @@ Textual uses the following scheme to map messages classes on to a Python method.
...
+    3. Matches the button with an id of "quit"
+
+=== "Output"
+
+    ```{.textual path="docs/examples/events/on_decorator02.py"}
+    ```
+
+While there are a few more lines of code, it is clearer what will happen when you click any given button.
+
+Note that the decorator requires that the message class has a `control` attribute which should the widget associated with the message.
```
davep, line 214:
```suggestion
Note that the decorator requires that the message class has a `control` attribute which should be the widget associated with the message.
```

tests/test_on.py
davep, line 1: I might be tempted to add a test here that `MyApp.my_custom_handler.__doc__` is what is expected, but that might be taking things too far.

davep (approved):
Couple of typo things and an extreme test suggestion but otherwise looking good! Looking forward to using this.

docs/examples/events/on_decorator.css
```diff
@@ -0,0 +1,8 @@
+Screen {
+    align: center middle;
+    layout: horizontal;
+}
+
+Button {
+    margin:2  4;
```
rodrigogiraoserrao, line 7:
```suggestion
    margin: 2 4;
```

docs/guide/events.md
```diff
@@ -155,6 +155,66 @@ Textual uses the following scheme to map messages classes on to a Python method.
...
+This is true of the builtin controls, but you may need to add `control` to any [custom messages](#custom-messages) you write.
```
rodrigogiraoserrao, line 215: The statement is "the **message** must have an attribute [...]", so we say it is true of the builtin _messages_:

```suggestion
This is true of the builtin messages, but you may need to add `control` to any [custom messages](#custom-messages) you write.
```

docs/guide/events.md
rodrigogiraoserrao, line 1: What if there is a generic handler `on_button_pressed` and a handler with the decorator `@on`?
What if multiple `@on` methods match?

src/textual/_on.py
```diff
@@ -0,0 +1,56 @@
...
+        ```python
+        @on(Button.Pressed, "#quit")
+        def quit_button(self) -> None:
+            self.app.quit()
+        ```
+
+    Args:
+        message_type: The message type (i.e. the class).
+        selector: An optional [selector](/guide/CSS#selectors). If supplied, the handler will only be called if `selector`
+            matches the sender of the message.
```
rodrigogiraoserrao, line 31: Might be a good idea to note that this sender is checked in the attribute `control`.

src/textual/_on.py
```diff
@@ -0,0 +1,56 @@
...
+
+from .css.parse import parse_selectors
+from .css.tokenizer import TokenError
+from .message import Message
+
+DecoratedType = TypeVar("DecoratedType")
+
+
+class OnDecoratorError(Exception):
+    """The selector in the 'on' decorator failed to parse."""
```
rodrigogiraoserrao, line 13: This error is also used when other things fail, so maybe the docstring should be a bit more generic?

rodrigogiraoserrao (approved):
A couple of minor suggestions.

PR #2454, opened 2023-05-02
Attempt to narrow CI/test trigger to just code changes

davep:
Looking at [the GitHib docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request) I *think* this might be the approach to take. Consider this an experimental change that will very likely get rolled back.

The thinking here is that we *don't* run testing on a PR unless one of the following things have happened in that PR:

- A `py` file has been modified.
- A `pyi` file has been modified.
- A `css` file has been modified.
- The snapshot data has been modified.
- The lock file has been modified.
- The `Makefile` has been modified.

See #2404

I suggest we give this a spin and see if it works out; if not it should be very simple to back this out.

---
One thing to keep in mind and keep an eye out for (it should be obvious): the ability to merge is normally down to having an approval and passing tests; so is having no test fail the same as passing tests?

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -1,6 +1,11 @@
 name: Test Textual module
 
-on: [pull_request]
+on:
+  pull_request:
+    paths:
+      - '**.py'
+      - '**.pyi'
+      - '**.css'
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


davep:
Right. That didn't attempt to run tests at all, which is what I expected. Next up I'm going to modify a `py` file and see if tests kick in.

Issue #2396, opened 2023-04-27
Switching focus with a(n unreasonably) large number of widgets can be slow


davep:
While testing out something unrelated, I happened to create a little application that creates a silly number of focusable widgets:

```python
from textual.app        import App, ComposeResult
from textual.binding    import Binding
from textual.containers import Horizontal, VerticalScroll
from textual.widgets    import Header, Footer, TabbedContent, TabPane, Tabs, Input

class StandardTabs( VerticalScroll ):

    def compose( self ) -> ComposeResult:
        with TabbedContent():
            for n in range( 10 ):
                with TabPane( f"Tab {n}" ):
                    for m in range(30):
                        yield Input( placeholder=f"{n:2}:{m:2} - Here's an input to make this more busy" )

class SelfSwitchTabs( StandardTabs ):

    BINDINGS = [
        Binding( "a", "previous", "Previous" ),
        Binding( "d", "next", "Next" ),
    ]

    def action_previous(self) -> None:
        self.query_one(Tabs).action_previous_tab()

    def action_next(self) -> None:
        self.query_one(Tabs).action_next_tab()

class TabbedContentTesterApp( App[ None ] ):

    CSS = """
    StandardTabs:focus-within {
        background: $panel;
    }
    """

    def compose( self ) -> ComposeResult:
        yield Header()
        with Horizontal():
            yield StandardTabs()
            yield SelfSwitchTabs()
        yield Footer()

if __name__ == "__main__":
    TabbedContentTesterApp().run()
```

When run, note that <kbd>Tab</kbd>bing between the `Input`s can be quite laggy; most likely because of how the focus chain is calculated each time.

Look into improving the speed here.

PR #2460, opened 2023-05-02
optimize focus

willmcgugan:
Optimize style changes when focus changes.

the `focus-within` selector would cause every widget underneath it to refresh its CSS, which was slow.

This change detects which widgets will change according to focus-within, and does it earlier on in the callstack.

Also updates some of the previews to have non-focusable containers.

Fixes https://github.com/Textualize/textual/issues/2396


davep (approved):

src/textual/screen.py
```diff
@@ -404,6 +405,32 @@ def _reset_focus(
...
+                    break
+        if blurred is not None:
+            for widget in reversed(blurred.ancestors_with_self):
+                if widget._has_focus_within:
+                    widgets.update(widget.walk_children(with_self=True))
+                    break
+        if widgets:
+            self.app.stylesheet.update_nodes(
+                [widget for widget in widgets if widget._has_focus_within], animate=True
+            )
```
rodrigogiraoserrao, line 432: Can the first two `if`s evaluate to `True` at the same time?

Also, the third `if` goes over `widgets` and checks `widget._has_focus_within`, which seems to overlap a bit with the inner `if widget._has_focus_within` in each of the first two `if`s.

If this is an optimisation, isn't there a way to short-circuit the conditional in the bottom list comprehension by doing some more work in each of the first two `if`s?
I'm not familiarised with the focus chain so I can't really propose something more specific.

rodrigogiraoserrao (approved):

src/textual/screen.py
willmcgugan: > Can the first two `if`s evaluate to `True` at the same time?

Yes, typically when a widget is focused another is blurred.

> 
> Also, the third `if` goes over `widgets` and checks `widget._has_focus_within`, which seems to overlap a bit with the inner `if widget._has_focus_within` in each of the first two `if`s.
> 
> If this is an optimisation, isn't there a way to short-circuit the conditional in the bottom list comprehension by doing some more work in each of the first two `if`s? I'm not familiarised with the focus chain so I can't really propose something more specific.

Focus-within is a weird one.  Focusing a widget can modify the style of an ancestor. Which is why it has to back-track to find a parent with focus-within, then look at the children. It's already a massive win. Don't think there are any significant wins remaning.

Issue #2462, opened 2023-05-03
Add the ability to remove `Tree` nodes


davep:
In part motivated by #2448, although I think the request has popped up before: it would be useful if `Tree` had a `delete_node` method to go along with `add_node`.

tscholak:
Is there a workaround? Afaik, children of a node are currently immutable. Is it therefore recommended that the tree is cleared and reconstructed whenever one needs to remove nodes?

davep:
@tscholak Short of messing around in the internals of the `Tree` (as it stands until this issue is handled), that's probably the way to do it right now, I'd imagine.

Issue #2456, opened 2023-05-02
DirectoryTree should load a directory in a Worker


willmcgugan:
Getting the current directory is a blocking operation.

It is typically virtually instant, but it can be slow for larger directories. We should load the directory within a thread worker to avoid suspending the UI.

davep:
Raised the above as #2535 to highlight this side-issue; this being the main issue and #2535 needing to be done to make sense of this.

Issue #2448, opened 2023-05-01
Add a way to load a new directory in DirectoryTree


willmcgugan:
- We need a way to change the directory root (and refresh) a DirectoryTree. 
- Related: we need a way to refresh the current directory.
- We have a `load_directory` method, which I think should be private.

davep:
This is pretty much a duplicate of #2056

davep:
Making a note of part of a conversation yesterday: for the moment the "refresh" will be for the `DirectoryTree` as a whole. Ideally we'd like to add the ability to refresh an individual directory within the tree. To do that, however, `Tree` itself will need to be updated to allow for a node's children to be removed (and, indeed, it would be useful if `Tree` supported node removal).

PR #2463, opened 2023-05-03
Allow changing the "root" of a `DirectoryTree`

davep:
This PR updates `DirectoryTree` to provide a `path` reactive that, when set, will change the "root" of the tree. See #2448 for context.

Other related changes include:

- `load_directory` has now been made private.
- From a public interface point of view `DirectoryTree` is now `Path`-first; that is: it'll accept `str` but will always work with and expose a `Path` value.
- Added a `node` attribute to the `FileSelected` message so that it matches the recent changes to `Tree` messages, and so that anyone using `DirectoryTree` can know which tree sent the message.
- `FileSelected.path` is now typed as `Path`.
- Added a `reload` method.

Right now the `reload` is for the content of the `DirectoryTree` as a whole. Ideally, at some point, it would be useful to allow for the reloading of a specific directory's contents within the displayed hierarchy; to achieve this really needs that the ability to delete nodes is added to `Tree` itself (raised #2462 to this end).

A logical follow-on issue from this PR is #2456.


src/textual/widgets/_directory_tree.py
```diff
@@ -91,18 +94,58 @@ def __init__(
...
+    def watch_path(self, new_path: str | Path) -> None:
+        """Watch for changes to the `path` of the directory tree.
+
+        Args:
+            new_path: The new path being set.
+
+        If the path is changed the directory tree will be repopulated using
+        the new value as the root.
+        """
+        self.path = Path(new_path)
```
willmcgugan, line 145: I'm confused about this. If we have a validator, won't this always be a `Path` ?

And if we're watching path, do we need to set path again?

src/textual/widgets/_directory_tree.py
```diff
@@ -91,18 +94,58 @@ def __init__(
...
+            path: The path to validate.
+
+        Returns:
+            The validated Path value.
+
+        Note:
+            The result will always be a Python `Path` object, regardless of
+            the value given.
+        """
+        return Path(path)
```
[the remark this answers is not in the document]
davep: Almost. You do get a fresh object.  I suppose this could be modified to be:

```python
return path if isinstance(path, Path) else Path(path)
```

but that didn't seem to save much of anything.

src/textual/widgets/_directory_tree.py
davep: Nah, good catch, looks like a complete thinko on my part. 100% no reason ti assign it again.

CHANGELOG.md
```diff
@@ -21,11 +21,17 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
...
+- Unknown psuedo-selectors will now raise a tokenizer error (previously they were silently ignored) https://github.com/Textualize/textual/pull/2445
+- Breaking change: `DirectoryTree.FileSelected.path` is now always a `Path` https://github.com/Textualize/textual/issues/2448
+- Breaking change: `Directorytree.load_directory` renamed to `Directorytree._load_directory` https://github.com/Textualize/textual/issues/2448
 - Unknown pseudo-selectors will now raise a tokenizer error (previously they were silently ignored) https://github.com/Textualize/textual/pull/2445
 
 ### Added
 
 - Watch methods can now optionally be private https://github.com/Textualize/textual/issues/2382
+- Added `DirectoryTree.path` reactive attribute https://github.com/Textualize/textual/issues/2448
+- Added `DirectoryTree.FileSelected.node`
```
rodrigogiraoserrao, line 33: Missing issue/PR.

CHANGELOG.md
```diff
@@ -21,11 +21,17 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
...
+- Added `DirectoryTree.reload` https://github.com/Textualize/textual/issues/2448
```
rodrigogiraoserrao, line 34: I usually group additions like this in a single list item to save some vertical space.
I've noticed you usually put one addition per list item. Why? I'm considering changing the way I list things in the changelog.

src/textual/widgets/_directory_tree.py
```diff
@@ -10,33 +9,27 @@
...
-        disabled: Whether the directory tree is disabled or not.
-    """
+    """A Tree widget that presents files and directories."""
 
     COMPONENT_CLASSES: ClassVar[set[str]] = {
-        "directory-tree--folder",
-        "directory-tree--file",
         "directory-tree--extension",
+        "directory-tree--file",
+        "directory-tree--folder",
```
rodrigogiraoserrao, line 32: ❤️

src/textual/widgets/_directory_tree.py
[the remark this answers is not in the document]
rodrigogiraoserrao: I think `return Path(path)` is a good idea to make sure that we don't get a path that the user mutates later.

src/textual/widgets/_directory_tree.py
```diff
@@ -119,6 +158,16 @@ def process_label(self, label: TextType):
         return first_line
 
     def render_label(self, node: TreeNode[DirEntry], base_style: Style, style: Style):
```
rodrigogiraoserrao, line 160: Type annotation is missing here, to go with the `Returns` in the docstring.

rodrigogiraoserrao (approved):
I left a question and a couple of comments, but after you resolve them this looks good to be merged!

CHANGELOG.md
davep: On purpose as it wasn't really an issue/PR-related thing, more of a drive-by change while I was looking at that code. But I guess linking to this PR won't hurt.

CHANGELOG.md
davep: For me, if all of the above were in one long single item, I'd be more prone to not pick up on the impact of the changes. Given the above changes aren't really related (so, it's not like "added `foo` support to `Bar`, `Baz` and `Wibble`) I feel it makes more sense and highlights they're different things with different impacts.

rodrigogiraoserrao: I get your point. Thanks.

src/textual/widgets/_directory_tree.py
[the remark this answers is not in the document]
davep: Fairly sure `Path` is immutable that that shouldn't be an issue.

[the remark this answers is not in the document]
willmcgugan: Yeah, I checked. Paths are indeed immutable.

src/textual/widgets/_directory_tree.py
willmcgugan: Yep. Looks like it slipped through a previous review. @davep would you mind?

willmcgugan (approved):

PR #2464, opened 2023-05-03
Scroll cursor into view.

### LINKED ISSUE #2459: DataTable - When cursor_coordinate updates, we should scroll_cursor_visible
When the cursor moves out of the screen, we should scroll the cursor into visibility.

It seems like there is already code to do this, but because of #2458 the scrolling doesn't work.

rodrigogiraoserrao:
This will fix #2459.
Scrolls the cursor into view when the cursor coordinate is changed programmatically.


willmcgugan (approved):
A test couldn't hurt.

PR #2469, opened 2023-05-03
Catch OSError when source code is not available

### LINKED ISSUE #2468: OSError when the file system is not available
Textual (specifically App and DOMNode) assumes that the file containing source code for Widget classes is available and raises an `OSError` when this is not the case. This information is used to render the location of CSS styles.

I'm trying to get a Textual app running in an environment in which `getpath` doesn't work (Pyodide) because there's no file system in WebAssembly. Obviously, this is a pretty extreme environment, but I think the bug remains reasonable to address.

The calls to `getfile` are already wrapped in a try/except for `TypeError` with fallback behavior, so I propose that `OSError` be handled here, too. I *can* provide a standalone test case, though it's a little involved, so I'd rather just open a PR shortly, if you'll consider it.

image

williballenthin:
closes #2468

with this fix, my use case works a bit more:
image

There's still an outstanding call to `getfile` but since I haven't hit this code path yet and therefore can't verify a fix, I haven't attempted it.
https://github.com/Textualize/textual/blob/c87a2b1b3731140a461b4f34bbd7b1c6de8d7e37/src/textual/_path.py#L30

- n/a Docstrings on all new or modified functions / classes 
- n/a Updated documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Fixed `!important` not applying to `overflow` https://github.com/Textualize/textual/issues/2420
 - Fixed `!important` not applying to `scrollbar-size` https://github.com/Textualize/textual/issues/2420
 - Fixed `outline-right` not being recognised https://github.com/Textualize/textual/issues/2446
+- Fixed OSError when a file system is not available https://github.com/Textualize/textual/issues/2468
 
 ### Changed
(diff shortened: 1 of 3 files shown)


willmcgugan:
Seems reasonable.

Do you know if it is possible to detect a FS free environment? i.e. an ENV var?

willmcgugan (approved):

willmcgugan (approved):

PR #2477, opened 2023-05-04
Datatable scrolling faq

rodrigogiraoserrao:
Write more concisely.


questions/datatable-doesnt-scroll.question.md
```diff
@@ -5,14 +5,13 @@ alt_titles:
   - "Datatable cursor goes off screen and doesn't scroll."
 ---
 
-If it looks like the scrolling in your `DataTable` is broken, it may be because your `DataTable` does not have its height set, which means it is using the default value of `height: auto`.
-In turn, this means that the `DataTable` itself does not have a scrollbar and, hence, it cannot scroll.
-
-If it looks like your `DataTable` has scrollbars, those might belong to the container(s) of the `DataTable`, which in turn makes it look like the scrolling of the `DataTable` is broken.
+If scrolling in your `DataTable` is _apparently_ broken, it may be because your `DataTable` is using the default value of `height: auto`.
+This means that the `DataTable` itself does not have a scrollbar and it cannot scroll itself.
+The scrollbar(s) you see belong to the container(s) of the `DataTable`.
```
willmcgugan, line 10:
```suggestion
This means that the table will be sized to fit its rows without scrolling, which may cause the *container* (typically the screen) to scroll.
If you would like the table itself to scroll, set the height to something other than `auto`. 
For instance, `100%` which will be the height of the container.
```

willmcgugan (changes requested):
Let's either lose the example entirely, or have a have an example simply shows a correctly sizing table.

The FAQ should contain fairly concise answers. It's not intended to be didactic.

willmcgugan (commented):
We probably still want to recommend  setting height `100%` as that is a quick fix that would work for most use cases.

willmcgugan (approved):
Thanks for sticking with it.

PR #2479, opened 2023-05-04
Add `DataTable.move_cursor`

### LINKED ISSUE #2471: Moving cursor sideways in the DataTable should animate.
With #2464, I broke the smooth animation you get when moving the cursor sideways in a `DataTable`.
This behaviour should be restored.

### LINKED ISSUE #2472: Add method `move_cursor` to `DataTable`
The method `move_cursor` moves the cursor to a given position and it takes an optional parameter `animate=False` to animate the change.

rodrigogiraoserrao:
This will add the method `DataTable.move_cursor`.
This also fixes the animation you get when going left/right in a data table.

- [x] Updated documentation

Closes #2471.
Closes #2472.

### DIFF
--- a/src/textual/widgets/_data_table.py
+++ b/src/textual/widgets/_data_table.py
@@ -955,7 +955,9 @@ class DataTable(ScrollView, Generic[CellType], can_focus=True):
             elif self.cursor_type == "column":
                 self.refresh_column(old_coordinate.column)
                 self._highlight_column(new_coordinate.column)
-            self._scroll_cursor_into_view()
+            # If the coordinate was changed via `move_cursor`, give priority to its
+            # scrolling because it may be animated.
+            self.call_next(self._scroll_cursor_into_view)
 
     def move_cursor(
         self,
--- a/src/textual/widgets/_data_table.py
+++ b/src/textual/widgets/_data_table.py
@@ -2113,7 +2113,6 @@ class DataTable(ScrollView, Generic[CellType], can_focus=True):
             self.cursor_coordinate = Coordinate(
                 row_index + rows_to_scroll - 1, column_index
             )
-            self._scroll_cursor_into_view()
         else:
             super().action_page_down()
 
@@ -2137,7 +2136,6 @@ class DataTable(ScrollView, Generic[CellType], can_focus=True):
             self.cursor_coordinate = Coordinate(
                 row_index - rows_to_scroll + 1, column_index
             )
-            self._scroll_cursor_into_view()
         else:
             super().action_page_up()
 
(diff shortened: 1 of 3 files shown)


src/textual/widgets/_data_table.py
```diff
@@ -950,7 +955,63 @@ def watch_cursor_coordinate(
             elif self.cursor_type == "column":
                 self.refresh_column(old_coordinate.column)
                 self._highlight_column(new_coordinate.column)
-            self._scroll_cursor_into_view()
+            # If the coordinate was changed via `move_cursor`, give priority to its
+            # scrolling because it may be animated.
+            self.call_next(self._scroll_cursor_into_view)
+
+    def move_cursor(
```
darrenburns, line 962: Not sure about the API here personally, because the parameters are kind of dependent on each other. (e.g. if you pass `coordinate` and `row`, things get weird). Do we need the `coordinate` parameter?

rodrigogiraoserrao: If you pass a coordinate and a row, you get an error.

I opted for this signature because if the cursor is a coordinate (`DataTable.cursor_coordinate`) it makes sense to accept the same type; then I added `row` and `column` as convenience parameters, which work similarly to the parameters `after` and `before` in methods like `Widget.mount` and `Widget.move_child`

willmcgugan: I concur with Darren here. The signature can be ambiguous. Given there can be row and column cursors, where only a single value makes sense, I think we should drop the `coordinate`.

src/textual/widgets/_data_table.py
```diff
@@ -950,7 +955,63 @@ def watch_cursor_coordinate(
...
+            row: The new row to move the cursor to.
+            column: The new column to move the cursor to.
+            animate: Whether to animate the change of coordinates.
+
+        Raises:
+            RuntimeError: If the parameter `coordinate` is specified together with any
+                of `row` or `column`, or if no new position is specified.
+        """
+        if coordinate is not None and (row is not None or column is not None):
+            raise RuntimeError("Can't specify `coordinate` and `row`/`column`.")
```
willmcgugan, line 1000: I gather this is trying to resolve ambiguous combination of parameters. But it would be better to not have to catch these at runtime.

willmcgugan (changes requested):

src/textual/widgets/_data_table.py
rodrigogiraoserrao: Even when the cursor type is row/column, the reactive attribute `cursor_coordinate` is a full coordinate with X _and_ Y coordinates, so this is aligned with that.
But if you all think it is a bad idea 🤷

willmcgugan (approved):

PR #2490, opened 2023-05-04
Add control to widget messages.

### LINKED ISSUE #2486: Fix documentation for OptionList
The documentation for `OptionList` looks odd, specifically the sub-subsection "Common message properties". [Here](https://textual.textualize.io/widgets/option_list/#common-message-properties).
Not only the TOC headers but also the contents of those H4 headers.

I think we're trying to use some `mkdocstrings` feature but we failed. It should be fairly straightforward to fix.

### LINKED ISSUE #2489: All widget events should have a control property
Following up on #2478, all widget events should have a `control` property for use with the decorator `on`.

rodrigogiraoserrao:
This adds the property `control` to all the widget messages.

Closes #2489.
Closes #2486.


docs/widgets/option_list.md
rodrigogiraoserrao, line 1: This may look like an unrelated change, but I needed to update this documentation to add `control`.
But this was the source of the issue with #2486 so I just went ahead and took care of that as well.

willmcgugan (approved):

src/textual/widgets/_directory_tree.py
```diff
@@ -64,19 +64,32 @@ class FileSelected(Message, bubble=True):
         `DirectoryTree` or in a parent widget in the DOM.
         """
 
-        def __init__(self, node: TreeNode[DirEntry], path: Path) -> None:
+        def __init__(
+            self, tree: DirectoryTree, node: TreeNode[DirEntry], path: Path
```
davep, line 68: I realise this is a bit late now, but this will have broken `DirectoryTree` as it didn't update the use of the message later on in the code.

src/textual/widgets/_directory_tree.py
```diff
@@ -64,19 +64,32 @@ class FileSelected(Message, bubble=True):
...
             """The path of the file that was selected."""
 
+        @property
+        def control(self) -> DirectoryTree:
+            """The `DirectoryTree` that had a file selected.
+
+            This is an alias for [`FileSelected.tree`][textual.widgets.DirectoryTree.FileSelected.tree]
+            which is used by the [`on`][textual.on] decorator.
+            """
+            return self.tree
```
davep, line 91: This could be done without the `Tree` reference by using:

```python
return self.node.tree
```

src/textual/widgets/_directory_tree.py
rodrigogiraoserrao: Thanks. Sorry for the mess-up D:

PR #2497, opened 2023-05-05
Fix TabbedContent __init__ signature.

rodrigogiraoserrao:
This fixes the signature of `TabbedContent.__init__` so that it is initialised like any other widget.

Of course, this PR only makes sense under the assumption that these `kwargs` were missing due to our oversight in implementing and reviewing `TabbedContent`.


rodrigogiraoserrao:
Once this has been merged, we should uncomment the middle sub-test in `tests/test_on.py::test_on_arbitrary_attributes` in 83b0ae7

willmcgugan (approved):

PR #2498, opened 2023-05-05
Extend `@on` decorator to filter arbitrary attributes

### LINKED ISSUE #2484: Extend decorator `on` to specify non-`control` attributes to match.
By default, the decorator `on` accepts a selector that applies to the attribute `control` of the message.

We want `on` to grow `**kwargs` so that we can specify arbitrary selectors for arbitrary attributes of the messages we handle.

E.g., to handle the message `Tabs.TabActivated`, we may want `@on(Tabs.TabActivated, tab="#first-tab")` to handle the activation of the tab `#first-tab`.

_**EDIT**_: instead of matching arbitrary attributes of the message, we can only try to match attributes that have been whitelisted in the message's classvar `ON_MATCHABLE_ATTRIBUTES`.

rodrigogiraoserrao:
The decorator `@on` now accepts arbitrary `**kwargs` that determine extra attributes for which selectors are provided.

This will close #2484.

### DIFF
--- a/tests/test_on.py
+++ b/tests/test_on.py
@@ -5,7 +5,7 @@ from textual._on import OnDecoratorError
 from textual.app import App, ComposeResult
 from textual.message import Message
 from textual.widget import Widget
-from textual.widgets import Button
+from textual.widgets import Button, TabbedContent, TabPane
 
 
 async def test_on_button_pressed() -> None:
@@ -102,3 +102,63 @@ def test_on_no_control() -> None:
         @on(CustomMessage, "#foo")
         def foo():
             pass
+
+
+async def test_on_arbitrary_attributes() -> None:
+    log: list[str] = []
+
+    class OnArbitraryAttributesApp(App[None]):
+        def compose(self) -> ComposeResult:
+            with TabbedContent():
+                yield TabPane("One", id="one")
+                yield TabPane("Two", id="two")
+                yield TabPane("Three", id="three")
+
+        def on_mount(self) -> None:
+            self.query_one(TabbedContent).add_class("tabs")
+
+        @on(TabbedContent.TabActivated, tab="#one")
+        def one(self) -> None:
+            log.append("one")
+
+        # @on(TabbedContent.TabActivated, ".tabs", tab="#two")
+        # def two(self) -> None:
+        #     log.append("two")
+
+        @on(TabbedContent.TabActivated, tabbed_content=".tabs", tab="#three")
+        def three(self) -> None:
+            log.append("three")
+
+    app = OnArbitraryAttributesApp()
+    async with app.run_test() as pilot:
+        await pilot.press("tab", "right", "right")
+
+    assert log == ["one", "three"]
+
+
+async def test_on_extra_attributes() -> None:
+    """@on shouldn't trigger when there are selectors for non-existing attributes."""
+    log: list[str] = []
+
+    class OnExtraAttributes(App[None]):
+        def compose(self) -> ComposeResult:
+            with TabbedContent():
+                yield TabPane("One", id="one")
+
+        @on(TabbedContent.TabActivated, tab="#one", hello="#world")
+        def one_hello_world(self) -> None:
+            log.append("one hello world")
+
+        @on(TabbedContent.TabActivated, hello="#world")
+        def hello_world(self) -> None:
+            log.append("hello world")
+
+        @on(TabbedContent.TabActivated, tab="#one")
+        def one(self) -> None:
+            log.append("control")
+
+    app = OnExtraAttributes()
+    async with app.run_test():
+        pass
+
+    assert log == ["control"]


rodrigogiraoserrao:
Tests will pass as soon as I merge #2490.

docs/guide/events.md
rodrigogiraoserrao, line 1: Still trying to strike a balance between my personal verbose style and the intended style for the docs.
Let me know how this looks.

docs/examples/events/on_decorator03.py
```diff
@@ -0,0 +1,38 @@
...
+    def toggle_dark(self):
+        """Called when the 'toggle dark' button is pressed."""
+        self.dark = not self.dark
+
+    @on(Button.Pressed, button="#quit")  # (3)!
+    def quit(self):
+        """Called when the quit button is pressed."""
+        self.exit()
+
+    @on(Button.Pressed, super_cool_button="#button")  # (4)!
```
willmcgugan, line 30: We can't have examples that are simply dead code. Examples should work out of the box. No "exercise left to the reader".

docs/examples/events/on_decorator03.py
```diff
@@ -0,0 +1,38 @@
...
+class OnDecoratorApp(App):
+    CSS_PATH = "on_decorator.css"
+
+    def compose(self) -> ComposeResult:
+        """Three buttons."""
+        yield Button("Bell", id="bell")
+        yield Button("Toggle dark", classes="toggle dark")
+        yield Button("Quit", id="quit")
+
+    @on(Button.Pressed, button="#bell")  # (1)!
```
willmcgugan, line 15: I don't think we should repeat annotation. Just annotate anything changed / added.

src/textual/_on.py
```diff
@@ -13,47 +14,70 @@ class OnDecoratorError(Exception):
...
-
     """
 
 
 def on(
-    message_type: type[Message], selector: str | None = None
+    message_type: type[Message], selector: str | None = None, **kwargs: str
 ) -> Callable[[DecoratedType], DecoratedType]:
-    """Decorator to declare method is a message handler.
+    """Decorator to declare that method is a message handler.
```
willmcgugan, line 23:
```suggestion
    """Decorator to declare that the method is a message handler.
```

I know that was missing in the original...

docs/guide/events.md
```diff
@@ -218,10 +218,31 @@ Messages from builtin controls will have this attribute, but you may need to add
...
 !!! note
 
-    If multiple decorated handlers match the `control`, then they will *all* be called in the order they are defined.
+    If multiple decorated handlers match the message, then they will *all* be called in the order they are defined.
 
     The naming convention handler will be called *after* any decorated handlers.
 
+#### Applying CSS selectors to arbitrary attributes
+
+The decorator `on` also accepts `**kwargs` that can be used to specify CSS selectors for the specified attributes of the message.
```
willmcgugan, line 227:
```suggestion
The `on` decorator also accepts selectors as keyword arguments that may be used to match other attributes in a Message.
```

src/textual/message_pump.py
```diff
@@ -563,14 +563,18 @@ def _get_dispatch_methods(
             decorated_handlers = cls.__dict__.get("_decorated_handlers")
             if decorated_handlers is not None:
                 handlers = decorated_handlers.get(type(message), [])
-                for method, selector in handlers:
-                    if selector is None:
+                _sentinel = object()
```
willmcgugan, line 566: Why do we need a sentinal? Why not `None`?

If we do need it, let's move it outside the loop.

willmcgugan (changes requested):
The dead code in the example is a blocker. Either we come up with a working example, or we don't bother with an example for this.

docs/examples/events/on_decorator03.py
rodrigogiraoserrao: This is showing what happens if you provide `kwargs` for attributes that don’t exist. The example works out of the box, it’s just that this method will never be called.

willmcgugan: Yes, but it doesn't do anything. Examples should show the user what to do, not what *not* to do.

src/textual/message_pump.py
rodrigogiraoserrao: The attribute might be there but it may be set to `None`, in which case I figured we’d like to error.

should I move the sentinel out and add a custom error if the attribute doesn’t provide a valid widget?

willmcgugan: The decorator does this for `control`, at import time. We should do this for the other attributes as well. In that case it is probably safe to assume the presence of the attribute, and that it is a widget.

rodrigogiraoserrao:
Changed the API so that now the `**kwargs` can only match attributes that have been whitelisted in the message's `ON_MATCHABLE_ATTRIBUTES` classvar.

rodrigogiraoserrao:
Changed the API so that now the `**kwargs` can only match attributes that have been whitelisted in the message's `ON_MATCHABLE_ATTRIBUTES` classvar.

src/textual/_on.py
```diff
@@ -13,47 +14,76 @@ class OnDecoratorError(Exception):
...
+
+
 def on(
-    message_type: type[Message], selector: str | None = None
+    message_type: type[Message], selector: str | None = None, **kwargs: str
 ) -> Callable[[DecoratedType], DecoratedType]:
-    """Decorator to declare method is a message handler.
+    """Decorator to declare that the method is a message handler.
+
+    The decorator can take a CSS selector that is applied to the attribute `control`
```
willmcgugan, line 29:
```suggestion
    The decorator accepts an optional CSS selector that is matched against a widget exposed by a `control` attribute on the message.
```

It's not "applied to" ...

src/textual/_on.py
```diff
@@ -13,47 +14,76 @@ class OnDecoratorError(Exception):
...
-
+    selectors: dict[str, str] = {}
     if selector is not None:
+        selectors["control"] = selector
+    if kwargs:
+        selectors.update(kwargs)
+
+    parsed_selectors: dict[str, tuple[SelectorSet, ...]] = {}
+    for attribute, css_selector in selectors.items():
+        if attribute not in message_type.ON_MATCHABLE_ATTRIBUTES and not hasattr(
```
willmcgugan, line 67: Still calling `hasattr`. Won't this break for instance attributes?

willmcgugan: May as well. Not need for two mechanisms to do the same thing.

willmcgugan (approved):
LGTM

src/textual/_on.py
```diff
@@ -64,7 +64,12 @@ def switch_to_home(self) -> None:
 
     parsed_selectors: dict[str, tuple[SelectorSet, ...]] = {}
     for attribute, css_selector in selectors.items():
-        if attribute not in message_type.ALLOW_SELECTOR_MATCH:
+        if attribute == "control":
+            if message_type.control is None:
```
rodrigogiraoserrao, line 68: This can raise an uncaught `AttributeError`.

PR #2501, opened 2023-05-05
Select widget

willmcgugan:
- Adds a Select widget.
- Adds two new CSS rules. `overlay:` and `constrain:`.
- Allows bindings to be set with two values.
- Adds `Region.translate_inside`

Setting `overlay: screen` display a widget above everything else. Like there is an implicit top level layer.

Setting `constrain:` to `x`, `y`, or `both` ensures that the widget doesn't get clipped of the edge of the screen.

This two rules are undocumented as they are experimental. They simplify the implementation of pop-up controls like this Select widget. In the future we can also use them for auto-complete lists and tool tips.

https://user-images.githubusercontent.com/554369/236685193-0f251785-0a0a-4f53-a055-41e7f58dc64b.mov


src/textual/message.py
```diff
@@ -38,6 +38,9 @@ class Message:
     namespace: ClassVar[str] = ""  # Namespace to disambiguate messages
 
     def __init__(self) -> None:
+        self.__post_init__()
```
willmcgugan, line 41: This allows creating messages from dataclasses.

docs/widgets/select.md
```diff
@@ -0,0 +1,90 @@
...
+
+A Select widget is a compact control to allow the user to select between a number of possible options.
+
+
+- [X] Focusable
+- [ ] Container
+
+
+The options in a select control may be passed in to the constructor or set later with [set_options][textual.widgets.Select.set_options].
+Options should be given as sequence of tuples consisting of two values: the first is the string (or [Rich Renderable](https://rich.readthedocs.io/en/latest/protocol.html)) to display in the control and list of options, the second is the value of option.
```
davep, line 13:
```suggestion
Options should be given as a sequence of tuples consisting of two values: the first is the string (or [Rich Renderable](https://rich.readthedocs.io/en/latest/protocol.html)) to display in the control and list of options, the second is the value of option.
```

src/textual/binding.py
```diff
@@ -74,7 +74,7 @@ def make_bindings(bindings: Iterable[BindingType]) -> Iterable[Binding]:
             for binding in bindings:
                 # If it's a tuple of length 3, convert into a Binding first
                 if isinstance(binding, tuple):
-                    if len(binding) != 3:
+                    if len(binding) not in (2, 3):
                         raise BindingError(
                             f"BINDINGS must contain a tuple of three strings, not {binding!r}"
```
davep, line 79: This error message needs an update.

src/textual/css/_styles_builder.py
```diff
@@ -1003,6 +1005,30 @@ def process_grid_size(self, name: str, tokens: list[Token]) -> None:
         else:
             self.error(name, tokens[0], "expected two integers here")
 
+    def process_overlay(self, name: str, tokens: list[Token]) -> None:
```
davep, line 1008: Missing a docstring.

src/textual/css/_styles_builder.py
```diff
@@ -1003,6 +1005,30 @@ def process_grid_size(self, name: str, tokens: list[Token]) -> None:
...
+        except StyleValueError:
+            self.error(
+                name,
+                tokens[0],
+                string_enum_help_text(name, VALID_OVERLAY, context="css"),
+            )
+        else:
+            self.styles._rules[name] = value  # type: ignore
+
+    def process_constrain(self, name: str, tokens: list[Token]) -> None:
```
davep, line 1020: Missing a docstring.

src/textual/widgets/_select.py
```diff
@@ -0,0 +1,376 @@
...
+        box-sizing: content-box;
+        width: 1;
+        height: 1;
+        padding: 0 0 0 1;
+        color: $text-muted;
+        background: transparent;
+    }
+    """
+
+    has_value = var(False)
```
davep, line 115: I believe this should be typed as it helps the docs?

src/textual/widgets/_select.py
```diff
@@ -0,0 +1,376 @@
...
+    A Select displays the current selection.
+    When activated with ++enter++ the widget displays an overlay with a list of all possible options.
+
+    """
+
+    BINDINGS = [("enter", "show_overlay")]
+    """
+    | Key(s) | Description |
+    | :- | :- |
+    | enter | Activate the overlay |
```
davep, line 176: <kbd>Space</kbd> is another common keystroke to active such a control (see macOS for example). I think I'd be tempted to support that too, perhaps even down arrow as well.

src/textual/widgets/_select.py
```diff
@@ -0,0 +1,376 @@
...
+    Select.-expanded > SelectOverlay {
+        display: block;
+    }
+
+    Select.-expanded > SelectCurrent {
+        border: tall $accent;
+    }
+    """
+
+    expanded = var(False, init=False)
```
davep, line 222: I suspect this needs a type to help the docs?

src/textual/widgets/_select.py
```diff
@@ -0,0 +1,376 @@
...
+    """The value of the select."""
+
+    class Changed(Message, bubble=True):
+        """Posted when the select value was changed.
+
+        This message can be handled using a `on_select_changed` method.
+
+        """
+
+        def __init__(
```
davep, line 236: Missing docstring.

src/textual/widgets/_select.py
```diff
@@ -0,0 +1,376 @@
...
+        event.stop()
+        self.expanded = not self.expanded
+
+    @on(SelectOverlay.Dismiss)
+    def _select_overlay_dismiss(self, event: SelectOverlay.Dismiss) -> None:
+        """Dismiss the overlay."""
+        event.stop()
+        self.expanded = False
+        if not event.lost_focus:
+            """If the overlay didn't lose focus, we want to re-focus the select."""
```
davep, line 354: String used as a comment? (some folk do, I know, but not usual for us).

src/textual/widgets/_select.py
davep, line 1: There's one slightly surprising thing about how this works, when it comes to changing focus. I can see why it works this way but I wonder if we can work around it? If the list is dropped down and you're active in the list and you press <kbd>tab</kbd> it closes the dropdown list and moves focus to the next control.

On the other hand, in the same situation, if you press <kbd>shift</kbd>+<kbd>tab</kbd> it closes the dropdown list and "remains" on the same control. Of course, really, it *is* moving to the previous control in the focus chain but I feel it should ideally move to the control that's prior to the dropdown's `SelectCurrent`.

davep (commented):
Second coffee of the morning read-through and code play, couple of questions.

src/textual/widgets/_select.py
willmcgugan: I'll think we'll live with that for now. Can't think of a simple workaround.

src/textual/css/_styles_builder.py
willmcgugan: None of the process methods do. Will leave it until we add all of them.

davep (approved):
Main questions resolved, looks good to go for me. Looking forward to making use of this.

I'm actually really taken by how good and smooth it all looks.

Issue #2261, opened 2023-04-11
Sort method in`DataTable` doesn't support `rich.Text` types


msempere:
Coming from https://github.com/Textualize/textual/pull/2243.

The `sort` method in`DataTable` doesn't support `rich.Text` types stored in the cells. When that is done a `TypeError: '<' not supported between instances of 'Text' and 'Text'` error is thrown because `rich.Text` have no [rich comparison operators](https://docs.python.org/2.7/reference/datamodel.html#object.__lt__) defined.

cc @rodrigogiraoserrao

rodrigogiraoserrao:
My original thought (from the PR linked above):

 > “Maybe we could add a key parameter to the sort methods, akin to how sorted(..., key=...) works.”

@darrenburns might be better suited to determine if this makes sense.

msempere:
> A `key` method makes sense. But I have no issue adding comparison methods to `Text` which just treat it like a string.

@willmcgugan adding them to `Text` is probably the best idea. I'll work on that.
Could you assign that issue to me? thanks!

candidtim:
Only having the ordering functions on the `rich.Text` won't be enough for all use cases (although I entirely support this change too). For example: I want to sort a column with humanized numbers (e.g., file sizes like "42 Bytes" and "2 MiB"). It needs to be sorted based on the raw values themselves, independently of their representation. 

Adding a `sort_fn` to the `add_column` method would not be helpful in **this** case either, because it implies that it should be possible to convert the textual representation back into whatever can be ordered. But otherwise I think it is a nice flexible solution too.

One solution to the above problem is to allow setting the values and their representation separately (and maybe use the `str(value)` as a default representation). Then, either use the values directly, or also allow a custom function to extract the sort keys, based on the values.

Meanwhile, for a workaround I extend the `rich.text.Text` to implement the ordering. For example:

```python
@functools.total_ordering
class Sortable(Text):
    """Like rich.text.Text, but with ordering based on a raw value"""

    def __init__(self, value, *args, **kwargs):
        self.value = value
        super().__init__(*args, **kwargs)

    def __lt__(self, other: "Sortable") -> bool:
        return self.value < other.value

    def __eq__(self, other: "Sortable") -> bool:
        return self.value == other.value
```

and later use it like so:

```python

    def watch_xyz(self, ...):
        ...
        self.table.add_row(Sortable(file_size, naturalsize(file_size), ...)

```

[...]

PR #2512, opened 2023-05-08
Support custom sort functions in 'DataTable'

Alphix:
This is somewhat related to issue #2261

- [x] Updated documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 
 - The DataTable cursor is now scrolled into view when the cursor coordinate is changed programmatically https://github.com/Textualize/textual/issues/2459
+- Added `key_function` as an optional argument for `DataTable.sort` https://github.com/Textualize/textual/pull/2512
 - run_worker exclusive parameter is now `False` by default https://github.com/Textualize/textual/pull/2470
 - Added `always_update` as an optional argument for `reactive.var`
 - Made Binding description default to empty string, which is equivalent to show=False https://github.com/Textualize/textual/pull/2501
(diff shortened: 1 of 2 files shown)


willmcgugan:
@darrenburns Thoughts on this?

darrenburns:
I like the idea but I'm wondering if sorting by key should be a separate method. With the approach in this PR,  "column" and "key" have an awkward dependency on each other. If you supply both, it's unclear what will happen without reading the docs.

Alphix:
> I like the idea but I'm wondering if sorting by key should be a separate method. With the approach in this PR, "column" and "key" have an awkward dependency on each other. If you supply both, it's unclear what will happen without reading the docs.

I could change the PR to follow that approach. I guess the alternative method (`sort_custom()`?) would have to duplicate most of the body of the current `sort()` method, but that's only 9 lines of duplication...

darrenburns:
Another option could be that the `sort` method signature is updated from

```python
    def sort(
        self,
        *columns: ColumnKey | str,
        reverse: bool = False,
    ) -> Self
```

To something like...

```python

    def sort(
        self,
        by: Iterable[ColumnKey | str] | Callable,
        reverse: bool = False,
    ) -> Self
```

So you could either supply an iterable of columns _or_ you can supply a function to `sort`.

TomJGooding:
I stumbled across this open PR looking for something else. 

Sorting a `DataTable` with a custom function was added in Textual v0.41.0. - https://textual.textualize.io/widgets/data_table/#sorting

Alphix:
> I stumbled across this open PR looking for something else.
> 
> Sorting a `DataTable` with a custom function was added in Textual v0.41.0. - https://textual.textualize.io/widgets/data_table/#sorting

Indeed, I'll close this...

Issue #2502, opened 2023-05-05
ZeroDivisionError: Fraction


romanin-rf:
When the window narrows, it flies out. I don't even know if this is a problem or not.
But just in case to check - searched and since I couldn't find anything I wrote this Issue.

**Everything ran on**:
- `Platform`: Windows 10 (64 bit)
- `Python Version`: 3.9.9
    - `Rich Version`: 13.3.5
    - `Textual Version`: 0.23.0

I made a video, but I don't think I can get the whole thing off the trackback.

- Links:
    - `Github Repo`: https://github.com/romanin-rf/SeaPlayer
    - `Video`: https://youtu.be/bVAafzzUK20 (сoming soon)

willmcgugan:
```
│ D:\Program Files\Python39\lib\site-packages\textual\_resolve.py:148 in resolve_fraction_unit                        │
│                                                                                                                     │
│   145 │                                                                                                             │
│   146 │   while True:                                                                                               │
│   147 │   │   remaining_space_changed = False                                                                       │
│ ❱ 148 │   │   resolve_fraction = Fraction(remaining_space, remaining_fraction)                                      │
│   149 │   │   for index, (scalar, min_value, max_value) in enumerate(resolve):                                      │
│   150 │   │   │   value = resolved[index]                                                                           │
│   151 │   │   │   if value is None:                                                                                 │
│                                                                                                                     │
│ ╭──────────────────────────────────────────────────── locals ─────────────────────────────────────────────────────╮ │
│ │                   index = 2                                                                                     │ │
│ │               max_value = None                                                                                  │ │
│ │               min_value = Fraction(16, 1)                                                                       │ │
│ │      remaining_fraction = Fraction(0, 1)                                                                        │ │
│ │         remaining_space = Fraction(-11, 1)                                                                      │ │
│ │ remaining_space_changed = False                                                                                 │ │
│ │                 resolve = [                                                                                     │ │
│ │                           │   (                                                                                 │ │
```

davep:
Managed to narrow a recreation of this down to this:

```python
from textual.app     import App, ComposeResult
from textual.widgets import Static, Button

class ResizeBlackHoleApp( App[ None ] ):

    CSS = """
    Screen {
        layout: horizontal;
    }

    .button {
        width: 1fr;
    }

    Static {
        width: 0;
    }
    """

    def compose( self ) -> ComposeResult:
        for _ in range(100):
            yield Button("", classes="button")
            yield Static()

[...]

PR #2519, opened 2023-05-09
Fix for `resolve_fraction_unit` zero division error (#2502)

davep:
This will require some serious review by especially @willmcgugan (as function author) and @rodrigogiraoserrao (as maths wizard), as I'm not confident I really follow what the code is doing. However, having hand-debugged this I *think* this, or something similar, is the solution to #2502.

The approach here is that there's no point in carrying on looping around if there is no remaining fraction, and on the way out, if the remaining space has gone negative, the fallback value (`Fraction(1)`) should be used (before now it was if the remaining space was `0`, so this is in effect clamping the lower bound).

### DIFF
--- a/src/textual/_resolve.py
+++ b/src/textual/_resolve.py
@@ -143,7 +143,7 @@ def resolve_fraction_unit(
     resolved: list[Fraction | None] = [None] * len(resolve)
     remaining_fraction = Fraction(sum(scalar.value for scalar, _, _ in resolve))
 
-    while remaining_fraction:
+    while remaining_fraction > 0:
         remaining_space_changed = False
         resolve_fraction = Fraction(remaining_space, remaining_fraction)
         for index, (scalar, min_value, max_value) in enumerate(resolve):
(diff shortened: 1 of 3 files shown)


rodrigogiraoserrao:
Your fix doesn't seem to have mathematical issues 😛 
But your fix also does beg the question about whether there might be a configuration that would take the `remaining_space`/`remaining_fraction` to be negative values during the loop, which wouldn't break the loop.

Hence, my comment in the review.

davep:
Breaking unit test has now been simplified. It's now been reduced down to a single widget that is `1fr` and which has a `min-width` bigger than the incoming `remaining_space`.

rodrigogiraoserrao (approved):
I'm happy with this.

src/textual/_resolve.py
```diff
@@ -166,7 +166,7 @@ def resolve_scalar(
 
     return (
         Fraction(remaining_space, remaining_fraction)
-        if remaining_space
+        if remaining_space > 0
```
darrenburns, line 169: Should this be `remaining_fraction > 0`?

Issue #1743, opened 2023-02-08
Make moving a child before or after itself, a no-op


rodrigogiraoserrao:
If we try to use `move_child` to move a widget to after/before itself, should we get an error or is that a no-op?
(Add tests to tests/test_widget_child_moving.py accordingly.)

Context: this arose from writing some functionality in a TODO app that sorts TODO items according to their due date and reorders them when the due date is edited. From this point of view, it makes sense for moving to after/before itself to be a no-op.
Otherwise, the sorting code needs to take extra steps to make sure we don't try to move things to where they already are.

davep:
The more I think about it, the more I think I'm inclined to agree with you that it makes sense to have it be a no-op. "Move child to where it is" seems like it shouldn't error even if it's a weird thing to ask, as it's not the same as calling `move_child` and not providing anywhere, or providing somewhere out of bounds. The request is a valid location and is within bounds.

willmcgugan:
No-op makes sense.

rodrigogiraoserrao:
Note to self: fix the docstrings in `move_child` as well.
Note to self x2: Dave owes me a favour.

davep:
> Note to self x2: Dave owes me a favour.

Let's call it payback for retconning my first blog post. 😛

willmcgugan:
Does this require work? What happens now if you move a child before/after itself?

I agree that should be a no-op and not an error.

willmcgugan:
And the problem is what, exactly? Moving a child before itself raises an error and we think it should be a no-op?

davep:
Yes and [yes](https://github.com/Textualize/textual/issues/1743#issuecomment-1422886317).

rodrigogiraoserrao:
This should also be a no-op if `before`/`after` are the indices of the widget we are trying to move.

PR #2522, opened 2023-05-09
Make the widget moving tests more granular and add more tests

davep:
Adds two (currently xfail) tests to illustrate #1743


rodrigogiraoserrao (approved):
I like this :)

PR #2524, opened 2023-05-09
Initial stab at creating a CheckBox

Alphix:

- [ ] Updated documentation


willmcgugan:
Looks great!

We've had a discussion in the office, and we think we want to go with the OptionList approach.

The reasoning is that lists of check boxes can grow quite large. Unlike RadioSets which tend to be a more manageable number. The OptionList is better at managing very large numbers of things.

Appreciate the effort, but we are going to tackle this one ourselves.

Alphix:
@willmcgugan ok, I'm not sure why/how OptionList would be better at managing very large number of options, but I'm looking forward to seeing the OptionList implementation :)

PR #2527, opened 2023-05-09
Add `auto_focus` to screens

### LINKED ISSUE #2457: Add option, possibly the default, to auto focus the first thing on a screen.
I'm thinking that screen could grow a `auto_focus` parameter which takes a selector. It would then focus the first matching node, if there is one.

This would need to be done when the screen is first made active, and when it is resumed. i.e. if a screen is popped off the stack, the new one should focus.

rodrigogiraoserrao:
This will close #2457.

I've added a docstring but I haven't added documentation elsewhere. Should I?


src/textual/screen.py
```diff
@@ -101,6 +101,12 @@ class Screen(Generic[ScreenResultType], Widget):
     }
     """
 
+    auto_focus: str | None = "*"
```
willmcgugan, line 104: Let's make this an instance var, which should be the default unless there is good rationale for it being a classvar.

willmcgugan (changes requested):

src/textual/screen.py
rodrigogiraoserrao: Mostly because it's a pain to set it as an instance attribute only for people creating their own screens:

```py
class FormScreen(Screen):
    auto_focus = "Input"

    ...
```

vs 

```py
class FormScreen(Screen):
    def __init__(
        self,
        name: str | None = None,
        id: str | None = None,
        classes: str | None = None,
    ) -> None:
        self.auto_focus = "Input"

    ...
```

rodrigogiraoserrao: Classvar it is.

willmcgugan (approved):
I'm actually rather surprised this didn't break any unit tests.

PR #2530, opened 2023-05-09
Moving child before/after self is a no-op.

### LINKED ISSUE #1743: Make moving a child before or after itself, a no-op

rodrigogiraoserrao:
This will close #1743.


willmcgugan (approved):
Just a request for a comment

willmcgugan (approved):

PR #2532, opened 2023-05-09
Allow setting (sub)title of any type.

### LINKED ISSUE #2521: App title and sub_title should be converted to strings
I'm thinking that `App.title` and `App.sub_title` should be converted to strings.

Balancing convenience with hiding errors, this is probably the right thing to do 90% of the time.

rodrigogiraoserrao:
This will close #2521.

App `title` and `sub_title` can now be set to values of arbitrary types but are always converted to strings.

### DIFF
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -409,9 +409,11 @@ class App(Generic[ReturnType], DOMNode):
         self.set_class(not self.dark, "-light-mode")
 
     def validate_title(self, title: Any) -> str:
+        """Make sure the title is set to a string."""
         return str(title)
 
     def validate_sub_title(self, sub_title: Any) -> str:
+        """Make sure the sub-title is set to a string."""
         return str(sub_title)
 
     @property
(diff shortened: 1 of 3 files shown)


src/textual/app.py
```diff
@@ -406,6 +408,12 @@ def __init__(
         self.set_class(self.dark, "-dark-mode")
         self.set_class(not self.dark, "-light-mode")
 
+    def validate_title(self, title: Any) -> str:
```
willmcgugan, line 411: Docstrings por favor

rodrigogiraoserrao: I tried adding docstrings _and_ making the validate methods private but that doesn't work.
I we had a similar issue recently but maybe we only added the ability for watcher methods to be private.
I think validators and compute methods should have the ability to also be private, for exactly the same reasons.
What do you think, @willmcgugan?

davep: > maybe we only added the ability for watcher methods to be private.

Correct, see #2382 and #2442

davep (approved):

PR #2536, opened 2023-05-10
Ensure that `TreeNode` messages are posted when via-API changes are made

davep:
Until now the `Tree.NodeExpanded` and `Tree.NodeCollapsed` messages were only sent out when changes were made to the tree by user interaction. This meant that if any changes were made with the `TreeNode` `expand`, `expand_all`, `collapse`, `collapse_all`, `toggle` or `toggle_all` API calls no messages would be sent.

This PR corrects this.

The work here is, in part, required for #2456 (`DirectoryTree` lazy-loads directory information on node expansion so if someone is expanding nodes under code control the `DirectoryTree` never gets to know that it should load a directory's content) and will build on #1644, essentially adding a missing aspect to the latter PR.

---
To answer the obvious question: yes, `my_tree.root.expand()` on a tree with 1,000s of non-leaf nodes will result in quite the message storm. We've decided that this is an inevitable feature.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
+- Fixed `ZeroDivisionError` in `resolve_fraction_unit` https://github.com/Textualize/textual/issues/2502
 - Fixed `TreeNode.expand` and `TreeNode.expand_all` not posting a `Tree.NodeExpanded` message https://github.com/Textualize/textual/issues/2535
 - Fixed `TreeNode.collapse` and `TreeNode.collapse_all` not posting a `Tree.NodeCollapsed` message https://github.com/Textualize/textual/issues/2535
 - Fixed `TreeNode.toggle` and `TreeNode.toggle_all` not posting a `Tree.NodeExpanded` or `Tree.NodeCollapsed` message https://github.com/Textualize/textual/issues/2535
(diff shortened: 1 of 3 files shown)


rodrigogiraoserrao:
So, `tree.root.expand()` doesn't send the message `NodeExpanded` only for the root?

Would it make sense to consider something like `tree.root.expand(suppress_child_messages=False)` that is `True` by default and that will not spam the user with all the messages that we'd get?

davep:
> So, tree.root.expand() doesn't send the message NodeExpanded only for the root?

It only sends it for the root; the same is true for every node. It's `expand_all` (and the other `_all` methods) that would result in many messages.

> Would it make sense to consider something like tree.root.expand(suppress_child_messages=False) that is True by default and that will not spam the user with all the messages that we'd get?

Not really. See above for `expand`, for example. But even if you were to do it on `expand_all`, that would have the effect of pretty much taking this PR back to where we were before this PR was added (see `DirectoryTree` -- because it lazy-loads directory information for every non-leaf node, only when that non-leaf node is expanded, it needs to receive a message about each non-leaf node to know to kick off the load process).

davep:
> LGTM, although the message storm on recursive expansions and collapses hurts my soul.

Heh. We'll see how it pans out. I went through this with @willmcgugan yesterday and he seemed fairly relaxed about the idea. I was apprehensive at first, but really Textual's message queues should handle it just fine and it also does raise the issue of, if your (for imaginary developer values of "your", not you) trees are that large, perhaps don't allow expanding the whole thing in one go in the first place?

rodrigogiraoserrao:
Alright, thanks for sharing your thoughts!

PR #2545, opened 2023-05-11
Load `DirectoryTree` contents in a worker

davep:
This PR updates `DirectoryTree` so that the contents of any given directory associated with a node in the tree are loaded in a [worker](https://textual.textualize.io/guide/workers/). In most situations this *should* appear to make no difference to the user. However, in environments where loading filesystem data is slow, or even when attempting to load a directory with a *lot* of content, this should keep the rest of the application responsive.

The worst case envisaged here is that someone is viewing a very large directory, which in turn contains lots of directories, which in turn contain lots of entries, and then an `expand_all` was performed[^1]; this PR is designed such that every request to load is placed in a queue, with the queue being processed by a worker thread that services the queue.

Note that this PR has been through at least 3 iterations. It's been a helpful shakedown of the `@work` decorator and we've managed to learn a couple of things that we should be aware of and warn people about (going to add a card to the backlog about making a note of these things in the docs).

So, anyway, here goes background-loading directory tree plan C ¯\\_(ツ)_/¯

PS: Also contains code to address #2564. Not strictly related to this but while I was working on loading code it seemed to make sense to roll it in here.

[^1]: Note that due to the nature of `DirectoryTree`'s lazy loading, an `expand_all` won't expand every single level in the filesystem below that node, because the underlying `Tree` code can't see non-leaf nodes that haven't been loaded yet. Technically a bug, I guess, but in this case likely best considered a feature because performing an `expand_all` on `/` in some filesystem is going to be painful in a few ways.


rodrigogiraoserrao (commented):
Makes sense to me and I couldn't find any issues... But I am not familiar with the `@work` decorator!

willmcgugan:
Some thread-safety issues here. The max concurrency is, in effect, a thread pool. But implemented not quite so optimally. Suggest we don't try to implement the max concurrency right now, and plan an extension to the worker system.

Chat about it on Monday.

davep:
> Some thread-safety issues here. The max concurrency is, in effect, a thread pool. But implemented not quite so optimally. Suggest we don't try to implement the max concurrency right now, and plan an extension to the worker system.

The main reason I did this is that, in conversation last week, we spoke about adding a version of this to `DirectoryTree` for now and then extending the way the worker API worked later and moving over to that. But happy for us to bump that up.

Would be good to hear the thread-safety issues you're seeing here; I tried as much as possible to follow all the advice in the worker API docs so if I've missed something or misunderstood something it'd be great to get that cleared up.

davep:
Right, post-chat change: we've decided that concurrency management really should just be done as part of the worker API from the off; so we'll revert from managing it here and swap to using whatever gets added to the worker API.

Meanwhile... we should try and still manage things here by attempting to have a single queue of nodes that need to be handled, and having a single worker that keeps an eye on the queue and does the work when it comes in. In other words: kick off a loader in `on_mount` that runs for the lifetime of the widget and services one load request at a time.

willmcgugan:
@rodrigogiraoserrao Happy to go over threading and concurrency with you guys at any point. It is something I have done a lot of in my career.

src/textual/widgets/_directory_tree.py
```diff
@@ -116,6 +118,7 @@ def __init__(
             classes: A space-separated list of classes, or None for no classes.
             disabled: Whether the directory tree is disabled or not.
         """
+        self._to_load: Queue[TreeNode[DirEntry]] = Queue()
```
willmcgugan, line 121: I don't find this name to be descriptive. Maybe `_load_queue` ?

src/textual/widgets/_directory_tree.py
```diff
@@ -229,37 +238,115 @@ def filter_paths(self, paths: Iterable[Path]) -> Iterable[Path]:
...
+        """Load the directory contents for a given node.
+
+        Args:
+            node: The node to load the directory contents for.
+
+        Returns:
+            The list of entries within the directory associated with the node.
+        """
+        assert node.data is not None
+        node.data.loaded = True
```
willmcgugan, line 306: This may be the wrong place to set the `loaded` flag. It looks like there is theoretical window where you may end up loading the same directory more than once. 

I think you can avoid that by checking and setting the loaded flag outside of the worker.

src/textual/widgets/_directory_tree.py
```diff
@@ -229,37 +238,115 @@ def filter_paths(self, paths: Iterable[Path]) -> Iterable[Path]:
...
+                # The worker was cancelled, that would suggest we're all
+                # done here and we should get out of the loader in general.
+                break
+            except WorkerFailed:
+                # This particular worker failed to start. We don't know the
+                # reason so let's no-op that (for now anyway).
+                pass
+            # We're still here and we have directory content, get it into
+            # the tree.
+            if content:
```
willmcgugan, line 337: Could this be an `else` clause?

willmcgugan (changes requested):
Looking good, just nitpicks.

src/textual/widgets/_directory_tree.py
davep: Makes sense.

src/textual/widgets/_directory_tree.py
davep: Aye, good cool, that can move up.

willmcgugan (approved):

src/textual/widgets/_directory_tree.py
```diff
@@ -229,37 +248,115 @@ def filter_paths(self, paths: Iterable[Path]) -> Iterable[Path]:
...
-        self._load_directory(self.root)
+    def _directory_content(self, location: Path, worker: Worker) -> Iterator[Path]:
+        """Load the content of a given directory.
+
+        Args:
+            location: The location to load from.
+            worker: The worker that the loading is taking place in.
+
+        Yields:
+            Path: A entry within the location.
```
rodrigogiraoserrao, line 295:
```suggestion
            Path: An entry within the location.
```

rodrigogiraoserrao (approved):
So, to make sure I follow, this iteration just gets a single queue of nodes to be loaded instead of pooling the load jobs?

davep:
@rodrigogiraoserrao 

> So, to make sure I follow, this iteration just gets a single queue of nodes to be loaded instead of pooling the load jobs?

Ish, kinda, sorta. So there's [an async Queue](https://github.com/Textualize/textual/blob/7ff205bc29d0ab1dd2704d25c9ee936fc47e5765/src/textual/widgets/_directory_tree.py#L146) into which nodes that need loading get dropped. There's there [an exclusive async worker](https://github.com/Textualize/textual/blob/7ff205bc29d0ab1dd2704d25c9ee936fc47e5765/src/textual/widgets/_directory_tree.py#L323) that loops around, waiting for stuff to show up in the queue. When something does turn up in the queue [a thread worker is created and waited upon])(https://github.com/Textualize/textual/blob/7ff205bc29d0ab1dd2704d25c9ee936fc47e5765/src/textual/widgets/_directory_tree.py#L335) that does the work of loading the data.

Issue #2480, opened 2023-05-04
Confusing `ValueError: No Tab with id 'tab-1'` when a `NoMatches` is expected


barisione:
Querying a non existing ID in an app using `TabbedCotent` causes a confusing error:

```
from textual import containers, widgets
from textual.app import App, ComposeResult


class MyApp(App):
    def compose(self) -> ComposeResult:
        with containers.Vertical():
            with widgets.TabbedContent():
                with widgets.TabPane("List #1"):
                    yield widgets.ListView(id="foo")

    def on_ready(self) -> None:
        self.query_one(f"#bar")  # Note the wrong ID.


MyApp().run()
```

This gives me `ValueError: No Tab with id 'tab-1'`. If you scroll up you can see  `NoMatches: No nodes match` but it's confusing  and sort of hidden.

# Textual Diagnostics

## Versions

| Name    | Value  |
|---------|--------|
| Textual | 0.23.0 |
| Rich    | 13.3.5 |

## Python

| Name           | Value                                                                     |
|----------------|---------------------------------------------------------------------------|
| Version        | 3.10.1                                                                    |
| Implementation | CPython                                                                   |
| Compiler       | Clang 13.0.0 (clang-1300.0.29.30)                                         |
| Executable     | /Users/bari/.pyenv/versions/3.10.1/envs/undo-core-venv3.10/bin/python3.10 |

## Operating System

| Name    | Value                                                                                             |
|---------|---------------------------------------------------------------------------------------------------|
| System  | Darwin                                                                                            |
| Release | 21.6.0                                                                                            |
| Version | Darwin Kernel Version 21.6.0: Mon Aug 22 20:17:10 PDT 2022; root:xnu-8020.140.49~2/RELEASE_X86_64 |

## Terminal

| Name                 | Value                   |
|----------------------|-------------------------|
| Terminal Application | iTerm.app (3.5.0beta10) |
| TERM                 | xterm-256color          |
| COLORTERM            | truecolor               |
| FORCE_COLOR          | *Not set*               |
| NO_COLOR             | *Not set*               |

## Rich Console options

| Name           | Value                |
|----------------|----------------------|
| size           | width=238, height=62 |
| legacy_windows | False                |
| min_width      | 1                    |
| max_width      | 238                  |
| is_terminal    | True                 |
| encoding       | utf-8                |
| max_height     | 62                   |
| justify        | None                 |
| overflow       | None                 |
| no_wrap        | False                |
| highlight      | None                 |
[...]

PR #2554, opened 2023-05-12
Exit debug

willmcgugan:
Some fatal errors will cause a cascade of additional errors. Generally the first one is most relevant.

- Show a single error on exit by default, and a message if there are any hidden.
- Show all messages in dev mode. If there are more than one, show the amount.

Fixes https://github.com/Textualize/textual/issues/2480

Screenshot 2023-05-12 at 18 05 39
Screenshot 2023-05-12 at 18 05 26


src/textual/app.py
```diff
@@ -1630,8 +1630,22 @@ def _fatal_error(self) -> None:
...
+                self.error_console.print(
+                    f"\n[b]NOTE:[/b] {error_count} errors show above.", markup=True
+                )
+        elif self._exit_renderables:
+            self.error_console.print(self._exit_renderables[0])
+            if error_count > 1:
+                self.error_console.print(
+                    f"\n[b]NOTE:[/b] 1 of {error_count} errors show. Run with [b]--dev[/] to see all errors.",
+                    markup=True,
+                )
```
rodrigogiraoserrao, line 1647: This makes sense. At the same time, the error(s) may show up unexpectedly and the user may not be able to reproduce them. Would it make even more sense to write the remaining errors to a log?

rodrigogiraoserrao (approved):

src/textual/app.py
willmcgugan: If logging is enabled, these will still go to the logs.

Issue #2557, opened 2023-05-13
inconsistent behavior Select Widged


dominikhoebert:
Getting very inconsistent behavior with the new Select Widged when using their .set_options. Often after the options stay empty, depending on if the Select was used before, an item was selected before or not, or the options before where empy ([]).

```python
from textual import on
from textual.app import App, ComposeResult
from textual.widgets import Select, Button


class SelectBug(App):

    def compose(self) -> ComposeResult:
        self.select = Select(options=[("empty", 0)])
        yield self.select
        yield Button("Change Options", id="change_options")

    @on(Button.Pressed, "#change_options")
    def change_options(self, event):
        self.select.set_options([("Four", 4), ("Five", 5), ("Six", 6)])


if __name__ == "__main__":
    app = SelectBug()
    app.run()
```
Keep the great work up!

willmcgugan:
@davep Suspect this is an issue with `virtual_size` in OptionList (but far from certain).

willmcgugan:
The following diff fixes it. AFAICT.

```
@@ -631,7 +631,7 @@ class OptionList(ScrollView, can_focus=True):
         self.highlighted = None
         self._mouse_hovering_over = None
         self.virtual_size = Size(self.scrollable_content_region.width, 0)
-        self.refresh()
+        self._request_content_tracking_refresh()
         return self

     def _set_option_disabled(self, index: int, disabled: bool) -> Self:
```

If this is the fix, it just needs tests.

davep:
Yup, looks good here.

willmcgugan:
@rodrigogiraoserrao I think Dave is doing this.

davep:
While looking at adding some useful testing for this, I've realised that [the suggested fix](https://github.com/Textualize/textual/issues/2557#issuecomment-1546883815), while it works, is slightly puzzling.

The change sets up a call to `_refresh_content_tracking` via `_on_idle`, which is fine; but beforehand in the code there's already a forced call to `_refresh_content_tracking`. On the surface this would seem to be the same thing, albeit a delayed refresh of the content tracking, rather than an immediate one.

willmcgugan:
You could compare what happens if you remove `overlay: screen` from the select.

davep:
@willmcgugan Good call! Okay, it's not that. And now I look at it it's possibly more obvious: I suspect the problem is that the clear/add is happening while it's `display: none`, which presumably means it has no size at that point.

davep:
[...]

PR #2567, opened 2023-05-15
Fix clearing an OptionList

davep:
See #2557, credit to Will: https://github.com/Textualize/textual/issues/2557#issuecomment-1546883815

After a diversion into #2574 and #2582, with the obvious need to dive into the latter much deeper, we're going with this fix for the moment. I don't think it's *the* fix here, but it's for sure a fix that works for now and this should get a revisit if and when #2582 gets a deep dive.

Long story short: because of the nature of how `Select` works, most of the times that the content refresh happens the `OptionList` has no size. No size means it's impossible to calculate the content as the `width` is needed to know how many lines is needed for each prompt. The obvious fix here would be to add `_on_show` to `OptionList` and refresh the content, much like it does with `_on_resize`; however I'm not seeing the expected events happen when it's used within `Select`.

This fix delays the refresh until there is size for the widget, having the desired effect, albeit not in the most ideal way.

### DIFF
--- a/src/textual/widgets/_option_list.py
+++ b/src/textual/widgets/_option_list.py
@@ -630,6 +630,13 @@ class OptionList(ScrollView, can_focus=True):
         self.highlighted = None
         self._mouse_hovering_over = None
         self.virtual_size = Size(self.scrollable_content_region.width, 0)
+        # TODO: See https://github.com/Textualize/textual/issues/2582 -- it
+        # should not be necessary to do this like this here; ideally here in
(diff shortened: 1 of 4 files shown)


rodrigogiraoserrao (approved):
It's always amusing how a problem that created +2 related issues and that probably took you a long time to study just ends up with a temporary fix that is just 2 lines of code. 🤷

willmcgugan (approved):

PR #2568, opened 2023-05-15
Fix empty paste

### LINKED ISSUE #2563: Pasting in Input with an empty string crashes the program
Pressing middle mouse button to paste causes this error in `Input._on_paste` if the clipboard is empty of text:

```
╭───────────────────────────────────────────────────────────────────────── Traceback (most recent call last) ─────────────────────────────────────────────────────────────────────────╮
│ /home/io/Projects/textual-paint/.venv/lib/python3.10/site-packages/textual/widgets/_input.py:335 in _on_paste                                                                       │
│                                                                                                                                                                                     │
│   332 │   │   │   event.prevent_default()                                                                                                                                           │
│   333 │                                                                                                                                                                             │
│   334 │   def _on_paste(self, event: events.Paste) -> None:                                                                                                                         │
│ ❱ 335 │   │   line = event.text.splitlines()[0]                                                                                                                                     │
│   336 │   │   self.insert_text_at_cursor(line)                                                                                                                                      │
│   337 │   │   event.stop()                                                                                                                                                          │
│   338                                                                                                                                                                               │
│                                                                                                                                                                                     │
│ ╭──────────────────────────────────────────────────────────── locals ────────────────────────────────────────────────────────────╮                                                  │
│ │ event = Paste(text='')                                                                                                         │                                                  │
│ │  self = CharInput(id='selected_color_char_input', classes={'color_well'}, pseudo_classes={'focus', 'focus-within', 'enabled'}) │                                                  │
│ ╰────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯                                                  │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
IndexError: list index out of range
```

To reproduce, use `printf "" | xsel` before pasting with MMB.

# Textual Diagnostics

## Versions

| Name    | Value  |
|---------|--------|
| Textual | 0.22.3 |
| Rich    | 13.3.5 |

## Python

| Name           | Value                                            |
|----------------|--------------------------------------------------|
| Version        | 3.10.6                                           |
| Implementation | CPython                                          |
| Compiler       | GCC 11.3.0                                       |
| Executable     | /home/io/Projects/textual-paint/.venv/bin/python |

## Operating System

| Release | 5.19.0-41-generic                                                |
| Version | #42~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Tue Apr 18 17:40:00 UTC 2 |

## Terminal

| Terminal Application | vscode (1.77.3) |

## Rich Console options

| size           | width=183, height=32 |
| max_width      | 183                  |
| max_height     | 32                   |

rodrigogiraoserrao:
Closes #2563.

### DIFF
--- a/src/textual/widgets/_input.py
+++ b/src/textual/widgets/_input.py
@@ -332,7 +332,7 @@ class Input(Widget, can_focus=True):
             event.prevent_default()
 
     def _on_paste(self, event: events.Paste) -> None:
-        line = event.text.splitlines()[0]
+        line = event.text.splitlines()[0] if event.text else ""
         self.insert_text_at_cursor(line)
         event.stop()
 
--- a/src/textual/widgets/_input.py
+++ b/src/textual/widgets/_input.py
@@ -335,7 +335,7 @@ class Input(Widget, can_focus=True):
         if event.text:
             line = event.text.splitlines()[0]
             self.insert_text_at_cursor(line)
-            event.stop()
+        event.stop()
 
     async def _on_click(self, event: events.Click) -> None:
         offset = event.get_content_offset(self)
--- a/src/textual/widgets/_input.py
+++ b/src/textual/widgets/_input.py
@@ -332,9 +332,10 @@ class Input(Widget, can_focus=True):
             event.prevent_default()
 
     def _on_paste(self, event: events.Paste) -> None:
-        line = event.text.splitlines()[0] if event.text else ""
-        self.insert_text_at_cursor(line)
-        event.stop()
+        if event.text:
+            line = event.text.splitlines()[0]
+            self.insert_text_at_cursor(line)
+            event.stop()
 
     async def _on_click(self, event: events.Click) -> None:
         offset = event.get_content_offset(self)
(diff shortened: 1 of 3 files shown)


src/textual/widgets/_input.py
```diff
@@ -332,7 +332,7 @@ async def _on_key(self, event: events.Key) -> None:
             event.prevent_default()
 
     def _on_paste(self, event: events.Paste) -> None:
-        line = event.text.splitlines()[0]
+        line = event.text.splitlines()[0] if event.text else ""
```
willmcgugan, line 335: Could we not skip the `insert_text_at_cursor` entirely when the pasted text is an empty string.

I'm assuming that inserting an empty string does nothing useful.

rodrigogiraoserrao: It doesn't do anything very useful, but that would short-circuit other things that might be expected, like an `Input.Changed` event.
Do you think it's worth adding that exception here?

willmcgugan: I don't think you would get a `Change` event, because nothing should have changed. Have you checked?

rodrigogiraoserrao: My bad, I thought `always_update` was `True` for reactives by default.

willmcgugan (approved):

PR #2577, opened 2023-05-16
Update workflows.

rodrigogiraoserrao:
This caches the virtual environment so we don't have to download it every time (the cache can be cleared from the repository > Actions > Caches (on the left)).

We also split black formatting into a separate workflow. This means we can run black ONLY when *.py files are changed. It also means all other testing jobs don't need to _also_ check formatting.

I think yesterday I skimmed over an action that could let us only trigger the testing jobs if the incremental changes affect code. Right now, if you have a PR that changes code and CI passes, and then you push a change to the changelog (for example), it will trigger the testing CI all over again.
Should I investigate this?

### DIFF
--- a/.github/workflows/black_format.yml
+++ b/.github/workflows/black_format.yml
@@ -4,6 +4,7 @@ name: Black format check
 on:
   pull_request:
     paths:
+      - '.github/workflows/black_format.yml'
       - '**.py'
 
 jobs:
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -3,6 +3,7 @@ name: Test Textual module
 on:
   pull_request:
     paths:
+      - '.github/workflows/pythonpackage.yml'
       - '**.py'
       - '**.pyi'
       - '**.css'


willmcgugan:
Seems solid. What does our resident Github Actions expert think? @darrenburns ?

rodrigogiraoserrao:
@willmcgugan so I should further investigate skipping testing CI on non-code changes, or are we happy with this for now?

darrenburns:
CI hasn't run on this PR - do we need to add `./github/` to the `paths` to ensure that PRs like this trigger CI in pull requests.

These run in parallel right since the Black and testing jobs aren't dependent, right?

rodrigogiraoserrao:
> CI hasn't run on this PR - do we need to add `./github/` to the `paths` to ensure that PRs like this trigger CI in pull requests?

So that CI runs once before merging these changes, kind of as a test?
Sure.

> These run in parallel right since the Black and testing jobs aren't dependent, right?
Yeah.
There should be an example below as soon as I update `paths` to include the jobs I changed.

darrenburns:
> > CI hasn't run on this PR - do we need to add `./github/` to the `paths` to ensure that PRs like this trigger CI in pull requests?
> 
> So that CI runs once before merging these changes, kind of as a test? Sure.

I mean we should keep it in there - we always want CI to run the new workflows any time they change.

> > These run in parallel right since the Black and testing jobs aren't dependent, right?
> 
> Yeah. There should be an example below as soon as I update `paths` to include the jobs I changed.

Nice 👌

darrenburns (approved):

PR #2581, opened 2023-05-16
AUTO_FOCUS targets first focusable widget.

### LINKED ISSUE #2578: AUTO_FOCUS has no effect in most cases
Testing with modal02.py in /docs/examples/guide/screens AUTO_FOCUS doesn't appear to work.

Looking at the code, there doesn't appear to be any check that the thing being auto focused is actually focusable. Suspect it is trying to focus some unrelated widget.

rodrigogiraoserrao:
Fixes #2578.


src/textual/screen.py
```diff
@@ -666,15 +666,18 @@ def _on_screen_resume(self) -> None:
         """Screen has resumed."""
         self.stack_updates += 1
         size = self.app.size
+        self._refresh_layout(size, full=True)
+        self.refresh()
         if self.AUTO_FOCUS is not None and self.focused is None:
             try:
-                to_focus = self.query(self.AUTO_FOCUS).first()
+                focus_candidates = self.query(self.AUTO_FOCUS)
```
willmcgugan, line 673: Since you aren't using `first()` now, this will never raise `NoMatches`

willmcgugan (commented):
I'm puzzled as to why so few tests broke.

I would have expected that every snapshot test with a button or other control will now render differently.

willmcgugan:
@rodrigogiraoserrao Have you checked auto focus works on the default screen?

src/textual/screen.py
rodrigogiraoserrao: Thanks!

willmcgugan (approved):

PR #2600, opened 2023-05-17
Validation

darrenburns:


willmcgugan (commented):
Just some comments. Will do a review tomorrow.

src/textual/validation.py
```diff
@@ -0,0 +1,478 @@
...
+        """Utility for extracting failure descriptions as strings.
+
+        Useful if you don't care about the additional metadata included in the `Failure` objects.
+
+        Returns:
+            A list of the string descriptions explaining the failing validations.
+        """
+        return [failure.description for failure in self.failures]
+
+    def __bool__(self):
```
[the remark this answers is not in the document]
darrenburns: Removed

src/textual/validation.py
```diff
@@ -0,0 +1,478 @@
...
+from urllib.parse import urlparse
+
+from textual._types import Pattern
+
+
+@dataclass
+class ValidationResult:
+    """The result of calling a `Validator.validate` method."""
+
+    valid: bool
```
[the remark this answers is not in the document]
darrenburns: Yup

src/textual/validation.py
```diff
@@ -0,0 +1,478 @@
...
+        return self.valid
+
+
+@dataclass
+class Failure:
+    """Information about a validation failure."""
+
+    value: str | None = None
+    """The value which resulted in validation failing."""
+    validator: Validator | None = None
```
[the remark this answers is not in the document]
darrenburns: Updated - this is a remnant from my initial approach

docs/widgets/input.md
```diff
@@ -20,10 +22,52 @@ The example below shows how you might create a simple form using two `Input` wid
     --8<-- "docs/examples/widgets/input.py"
     ```
 
+### Validating Input
+
+You can supply one or more *validators* to the `Input` widget to validate the value.
```
willmcgugan, line 27: Can you link to the Validator base class here?

src/textual/validation.py
```diff
@@ -0,0 +1,505 @@
...
+
+@dataclass
+class ValidationResult:
+    """The result of calling a `Validator.validate` method."""
+
+    failures: Sequence[Failure] = field(default_factory=list)
+    """A list of reasons why the value was invalid. Empty if valid=True"""
+
+    @staticmethod
+    def merge(results: Sequence["ValidationResult"]) -> "ValidationResult":
```
willmcgugan, line 25: Does the Textual dev need this method? I not, could we make it private.

src/textual/validation.py
```diff
@@ -0,0 +1,505 @@
...
+    validator: Validator
+    """The Validator which produced the failure."""
+    value: str | None = None
+    """The value which resulted in validation failing."""
+    description: str | None = None
+    """An optional override for describing this failure. Takes precedence over any messages set in the Validator."""
+
+    def __post_init__(self) -> None:
+        # If a failure message isn't supplied, try to get it from the Validator.
+        if self.description is None and self.validator is not None:
```
willmcgugan, line 95: `self.validator` can never be `None` according the the type.

I gather this sets `description` if its None. But that means that `description` is still type as being `| None`, which it never will be from the POV of the Textual developer.

Can this logic be moved to where a `Failure` is constructed? Thus removing the need to type `description` as potentially being `None` ?

src/textual/validation.py
```diff
@@ -0,0 +1,505 @@
...
+        yield self.value
+        yield self.validator
+        yield self.description
+
+
+class Validator(ABC):
+    """Base class for the validation of string values.
+
+    Commonly used in conjunction with the `Input` widget, which accepts a
+    list of validators via its constructor. This validation framework can also be used to validate any 'stringly-typed'
```
willmcgugan, line 111: `stringly-typed` ? Is this an established term?

willmcgugan (changes requested):
Great work. Some comments, and maybe-issues.

src/textual/validation.py
darrenburns: They might - if you have a few inputs and you want to know "was everything successful?", and the descriptions for any failures that occurred.

src/textual/validation.py
darrenburns: It's a pretty well-known term referring to data that shouldn't be a string but is a string (for whatever reason). e.g. `"10"` is a stringly-typed integer.

willmcgugan: Fair enough. New one on me!

src/textual/validation.py
darrenburns: I'm not sure this is possible without having to duplicate this logic across all validators (meaning users writing custom validators would also have to duplicate this logic). I think things will become much messier and the trade-off isn't worth it unless you can suggest another approach. 

Is the description being `None` bad? Maybe as a developer who wants a custom `Validator`, I just have no need to describe my validation failure because I won't be using the description anyway.

src/textual/validation.py
```diff
@@ -0,0 +1,505 @@
...
+        """Ensure that the regex is found inside the value.
+
+        Args:
+            value: The value to search in.
+
+        Returns:
+            The result of the validation.
+        """
+        regex = self.regex
+        has_match = re.search(regex, value, flags=self.flags) is not None
```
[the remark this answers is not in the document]
darrenburns: `fullmatch` is what I want 😄

docs/widgets/input.md
darrenburns: Done.

src/textual/validation.py
```diff
@@ -0,0 +1,508 @@
...
+            return number_validation_result
+
+        # We know it's a number, but is that number an integer?
+        is_integer = float(value).is_integer()
+        if not is_integer:
+            return ValidationResult.failure([Integer.NotAnInteger(self, value)])
+
+        return self.success()
+
+    def describe_failure(self, failure: Failure) -> str | None:
```
willmcgugan, line 358: Same comment as base class...

src/textual/validation.py
```diff
@@ -0,0 +1,508 @@
...
+        """Describes why the validator failed.
+
+        Args:
+            failure: Information about why the validation failed.
+
+        Returns:
+            A string description of the failure.
+        """
+        if isinstance(failure, Length.TooShort):
```
willmcgugan, line 424: The error message only reveals part of the requirements. Could we have a single error message that includes the minimum *and* maximum? Should be obvious in context wether the user was too high or too low.

willmcgugan (changes requested):
Some requests re error messages

src/textual/validation.py
```diff
@@ -0,0 +1,508 @@
...
+
+        Returns:
+            A string description of the failure.
+        """
+        if isinstance(failure, Number.NotANumber):
+            return f"Must be a valid number."
+        elif isinstance(failure, Number.NotInRange):
+            minimum = self.minimum if self.minimum is not None else "-∞"
+            maximum = self.maximum if self.maximum is not None else "∞"
+            return f"Must be in range [{minimum}, {maximum}]."
```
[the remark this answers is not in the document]
darrenburns: I need guidance on exactly what you want the messages to be here.

`Must be greater than or equal to A, and less than or equal to B` sounds quite verbose.

`Must be between A and B inclusive`?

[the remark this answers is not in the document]
willmcgugan: Second works works for me. 

How about this to cover all combinations:

"Must be between A and B"
"Must be greater than or equal to X"
"Must be less than or equal to Y"

willmcgugan (approved):
LGTM

PR #2604, opened 2023-05-18
Input completion suggestions

### LINKED ISSUE #2330: Input auto-complete
We need a way to suggest auto-complete for the Input widget.

The auto-complete I'm thinking of is when you start typing and a the rest of the word or phrase appears in front of the cursor. Something like this:

You hit `T` and the text `extualize` appears with a different style in front of the cursor:

T*extualize*

Hitting `e` adds *xtualize*

Te*xtualize*

Pressing escape cancels the auto complete, pressing return accept it.

## The tricky part

The obvious solution would be to have a callback that takes the text in the input, and returns a potential auto-complete. However, we want this to be asynchronous. i.e. the callback should be able to do non-blocking work without making typing slow.

The probably probably requires the use of workers. Please discuss with @willmcgugan re the proposed solution.

rodrigogiraoserrao:
Adds suggested completions to the `Input`

- `Input` accepts a parameter `suggestions` with a list of suggestions that are shown while the user types.
- `Input.suggestions` reactive can be used to change the possible suggestions for a given input.
- `Input` has new component class `input--suggestion`.
- Keybinding <kbd>right</kbd> in `Input` is also used to accept a suggestion if the cursor is at the end of the input.

Closes #2330.


willmcgugan:
Video please!

src/textual/widgets/_input.py
```diff
@@ -492,3 +531,23 @@ def action_delete_left_all(self) -> None:
...
+
+    def validate_suggestions(
+        self, suggestions: Iterable[str] | None
+    ) -> list[str] | None:
+        """Convert suggestions iterable into a list."""
+        if suggestions is None:
+            return None
+        return list(suggestions)
+
+    @work(exclusive=True)
```
willmcgugan, line 543: This worker isn't going to do anything useful...

rodrigogiraoserrao:
> Think we need to go over this one. Chat about it today?

Ping me when you're available.

rodrigogiraoserrao:
Thanks for your guidance @willmcgugan.
This is ready for a second review.
I haven't written any docs yet. Should I add something to the `Input` reference, for example?

src/textual/suggester.py
```diff
@@ -0,0 +1,84 @@
...
+
+class Suggester(ABC):
+    """Defines how [inputs][textual.widgets.Input] generate completion suggestions.
+
+    To define a custom suggester, subclass `Suggester` and implement the async method
+    `get_suggestion`.
+    See [`SuggestFromList`][textual.suggester.SuggestFromList] for an example.
+    """
+
+    async def get(self, input: "Input", value: str) -> None:
```
willmcgugan, line 29: Afraid I would veto the name `get`. Too generic. Is this going to be called by the dev? If not, it could be private.

Also we probably don't want to limit ourselves to `Input` widgets. We might want to offer suggestions elsewhere. Since we're posting messages, it could just be a `MessageTarget`

src/textual/suggester.py
```diff
@@ -0,0 +1,84 @@
...
+
+        Custom suggesters should implement this method.
+
+        Args:
+            value: The current value of the input widget.
+
+        Returns:
+            A valid suggestion or `None`.
+        """
+        raise NotImplementedError()
```
willmcgugan, line 58: I think the raise may be redundant. Since its an abstractmethod, I don't think it will even let you subclass without an implementation.

willmcgugan:
Looking good so far! Will do a review post lunch. Can you add a video?

src/textual/suggester.py
```diff
@@ -0,0 +1,115 @@
...
+
+class Suggester(ABC):
+    """Defines how widgets generate completion suggestions.
+
+    To define a custom suggester, subclass `Suggester` and implement the async method
+    `get_suggestion`.
+    See [`SuggestFromList`][textual.suggester.SuggestFromList] for an example.
+    """
+
+    cache: dict[str, str | None] | None
```
willmcgugan, line 29: Just a thought. You might want to use a `FIFOCache` rather than a dict. So it doesn't have the potential to grow indefinately.

willmcgugan (commented):
This is great! But we've had a discussion in the office, and I think we need to handle both case sensitive and case insensitive suggestions.

Let's have a chat about it tomorrow.

rodrigogiraoserrao:
@willmcgugan Case-insensitive logic has been added.

darrenburns (commented):
I've had a play around with it and it feels good to use :) my only suggestion would be to have case_sensitive=False by default as I think that's more common.

willmcgugan (approved):
Nice work.

PR #2607, opened 2023-05-18
Placeholder color cycle

### LINKED ISSUE #2590: Placeholder cycle per app
The placeholder stores the current color in a classvar.

This means that a second app will start where the previous app left off. And may return a different color from the first iteration.

Not much of an issue in practice, but in the docs the colors change every screenshot. I think we need a separate cycle per app. Suggest we have an independent sequence for each app instance.

rodrigogiraoserrao:
Closes #2590.


willmcgugan (approved):
Good solution!

PR #2608, opened 2023-05-18
DirectoryTree + universal-pathlib

juftin:
## `DirectoryTree` compatibility with different filesystems

I wanted to gauge team Textual's interest in this feature before going down the documentation route - please let me know if you'd be interested in this. This PR swap's `pathlib.Path` for [upath.UPath](https://github.com/fsspec/universal_pathlib) on the `DirectoryTree` widget when installed.

`upath.Path` is a subclass of `Pathlib` that allows it to be a drop in replacement with different filesystems, backed by [fsspec](https://github.com/fsspec/filesystem_spec). When a `upath.UPath` instance refers to a local filesystem it resolves back down to a `pathlib.Path` object so all existing implementations remain the same. When the relevant `fsspec` backend is available `upath.UPath` supports SSH, AWS S3, Google GCS, and Azure for textual trees.  I've also added an implementation for GitHub as well which works out of the box with `requests` installed.

Try out the example app:

```shell
make setup
```

```shell
poetry run textual run examples/universal_directory_tree.py .
```

```shell
poetry run textual run examples/universal_directory_tree.py github://textualize:textual@main/
```

image

```shell
poetry run pip install s3fs
poetry run textual run examples/universal_directory_tree.py s3://my-bucket/
```

- [ ] Updated documentation

## Considerations

I'm happy to make this a plugin as well, I originally tried to subclass the `DirectoryTree `class to achieve this but ran into issues with how embedded `pathlib.Path` is in the `__init__` method of `DirectoryTree`. Ultimately I thought the backwards compatibility for local filesystems in `upath.UPath` made it a nice feature for implement on textual itself.

As a plugin, the cleanest solution I've come up to handle inheritance is something like this without duplicating a bunch of code, but I don't like the hacky way I'm injecting `upath` here.

```python
import textual
import upath
from textual.widgets import DirectoryTree

textual.widgets._directory_tree.Path = upath.UPath


class UniversalDirectoryTree(DirectoryTree):
    """
    A DirectoryTree that can handle any filesystem.
    """

    DEFAULT_CSS = DirectoryTree.DEFAULT_CSS.replace("DirectoryTree", "UniversalDirectoryTree")
```


davep:
Ultimately it's @willmcgugan's call when it comes to adding fresh dependencies to Textual, optional or not, so that's not my call. However, my thinking here is this:

This absolutely feels like something better done as a third-party library. I take your point on the "hacky" way of doing it, but given that `DirectoryTree` only actually calls on `Path` in 4 locations, I'm thinking such a third-party library could be made easier to produce and maintain by having either the `Path` "constructor" be a class variable of `DirectoryTree`, or have some sort of class method that is a "path factory".

```python
DEFAULT_CSS = DirectoryTree.DEFAULT_CSS.replace("DirectoryTree", "UniversalDirectoryTree")
```

By the way, this shouldn't be necessary. The `DEFAULT_CSS` should apply fine to your subclass. For example, try this on the `code_browser.py` example:

```diff
diff --git a/examples/code_browser.py b/examples/code_browser.py
index 025f99f6..00cb919a 100644
--- a/examples/code_browser.py
+++ b/examples/code_browser.py
@@ -18,6 +18,9 @@ from textual.reactive import var
 from textual.widgets import DirectoryTree, Footer, Header, Static
 
 
+class MyDir(DirectoryTree):
+    pass
+
 class CodeBrowser(App):
     """Textual code browser app."""
 
@@ -38,7 +41,7 @@ class CodeBrowser(App):
         path = "./" if len(sys.argv) < 2 else sys.argv[1]
         yield Header()
         with Container():
-            yield DirectoryTree(path, id="tree-view")
+            yield MyDir(path, id="tree-view")
             with VerticalScroll(id="code-view"):
                 yield Static(id="code", expand=True)
         yield Footer()
```

---
Linking #1719 as a related discussion about this.

juftin:
Closing in favor of https://github.com/Textualize/textual/pull/2610

PR #2610, opened 2023-05-19
Let child classes of DirectoryTree override Path creation

davep:
With #1719 in mind, and as an alternative to #2608, this allows for a child class of DirectoryTree to specify how a fresh `Path` should be created. The idea here being that whatever is created should be of the `Path` type, but can have other abilities.

With this change, and using the `UPath` class mentioned elsewhere, and taking `code_browser.py` as an example (note this diff isn't intended as something we'd do to Textual itself, just to illustrate how a third party could make use of this change to `DirectoryTree`):

```diff
--- /Users/davep/develop/python/textual/examples/code_browser.py	2023-05-19 10:05:31
+++ code_browser.py	2023-05-19 10:11:03
@@ -17,7 +17,11 @@
 from textual.reactive import var
 from textual.widgets import DirectoryTree, Footer, Header, Static
 
+from upath import UPath
 
+class UDirectoryTree( DirectoryTree ):
+    PATH = UPath
+
 class CodeBrowser(App):
     """Textual code browser app."""
 
@@ -38,7 +42,7 @@
         path = "./" if len(sys.argv) < 2 else sys.argv[1]
         yield Header()
         with Container():
-            yield DirectoryTree(path, id="tree-view")
+            yield UDirectoryTree(path, id="tree-view")
             with VerticalScroll(id="code-view"):
                 yield Static(id="code", expand=True)
         yield Footer()
```

I can then point it at our GitHub repo with this result:

```sh
$ python code_browser.py github://textualize:textual@main/ 
```

resulting in:

Screenshot 2023-05-19 at 10 21 36

With this small change it would allow someone to make a `textual-universal-directorytree` widget library.

### DIFF
--- a/src/textual/widgets/_directory_tree.py
+++ b/src/textual/widgets/_directory_tree.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 from asyncio import Queue
 from dataclasses import dataclass
 from pathlib import Path
-from typing import ClassVar, Iterable, Iterator
+from typing import Callable, ClassVar, Iterable, Iterator
 
 from rich.style import Style
 from rich.text import Text, TextType
@@ -59,6 +59,9 @@ class DirectoryTree(Tree[DirEntry]):
     }
     """
 
+    PATH: Callable[[str | Path], Path] = Path
+    """Callable that returns a fresh path object."""
+
     class FileSelected(Message, bubble=True):
         """Posted when a file is selected.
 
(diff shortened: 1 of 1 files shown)


juftin:
Oh this is a great way to implement this - watching this so I can create a `UniversalDirectoryTree` plugin once ready. Thanks Dave!

juftin:
Alright, assuming team textual is on board with this change I've implemented a plugin ready to be published to PyPI: [textual-universal-directorytree](https://github.com/juftin/textual-universal-directorytree).

If this isn't accepted I can go back to doing some hackery with
```python
textual.widgets._directory_tree.Path = upath.UPath
```

But this PR is a great way to enable the `UniversalDirectoryTree` without actually materially changing anything on the textual implementation. Thanks for your thoughts on a nice solution Dave!

rodrigogiraoserrao (approved):
Makes sense to me.

PR #2620, opened 2023-05-21
Fix App.BINDINGS type

flupke:
The implicit type was creating mypy errors when defining bindings with tuples. For example:

    class MyApp(App):
        BINDINGS = [("q", "quit", "Quit")]

Would give the error:

    error: List item 0 has incompatible type "Tuple[str, str, str]"; expected "Binding"  [list-item]


willmcgugan (approved):

willmcgugan:
THanks

PR #2623, opened 2023-05-22
remove markup, simplify repr

willmcgugan:
- Remove markup from log statements (no longer supported)
- Simplified widget default repr by removing classes an psuedo-classes. 99% of the time it is noise.


davep (approved):
The repr change *might* benefit from a mention in the ChangeLog?

PR #2628, opened 2023-05-23
Add auto focus to app.

### LINKED ISSUE #2594: AUTO_FOCUS on App should apply to screen
If you don't define a screen, there is no way to set `AUTO_FOCUS`. We should make `AUTO_FOCUS` on App apply to the default screen.

rodrigogiraoserrao:
Closes #2594.

- [x] Updated documentation

### DIFF
--- a/src/textual/screen.py
+++ b/src/textual/screen.py
@@ -94,11 +94,12 @@ class ResultCallback(Generic[ScreenResultType]):
 class Screen(Generic[ScreenResultType], Widget):
     """The base class for screens."""
 
-    AUTO_FOCUS: ClassVar[str | None] = "*"
+    AUTO_FOCUS: ClassVar[str | None] = None
     """A selector to determine what to focus automatically when the screen is activated.
 
     The widget focused is the first that matches the given [CSS selector](/guide/queries/#query-selectors).
-    Set to `None` to disable auto focus.
+    Set to `None` to inherit the value from the screen's app.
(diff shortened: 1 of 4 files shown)


src/textual/screen.py
```diff
@@ -94,11 +94,12 @@ def __call__(self, result: ScreenResultType) -> None:
 class Screen(Generic[ScreenResultType], Widget):
     """The base class for screens."""
 
-    AUTO_FOCUS: ClassVar[str | None] = "*"
+    AUTO_FOCUS: ClassVar[str | None] = None
```
darrenburns, line 97: Feel like I'm very late to the party with this, but out of curiosity - since this is a ClassVar do we not have a way of having 2 Screen instances have different auto-focus depending on context?

For example, when you push an EditForm screen, maybe the initially focused form control differs based on how you arrived at that screen. If you clicked "edit username", the username input would be auto-focused when the screen is pushed. If you clicked "edit date of birth", the date of birth input would be auto-focused when the screen is pushed.

willmcgugan: In that case you could set `AUTO_FOCUS` to `""` (to disable it), then handle focusing manually.

rodrigogiraoserrao: You can also set `AUTO_FOCUS` on the instance and it'll be picked up. 🤷 
Or what am I missing?

willmcgugan (approved):

PR #2631, opened 2023-05-23
Add sparkline widget.

rodrigogiraoserrao:
Implementation of the sparkline widget.

I'd like to validate the API before writing tests and docs.

I used `$success` and `$error` for the two default colours but their interpolation doesn't look great, IMO.
The example below is pretty much the example from the sparkline renderable file.

Untitled

### DIFF
--- a/docs/examples/widgets/sparkline_basic.py
+++ b/docs/examples/widgets/sparkline_basic.py
@@ -1,7 +1,7 @@
 from textual.app import App, ComposeResult
 from textual.widgets._sparkline import Sparkline
 
-data = [1, 2, 2, 1] + [1, 4, 3, 1] + [1, 8, 8, 2]  # (1)!
+data = [1, 2, 2, 1, 1, 4, 3, 1, 1, 8, 8, 2]  # (1)!
 
 
 class SparklineBasicApp(App[None]):


src/textual/renderables/sparkline.py
```diff
@@ -15,7 +15,7 @@
 SummaryFunction = Callable[[Sequence[T]], float]
 
 
-class Sparkline(Generic[T]):
+class SparklineRenderable(Generic[T]):
```
willmcgugan, line 18: Not keen on the redundant `Renderable` in the name. Suggest when you import it you can do `from renderables.sparkline import Sparkline as SparklineRenderable`

willmcgugan:
Agree re color. How about $accent as the peak color, and `$accent 30%` as the base?

Might need a little experimenting to find something that works.

rodrigogiraoserrao:
@willmcgugan some colour experiments here:

Untitled

<details>
<summary>App with experiments</summary>

```py
from textual.app import App
from textual.widgets._sparkline import Sparkline

import statistics


class MyApp(App):
    CSS = """
    Sparkline {
        width: 100%;
        margin: 2 0;
    }
    #fst > .sparkline--max-color {
        color: $success;
    }
    #fst > .sparkline--min-color {
        color: $warning;
    }
    #snd > .sparkline--max-color {
        color: $warning;
    }
    #snd > .sparkline--min-color {
        color: $success;
    }
    #trd > .sparkline--max-color {
        color: $error;
    }
    #trd > .sparkline--min-color {
        color: $warning;
    }
    #frt > .sparkline--max-color {
        color: $warning;
    }
    #frt > .sparkline--min-color {
        color: $error;
    }
    #fft > .sparkline--max-color {
        color: $accent;
    }
    #fft > .sparkline--min-color {
        color: $accent 30%;
    }
    #sxt > .sparkline--max-color {
        color: $accent 30%;
    }
    #sxt > .sparkline--min-color {
        color: $accent;
    }
    #svt > .sparkline--max-color {
        color: $error;
    }
    #svt > .sparkline--min-color {
        color: $error 30%;
    }
    #egt > .sparkline--max-color {
        color: $error 30%;
    }
    #egt > .sparkline--min-color {
        color: $error;
    }
    #nnt > .sparkline--max-color {
        color: $success;
    }
    #nnt > .sparkline--min-color {
        color: $success 30%;
    }
    #tnt > .sparkline--max-color {
        color: $success 30%;
    }
    #tnt > .sparkline--min-color {
        color: $success;
    }
    """

    def compose(self):
        nums = [10, 2, 30, 60, 45, 20, 7, 8, 9, 10, 50, 13, 10, 6, 5, 4, 3, 7, 20]
        yield Sparkline(nums, summary_function=max, id="fst")
        yield Sparkline(nums, summary_function=max, id="snd")
        yield Sparkline(nums, summary_function=max, id="trd")
        yield Sparkline(nums, summary_function=max, id="frt")
        yield Sparkline(nums, summary_function=max, id="fft")
        yield Sparkline(nums, summary_function=max, id="sxt")
        yield Sparkline(nums, summary_function=max, id="svt")
        yield Sparkline(nums, summary_function=max, id="egt")
        yield Sparkline(nums, summary_function=max, id="nnt")
        yield Sparkline(nums, summary_function=max, id="tnt")


app = MyApp()
if __name__ == "__main__":
    app.run()
```

</details>

davep:
Looks like roadmap should get a tweak in this PR too.

davep:
Looking over this, I wonder if the documentation for the widget needs fleshing out a little? I've never personally used a sparkline for anything, ever, so perhaps my ignorance of what they are and why I'd use them comes into play here. But the examples don't really give me a sense of *why* I'd use this, and to some degree I'm a little confused about the how too.

I can see that I feed it data, which will be numeric, but then it's not clear to me what the summary function actually does and why I'd want to use it. Also, there's talk of "buckets" but no explanation of what a bucket it is and how it comes into play here.

I feel that perhaps this widget deserves a good, simple, "real world" example that walks someone though the how and why of using the widget. Why would I use it? What sort of data is it good for? What is a bucket? What choices make sense for the summary function and how do I evaluate those choices?

src/textual/widgets/_sparkline.py
```diff
@@ -0,0 +1,91 @@
...
+    }
+    Sparkline > .sparkline--max-color {
+        color: $accent;
+    }
+    Sparkline > .sparkline--min-color {
+        color: $accent 30%;
+    }
+    """
+
+    data = reactive[Optional[Sequence[Union[int, float]]]](None)
```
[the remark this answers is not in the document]
willmcgugan: I think we should stick to floats. The dev can easy convert those to floats as required.

@rodrigogiraoserrao I think type checkers allow passing an int to something typed as `float`, so you probably don't need the union.

rodrigogiraoserrao:
@willmcgugan @davep I've improved the docs but I still wasn't _very_ verbose.
Let me know where you stand.

davep:
@rodrigogiraoserrao The expansion to the docs makes a lot more sense to be; I feel it's a lot more clear how it works and the role the functions play. I think the only thing I'd mention from these changes is that, rather than mention a "bucket" then go on to explain what it is in a tip, I think I'd start out in the main text talking about how the data gets portioned up to fit in the widget's width.

Maybe? Perhaps?

docs/widgets/sparkline.md
```diff
@@ -0,0 +1,112 @@
...
+    ```sass
+    --8<-- "docs/examples/widgets/sparkline_colors.css"
+    ```
+
+
+## Reactive Attributes
+
+| Name      | Type  | Default     | Description                                        |
+| --------- | ----- | ----------- | -------------------------------------------------- |
+| `data` | `Sequence[int | float] | None` | `None` | The data represented by the sparkline. |
```
davep, line 99:
```suggestion
| `data` | `Sequence[float] | None` | `None` | The data represented by the sparkline. |
```

docs/widgets/sparkline.md
```diff
@@ -0,0 +1,112 @@
...
+| `summary_function` | `Callable[[Sequence[int | float]], float]` | `max` | The function used to summarise each of the bins. |
```
davep, line 100:
```suggestion
| `summary_function` | `Callable[[Sequence[float]], float]` | `max` | The function used to summarise each of the bins. |
```

willmcgugan:
I don't think the term "bucket" is relevant as far as the textual dev is concerned.

All they need to know is that they can assign some data, and it will be stretched to fit the space allocated to it.

src/textual/widgets/_sparkline.py
```diff
@@ -0,0 +1,91 @@
...
+        super().__init__(name=name, id=id, classes=classes, disabled=disabled)
+        self.data = data
+        self.summary_function = summary_function
+
+    def render(self) -> RenderResult:
+        """Renders the sparkline when there is data available."""
+        if not self.data:
+            return "<empty sparkline>"
+        _, base = self.background_colors
+        assert self.summary_function is not None  # Sanity check.
```
davep, line 80: Given that `summary_function` is typed as `None` being a permitted value, and documented as `max` being the default, wouldn't it make more sense to have `None` imply `max` rather than result in an assert fail?

rodrigogiraoserrao: Yeah, I kind of get what you mean...
What I really wanted was to initialise the reactive with `max` as the default value, but if I pass in a callable to `reactive`, it'll think it is the default value factory and it will call `max` to generate the default.
So, the only thing I came up with was setting it to `None` in the beginning and then setting it to a function ASAP.
I don't really ever want `summary_function` to be `None`...

davep: > it'll think it is the default value factory and it will call max to generate the default

No chance of fooling it by initially setting it to `lambda: max` or `partial(max)`?

rodrigogiraoserrao: `lambda: max` breaks typing and `partial(max)` is _not_ a callable that returns `max`, so it doesn't work.
But you gave me another idea.

rodrigogiraoserrao:
@willmcgugan @davep I got rid of the word "bucket".

src/textual/widgets/_sparkline.py
davep: I did imagine that a `cast` would be needed along the way, which is what you kind of did anyway. ;-)

docs/examples/widgets/sparkline_basic.py
```diff
@@ -0,0 +1,19 @@
+from textual.app import App, ComposeResult
+from textual.widgets._sparkline import Sparkline
+
+data = [1, 2, 2, 1] + [1, 4, 3, 1] + [1, 8, 8, 2]  # (1)!
```
davep, line 4: I feel that creating this list of 12 data points, by joining 3 lists, confuses the whole "bucketing" thing if you look at the code. Normally folk wouldn't create a list this way so the fact that it's done like this seems like it's significant or a callout to something I need to be aware of...

I'd just make this one single list to better-emphasise the fact that the "bucketing" is unrelated to the content of the list.

davep (approved):
That last comment about the `sparkline_basic.py` example aside... looking good to me.

docs/examples/widgets/sparkline_basic.py
rodrigogiraoserrao: Thanks for the suggestion.

PR #2636, opened 2023-05-24
Fix Select reactives table layout

davep:
The escaped `|` wasn't being rendered correctly as it was inside back-ticks. Noticed it while passing by.

### DIFF
--- a/docs/widgets/select.md
+++ b/docs/widgets/select.md
@@ -66,10 +66,10 @@ The following example presents a `Select` with a number of options.
 ## Reactive attributes
 
 
-| Name       | Type                 | Default | Description                         |
-| ---------- | -------------------- | ------- | ----------------------------------- |
-| `expanded` | `bool`               | `False` | True to expand the options overlay. |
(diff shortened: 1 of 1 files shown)


docs/widgets/select.md
```diff
@@ -66,10 +66,10 @@ The following example presents a `Select` with a number of options.
...
 
 
-| Name       | Type                 | Default | Description                         |
-| ---------- | -------------------- | ------- | ----------------------------------- |
-| `expanded` | `bool`               | `False` | True to expand the options overlay. |
-| `value`    | `SelectType \| None` | `None`  | Current value of the Select.        |
+| Name       | Type                   | Default | Description                         |
+|------------|------------------------|---------|-------------------------------------|
+| `expanded` | `bool`                 | `False` | True to expand the options overlay. |
+| `value`    | `SelectType` \| `None` | `None`  | Current value of the Select.        |
```
darrenburns, line 72: Does `SelectType | None` not work? (without the `\`)

PR #2670, opened 2023-05-27
Tooltips

willmcgugan:
- Adds Widget.tooltip
- Added Region.inflect
- Added `constrain: inflect;` Currently undocumented, as it is not stable.

https://github.com/Textualize/textual/assets/554369/a0b8bc84-4ca2-4ff0-930b-0696e9829b7e


src/textual/_doc.py
```diff
@@ -115,7 +119,11 @@ async def auto_pilot(pilot: Pilot) -> None:
             result = run_before(pilot)
             if inspect.isawaitable(result):
                 await result
+        await pilot.pause()
         await pilot.press(*press)
+        if hover:
+            await pilot.hover(hover)
+            await pilot.pause(0.5)
```
rodrigogiraoserrao, line 126: For the purposes of testing and auto-pilots, isn't there something else we can call to trigger the tooltip immediately? Waiting half a second for a tooltip to show up in CI can compound to a lot of time.

src/textual/geometry.py
```diff
@@ -905,6 +905,47 @@ def translate_inside(
...
+                        │    ┌──────────┐
+                             │          │
+                        │    │  Result  │
+                             │          │
+                        │    └──────────┘
+        ```
+
+        Args:
+            x_axis: +1 to infect in the positive direction, -1 to inflect in the negative direction.
+            y_axis: +1 to infect in the positive direction, -1 to inflect in the negative direction.
```
rodrigogiraoserrao, line 935: 🤧 😷 

```suggestion
            x_axis: +1 to inflect in the positive direction, -1 to inflect in the negative direction.
            y_axis: +1 to inflect in the positive direction, -1 to inflect in the negative direction.
```

src/textual/geometry.py
```diff
@@ -905,6 +905,47 @@ def translate_inside(
...
         )
 
+    def inflect(
+        self, x_axis: int = +1, y_axis: int = +1, margin: Spacing | None = None
+    ) -> Region:
+        """Inflect a region around one or both axis.
+
+        The `x_axis` and `y_axis` parameters define which direction to move the region.
+        A positive value will move the region right or down, a negative value will move
+        the region left or up.
```
rodrigogiraoserrao, line 915: Should we specify that the value `0` does _not_ inflect the given axis?

src/textual/screen.py
```diff
@@ -169,6 +176,18 @@ def _update_timer(self) -> Timer:
             )
         return self.__update_timer
 
+    @property
+    def layers(self) -> tuple[str, ...]:
+        """Layers of from parent.
```
rodrigogiraoserrao, line 181: Not sure what you meant here?

```suggestion
        """Layers of names from parent.
```

```suggestion
        """Layers from parent.
```

src/textual/widget.py
```diff
@@ -367,7 +370,7 @@ def __init__(
     """Show a horizontal scrollbar?"""
 
     show_horizontal_scrollbar: Reactive[bool] = Reactive(False, layout=True)
-    """SHow a vertical scrollbar?"""
+    """Show a vertical scrollbar?"""
```
rodrigogiraoserrao, line 373:
```suggestion
    """Show a horizontal scrollbar?"""
```

src/textual/widget.py
```diff
@@ -3103,8 +3120,16 @@ async def _on_compose(self) -> None:
...
+            self._extend_compose(widgets)
             await self.mount(*widgets)
 
+    def _extend_compose(self, widgets: list[Widget]) -> None:
+        """Hook to extend composed widgets.
+
+        Args:
+            widgets: Widgets to be mounted.
+        """
+
```
rodrigogiraoserrao, line 3132: I don't get the overarching idea of adding `_extend_compose`. 😞

rodrigogiraoserrao (changes requested):
Should we have a snapshot test of a tooltip?

src/textual/_doc.py
willmcgugan: It is an issue, but anything else wouldn't be a fair test.

I did start implementing a virtual clock of sorts, that could run at 10x speed when testing. But that didn't work out.

src/textual/widget.py
willmcgugan: It's to allow the Screen to add a Tooltip to the DOM, without requiring it to be explicitly added to `compose()`

src/textual/geometry.py
willmcgugan: I guess so.

willmcgugan:
We don't have a way of snapshot testing tooltips. It requires that you move the (virtual) mouse over a button and wait for the timeout. Will revisit when we have updated the snapshot API to support that.

tests/snapshot_tests/snapshot_apps/scroll_to_center.py
```diff
@@ -21,6 +21,8 @@ class MyApp(App[None]):
     }
     """
 
+    # AUTO_FOCUS = ""
+
```
rodrigogiraoserrao, line 25:
```suggestion
```

rodrigogiraoserrao (approved):

PR #2672, opened 2023-05-28
docs(button): update note about button spacing/padding

### LINKED ISSUE #2671: Misleading note in Buttons docs about spacing
Following on from discussion #2633, this note in the [Button docs](https://textual.textualize.io/widgets/button/#additional-notes) is a little misleading:

> The spacing between the text and the edges of a button are due to border, not padding. To create a button with zero visible padding, use the border: none; declaration.

The [default CSS for the Button ](https://github.com/Textualize/textual/blob/main/src/textual/widgets/_button.py#L29-L36) also has a `height` and a `min-width`, so actually to "create a button with zero visible padding" you would  also need to change these values.

TomJGooding:
Closes #2671.

I'm not sure about the wording for this, feel free to make any changes!

### DIFF
--- a/docs/widgets/button.md
+++ b/docs/widgets/button.md
@@ -43,7 +43,7 @@ Clicking any of the non-disabled buttons in the example app below will result in
 
 ## Additional Notes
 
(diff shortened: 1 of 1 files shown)


rodrigogiraoserrao (approved):
Looks good to me.

PR #2675, opened 2023-05-28
handler name in classvar

willmcgugan:
Makes the handler name a class var. This is a small optimization win, but it also allows you to easily inspect the handler name:

```py
>>> from textual.widgets import Input
>>> Input.Changed.handler_name
'on_input_changed'
```

Also fixed the docs issue reported by @darrenburns Details here: https://mkdocstrings.github.io/python/changelog/?h=format_signature#breaking-changes


davep (approved):
Looks good. I wonder if we can make use of this in the docs somehow, to auto-document the handler name?

willmcgugan:
> Looks good. I wonder if we can make use of this in the docs somehow, to auto-document the handler name?

That would be nice, if we could figure out how.

PR #2684, opened 2023-05-29
Ensure reflow reflows non-visible widgets, fix scroll_to_center issue

### LINKED ISSUE #2254: Compositor map not aware of widgets nested deep into the DOM on app startup
When working on #2031 we realised that there is an issue with the compositor `_full_map` on app startup, in that it does not contain the widgets that are nested within the DOM.

To reproduce:
 1. Run the app below and connect to the console.
 2. Press `P` after app startup and note the size of the map that is printed.
 3. Scroll the app to the very bottom.
 4. Press `P` again.
 5. Notice that the map that was printed now is *significantly* bigger.

<details>
<summary>App</summary>

```py


class MyApp(App[None]):
        with VerticalScroll():
            yield Label("v\n" * 500)
            with VerticalScroll():
                for i in range(15):
                    yield Label(str(i))
                with VerticalScroll():
                    for i in range(35):
                        yield Label(str(i))

    def key_p(self) -> None:
        print(self.screen._compositor._full_map)


if __name__ == "__main__":
    MyApp().run()
```

</details>

### LINKED ISSUE #2268: Scroll to center problem
There's an issue with `scroll_to_center` that can be seen in the `scroll_to_center.py` snapshot.

Essentially it doesn't seem to scroll the inner-most container the first time, but does on the second time.

I did some debugging, and it appears that `max_scroll_x` and `max_scroll_y` are erroneously reporting zero, until the container becomes visible for the first time, then they have the correct values. The root cause seems to be that the `virtual_size` is the same as `size`, which is incorrect.

When fixing this, be careful not to break `textual colors`. If you click the last button it should scroll to the end.

This issue likely does touch on the compositor. I suspect it is something related to the `_full_map` structure.

darrenburns:
Closes #2268 
Closes #2254 

When `reflow` was called, we were only reflowing visible widgets. This was preventing widgets that were not visible from being added into `full_map`.

`reflow_visible` was explicitly passing `visible_only=True` to `arrange_root`, but that was actually already the default value. `reflow` was not passing `visible_only` at all, so it was defaulting to `True`.

This broke `scroll_to_center`, because widgets that were scrolled out of view were not present in the compositors full map.

With this change, the `_full_map` is now fully populated by `reflow`, and `scroll_to_center` works as expected.

image

I've removed the `xfail` the `scroll_to_center` snapshot test and it seems to pass consistently now.

- [ ] Updated documentation

### DIFF
--- a/tests/snapshot_tests/snapshot_apps/scroll_to_center.py
+++ b/tests/snapshot_tests/snapshot_apps/scroll_to_center.py
@@ -5,32 +5,33 @@ from textual.widgets import Label
 
 class MyApp(App[None]):
     CSS = """
-    VerticalScroll, HorizontalScroll {
+    VerticalScroll {
         border: round $primary;
     }
     #vertical {
         height: 21;
     }
     HorizontalScroll {
+        border: round $secondary;
         height: auto;
+    }
+    Label {
+        height: auto;
+        width: auto;
     }
     """
 
     def compose(self) -> ComposeResult:
-        with VerticalScroll():
-            yield Label(("SPAM\n" * 205)[:-1])
-            with VerticalScroll():
-                yield Label(("SPAM\n" * 53)[:-1])
-                with VerticalScroll(id="vertical"):
-                    yield Label(("SPAM\n" * 78)[:-1])
-                    with HorizontalScroll():
-                        yield Label(("v\n" * 17)[:-1])
-                        yield Label("@" * 302)
-                        yield Label("[red]>>bullseye<<[/red]", id="bullseye")
-                        yield Label("@" * 99)
-                    yield Label(("SPAM\n" * 49)[:-1])
-                yield Label(("SPAM\n" * 51)[:-1])
-            yield Label(("SPAM\n" * 59)[:-1])
+        yield Label(("SPAM\n" * 53)[:-1])
+        with VerticalScroll(id="vertical"):
+            yield Label(("SPAM\n" * 78)[:-1])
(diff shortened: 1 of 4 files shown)


willmcgugan:
Is that all it was?! I thought it would be a much more complex issue than that.

darrenburns:
@willmcgugan Yup, seems like it 😄

tests/snapshot_tests/snapshot_apps/scroll_to_center.py
```diff
@@ -5,32 +5,33 @@
...
-                    yield Label(("SPAM\n" * 78)[:-1])
-                    with HorizontalScroll():
-                        yield Label(("v\n" * 17)[:-1])
-                        yield Label("@" * 302)
-                        yield Label("[red]>>bullseye<<[/red]", id="bullseye")
-                        yield Label("@" * 99)
-                    yield Label(("SPAM\n" * 49)[:-1])
-                yield Label(("SPAM\n" * 51)[:-1])
-            yield Label(("SPAM\n" * 59)[:-1])
+        yield Label(("SPAM\n" * 53)[:-1])
```
darrenburns, line 25: Simplified this a little because I was struggling to understand it vs what was on screen. It still has multiple levels of nested scrolling containers, both vertical and horizontal.

rodrigogiraoserrao (approved):
Nice 🚀

willmcgugan (approved):

PR #2686, opened 2023-05-29
Increase worker coverage and fix bug.

rodrigogiraoserrao:


tests/test_worker.py
```diff
@@ -1,11 +1,15 @@
 import asyncio
+from re import A
```
willmcgugan, line 2: What the heck is A?

rodrigogiraoserrao: I have no freaking idea.
I really have to check my IDE settings. I keep importing weird things.

Issue #2630, opened 2023-05-23
Consider the idea of having the `@on` message handler capture and fire on base messages


davep:
Noting this as a reminder about a conversation for later consideration.

Right now the `@on` decorator only fires on a specific message type. So imagine having this sort of message hierarchy:

```python
class PersonChanged(Message):
    ...

class PersonAdded(PersonChanged):
    ...

class PersonEdited(PersonChanged):
    ...

class PersonRemoved(PersonChanged):
    ...
```

To have an event handler that handled any of the latter three messages, you'd currently need to write:

```python
@on(PersonAdded)
@on(PersonEdited)
@on(PersonRemoved)
def person_changed(self, event ):
    ...
```

Perhaps we should support being able to do this:

```python
@on(PersonChanged)
def person_changed(self, event ):
    ...
```

such that the `person_changed` method is called if any of the above 4 messages are posted.

davep:
Reminder to self that #2453 is the place to look for the original implementation PR.

davep:
My current thinking on how this would work would be something [like this](https://github.com/davep/textual/commit/283a8b65299e47d92e417047dde5fb8e9e543e29?diff=unified). Applying this, all but one test passes in Textual, and the addition of tests for the expected new behaviour pass too.

There's just one fail: the snapshot test for [the example of `SelectionList.selected` changing](https://github.com/Textualize/textual/blob/a40300a6f5f78c5b2dd41bf36e282cbea2af7ff1/docs/examples/widgets/selection_list_selected.py). When the change in place, for reasons I can't quite get the the bottom of yet, the `@on(Mount)` seems to fire before the DOM is up and running, but it also seems to eventually work -- as in the app doesn't crash when run standalone, and the initial population of the `Pretty` widget takes place, but when you exit the app there's an error that `query_one` couldn't find the `Pretty`.

Without the change I'm experimenting with, that example code works fine with no error (which, as sneaky as it is -- splitting `on_mount` behaviour between two methods -- it should work). I can make the error go away by changing the `on_mount` work to a more traditional approach, but that would be just ignoring the problem. This should work.

So, noting this here as a reminder to tomorrow morning me; and I think I might also need to get a walkthrough of the way `@on` works under the hood and what `_get_dispatch_methods` is doing within the greater scheme of things; I suspect I'm misreading something here.

davep:
[...]

PR #2691, opened 2023-05-30
Extend the `@on` decorator so that it will match superclasses

davep:
This PR seeks to implement the requirement laid out in #2630. In short, the idea is that the the `@on` decorator should match a given message and any message that inherits from it. To borrow from the issue this stems from, it means that this sort of hierarchy:

```python
class PersonChanged(Message):
    ...

class PersonAdded(PersonChanged):
    ...

class PersonEdited(PersonChanged):
    ...

class PersonRemoved(PersonChanged):
    ...
```

no longer needs something like this:

```python
@on(PersonAdded)
@on(PersonEdited)
@on(PersonRemoved)
def person_changed(self, event ):
    ...
```

but can be handles something like this:

```python
@on(PersonChanged)
def person_changed(self, event ):
    ...
```

---
One other small change included in this PR is to `events.Unmount`, which used to inherit from `events.Mount`; as of this PR it inherits from `events.Event`; just like all other events. The point of this change being that a developer van safely `@on(Mount)` without needing to handle fact that `events.Unmount` would trigger the handler too.

### DIFF
--- a/src/textual/message_pump.py
+++ b/src/textual/message_pump.py
@@ -578,15 +578,16 @@ class MessagePump(metaclass=_MessagePumpMeta):
             # Try decorated handlers first
             decorated_handlers = cls.__dict__.get("_decorated_handlers")
             if decorated_handlers is not None:
-                handlers = []
+                handlers: list[tuple[Callable, dict[str, tuple[SelectorSet, ...]]]] = []
                 add_handlers = handlers.extend
                 # Allow for firing of handlers bound to a message higher up
                 # the inheritance tree.
                 for check_message in type(message).__mro__:
-                    add_handlers(decorated_handlers.get(check_message, []))
-                    # If we've hit message, that's far enough.
-                    if check_message == Message:
+                    # If we've hit something that isn't derived from Message
+                    # we can give up.
+                    if not issubclass(check_message, Message):
                         break
+                    add_handlers(decorated_handlers.get(check_message, []))
                 from .widget import Widget
 
                 for method, selectors in handlers:
--- a/src/textual/message_pump.py
+++ b/src/textual/message_pump.py
@@ -582,7 +582,7 @@ class MessagePump(metaclass=_MessagePumpMeta):
                 add_handlers = handlers.extend
                 # Allow for firing of handlers bound to a message higher up
                 # the inheritance tree.
-                for check_message in type(message).__mro__:
+                for check_message in message.__class__.__mro__:
                     # If we've hit something that isn't derived from Message
                     # we can give up.
                     if not issubclass(check_message, Message):
--- a/src/textual/message_pump.py
+++ b/src/textual/message_pump.py
@@ -578,7 +578,15 @@ class MessagePump(metaclass=_MessagePumpMeta):
             # Try decorated handlers first
             decorated_handlers = cls.__dict__.get("_decorated_handlers")
             if decorated_handlers is not None:
-                handlers = decorated_handlers.get(type(message), [])
+                handlers = []
+                add_handlers = handlers.extend
+                # Allow for firing of handlers bound to a message higher up
+                # the inheritance tree.
+                for check_message in type(message).__mro__:
+                    add_handlers(decorated_handlers.get(check_message, []))
+                    # If we've hit message, that's far enough.
+                    if check_message == Message:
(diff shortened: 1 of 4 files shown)


rodrigogiraoserrao (approved):

rodrigogiraoserrao:
Hold on, shouldn't the title say "[...] it will match subclasses"?

src/textual/message_pump.py
```diff
@@ -578,7 +578,16 @@ def _get_dispatch_methods(
...
             if decorated_handlers is not None:
-                handlers = decorated_handlers.get(type(message), [])
+                handlers: list[tuple[Callable, dict[str, tuple[SelectorSet, ...]]]] = []
+                add_handlers = handlers.extend
+                # Allow for firing of handlers bound to a message higher up
+                # the inheritance tree.
+                for check_message in message.__class__.__mro__:
+                    # If we've hit something that isn't derived from Message
+                    # we can give up.
+                    if not issubclass(check_message, Message):
```
willmcgugan, line 588: This may fail in the case of multiple inheritance or mixins. I would hope that nobody uses those in the context of messages, but maybe we should continue rather than break.

src/textual/message_pump.py
```diff
@@ -578,7 +578,16 @@ def _get_dispatch_methods(
...
+                        break
+                    add_handlers(decorated_handlers.get(check_message, []))
```
willmcgugan, line 590: Could this lead to a handler being called more than once for a message?

i.e if you have an inheritance chain Foo -> Bar -> Baz (where Foo is the base class).  It looks like if the messages was a `Baz` it would call the handler 3 times?

Could very well have misinterpreted the code.

davep: I *might* be misinterpreting the question, but isn't the situation mentioned here what this test is about?

```python
class MessageSender(Widget):
    class Parent(Message):
        pass

    class Child(Parent):
        pass

    def post_parent(self) -> None:
        self.post_message(self.Parent())

    def post_child(self) -> None:
        self.post_message(self.Child())


async def test_fire_on_inherited_message() -> None:
    """Handlers should fire when descendant messages are posted."""

    posted: list[str] = []

    class InheritTestApp(App[None]):
            yield MessageSender()

        @on(MessageSender.Parent)
        def catch_parent(self) -> None:
            posted.append("parent")

        @on(MessageSender.Child)
        def catch_child(self) -> None:
            posted.append("child")

        def on_mount(self) -> None:
            self.query_one(MessageSender).post_parent()
            self.query_one(MessageSender).post_child()

    async with InheritTestApp().run_test():
        pass

    assert posted == ["parent", "child", "parent"]
```

If the message is `MessageSender.Child` then `InheritTestApp.catch_child` and `InheritTestApp.catch_parent` get triggered; which is what I'd expect here and actually desire.

src/textual/message_pump.py
davep: Fair point, I'll add some testing for this first. The intent here was to not go further up the MRO than `Message`, to save looping more than necessary (although of course, at the moment, that's only really going to save one more trip round).

davep: Test added and the "get out as soon as possible" code changed to "wander up the MRO and check anything that looks like a `Message`): 02b862c28c57fff6186a35a1ed611cd1d0f90860

src/textual/message_pump.py
davep: Okay, after a wee chat this morning, the (valid) concern here is what happens in *this* situation (that is, someone decorates the *same* handler with messages that are parent/child related):

```python
async def test_fire_inherited_on_single_handler() -> None:
    """Test having parent/child messages on a single handler."""

    posted: list[str] = []

    class InheritTestApp(App[None]):
            yield MessageSender()

        @on(MessageSender.Parent)
        @on(MessageSender.Child)
        def catch_either(self, event: MessageSender.Parent) -> None:
            posted.append(f"either {event.__class__.__name__}")

        def on_mount(self) -> None:
            self.query_one(MessageSender).post_parent()
            self.query_one(MessageSender).post_child()

    async with InheritTestApp().run_test():
        pass

    # TODO: Likely not this!
    assert posted == ["either Parent", "either Child", "either Child"]
```

PR #2700, opened 2023-05-31
Split up the OptionList snapshot tests

davep:
Don't run 3 snapshot tests in one test function when it's easier to see what broke and where if each is in their own function.


rodrigogiraoserrao:
> Don't run 3 snapshot tests in one test function when it's easier to see what broke and where if each is in their own function.

I seem to recall this having been done earlier...
Or was it another set of 3 tests?

PR #2701, opened 2023-05-31
Improve `OptionList` test coverage

davep:
~~Almost 100%, but that last line is going to need changes to `snap_compare` so that we can pilot/mouse things for snapshot tests.~~

Got this to 100% via a slightly different route/test. Sadly I can't get this to work via snapshot tests as it looks like `Pilot.hover` doesn't quite tickle `_mouse_move` in the way you'd expect.


rodrigogiraoserrao:
> Almost 100%, but that last line is going to need changes to `snap_compare` so that we can pilot/mouse things for snapshot tests.

Doesn't the `run_before` parameter cut it?

davep:
@rodrigogiraoserrao Good call. Didn't even know that was a thing.

Issue #2662, opened 2023-05-26
Setting `Select.value` in `on_mount` for a `Select` in a container raises an exception


davep:
Initially stemming from [a question on Discord](https://discord.com/channels/1026214085173461072/1033754296224841768/1111680440709947533), it seems that it isn't possible to set `Select.value` in `on_mount` if the `Select` is inside a container. That is to say, this code works fine:

```python
from textual.app        import App, ComposeResult
from textual.widgets    import Select

class SelectBugApp( App[ None ] ):

    def on_mount( self ) -> None:
        self.query_one( Select ).value = 1

    def compose( self ) -> ComposeResult:
        yield Select( ( ( str( n ), n ) for n in range( 10 ) ) )

if __name__ == "__main__":
    SelectBugApp().run()
```

but this code:

```python
from textual.app        import App, ComposeResult
from textual.containers import Horizontal
from textual.widgets    import Select

class SelectBugApp( App[ None ] ):

    def on_mount( self ) -> None:
        self.query_one( Select ).value = 1

    def compose( self ) -> ComposeResult:
        yield Horizontal( Select( ( ( str( n ), n ) for n in range( 10 ) ) ) )

if __name__ == "__main__":
    SelectBugApp().run()
```

raises this error:

```
╭───────────────────────────────────────────────────── Traceback (most recent call last) ──────────────────────────────────────────────────────╮
│ /Users/davep/develop/python/textual-sandbox/select_value.py:8 in on_mount                                                                    │
│                                                                                                                                              │
│    5 class SelectBugApp( App[ None ] ):                                                                                                      │
│    6 │                                                                                                                                       │
│    7 │   def on_mount( self ) -> None:                                                                                                       │
│ ❱  8 │   │   self.query_one( Select ).value = 1                                                                                              │
│    9 │                                                                                                                                       │
│   10 │   def compose( self ) -> ComposeResult:                                                                                               │
│   11 │   │   yield Horizontal( Select( ( ( str( n ), n ) for n in range( 10 ) ) ) )                                                          │
│                                                                                                                                              │
│ ╭───────────────────────────── locals ──────────────────────────────╮                                                                        │
│ │ self = SelectBugApp(title='SelectBugApp', classes={'-dark-mode'}) │                                                                        │
│ ╰───────────────────────────────────────────────────────────────────╯                                                                        │
│                                                                                                                                              │
[...]

PR #2702, opened 2023-05-31
mount order

willmcgugan:
Fix an issue with mount order


davep:
Note to add that this fixes the underlying cause of https://github.com/Textualize/textual/issues/2662

davep (approved):

PR #2708, opened 2023-05-31
Private validate compute

rodrigogiraoserrao:

- [ ] Updated documentation

### DIFF
--- a/src/textual/message_pump.py
+++ b/src/textual/message_pump.py
@@ -79,13 +79,13 @@ class _MessagePumpMeta(type):
                     )
 
         # Look for reactives with public AND private compute methods.
-        for name, value in class_dict.items():
-            if name.startswith("compute_") and callable(value):
-                reactive_name = name[len("compute_") :]
+        for attr_name, value in class_dict.items():
+            if attr_name.startswith("compute_") and callable(value):
+                reactive_name = attr_name[len("compute_") :]
                 if (
                     reactive_name in class_dict
                     and isinstance(class_dict[reactive_name], Reactive)
-                    and f"_{name}" in class_dict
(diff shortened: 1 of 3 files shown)


src/textual/message_pump.py
```diff
@@ -75,6 +75,19 @@ def __new__(
                         f"on_{namespace}_{camel_to_snake(value.__name__)}"
                     )
 
+        # Look for reactives with public AND private compute methods.
+        for attr_name, value in class_dict.items():
+            if attr_name.startswith("compute_") and callable(value):
+                reactive_name = attr_name[len("compute_") :]
```
davep, line 81: Possibly taking things too far, but I think I'd be tempted to take the `len` of the string outside of the loop and use that, so you're not calling `len` every time around.

davep (approved):
Looks good to me. One very minor suggestion.

src/textual/message_pump.py
rodrigogiraoserrao: So this is what it feels like to be Rodrigoed!

davep: I learnt from the best. 🤷‍♀️

src/textual/message_pump.py
```diff
@@ -76,9 +76,11 @@ def __new__(
                     )
 
         # Look for reactives with public AND private compute methods.
+        prefix = "compute_"
+        prefix_len = len("compute_")
```
davep, line 80: So... you know what I'm going to mention here, right? ;-)

rodrigogiraoserrao: I'm guessing:

```suggestion
        prefix_len = len(prefix)
```

willmcgugan (approved):

PR #2709, opened 2023-05-31
Delete Reactive.var.

### LINKED ISSUE #2706: Remove `Reactive.var` in favour of `var`?
In `reactive.py` we have the classmethod `Reactive.var` and the class `var` that subclasses `Reactive`.
They are not strictly equivalent, but they seem to have the same purpose.
Should we delete `Reactive.var`?

rodrigogiraoserrao:
Closes #2706.

Deletes `Reactive.var` in favour of `var`.


davep (approved):
Assuming this isn't ever used anywhere, makes sense to me. It's worth noting too that it never passed on `always_update`.

I'd be tempted to mark this in the CHANGELOG as a breaking change though given it's removing something.

rodrigogiraoserrao:
> I'd be tempted to mark this in the CHANGELOG as a breaking change though given it's removing something.

Aren't all deletions breaking changes by default?

davep:
Sure, but I feel it makes it stand out and easier to search for - would also fit in with other entries marked this way.

willmcgugan (approved):

Issue #2698, opened 2023-05-31
Updating a `TreeNode.label` doesn't refresh the tree right away


davep:
In this little illustration of how a `Tree`'s nodes' labels can be updated on the fly, the actual updates don't show until some other interaction with the `Tree` takes place (moving the cursor, causing a mouse hover event, etc):

```python
from textual.app     import App, ComposeResult
from textual.widgets import Header, Footer, Tree

class TreeNodeUpdateApp( App[ None ] ):

    BINDINGS = [
        ( "a", "add", "" ),
    ]

    def compose( self ) -> ComposeResult:
        yield Header()
        yield Tree( "100" )
        yield Footer()

    def on_mount( self ) -> None:
        for n in range( 10 ):
            node = self.query_one( Tree ).root.add( str( n ), expand=True )
            for m in range( 10 ):
                node.add_leaf( str( m ) )

    def action_add( self ):
        node = self.query_one( Tree ).cursor_node
        node.label = str( int( str( node.label ) ) + 1 )

if __name__ == "__main__":
    TreeNodeUpdateApp().run()
```

adding a `self.query_one( Tree ).refresh()` to the end of `action_add` fixes it so the update of the label is reflected right away.

[`TreeNode.set_label`](https://github.com/Textualize/textual/blob/149c39c86c083772d3249c8bbd5c1fa7923a8d55/src/textual/widgets/_tree.py#L301-L309) should probably be changed to cause the `Tree` to perform a `refresh`.

PR #2713, opened 2023-06-01
Ensure that a tree's node is refreshed right away when its label is changed

davep:
See #2698.


src/textual/widgets/_tree.py
```diff
@@ -307,6 +307,7 @@ def set_label(self, label: TextType) -> None:
         self._updates += 1
         text_label = self._tree.process_label(label)
         self._label = text_label
+        self._tree.call_later(self._tree._refresh_node, self)
```
rodrigogiraoserrao, line 310: I never know whether I should use `call_later`, `call_soon`, or the other one.
How do you decide?

rodrigogiraoserrao, line 310: Why can't you refresh right here?

davep: Calling refresh directly broke much of `test_disabled.py` in a really fun way (try it and see, it's one of the weirdest breaks I've seen so far); some tests taking an age to complete, one just flat out breaking. I suspect some sort of deadly embrace going on; seemed to happen as the `Tree` in the tests was being spun up.

davep: Did you mean `call_next` rather than `call_soon`? I generally go with the rule that `call_next` is the one to go with unless it's obvious otherwise because jumping the queue should be an exceptional thing to do.

rodrigogiraoserrao: Interesting!

rodrigogiraoserrao: That makes sense, thanks.

rodrigogiraoserrao (approved):

PR #2715, opened 2023-06-01
Remove duplicated modes.

rodrigogiraoserrao:
See https://github.com/Textualize/textual/pull/2540#discussion_r1213230392


rodrigogiraoserrao:
> Merge error?

Probably. But we defo don't need two of the exact same class!

Issue #2534, opened 2023-05-09
Textual Console should report working CSS after reporting an error


lolbat:
Currently when there is a CSS error in Textual you get a very good error report. But when you fix the issue you don't get a report that the CSS processed correctly. 

This may be related to #2533. I have a lot of information disabled and the CSS "all clear" might be getting sent to a category that I have shut off.

willmcgugan:
Suggest we write a green panel to the console whenever CSS has gone from a non-working state to a working state.

PR #2716, opened 2023-06-01
In devtools show confirmation of successfully loading CSS after a previous fail

davep:
See #2534

Screenshot 2023-06-01 at 16 00 17

### DIFF
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -414,6 +414,7 @@ class App(Generic[ReturnType], DOMNode):
 
         self.design = DEFAULT_COLORS
 
+        self._css_has_errors = False
         self.stylesheet = Stylesheet(variables=self.get_css_variables())
 
         css_path = css_path or self.CSS_PATH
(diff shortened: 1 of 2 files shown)


src/textual/app.py
```diff
@@ -1247,14 +1248,27 @@ async def _on_css_change(self) -> None:
...
+                    from rich.panel import Panel
+
+                    self.log.system(
+                        Panel(
+                            "CSS files successfully loaded after previous error:\n\n- "
+                            + "\n- ".join(str(path) for path in css_paths),
+                            style="green",
+                            border_style="green",
+                        )
+                    )
```
rodrigogiraoserrao, line 1261: Shouldn't this be in the `else` statement already?

davep: Already? Not sure I follow. This is something new. Extra bit of logging to go with the existing logging.

willmcgugan (approved):

src/textual/app.py
rodrigogiraoserrao: Ignore the "already", then.
Shouldn't this be the first piece of code in the `else` instead of the last piece of code in the `try`?

davep: Ahh I see. That would work too, but I was keeping the logging with the existing logging.

rodrigogiraoserrao: Maybe both logs should be in the `else` statement. 🤷

davep: That would work too, of course, but I feel keeping the reporting of the elapsed time closer to the calculation of it makes for an easier read.

Issue #2720, opened 2023-06-01
Some inline screenshots in the docs have been broken by `AUTO_FOCUS`


davep:
Following on from #2718: it looks like some other widget examples in the docs are broken after #2527. After a quick `ripgrep` of the docs to find likely candidates, followed by a quick visual inspection of those examples, the following seem to be in need of fixing:

- [x] [`ProgressBar` widget examples](https://textual.textualize.io/widgets/progress_bar/#complete-app-example)
- [x] [`Input` widget example](https://textual.textualize.io/widgets/input/#a-simple-example) (albeit it still works, but the input is in the wrong field now)
- [x] [`Input` gallery entry](https://textual.textualize.io/widget_gallery/#input) (ditto)
- [x] [`ProgressBar` gallery entry](https://textual.textualize.io/widget_gallery/#progressbar)
- [x] [The reactivity examples](https://textual.textualize.io/guide/reactivity/) all still work, but have a <kbd>Tab</kbd> that isn't needed (it works because they're just the one widget that can receive focus anyway)
- [x] [The input focus example](https://textual.textualize.io/guide/input/#input-focus) is likely off from what it used to be, although I think it still sort of works
- [x] [`byte02.py` in the compound widgets example](https://textual.textualize.io/guide/widgets/#__tabbed_16_1) is likely different from what it used to be.
- [x] [`byte03.py` in the compound widgets example](https://textual.textualize.io/guide/widgets/#attributes-down) is very obviously not working any more.

This is likely not an exhaustive list, it's just the result of a very quick `rg "textual.*press.*tab"` within the `docs/`, so I'd say a more comprehensive search would be worth it when it comes to completing this issue.

willmcgugan:
Good catch

davep:
A bit of a wider search turns up nothing else; except for the tutorial. Checking the "blame" for `tutorial.md`, I'm not seeing any edits to the relevant lines post-add-`AUTO_FOCUS`, but on the other hand it's not exactly obvious to me if there is a problem with any of the tutorial screenshots.

@willmcgugan: as the author of the tutorial, would you mind glancing over the published version when you get a moment and check if anything looks wrong with the generated screenshots? I feel they should have problems but it might be the nature of each one means it doesn't stand out like some of the others I've fixed.

PR #2722, opened 2023-06-02
Fix various screenshots in the docs, broken by the addition of `AUTO_FOCUS`

davep:
A small number of screenshots in the documentation, that used `press=` to get stuff done in the code, and which performed a <kbd>tab</kbd> to get focus to a particular control, were either different from intended or just plain not working since the addition of `AUTO_FOCUS` (because, of course, focus position was off by one from what was originally the case).

This PR fixes up all the ones I could find.

There is a question left hanging about the screenshots in the tutorial, which I've asked @willmcgugan to cast his eye over. I *think* they are likely wrong but because of the nature of them it's not obvious to me how they should look (checking the blame for `tutorial.md`, I can't see any evidence of a recent fix so it follows they must be wrong).

If the tutorial is shown to be off too that can be rolled into this, or done as a followup.

See #2720.

### DIFF
--- a/docs/guide/input.md
+++ b/docs/guide/input.md
@@ -102,7 +102,7 @@ The following example shows how focus works in practice.
 
 === "Output"
 
-    ```{.textual path="docs/examples/guide/input/key03.py", press="tab,H,e,l,l,o,tab,W,o,r,l,d,!"}
+    ```{.textual path="docs/examples/guide/input/key03.py", press="H,e,l,l,o,tab,W,o,r,l,d,!"}
     ```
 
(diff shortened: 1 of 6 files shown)


rodrigogiraoserrao:
Should all screenshots in the docs be also snapshot tests? So that we know when they change..?
Or, at least, hashed or something, so we can be notified of changes?

davep:
> Should all screenshots in the docs be also snapshot tests? So that we know when they change..?

Ideally, yes; although I think we'd want to do this in a way that's fully automated so we don't end up with duplicated stuff that needs maintaining in two places. Dragging in all the code for snapshots should be trivial enough, but this issue here is about also taking into account the `press=` that's inline in the Markdown.

I wonder if we could have something that runs through all the `md` files and pulls out all the `{.textual ...}` and snapshots them?

That feels like a good fresh issue.

PR #2737, opened 2023-06-05
Add more unit testing for the `Input` widget

davep:
Adding more unit tests for the `Input` widget, trying to cover as much code as possible.

This gets the `Input` testing as close to 100% coverage as possible. `get_content_height` isn't being used here (and really I'm not sure what sort of testing -- short of just calling it and seeing it it's equal to 1 -- would be sensible here) and there's handling of bindings in `_on_key` that is never done, even if a bound key is pressed.

On the latter point see #2742 -- I'm pretty sure that that code can be removed (I strongly suspect it's a hangover from before some changes were made last year to how bindings are processed).

I'll nuke that bit of code once these tests are in.

### DIFF
--- a/tests/input/test_input_properties.py
+++ b/tests/input/test_input_properties.py
@@ -27,7 +27,7 @@ async def test_internal_value_password():
         assert pilot.app.query_one(Input)._value == Text("•" * len(pilot.app.TEST_TEXT))
 
 
-async def test_internal_value_hilighted():
+async def test_internal_value_highlighted():
     async with InputApp().run_test() as pilot:
         pilot.app.query_one(Input).highlighter = JSONHighlighter()
         test_text = f'{{"test": "{pilot.app.TEST_TEXT}"}}'


tests/input/test_input_mouse.py
```diff
@@ -0,0 +1,40 @@
...
+            (10, 10),
+            (len(InputApp.TEST_TEXT) - 1, len(InputApp.TEST_TEXT) - 1),
+            (len(InputApp.TEST_TEXT), len(InputApp.TEST_TEXT)),
+            (len(InputApp.TEST_TEXT) * 2, len(InputApp.TEST_TEXT)),
+        ):
+            # Note the offsets to take into account the decoration around an
+            # Input.
+            await pilot.click(Input, Offset(click_at + 3, 1))
+            await pilot.pause()
+            assert pilot.app.query_one(Input).cursor_position == should_land
```
rodrigogiraoserrao, line 30: What's the difference between having the test like this or using `pytest.mark.parametrize` to parametrise the `click_at` and `should_land` coordinates?

rodrigogiraoserrao (approved):

tests/input/test_input_mouse.py
davep: Most likely my ignorance of `pytest` and a love of writing clear code that doesn't use magic. ;-)

rodrigogiraoserrao: Well, I asked because there might be a genuine advantage to writing it like this.
What I _know_ is good about parametrising the tests is that if only part of these assertions fail, it is easier to figure out which one(s) did fail.

davep: I just tried it out after you asked and you GET MORE TESTS FROM YOUR TESTS so I'm sold. ¯\\_(ツ)_/¯ 

(technically I guess one advantage of the ignorant-of-pytest approach would be be a bit of extra speed, as you're not spinning up and ripping down an application for every parameter).

rodrigogiraoserrao: Yeah x2
You get more tests for the same number of tests! 😆 and you also spend some more time, true.

PR #2740, opened 2023-06-05
Remove unneeded arguments from `DataTable.__init__`

mrossinek:
This commit removes the following arguments from `DataTable.__init__`:
  - `show_header`
  - `show_row_labels`
  - `fixed_rows`
  - `fixed_columns`
  - `zebra_stripes`
  - `header_height`
  - `show_cursor`

This allows subclasses of `DataTable` to simply change the default reactive value on the class level instead of changing the `__init__` signature or overwriting its behavior at all (which often is not needed).
Furthermore, this makes the API of the `DataTable` more consistent with other textual widgets, where reactive attributes such as these ones are not really exposed via the `__init__` (see for example `Tree.show_root`).

And finally, these arguments seem not to have been used at all anyways, since I did not need to update any tests for this to work.

- [x] Updated documentation

<hr/>

This idea was briefly discussed with @davep yesterday on Discord. I figured I will simply go ahead with a PR to see what the dev's team stance is on this. If you find this to be an unnecessary breaking change, feel free to close this PR.

### DIFF
--- a/src/textual/widgets/_data_table.py
+++ b/src/textual/widgets/_data_table.py
@@ -299,12 +299,19 @@ class DataTable(ScrollView, Generic[CellType], can_focus=True):
     """
 
     show_header = Reactive(True)
+    """Show/hide the header row (the row of column labels)."""
     show_row_labels = Reactive(True)
+    """Show/hide the column containing the labels of rows."""
     fixed_rows = Reactive(0)
+    """The number of rows to fix (prevented from scrolling)."""
     fixed_columns = Reactive(0)
+    """The number of columns to fix (prevented from scrolling)."""
     zebra_stripes = Reactive(False)
+    """Apply zebra effect on row backgrounds (light, dark, light, dark, ...)."""
     header_height = Reactive(1)
+    """The height of the header row (the row of column labels)."""
     show_cursor = Reactive(True)
+    """Show/hide both the keyboard and hover cursor."""
     cursor_type: Reactive[CursorType] = Reactive[CursorType]("cell")
     """The type of the cursor of the `DataTable`."""
 
@@ -572,13 +579,6 @@ class DataTable(ScrollView, Generic[CellType], can_focus=True):
     def __init__(
         self,
         *,
-        show_header: bool = True,
-        show_row_labels: bool = True,
-        fixed_rows: int = 0,
-        fixed_columns: int = 0,
-        zebra_stripes: bool = False,
-        header_height: int = 1,
-        show_cursor: bool = True,
         name: str | None = None,
         id: str | None = None,
         classes: str | None = None,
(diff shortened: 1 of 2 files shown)


src/textual/widgets/_data_table.py
davep, line 1: I think I'm tentatively in favour of this. `DataTable`, possibly more than any other widget in Textual, tends to need some "keep a reference and do more stuff to it" work anyway so folk will likely set these things on the reference, not when composing. Having keyword parameters that duplicate the defaults of the reactives they're related to does feel a little heavyweight.

This would of course need marking as a breaking change and attention drawn to it. I've also not really thought through any implications beyond the deduplication of defaults.

But... chalk me up as a tentative "aye".

mrossinek: When you say:

>  need marking as a breaking change

Does that "simply" amount to prefixing the Changelog entry with `Breaking change:` or is there anything else that needs to be done?

willmcgugan:
We tend to get asked to add reactives to `__init__`. Removing them is likely to inconvenience more people than it pleases.

Granted the signature is verbose, but you could cut and paste or use `**kwargs`.

mrossinek:
> We tend to get asked to add reactives to `__init__`. Removing them is likely to inconvenience more people than it pleases.

Okay I see, thanks! Does this also mean that overwriting the default value of a reactive variable via the class attribute is not the intended way? For example what I do right now in my subclass of `Tree` is the following
```python
class MyTree(Tree[Text]):

    show_root = reactive(False)
```

If at some point you will add `show_root` to `Tree.__init__` with your default, this will break my implementation since the value of `__init__` will overwrite my class attribute.

What would be the intended way of implementing this? Always overwriting an `__init__` either with the full API at the time of writing (which might become out-of-date) or catching via `**kwargs` (and then guarding against the values I want to use a different default for, but not overwrite a user provided value):
```python
class MyTree(Tree[Text]):

        self.show_root = kwargs.get("show_root", False)
```

Can the intended way of doing this be documented somewhere by textual?

mrossinek:
> I wouldn't recommend changing the reactive defaults. There could be unintended consequences.

Thanks for clarifying :+1: I will refrain from this approach then. It might be good to document this somehow if you plan to write some sort of guide/docs for a target audience of custom widget developers.

> There are other options in Textual. If you implement an `on_mount` you can update the reactives there, and you wouldn't need to implement `__init__`.
> 
> I'd also consider not extending it at all. If you want a tree with no root, you could write a factory function, which returns a `Tree` with `show_root` set to False.

Yes, I had considered `on_mount`, too. I think I will be doing this then.
The reason I want to extend the widget is that I want to add other application specific functions. Here are two examples:

https://github.com/mrossinek/cobib/blob/fff5b67ad06cf077c36241f8282406ce9e072d1e/src/cobib/ui/components/list_view.py#L82-L89

https://github.com/mrossinek/cobib/blob/fff5b67ad06cf077c36241f8282406ce9e072d1e/src/cobib/ui/components/search_view.py#L65-L79

<hr/>

Anyways... Thanks for the feedback! Closing this PR now :+1:

PR #2744, opened 2023-06-06
Screen css

### LINKED ISSUE #2137: Add CSS to Screens
- Add CSS to Screen (in addition to default_css)
- Raise an error if CSS is defined on other widgets

rodrigogiraoserrao:
Closes #2137.
Adds `CSS` and `CSS_PATH` class variables to screens.

- [ ] Updated documentation

@willmcgugan inside `App._load_screen_css`, why don't I traverse all the rules that were read and dynamically add a CSS selector to match the screen so that screen CSS only matches the screen it was imported from?

### DIFF
--- a/src/textual/_path.py
+++ b/src/textual/_path.py
@@ -2,6 +2,34 @@ from __future__ import annotations
 
 import inspect
 from pathlib import Path, PurePath
+from typing import List, Union
+
+from typing_extensions import TypeAlias
+
+CSSPathType: TypeAlias = Union[
+    str,
+    PurePath,
+    List[Union[str, PurePath]],
+]
+"""Valid ways of specifying paths to CSS files."""
+
+
+class CSSPathError(Exception):
+    """Raised when supplied CSS path(s) are invalid."""
+
+
+def _css_path_type_as_list(css_path: CSSPathType) -> list[PurePath]:
+    paths: list[PurePath] = []
+    if isinstance(css_path, str):
+        paths = [Path(css_path)]
+    elif isinstance(css_path, PurePath):
+        paths = [css_path]
+    elif isinstance(css_path, list):
+        paths = [Path(path) for path in css_path]
+    else:
+        raise CSSPathError("Expected a str, Path or list[str | Path] for the CSS_PATH.")
+
+    return paths
 
 
 def _make_path_object_relative(path: str | PurePath, obj: object) -> Path:
--- a/src/textual/screen.py
+++ b/src/textual/screen.py
@@ -14,6 +14,7 @@ from typing import (
     Generic,
     Iterable,
     Iterator,
+    List,
     Type,
     TypeVar,
     Union,
@@ -23,13 +24,12 @@ from typing import (
 import rich.repr
 from rich.console import RenderableType
 from rich.style import Style
-from rich.traceback import Traceback
 
 from . import errors, events, messages
 from ._callback import invoke
-from ._compose import compose
 from ._compositor import Compositor, MapGeometry
 from ._context import visible_screen_stack
+from ._path import CSSPathType, _css_path_type_as_list, _make_path_object_relative
 from ._types import CallbackType
 from .binding import Binding
 from .css.match import match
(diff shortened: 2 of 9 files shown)


src/textual/_path.py
```diff
@@ -2,6 +2,34 @@
...
+    List[Union[str, PurePath]],
+]
+"""Valid ways of specifying paths to CSS files."""
+
+
+class CSSPathError(Exception):
+    """Raised when supplied CSS path(s) are invalid."""
+
+
+def _css_path_type_as_list(css_path: CSSPathType) -> list[PurePath]:
```
willmcgugan, line 21: docstring please

src/textual/screen.py
```diff
@@ -106,6 +106,19 @@ class Screen(Generic[ScreenResultType], Widget):
     Set to `""` to disable auto focus.
     """
 
+    CSS: ClassVar[str] = ""
+    """Inline CSS, useful for quick scripts. Rules here take priority over CSS_PATH.
+
+    Note:
+        This CSS is _not_ scoped to only apply to the screen.
```
willmcgugan, line 113: Can we rephrase to not use a negative? i.e. Say what it does apply to, not what it doesn't?

willmcgugan (commented):
If you were to push and pop the same screen multiple times, will it parse the CSS each time?

rodrigogiraoserrao:
> If you were to push and pop the same screen multiple times, will it parse the CSS each time?

No. That's why there is a `_css_screens_loaded` set and we check it before calling `_load_screen_css`.

What do you think of this:

> inside App._load_screen_css, why don't I traverse all the rules that were read and dynamically add a CSS selector to match the screen so that screen CSS only matches the screen it was imported from?

rodrigogiraoserrao:
> That's a weak set, right? [...]

Right, in some cases it may reparse CSS.

> Sounds like it may be worth exploring. Do you have an idea how you would implement it?

I'd load and parse the styles into a separate stylesheet.
Then I'd traverse the rules there and modify the selectors to include the screen in the beginning.
Then, I'd add the new stylesheet to the app's stylesheet.

willmcgugan:
> > Sounds like it may be worth exploring. Do you have an idea how you would implement it?
> 
> I'd load and parse the styles into a separate stylesheet. Then I'd traverse the rules there and modify the selectors to include the screen in the beginning. Then, I'd add the new stylesheet to the app's stylesheet.

Clever. Lets consider that for a future update.

tests/css/test_screen_css.py
```diff
@@ -0,0 +1,307 @@
...
+    class MyApp(BaseApp):
+        def key_p(self):
+            self.push_screen(ScreenWithCSS())
+
+        def key_o(self):
+            self.pop_screen()
+
+    counter = 0
+
+    def reparse_wrapper(reparse):
```
darrenburns, line 83: Nice 👍

src/textual/screen.py
```diff
@@ -106,6 +105,19 @@ class Screen(Generic[ScreenResultType], Widget):
...
+    """Inline CSS, useful for quick scripts. Rules here take priority over CSS_PATH.
+
+    Note:
+        This CSS applies to the whole app.
+    """
+    CSS_PATH: ClassVar[CSSPathType | None] = None
+    """File paths to load CSS from.
+
+    Note:
+        This CSS applies to the whole app.
```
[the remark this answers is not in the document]
rodrigogiraoserrao: I know. I wanted to try and restrict this via [this suggestion](https://github.com/Textualize/textual/pull/2744#issuecomment-1580354629) but I don't think I'm supposed to go ahead and do it now... :(

darrenburns (approved):
This looks good to me. Small suggestion: this feels like a good candidate for a snapshot test which could make sure that screen CSS and CSS_PATH works and the precedence is as expected.

darrenburns:
Possibly, but you don't need to change them now. When I see tests that query the DOM and check styles it feels like it's trying to programmatically do something that snapshot tests give you for free.

rodrigogiraoserrao:
> Possibly, but you don't need to change them now. When I see tests that query the DOM and check styles it feels like it's trying to programmatically do something that snapshot tests give you for free.

Ahhh ok.
When I can, I do things programmatically because it is faster than running a snapshot test. (Right?)
Also, I had multiple variations of pretty much the same thing, so that's why I thought code was better 🤷 
Thanks for your input!

willmcgugan (approved):

src/textual/screen.py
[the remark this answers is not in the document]
ndevenish: I just shot myself in the foot with this footgun.

PR #2746, opened 2023-06-06
on super class

willmcgugan:
@on decorator now matches base classes. 

Solution is much the same as https://github.com/Textualize/textual/pull/2691 with optimizations.


davep (commented):
I think this deserves an entry in the ChangeLog, and should be marked as a breaking change too. While it's unlikely someone will have negative effects from this change, it isn't impossible.

Some mention of this should likely go into the documentation too.

PR #2747, opened 2023-06-06
docs: import `Message` from textual.message not messages

### LINKED ISSUE #2745: docs: `Message` not exported from `textual.messages`
Just something minor I spotted while playing with the [Messages up](https://textual.textualize.io/guide/widgets/#messages-up) example in the docs. The code still works but my editor complains that:

> "Message" is not exported from module "textual.messages"
> Import from "textual.message" instead [reportPrivateImportUsage]

Happy to submit a quick PR if deemed worth fixing!

TomJGooding:
Closes #2745. It looks like only a couple of occurrences in the docs after searching with ripgrep.

Also appears in a couple of tests but I left those alone!

### DIFF
--- a/docs/examples/guide/compound/byte02.py
+++ b/docs/examples/guide/compound/byte02.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 from textual.app import App, ComposeResult
 from textual.containers import Container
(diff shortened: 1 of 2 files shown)


willmcgugan:
Thanks!

PR #2427, opened 2023-04-28
Add support for adding new TabPane to TabbedContent

Ljzd-PRO:
**Please review the following checklist.**

- [x] Docstrings on all new or modified functions / classes 
- [x] Updated documentation
- [x] Updated CHANGELOG.md (where appropriate)

Issue #2750, opened 2023-06-07
Preserve active message pump in call_after_refresh callback


davep:
As of the time of posting this, I'm not sure yet if this is a bug so won't mark it as such; to start with I'm looking to check the how and why of this and to see if it's expected behaviour. The code that follows is a bit involved, but it's designed to illustrate three different scenarios for posting a message and how one of them behaves different from the other two. Consider this code as a playground to test this with, not the primary illustration of the issue. See after the code for some detail.

## The Code

```python
from __future__ import annotations

from dataclasses import dataclass

from textual            import on
from textual.app        import App, ComposeResult
from textual.containers import Vertical, Horizontal
from textual.message    import Message
from textual.widget     import Widget
from textual.widgets    import Button, TextLog

class MessageSource( Widget ):

    @dataclass
    class Boop( Message ):
        source: MessageSource

        def control( self ) -> MessageSource:
            return self.source

    def send_now( self ) -> None:
        self.post_message( self.Boop( self ) )

    def send_after_refresh( self ) -> None:
        self.call_after_refresh( self.post_message, self.Boop( self ) )

    def indirect_sender( self ):
        self.post_message( self.Boop( self ) )

    def send_indirectly_after_refresh( self ) -> None:
        self.call_after_refresh( self.indirect_sender )

class ReportingContainer( Vertical ):

    @on( MessageSource.Boop )
    def report( self ) -> None:
        self.app.query_one( TextLog ).write( "Container got boop" )

class DelayedMessageApp( App[ None ] ):

    CSS = """
    Screen > Horizontal > * {
        width: 1fr;
    }

    #buttons {
        height: auto;
    }

    ReportingContainer {
        display: none;
    }
    """

    def compose( self ) -> ComposeResult:
        with Horizontal(id="buttons"):
            yield Button( "Send Now", id="now" )
            yield Button( "Send After Refresh", id="refresh" )
            yield Button( "Send Indirectly After Refresh", id="indirectly" )
        with ReportingContainer():
            yield MessageSource()
        yield TextLog()

    @on( Button.Pressed, "#now" )
    def send_now( self ) -> None:
        self.query_one( TextLog ).write( "-" * 50)
        self.query_one( MessageSource ).send_now()

    @on( Button.Pressed, "#refresh" )
    def send_after_refresh( self ) -> None:
        self.query_one( TextLog ).write( "-" * 50)
        self.query_one( MessageSource ).send_after_refresh()

[...]

PR #2751, opened 2023-06-07
Update `TabbedContent` with methods to add, remove and clear tabs/panes

davep:
This PR extends `TabbedContent` with some new methods, and also adds a new message. The new methods are:

- `add_pane` to add a new `TabPane` to the widget
- `remove_pane` to remove an existing `TabPane` from the widget
- `clear_panes` to remove all existing panes from the widget
- `tab_count` (okay, actually a property) to give easy access to the underlying `Tabs.tab_count`

The new message is `TabbedContent.Cleared`, which, like with `Tabs`, is posted when the number of tabs in the widget falls to zero.

In the process of working on this, a small change has also been made to `Underline` and `Tabs` to counter the case where an animation of the underline is still taking place when the target tab has been removed.

---
Work on this was the cause for #2750 being noted.

### DIFF
--- a/src/textual/widgets/_tabbed_content.py
+++ b/src/textual/widgets/_tabbed_content.py
@@ -74,7 +74,7 @@ class TabPane(Widget):
 
 
 class AwaitTabbedContent:
-    """An awaitable return by [`TabbedContent`][textual.widgets.TabbedContent] methods that modify the tabs."""
+    """An awaitable returned by [`TabbedContent`][textual.widgets.TabbedContent] methods that modify the tabs."""
 
     def __init__(self, *awaitables: AwaitMount | AwaitRemove) -> None:
         """Initialise the awaitable.
--- a/src/textual/widgets/_tabbed_content.py
+++ b/src/textual/widgets/_tabbed_content.py
@@ -117,6 +117,28 @@ class TabbedContent(Widget):
             yield self.tabbed_content
             yield self.tab
 
+    class Cleared(Message):
+        """Posted when there are no more tab panes."""
+
+        def __init__(self, tabbed_content: TabbedContent) -> None:
+            """Initialize message.
+
+            Args:
+                tabbed_content: The TabbedContent widget.
+            """
+            self.tabbed_content = tabbed_content
+            """The `TabbedContent` widget that contains the tab activated."""
+            super().__init__()
+
+        @property
+        def control(self) -> TabbedContent:
+            """The `TabbedContent` widget that was cleared of all tab panes.
+
+            This is an alias for [`Cleared.tabbed_content`][textual.widgets.TabbedContent.Cleared.tabbed_content]
+            and is used by the [`on`][textual.on] decorator.
+            """
+            return self.tabbed_content
+
     def __init__(
         self,
         *titles: TextType,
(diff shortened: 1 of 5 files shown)


src/textual/widgets/_tabbed_content.py
```diff
@@ -70,6 +73,26 @@ def __init__(
...
+
+        Args:
+            *awaitables: The collection of awaitables to await.
+        """
+        super().__init__()
+        self._awaitables = awaitables
+
+    def __await__(self) -> Generator[None, None, None]:
+        async def await_tabbed_content() -> None:
+            for awaitable in self._awaitables:
```
willmcgugan, line 90: Suspect we should use `asyncio.gather` here. Effect is the same, but it will be a little more efficient.

src/textual/widgets/_tabbed_content.py
```diff
@@ -151,37 +196,37 @@ def validate_active(self, active: str) -> str:
             Value of `active`.
 
         Raises:
-            ValueError: If the active attribute is set to empty string.
+            ValueError: If the active attribute is set to empty string when there are tabs available.
         """
-        if not active:
-            raise ValueError("'active' tab must not be empty string.")
+        if not active and self.get_child_by_type(ContentSwitcher).current:
```
willmcgugan, line 201: Is there any chance the dev might attempt to set `active` earlier, and there won't be a `ContentSwitcer` yet?

src/textual/widgets/_tabbed_content.py
```diff
@@ -151,37 +196,37 @@ def validate_active(self, active: str) -> str:
...
+            raise ValueError(f"'active' tab must not be empty string.")
```
willmcgugan, line 202: Superfluous f string?

src/textual/widgets/_tabbed_content.py
```diff
@@ -70,6 +73,26 @@ def __init__(
         )
 
 
+class AwaitTabbedContent:
+    """An awaitable return by [`TabbedContent`][textual.widgets.TabbedContent] methods that modify the tabs."""
```
rodrigogiraoserrao, line 77:
```suggestion
    """An awaitable returned by [`TabbedContent`][textual.widgets.TabbedContent] methods that modify the tabs."""
```

src/textual/widgets/_tabbed_content.py
```diff
@@ -70,6 +73,26 @@ def __init__(
...
+        """
+        super().__init__()
+        self._awaitables = awaitables
+
+    def __await__(self) -> Generator[None, None, None]:
+        async def await_tabbed_content() -> None:
+            for awaitable in self._awaitables:
+                await awaitable
+
+        return await_tabbed_content().__await__()
```
rodrigogiraoserrao, line 93: I need to learn more about `__await__`.
Is it a "standard" approach to implement `__await__` with a wrapper around what we care about and then call `__await__` on the wrapper?

willmcgugan (commented):
Not a full review, just a few comments

src/textual/widgets/_tabbed_content.py
davep: Good catch, likely a slip of the thinger or a hangover from a version that said more.

src/textual/widgets/_tabbed_content.py
davep: I *think* the only way that could arise is if they do that before the widget is composed, but even without this change that would end up with a similar issue due to the triggering of `_watch_active`, right? Attempting to set the active tab to a pane that doesn't even exist yet is never going to go well.

src/textual/widgets/_tabbed_content.py
davep: What sort of efficiency gain will we see here, and will it guarantee the order that they're awaited?

tests/test_tabbed_content.py
```diff
@@ -156,3 +157,104 @@ def on_tabbed_content_tab_activated(
...
+    class TabbedApp(App[None]):
+        def compose(self) -> ComposeResult:
+            yield TabbedContent()
+
+    async with TabbedApp().run_test() as pilot:
+        tabbed_content = pilot.app.query_one(TabbedContent)
+        assert tabbed_content.active == ""
+        assert tabbed_content.tab_count == 0
+        await tabbed_content.add_pane(TabPane("Test 1", id="test-1"))
+        await pilot.pause()
```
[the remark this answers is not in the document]
davep: To give the subsequent messages that end up bubbling up a chance to actually turn up. Not really an issue on GNU/Linux or macOS from what I've seen, but poor old Windows needs a moment.

src/textual/widgets/_tabbed_content.py
rodrigogiraoserrao: I suspect that `asyncio.gather` _doesn't_ guarantee waiting order.

willmcgugan: If these are `AwaitMount` and `AwaitRemove` objects, then they are running concurrently anyway. The order you await them won't change the order they are completed.

`gather` will await multiple things with less context switches. Doubt it is a tangible difference for 2, but it is the prefered way to await concurrent tasks.

davep: > If these are AwaitMount and AwaitRemove objects, then they are running concurrently anyway. The order you await them won't change the order they are completed.

Ahh, interesting; this is a possible wrinkle to worry about here then. Might need to rethink this perhaps.

davep:
This PR can also close #2427 which does something similar, although as a subset of what this PR does.

willmcgugan (approved):

Issue #2582, opened 2023-05-16
Inconsistent events when revealing a widget


davep:
This is related to #2557 in that it's something noticed while trying to get to the core of what was going on in that issue. The code that follows is a little convoluted, but it's done in a way to try and mimic how `Select` works. While possibly not the root of the issue, the idea here is that a widget is shown/hidden by having the classes of its container manipulated (this is how `Select` works). It might be that this aspect of the issue can be eliminated.

Also note that this isn't being recorded for a deep dive right now, this is more a case of recording a set of symptoms and some code for reproducing them, and then it can be looked at later.

So, the code:

```python
from textual.app        import App, ComposeResult
from textual.containers import Vertical, Horizontal
from textual.widgets    import OptionList, Button, TextLog, Label

class Wrapper( Vertical ):
    pass

class ReportingLabel( Label ):

    def _on_show( self ) -> None:
        self.app.screen.query_one(TextLog).write(f"on_show - {self.size}")

    def _on_resize( self ) -> None:
        self.app.screen.query_one(TextLog).write(f"on_resize - {self.size}")

    def _on_hide( self ) -> None:
        self.app.screen.query_one(TextLog).write(f"on_hide - {self.size}")

class ReportingList( OptionList ):

    def _on_show( self ) -> None:
        self.app.screen.query_one(TextLog).write(f"on_show - {self.size}")

    def _on_resize( self ) -> None:
        self.app.screen.query_one(TextLog).write(f"on_resize - {self.size}")

    def _on_hide( self ) -> None:
        self.app.screen.query_one(TextLog).write(f"on_hide - {self.size}")

class ShowByParentClassApp( App[ None ] ):

    CSS = """
    Wrapper > Widget {
        width: 1fr;
        display: none;
        height: auto;
    }

    Wrapper.show > Widget {
        display: block;
    }
    """

    def compose( self ) -> ComposeResult:
        with Horizontal():
            with Vertical():
                yield Button( "Toggle Option List", id="list" )
                with Wrapper( id="target-list" ):
                    yield ReportingList( "Foo", "Bar", "Baz", "Wibble", "Wobble" )
            with Vertical():
                yield Button( "Toggle Label", id="label" )
                with Wrapper( id="target-label" ):
                    yield ReportingLabel( "This is a label" )
            yield TextLog()

    def on_button_pressed( self, event: Button.Pressed ) -> None:
        self.query_one( f"#target-{event.button.id}", Wrapper ).toggle_class( "show" )

[...]

PR #2754, opened 2023-06-08
Re-raise encountered exceptions in pilot tests

darrenburns:
When an exception was raised in a pilot test, it prevented logic in `_wait_for_screen` from working correctly (counters would never decrement to zero). The result was that when exceptions occurred, it always timed out.

We now record the first exception that occurs in an app, and we re-raise that exception in the `__aexit__` of the context manager returned by `run_test`, so that tests fail and test frameworks will print the stack trace.

We now detect if an exception occurs and set an event. If the event is set, then `_wait_for_screen` won't wait for the timeout to be reached allowing it to exit early.

- [ ] Updated documentation

### DIFF
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -370,6 +370,9 @@ class App(Generic[ReturnType], DOMNode):
         """The unhandled exception which is leading to the app shutting down,
         or None if the app is still running with no unhandled exceptions."""
 
+        self._exception_event: asyncio.Event | None = None
+        """An event that will be set when the first exception is encountered."""
+
         self.title = (
             self.TITLE if self.TITLE is not None else f"{self.__class__.__name__}"
         )
@@ -1086,6 +1089,7 @@ class App(Generic[ReturnType], DOMNode):
                 message_hook_context_var.set(message_hook)
             app._loop = asyncio.get_running_loop()
             app._thread_id = threading.get_ident()
+            app._exception_event = asyncio.Event()
             await app._process_messages(
                 ready_callback=on_app_ready,
                 headless=headless,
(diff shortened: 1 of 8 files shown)


davep:
To keep in mind for when this PR settles and/or is ready to merge: the fix that's in here for `OptionList` needs to keep #2567 in mind. The fix applied here pretty much reverts the change made for #2567 -- which is fair as that change was questionable at best but solved the immediate problem.

So when this lands we need to be 100% sure that it doesn't reintroduce #2557 and if it does (I suspect it might actually), thanks to #2582 being fixed, we can likely actually fix #2557 using the rebuild in `_on_show` approach.

darrenburns:
I've confirmed that this doesn't break `Select` and doesn't reintroduce #2557.

src/textual/app.py
```diff
@@ -1782,9 +1791,17 @@ def render(renderable: RenderableType) -> list[Segment]:
...
         """Called with an unhandled exception.
 
+        Always results in the app exiting.
+
         Args:
             error: An exception instance.
         """
+        # If we're running via pilot and this is the first exception encountered,
+        # take note of it so that we can re-raise for test frameworks later.
+        if self.is_headless and self._exception is None:
```
willmcgugan, line 1801: Why the `is_headless` guard here? There are a few reasons why `is_headless` may be True, unrelated to testing.

willmcgugan (commented):
Looking good, just one question.

src/textual/app.py
darrenburns: Because I didn't know that 😄 I was looking for a "if running via pilot" flag I could use for this guard.

What are the other reasons?

willmcgugan: Docs screenshots, automation, etc.

darrenburns: Ok, well we probably want exceptions to be raised up to any code which does automated things such as the docs screenshots. If an exception occurs during a screenshot rendering, we probably want to know about it rather than continuing silently with a broken screenshot.

willmcgugan: Agree. But we want to be sure not to raise exceptions if we have generated a nice error message.

darrenburns: I think we still want to raise exceptions to prevent silent issues

willmcgugan (approved):

PR #2755, opened 2023-06-08
Pull `Tooltip` out of the main DOM

### LINKED ISSUE #2723: A lone `Static` results in a `TooManyMatches` error when using `query_one`
I've not dived into this beyond knocking up this example to isolate what I saw (about to head out of the door but wanted to record this as a reminder). With 0.27.0 (perhaps before too, just noting the version here for the record), this code:

```python
from textual.app     import App, ComposeResult

class OneStatic( App[ None ] ):

        yield Static()

    def on_mount( self ) -> None:
        self.query_one( Static ).update( "Hello, World!" )

if __name__ == "__main__":
    OneStatic().run()
```

results in a `TooManyMatches` error being raised from the `query_one`. With very early testing this only seems to be the case with `Static` (at least, I tested with `Label` and `Button` and they're fine).

I think most people would rightly find this surprising.

davep:
Fixes #2723

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Fixed setting `TreeNode.label` on an existing `Tree` node not immediately https://github.com/Textualize/textual/pull/2713
 - Correctly implement `__eq__` protocol in DataTable https://github.com/Textualize/textual/pull/2705
+- Fixed `Tooltip` causing a `query_one` on a lone `Static` to fail https://github.com/Textualize/textual/issues/2723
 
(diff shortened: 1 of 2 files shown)


willmcgugan:
Well that was easier than expected!

Makes me wonder if we shouldn't do that for other widgets.

davep:
> Makes me wonder if we shouldn't do that for other widgets.

Yeah, I'm wonder the same thing too; especially for the "top level" widgets that start out inheriting from `Widget` or `Static`; perhaps even from a container (although inheriting from one of the containers while not inheriting its CSS would make that a bit pointless, I guess).

willmcgugan (approved):

PR #2762, opened 2023-06-11
tabs: set position of added tab with add_tab

blob42:
Just added `before` and `after` passed down to the mount operation to customize the position of the added tab.

### DIFF
--- a/src/textual/widgets/_tabs.py
+++ b/src/textual/widgets/_tabs.py
@@ -312,11 +312,20 @@ class Tabs(Widget, can_focus=True):
                 pass
         return None
 
-    def add_tab(self, tab: Tab | str | Text) -> None:
+    def add_tab(self,
+                tab: Tab | str | Text,
+                before: int | str | Widget | None = None,
+                after: int | str | Widget | None = None) -> None:
         """Add a new tab to the end of the tab list.
 
         Args:
             tab: A new tab object, or a label (str or Text).
+            before: Optional location to mount before. An `int` is the index
+                of the child to mount before, a `str` is a `query_one` query to
+                find the widget to mount before.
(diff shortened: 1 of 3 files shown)


src/textual/widgets/_tabs.py
```diff
@@ -312,19 +312,30 @@ def _next_active(self) -> Tab | None:
...
+                find the widget to mount after.
         """
         from_empty = self.tab_count == 0
         tab_widget = (
             Tab(tab, id=f"tab-{self._new_tab_id}")
             if isinstance(tab, (str, Text))
             else self._auto_tab_id(tab)
         )
-        mount_await = self.query_one("#tabs-list").mount(tab_widget)
+        mount_await = self.query_one("#tabs-list").mount(tab_widget,
```
davep, line 336: With the other changes above in mind, I'd also probably do a pre-mount check here; if given a `Tab` to mount `before` or `after`, I'd check that its `parent == self` and if not raise a `Tabs`-specific exception. Likewise, if given an `id`, I'd `query_one` for it being an immediate child of the `Tabs` widget and if it can't be found raise a similar (or likely the same) exception.

davep (changes requested):
This is a good idea, makes a lot of sense. I've made some suggestions for what I think are sensible changes.

Could we also get some tests too once done? (happy to help out with that).

davep:
I am currently making some changes to all things tabs, so if you don't get the chance (we'd merge the work anyway) I'm happy to roll the idea here into there - it makes a lot of sense as an enhancement.

blob42:
Thanks 👍 looks very clean.

davep:
My pleasure, and thanks for your contribution!

PR #2767, opened 2023-06-12
Don't return anything from `AwaitMount.__call__`

davep:
Possibly the least-useful PR I've made yet; but once seen it could not be unseen.


rodrigogiraoserrao (approved):
I approve. 🚀

Issue #2658, opened 2023-05-25
A macOS Terminal mode


willmcgugan:
macOS terminal doesn't render Textual apps well. It does a poor job of borders. I think we should consider a macOS terminal mode which override border choices to make it look better on macOS.

viega:
Agreed on this; lots of people don't install more capable terminals.

NfNitLoop:
+1 to this. My context:

I'm investigating Textual to use for a project. I'd like to propose using it. But if the out-of-the-box experience on Terminal.app is as it is today, it's not a great look for our product. Telling users to [configure fonts and font spacing in Terminal.app][1], or install a different terminal to use our product is not something we want to take on. I'll probably be recommending a much simpler prompt-based UI instead of Textual as a result.

[1]: https://github.com/Textualize/textual/blob/main/FAQ.md#why-doesn't-textual-look-good-on-macos

NfNitLoop:
Just the same sort of thing mentioned in the [link] from my comment.

[link]: https://github.com/Textualize/textual/blob/main/FAQ.md#why-doesn't-textual-look-good-on-macos

ex: ![img](https://user-images.githubusercontent.com/554369/246795432-e61f3876-3dd1-4ac8-b380-22922c89c7d6.png)

In particular, border styles are a bit odd.  I think the top/bottom borders are misaligned w.r.t. the background they're drawn in? It looks worse when there are multiple on screen.  
<img width="735" alt="Screenshot 2023-06-20 at 10 04 28 AM" src="https://github.com/Textualize/textual/assets/339075/8f962027-aed8-43b4-8edb-0b5c86b9db80">

---

As an example of what I'd expect, [in python_toolkit][docs], they show this: 

![img2](https://python-prompt-toolkit.readthedocs.io/en/master/_images/ptpython-history-help.png)

Which in Terminal.app gets simplified to this (delta some themes, I think):

<img width="476" alt="Screenshot 2023-06-20 at 10 06 31 AM" src="https://github.com/Textualize/textual/assets/339075/d17c88ae-cd6e-4f37-9b40-2b536306982f">

Their [dialogs] also use simpler rendering.

[dialogs]: https://python-prompt-toolkit.readthedocs.io/en/master/_images/messagebox.png
[docs]: https://python-prompt-toolkit.readthedocs.io/en/master/pages/gallery.html

willmcgugan:
[...]

PR #2770, opened 2023-06-12
Replace some borders in default mac terminal.

rodrigogiraoserrao:
Related issues: #2658

Replaces 3 terribly-looking borders with decent-looking alternatives.

What all borders look like in the default mac terminal with the default font `Menlo Regular 13`:

Untitled

### DIFF
--- a/src/textual/_border.py
+++ b/src/textual/_border.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 from functools import lru_cache
+from os import environ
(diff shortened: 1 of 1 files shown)


rodrigogiraoserrao:
We're parking this for now.
A possible fix seems to be to set the line spacing to a smaller value.

willmcgugan:
Closing in favor of a new FAQ entry. https://github.com/Textualize/textual/blob/main/FAQ.md#why-doesn't-textual-look-good-on-macos

Issue #2771, opened 2023-06-12
Demo doesn't scroll DataTable in to view when focused


willmcgugan:
If you tab trough the widgets on `python -m textual`, the DataTable doesn't scroll in to view when it is focused. I'm certain it used to.

darrenburns:
I can't reproduce this on `main`.

willmcgugan:
Dave is seeing the same behaviour. You have to adjust to terminal size.

It doesn't actually appear to be data table related. Can occur in any control.

darrenburns:
My `main` was outdated - I'm seeing it now. This means it must be a really recent thing.

What do you mean by "can occur in any control"? Do you mean there's other widgets impacted by this?

darrenburns:
Ah ok - so it seems like the resizing is interfering in some way.

willmcgugan:
You were right. I bisected (all by myself) to the following:

```
7049014faa79c423b8fa220731e65c04509882f0 is the first bad commit
commit 7049014faa79c423b8fa220731e65c04509882f0
Author: darrenburns <darrenburns@users.noreply.github.com>
Date:   Wed May 31 16:13:21 2023 +0100

    Option to ensure scroll_to_center doesn't scroll so as to hide the top left corner of the widget (#2682)

    * Option to ensure origin of widget is visible when calling scroll to center

    * Update CHANGELOG.md

    ---------

    Co-authored-by: Will McGugan <willmcgugan@gmail.com>

 CHANGELOG.md          |  1 +
 src/textual/screen.py |  2 +-
 src/textual/widget.py | 40 +++++++++++++++++++++++++++-------------
 3 files changed, 29 insertions(+), 14 deletions(-)
```

willmcgugan:
Makes sense... https://github.com/Textualize/textual/commit/7049014faa79c423b8fa220731e65c04509882f0

willmcgugan:
Fixed in 0.28.0

PR #2773, opened 2023-06-12
Fix for scroll to center

willmcgugan:
Possible fix for https://github.com/Textualize/textual/issues/2771

This seems to fix the issue, but I'm not sure it broken the `origin_visible` functionaly, which I can't see tests for.

### DIFF
--- a/src/textual/widget.py
+++ b/src/textual/widget.py
@@ -1767,7 +1767,7 @@ class Widget(DOMNode):
         duration: float | None = None,
         easing: EasingFunction | str | None = None,
         force: bool = False,
-    ) -> None:
+    ) -> Offset:
         """Scroll to a given (absolute) coordinate, optionally animating.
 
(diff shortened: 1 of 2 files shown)


darrenburns:
From memory the origin visible is covered by a snapshot test where it scrolls the bullseye to the center.

src/textual/widget.py
```diff
@@ -2459,6 +2453,7 @@ async def _scroll_widget_to_center_of_self(
                 duration=duration,
                 easing=easing,
                 force=force,
+                top=origin_visible and widget.region.height > container.region.height,
```
darrenburns, line 2456: The reason I didn't do this approach was that I didn't think it worked for horizontal scrolling.

willmcgugan:
Replaced by https://github.com/Textualize/textual/pull/2787

PR #2776, opened 2023-06-13
Disabled container

### LINKED ISSUE #2772: Shouldn't be able to type in to a disabled input
If a widget is *enabled* and has focus, then subsequently *disabled*, you are still able to type.

When an input is disabled, it should ignore all interactions.

rodrigogiraoserrao:
This fixes #2772 by blurring the focused widget if it is a descendant of the widget that was disabled.

- [x] Updated documentation

### DIFF
--- a/src/textual/widget.py
+++ b/src/textual/widget.py
@@ -2743,12 +2743,17 @@ class Widget(DOMNode):
 
     def watch_disabled(self) -> None:
         """Update the styles of the widget and its children when disabled is toggled."""
-        if (
-            self.disabled
-            and self.app.focused is not None
-            and self in self.app.focused.ancestors_with_self
-        ):
-            self.app.focused.blur()
+        from .app import ScreenStackError
+
+        try:
+            if (
+                self.disabled
+                and self.app.focused is not None
+                and self in self.app.focused.ancestors_with_self
+            ):
+                self.app.focused.blur()
+        except ScreenStackError:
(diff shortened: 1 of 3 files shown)


src/textual/widget.py
```diff
@@ -2743,7 +2743,12 @@ def watch_has_focus(self, value: bool) -> None:
 
     def watch_disabled(self) -> None:
         """Update the styles of the widget and its children when disabled is toggled."""
-        self.blur()
+        if (
+            self.disabled
+            and self.app.focused is not None
+            and self.app.focused in self.walk_children()
```
willmcgugan, line 2749: It is somewhat inefficient to walk the DOM every time something is disabled.

How about we just check if `self.app.focused` is disabled or has a disabled ancestor?

rodrigogiraoserrao: 🤔 as in a45b156?

willmcgugan: Yeah, that looks right to me.

rodrigogiraoserrao: But the tests don't like it. Why does this require a screen on the stack..?

rodrigogiraoserrao: @willmcgugan I'd say you have to check if the optimisation is worth the extra mess created to handle the cases where the `disabled` status is set outside the context of a running app.

willmcgugan (approved):

PR #2777, opened 2023-06-14
Style review

willmcgugan:
This is an attempt to harmonize our widgets, form widgets in particular. Consider this a first pass, there is still work to be done.

I almost replaced the "3D" buttons with flat buttons, but I'm not quite ready to give them up.
Screenshot 2023-06-14 at 10 56 42

Screenshot 2023-06-14 at 10 56 46


davep (approved):
I think this needs a huge "BREAKING CHANGE" entry in the ChangeLog, otherwise looking good.

PR #2785, opened 2023-06-15
docs: fix typos in geometry

TomJGooding:
Just a couple of typos spotted while delving into Textual geometry!


davep (approved):

davep:
Thanks for those Tom, good catch as always.

PR #2788, opened 2023-06-15
Ensure column width change results in virtual size update

### LINKED ISSUE #2782: DataTable resize not happening on column resize and also resizing oddly
When updating a cell inside of a DataTable, if the width of the updated cell goes outside of the window bounds the DataTable container doesn't resize to allow scrolling so you can't see the updated content. Oddly enough, if you update the cell a second time the DataTable container does resize and the scrollbars are drawn. You can see the Resize method being called from the console on the second update but not the first.

And a weird side-effect of the resize (from a user view/interaction stand point), is the last row of the table which was previously viewable is "covered" by the horizontal scrollbar after the resize. I tried digging into the Widget `refresh()` method, but it seemed like somewhere I shouldn't be messing around without more knowledge of the API.

CleanShot 2023-06-14 at 22 04 42

CleanShot 2023-06-14 at 22 07 21@2x

CleanShot 2023-06-14 at 22 07 32@2x

CleanShot 2023-06-14 at 22 07 42@2x

The script for this demo is based on your DataTable example script from the docs.

```python
from textual.widgets import DataTable, Footer

ROWS = [
    (1, "Aleksandr Sadovnikov", "Russia", 51.84),
    (10, "Darren Burns", "Scotland", 51.84),
]


class TableApp(App):
    BINDINGS = [
        Binding("ctrl+u", "update_records", "Update Records", show=True),
    ]

        yield DataTable()
        yield Footer()

    def on_mount(self) -> None:
        table.cursor_type = "row"
        table.zebra_stripes = True
        table.add_columns(*ROWS[0])
        table.add_rows(ROWS[1:])

    def action_update_records(self):
        table.update_cell_at(
            (0, 3),
            "[Disqualified]",
            update_width=True,
        )


app = TableApp()
if __name__ == "__main__":
    app.run()
```

# Textual Diagnostics

## Versions

| Name    | Value  |
|---------|--------|
| Textual | 0.27.0 |
| Rich    | 13.3.5 |

## Python

| Name           | Value                                                       |
|----------------|-------------------------------------------------------------|
| Version        | 3.11.3                                                      |
| Implementation | CPython                                                     |
| Compiler       | Clang 14.0.0 (clang-1400.0.29.202)                          |
| Executable     | /Users/jbd/Dropbox/DEV/projects/renamer/venv/bin/python3.11 |

## Operating System

| Name    | Value                                                                                                 |
|---------|-------------------------------------------------------------------------------------------------------|
| System  | Darwin                                                                                                |
| Release | 22.5.0                                                                                                |
| Version | Darwin Kernel Version 22.5.0: Mon Apr 24 20:52:24 PDT 2023; root:xnu-8796.121.2~5/RELEASE_ARM64_T6000 |

## Terminal

| Name                 | Value              |
|----------------------|--------------------|
| Terminal Application | iTerm.app (3.4.19) |
| TERM                 | xterm-256color     |
| COLORTERM            | truecolor          |
| FORCE_COLOR          | *Not set*          |
| NO_COLOR             | *Not set*          |

## Rich Console options

| size           | width=79, height=36 |
| max_width      | 79                  |
| max_height     | 36                  |

darrenburns:
Fixes #2782 

- [ ] Updated documentation

### DIFF
--- a/src/textual/widgets/_data_table.py
+++ b/src/textual/widgets/_data_table.py
@@ -1125,6 +1125,8 @@ class DataTable(ScrollView, Generic[CellType], can_focus=True):
             else:
                 column.content_width = max(new_content_width, label_width)
 
+        self._require_update_dimensions = True
+
     def _update_dimensions(self, new_rows: Iterable[RowKey]) -> None:
         """Called to recalculate the virtual (scrollable) size."""
         console = self.app.console
@@ -1425,6 +1427,13 @@ class DataTable(ScrollView, Generic[CellType], can_focus=True):
         whole DataTable and re-computing column widths after some cells
         have been updated. This is more efficient in the case of high
         frequency updates, ensuring we only do expensive computations once."""
+        if self._updated_cells:
+            # Cell contents have already been updated at this point.
+            # Now we only need to worry about measuring column widths.
+            updated_columns = self._updated_cells.copy()
+            self._updated_cells.clear()
+            self._update_column_widths(updated_columns)
+
         if self._require_update_dimensions:
             # Add the new rows *before* updating the column widths, since
             # cells in a new row may influence the final width of a column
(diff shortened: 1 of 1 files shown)


src/textual/widgets/_data_table.py
```diff
@@ -1125,6 +1125,8 @@ def _update_column_widths(self, updated_cells: set[CellKey]) -> None:
             else:
                 column.content_width = max(new_content_width, label_width)
 
+        self._require_update_dimensions = True
```
darrenburns, line 1128: We're updating the width of a column, so we need to set this flag to ensure the virtual size gets updated next time we go on_idle.

Issue #2789, opened 2023-06-15
Tooltips don't work as expected on compound widgets


davep:
This stems from someone [reporting on Discord that tooltips don't seem to work on `ProgressBar`](https://discord.com/channels/1026214085173461072/1033754296224841768/1118918219823063213). Turns out they're right and it doesn't work for reasons that make sense, but I think we should allow for.

Consider this code:

```python
from textual.app     import App, ComposeResult
from textual.widgets import ProgressBar

class ProgressTipApp( App[ None ] ):

    def compose( self ) -> ComposeResult:
        yield ProgressBar( 100 )

    def on_mount( self ) -> None:
        self.query_one( ProgressBar ).tooltip = "Hello from the Progress bar!"

if __name__ == "__main__":
    ProgressTipApp().run()
```

When run, no matter what part of the `ProgressBar` you hover over, you don't see the tooltip. People would, of course, expect that you do. The problem stems from the fact that the thing with the tooltip is the container, but the visible widget is really all of the component parts, none of which have a tooltip.

To get the desired effect a developer would have to do something like this:

```python
from textual.app     import App, ComposeResult
from textual.widgets import ProgressBar

class ProgressTipApp( App[ None ] ):

    def compose( self ) -> ComposeResult:
        yield ProgressBar( 100 )

    def on_mount( self ) -> None:
        for part in self.query_one( ProgressBar ).query( "*" ):
            part.tooltip = "Hello from the Progress bar!"

if __name__ == "__main__":
    ProgressTipApp().run()
```

Perhaps the tooltip system should, if the widget under the cursor has no tooltip, check for one in its ancestors in the DOM?

PR #2790, opened 2023-06-15
Inherited tooltips

davep:
If a widget's tooltip is checked for, and it doesn't have one, one will be looked for within its ancestors. This allows for the setting of tooltips on compound widgets and solves #2789.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Fixed issue where internal data of `OptionList` could be invalid for short window after `clear_options` https://github.com/Textualize/textual/pull/2754
 - Fixed `Tooltip` causing a `query_one` on a lone `Static` to fail https://github.com/Textualize/textual/issues/2723
 - Nested widgets wouldn't lose focus when parent is disabled https://github.com/Textualize/textual/issues/2772
+- Tooltips now work with compound widgets https://github.com/Textualize/textual/issues/2789
 
 ### Changed
 
--- a/src/textual/widget.py
+++ b/src/textual/widget.py
@@ -447,6 +447,8 @@ class Widget(DOMNode):
     @property
     def tooltip(self) -> RenderableType | None:
         """Tooltip for the widget, or `None` for no tooltip."""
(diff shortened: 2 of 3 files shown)


1j01:
`child.tooltip = None; child.tooltip != None` seems unintuitive to me.
Wouldn't it be better handled outside of the getter, where the tooltip code tests whether widgets have tooltips?

davep:
> `child.tooltip = None; child.tooltip != None` seems unintuitive to me.

I don't recognise the code you're quoting there, so I'm struggling to follow the question.

willmcgugan:
I think @1j01 is pointing out that the property may return a different value from the getter, than was set via the setter.

To avoid this, we could do the DOM inheritance logic outside of the property.

davep:
> I think @1j01 is pointing out that the property may return a different value from the getter, than was set via the setter.

But this is what we want here: setting a widget's tooltip to `None` would mean "inherit if possible or failing that don't have a tooltip". It would seem odd to have to repeat the logic of finding an appropriate tip, in multiple places, if tooltips were used for multiple purposes.

If we don't want that, that's fair enough: but what I've suggested here would be in keeping with what I'd expect from a DOM-based tooltip property; in my own app's code I'd expect to be able to ask a widget what its tooltip is and have it provide the value that would appear on the screen; without personally having to repeat that or call some other method.

1j01:
Sorry, I should've been clearer. I was trying to be too pithy.

I think the behavior of looking up ancestor elements that have a tooltip set within the property getter would be surprising, and it doesn't match e.g. the DOM in web browsers.
Imagine trying to write a function to toggle a tooltip:
```py
def toggle_tooltip(widget) -> None:
  widget.tooltip = widget.css_identifier_styled if widget.tooltip is None else None

toggle_tooltip(parent) # should work
toggle_tooltip(parent) # should work
toggle_tooltip(child) # should work
toggle_tooltip(child) # should work
toggle_tooltip(parent) # should work
toggle_tooltip(child) # would fail
```

If the parent's tooltip is set, toggling the child will fail, right?
(I will try to build a demo app to show this properly.)

davep:
Eh. Don't go to that trouble. I'm happy to nuke this approach if folk don't see the utility or elegance of the approach. More than happy to defer to something else.

Issue #2781, opened 2023-06-14
`Markdown`: Sections missing in the rendering


nkh:
50% of this file is missing in the rendering https://github.com/nkh/ftl/blob/main/config/ftl/man/ftl.md, all the other renderes I use renders the missing sections, the link above also shows a rendering.

davep:
Thanks; given that this really comes down to how the Textual `Markdown` widget sees the file you point to, I've moved the issue to here.

nkh:
There is no MD handling on my side, I just run: "frogmouth file-name" on my command line, and 50% is not rendered.

I've given the same input to 4 different renderers and it's the same input I give to pandoc to generate a man page.

nkh:
here's a minimum markdown that doesn't work:


## General *ftl* Commands

        «?»             Show this man page

nkh:
'''
## General *ftl* Commands

        «?»             Show this man page 
'''

nkh:
## General *ftl* Commands
    
            «?»             Show this man page

davep:
Nice job! Thanks. As minimal examples go it doesn't get better than that! 👍

davep:
Thanks for diving in and checking this @TomJGooding, very helpful (good recollection of the other issue too). Looking like `Markdown` does need a bit more love soon (there's a couple or so Frogmouth issues that would benefit from `Markdown` being extended).

willmcgugan:
Going to put this on the ToDo for next week.

Issue #2676, opened 2023-05-28
`Markdown` widget: Code blocks in lists are incorrectly ordered, sometimes don't appear


jamescooke:
## Current behaviour

Very similar to this old issue in Rich: https://github.com/Textualize/rich/issues/1036

When we use the same content (a list with a code block):

```sh
cat > order.md
```

```md
This is a test.

1. first item
2. second item

       code block

Normal text
```

And then render it with Frogmouth:

```sh
frogmouth order.md
```

Then the code block does not appear:

![Screenshot from 2023-05-28 16-34-15](https://github.com/Textualize/frogmouth/assets/781059/8f664ba9-2607-47b8-8c3f-1940da95d03b)

If we tweak the code block to use triple backticks, then the block appears, but out of order:

```sh
cat > order2.md
```

```md
This is a test.

1. first item
2. second item

   ``
   code block
   ``

Normal text
```

> Please forgive my use of a double backtick above - I wasn't able to get a triple backtick to render in a code block in this markdown issue body 😬 . I've attached the file for confirmation: [order2.md](https://github.com/Textualize/frogmouth/files/11584996/order2.md)

```sh
frogmouth order2.md
```

![Screenshot from 2023-05-28 16-36-40](https://github.com/Textualize/frogmouth/assets/781059/969241ca-d0ce-40d6-902a-e350b9e3c286)

## Expected behaviour

Render code block as part of the second item consistently.

Rich appears to render both markdown excerpts OK (and they appear the same). Tested with `rich==13.3.5`
![Screenshot from 2023-05-28 18-50-20](https://github.com/Textualize/frogmouth/assets/781059/b616c5e4-e964-4412-9abd-f0785ec4982b)

davep:
Given `Markdown` is based on `markdown-it`, the first thing to do here is be sure how it handles things. Looks like it renders this as expected:

![Screenshot 2023-05-28 at 20 55 32](https://github.com/Textualize/textual/assets/28237/cca94d19-260a-4fd5-beb7-26d8dc895805)

As such, at first glance, this would appear to be a bug in `Markdown`.

jamescooke:
Thanks for moving the issue from Frogmouth @davep 

I've confirmed the behaviour is in the Textual `Markdown` widget as follows:

* Used the `markdown.py` example file [from the docs](https://textual.textualize.io/widgets/markdown/#__tabbed_1_2)
[...]

Issue #2380, opened 2023-04-26
List items in a `Markdown` can receive focus


davep:
It's possible that there are other items within `Markdown` for which this is true, and when I get a moment to test this more I'll update this issue if this is the case, but right now a `Markdown` that is showing markdown text that has ordered or unordered lists within it will have the focus chain work through each of the list items.

This can be seen by running this code and pressing <kbd>Tab</kbd> often:

```python
from textual.app        import App, ComposeResult
from textual.containers import VerticalScroll
from textual.widgets    import Header, Footer, Markdown

MARKDOWN = """\
# Header

## Sub Header

- List
- List
- List
- List
- List
- List

1. Another List
1. Another List
1. Another List
1. Another List
1. Another List
1. Another List
"""

class MarkdownFocusApp( App[ None ] ):

    CSS = """
    *:focus {
        border: double red;
    }
    """

    def compose( self ) -> ComposeResult:
        yield Header()
        with VerticalScroll():
            yield Markdown(MARKDOWN)
        yield Footer()

if __name__ == "__main__":
    MarkdownFocusApp().run()
```

PR #2803, opened 2023-06-19
`Markdown` improvements

davep:
A small selection of `Markdown` QoL changes and fixes, including:

1. Adds an `unhandled_token` method, which in `Markdown` itself is a no-op (giving the effect of what happened before this was added), but also lets anyone inherit from `Markdown` and handle other MarkdownIT tokens -- possibly those added via extensions that aren't included in core Textual. This is also useful for testing purposes too as the method can be overridden to raise an error rather than ignore unknown tokens.
2. Exports `MarkdownBlock` from `textual.widgets.markdown`; pretty necessary for anyone wanting to use the above[^1].
3. Starts some `Markdown`-specific unit testing. Just the basics for now, but this at least provides some baseline tests and also a framework for further testing.
4. Adds a fix for #2781 (code blocks not appearing; not to be confused with fenced blocks).
5. By extension adds a fix for half of #2676.
6. Adds a fix for the main half of #2676 -- ensuring that code blocks in lists are rendered inline.
7. Adds a fix for #2380 

[^1]: I am thinking it might be a good idea to export all the major Markdown widgets in this way.

### DIFF
--- a/src/textual/widgets/_markdown.py
+++ b/src/textual/widgets/_markdown.py
@@ -654,7 +654,6 @@ class Markdown(Widget):
         Returns:
             Either a widget to be added to the output, or `None`.
         """
-        del token
         return None
 
     def update(self, markdown: str) -> None:
--- a/src/textual/widgets/_markdown.py
+++ b/src/textual/widgets/_markdown.py
@@ -790,7 +790,7 @@ class Markdown(Widget):
                             style_stack.pop()
 
                 stack[-1].set_content(content)
-            elif token.type == "fence":
+            elif token.type in ("fence", "code_block"):
                 output.append(
                     MarkdownFence(
                         self,
--- a/src/textual/widgets/_markdown.py
+++ b/src/textual/widgets/_markdown.py
@@ -801,7 +801,7 @@ class Markdown(Widget):
             else:
                 external = self.unhandled_token(token)
                 if external is not None:
-                    output.append(external)
+                    (stack[-1]._blocks if stack else output).append(external)
 
         self.post_message(Markdown.TableOfContentsUpdated(self, table_of_contents))
         with self.app.batch_update():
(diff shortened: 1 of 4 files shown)


src/textual/widgets/_markdown.py
```diff
@@ -644,6 +645,18 @@ async def load(self, path: Path) -> bool:
...
+    def unhandled_token(self, token: Token) -> MarkdownBlock | None:
+        """Process an unhandled token.
+
+        Args:
+            token: The token to handle.
+
+        Returns:
+            Either a widget to be added to the output, or `None`.
+        """
+        del token
```
willmcgugan, line 657: Why the `del` here? Pretty sure it won't do anything since there will be a reference in the caller.

davep: Just silence linters about the unused parameter; fairly normal idiom for that. It keeps the name for the documentation but tells linters and the like that it's not mistakenly unused.

willmcgugan: Haven't see that one. I'm not keen on code which does something pointless to keep a linter happy. Is there another way to quell the error?

davep: It's the best one I've seen, similar to using `assert` to give a big hint for typing. The alternatives I'm aware of are just not paying attention to the linter, doing `_ = that_unused_argument`, or including linter-choice-specific pragma comments.

Happy to go with an alternative you're more comfortable with.

willmcgugan: I think I'd just go for a pragma comment here.

davep: Thinking about it, I don't think there's a pragma that covers pylint, pyright and pylance (to pick 3 that will complain about an unused argument). I guess given it's happening a lot in the rest of Textual, it's preferable to just let them complain.

willmcgugan (approved):

Issue #2616, opened 2023-05-19
Strange behaviour trying to work with opacity in screens


TomJGooding:
Turning discussion #2611 into an issue to be looked at as a possible bug. There seems to be some strange behaviour when trying to work with opacity in screens - if this functionality should be supported in Textual? 

I've created a simple example below which hopefully demonstrates the issue. The screen only contains a single widget, and the opacity of both the screen and the widget are initially set to 0%. But when you start the app, you'll see there are **two** red boxes (the background colour of the screen) until the actual widget "fades in" to green. When you hit the space bar to trigger the animation to fade in the screen, the second box disappears.

The example from @xyxz-web in the original discussion also shows the duplicated widget issue with a more complex example. 

```python
from textual.app import App, ComposeResult
from textual.screen import ModalScreen, Screen
from textual.widgets import Footer, Static


class OpacityScreen(Screen):
    DEFAULT_CSS = """
    OpacityScreen {
        align: center middle;
        opacity: 0%;
        background: red;
    }

    OpacityScreen > Static {
        opacity: 0%;
        background: green;
        width: auto;
        content-align: center middle;
    }
    """

    BINDINGS = [("space", "fade_in", "Fade in screen")]

    def compose(self) -> ComposeResult:
        yield Static("Widget with fade in animation")
        yield Footer()

    def on_mount(self) -> None:
        widget = self.query_one(Static)
        widget.styles.animate("opacity", value=1.0, duration=2.0)

    def action_fade_in(self) -> None:
        self.styles.animate("opacity", value=1.0, duration=2.0)


class TestApp(App):
    def on_mount(self) -> None:
        self.push_screen(OpacityScreen())


if __name__ == "__main__":
    app = TestApp()
    app.run()
```

willmcgugan:
The issue there is that you have opacity 0% on the *screen*. Which would normally allow content beneath it to show through. There is nothing below the screen so it won't refresh properly.

That's something we should special case for the screen.

If you don't set the screen opacity to 0%, I think the behaviour makes sense?

xyxz-web:
If you disable setting the screen opacity:
1. The duplicate element on the left side still appears
2. The Screen doesn't fade in, just the widget changes colors

TomJGooding:
> The issue there is that you have opacity 0% on the _screen_. Which would normally allow content beneath it to show through. There is nothing below the screen so it won't refresh properly.
> 
> That's something we should special case for the screen.

[...]

PR #2814, opened 2023-06-20
fix for opacity

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/2616 and more.

Opacity can now be layered. Which means you can have a translucent thing on top of another translucent thing. This created an issue with disabling widgets. If you have a disabled container of widgets, the default styling made them doubly transparent because the default rule target the containers. To solve this, I added a `can-focus` pseudo class so that you can target only widgets that can possibly be focused.

### DIFF
--- a/src/textual/css/_style_properties.py
+++ b/src/textual/css/_style_properties.py
@@ -1023,8 +1023,9 @@ class FractionalProperty:
     string percentage (e.g. '10%'). Values will be clamped to the range (0, 1).
     """
 
-    def __init__(self, default: float = 1.0):
+    def __init__(self, default: float = 1.0, children: bool = False):
         self.default = default
+        self.children = children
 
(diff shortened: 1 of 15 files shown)


src/textual/css/_style_properties.py
```diff
@@ -1023,8 +1023,9 @@ class FractionalProperty:
     string percentage (e.g. '10%'). Values will be clamped to the range (0, 1).
     """
 
-    def __init__(self, default: float = 1.0):
+    def __init__(self, default: float = 1.0, children: bool = False):
```
darrenburns, line 1026: Could you add a docstring for this please?

darrenburns (approved):
Looks good. One request for docstring.

davep (approved):
What @darrenburns said.

PR #2850, opened 2023-06-27
Further `textual-dev` changes

davep:
Following on from the initial tests of the `textual-dev` branch...

**NOTE:** Set up for merging into `textual-dev` branch again.


README.md
```diff
@@ -43,10 +43,10 @@ Textual runs on Linux, macOS, and Windows. Textual requires Python 3.7 or above.
 Install Textual via pip:
 
 ```
-pip install "textual[dev]"
+pip install "textual"
```
willmcgugan, line 46: Think we can drop the quotes now.

docs/guide/animation.md
```diff
@@ -72,7 +72,7 @@ You can specify which easing method to use via the `easing` parameter on the `an
 
 !!! note
 
-    The `textual easing` preview requires the `dev` extras to be installed (using `pip install textual[dev]`).
+    The `textual easing` preview requires the development extras to be installed (using `pip install textual-dev`).
```
willmcgugan, line 75: The `textual` command is not in "extras" any more. Suggest we refer to it as "textual devtools" package. Or explicitly as "textual-dev"

docs/getting_started.md
```diff
@@ -22,13 +22,13 @@ Textual requires Python 3.7 or later (if you have a choice, pick the most recent
 
 You can install Textual via PyPI.
 
-If you plan on developing Textual apps, then you should install `textual[dev]`. The `[dev]` part installs a few extra dependencies for development.
+If you plan on developing Textual apps, then you should install `textual-dev`; this installs a few extra dependencies for development.
```
willmcgugan, line 25: I think we should spell out the two imports in two code fences.

Start with a suggestion to `pip install textual`. Then a recommendation to `pip install textual-dev` for the dev tools.

I know that technically only `textual-dev` is required, but I think this will help to keep the two packages distinct in the developers mind. We want to avoid devs thinking they need to add `textual-dev` to their package dependancies.

willmcgugan (changes requested):

willmcgugan (approved):

PR #2864, opened 2023-06-29
Fix race condition during shutdoen under Windows

paul-ollis:
The terminal writer thread could send escape sequences when the terminal was unable to process then; i.e. when not in virtual mode.

The following fixes have been made.

- Switch the terminal to virtual mode before the writer thread is started and any control sequences are queued to the writer thread.

- Wait for the writer thread to finish before switching the terminal out of virtual mode.

I have seen this cause problems at application start up and shut down. The changed code appears to provide stable behaviour. I have not attempted to add any new tests due to the difficulty of reliably reproducing the problem.

- [ ] Updated documentation

### DIFF
--- a/src/textual/drivers/windows_driver.py
+++ b/src/textual/drivers/windows_driver.py
@@ -76,10 +76,10 @@ class WindowsDriver(Driver):
         """Start application mode."""
         loop = asyncio.get_running_loop()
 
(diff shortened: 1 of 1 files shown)


willmcgugan (approved):

willmcgugan:
THanks

Issue #2879, opened 2023-07-02
stdout  redirect shouldn't set sys.stdout to None


willmcgugan:
It looks as though the `redirect_stdout(None)` (and stderr version) sets `sys.stdout` to `None`.

Nothing has broken up until now, which suggest that the `print` function treats `None` as a special case. However, this won't be the case where `sys.stdout` is expected to contain a file like object.

This is the root cause of https://github.com/Textualize/textual/issues/2877 and https://github.com/Textualize/textual/issues/2878

The redirected stdout should go to dev/null. See `os.devnull`

rodrigogiraoserrao:
Closed with #2880.

alexiri:
@willmcgugan This fix seems to be missing from the changelog of 0.29.0, but it's included, right?

rodrigogiraoserrao:
> This fix seems to be missing from the changelog of 0.29.0, but it's included, right?

Wooops, when we reviewed your PR #2880 we should've asked you to add the fix entry to the changelog, but we forgot and we didn't add it later.
Your fix should be in 0.29.0 regardless of this omission from the changelog.

Issue #2878, opened 2023-07-02
Textual does not play nicely with tqdm


alexiri:
I'm trying to develop a Textual interface for some libraries that use [`tqdm`](https://github.com/tqdm/tqdm) to display a progress bar while loading some data at startup. The progress bar can't be disabled, and it's making the Textual app crash.

Here's a minimal example that demonstrates the issue:

```python
from textual.app import App
from tqdm import tqdm

class OtherClass:
    def __init__(self):
        for i in tqdm(range(10000)):
            pass

class MyApp(App):
    def on_mount(self) -> None:
        self.other = OtherClass()

if __name__ == "__main__":
    app = MyApp()
    app.run()
```

And here's the crash:
```
AttributeError: 'NoneType' object has no attribute 'write'
Exception ignored in: <function tqdm.__del__ at 0x7fd916532d40>
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/site-packages/tqdm/std.py", line 1145, in __del__
    self.close()
  File "/usr/local/lib/python3.10/site-packages/tqdm/std.py", line 1274, in close
    if self.last_print_t < self.start_t + self.delay:
AttributeError: 'tqdm' object has no attribute 'last_print_t'
```

I wouldn't mind so much not being able to see the progress bar, but the app really shouldn't crash.

alexiri:
Fixed by #2880.

PR #2880, opened 2023-07-02
Redirect output to /dev/null

### LINKED ISSUE #2877: Textual interferes with other libraries' use of logging
I'm trying to create a Textual app that creates an instance of an object that eventually loads the [`transformers` module](https://pypi.org/project/transformers/), and this fails due to something that module is doing with the logging module. Here's a minimal example that demonstrates the issue:

```python
import logging, sys
from textual.app import App
from textual.logging import TextualHandler

logging.basicConfig(
    level="NOTSET",
    handlers=[TextualHandler()],
)

class OtherClass:
    def __init__(self):
        # https://github.com/huggingface/transformers/blob/main/src/transformers/utils/logging.py#L87
        _default_handler = logging.StreamHandler()  # Set sys.stderr as stream.                                                        │
        _default_handler.flush = sys.stderr.flush

class MyApp(App):
    def on_mount(self) -> None:
        self.other = OtherClass()

if __name__ == "__main__":
    app = MyApp()
    app.run()
```

This fails with the error `AttributeError: 'NoneType' object has no attribute 'flush'`.

`TextualHandler` doesn't alter this behavior. The workaround I'm currently using is defining the instance of OtherClass in the main, before creating MyApp, but that does make some of the code a lot more cumbersome and it shouldn't be necessary.

alexiri:
Fixes #2877, #2878 and #2879.

- [ ] Updated documentation

### DIFF
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -1978,8 +1978,8 @@ class App(Generic[ReturnType], DOMNode):
                                 with redirect_stdout(redirector):  # type: ignore
                                     await run_process_messages()
                         else:
-                            with redirect_stderr(None):
-                                with redirect_stdout(None):
+                            with open(os.devnull, "w") as thevoid:
+                                with redirect_stdout(thevoid), redirect_stderr(thevoid):
                                     await run_process_messages()
 
                 finally:
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -1978,8 +1978,8 @@ class App(Generic[ReturnType], DOMNode):
                                 with redirect_stdout(redirector):  # type: ignore
                                     await run_process_messages()
                         else:
-                            with open(os.devnull, "w") as thevoid:
-                                with redirect_stdout(thevoid), redirect_stderr(thevoid):
+                            with open(os.devnull, "w") as null_file:
+                                with redirect_stdout(null_file), redirect_stderr(null_file):
                                     await run_process_messages()
 
                 finally:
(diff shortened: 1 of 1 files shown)


src/textual/app.py
```diff
@@ -1978,8 +1978,8 @@ async def invoke_ready_callback() -> None:
                                 with redirect_stdout(redirector):  # type: ignore
                                     await run_process_messages()
                         else:
-                            with redirect_stderr(None):
-                                with redirect_stdout(None):
+                            with redirect_stderr(open(os.devnull, 'w')):
```
willmcgugan, line 1981: The file handle should really be explicitly closed. You could probably get away with opening a single file.

alexiri: You're absolutely right, sorry about that.

willmcgugan: No worries!

src/textual/app.py
```diff
@@ -1978,8 +1978,8 @@ async def invoke_ready_callback() -> None:
                                 with redirect_stdout(redirector):  # type: ignore
                                     await run_process_messages()
                         else:
-                            with redirect_stderr(None):
-                                with redirect_stdout(None):
+                            with open(os.devnull, "w") as thevoid:
```
willmcgugan, line 1981: We prefer camel case words. I think there is a more apt term than "void" here. Suggest `null_file`.

willmcgugan (changes requested):

src/textual/app.py
alexiri: `null_file` makes the line too long and `black` complains. Should I reflow the line even though it's ugly, or should I use the "standard" `f` as a file handle?

willmcgugan: I'd break it out on to another line. Don't mind that. No single letter variables though...

alexiri: Ok, done.

willmcgugan (approved):
Thanks

PR #2881, opened 2023-07-03
Added suggester to complete input multiply times by last entered word

lakdemon:

- [ ] Updated documentation

### DIFF
--- a/src/textual/suggester.py
+++ b/src/textual/suggester.py
@@ -104,7 +104,7 @@ class SuggestFromList(Suggester):
     """
 
     def __init__(
-            self, suggestions: Iterable[str], *, case_sensitive: bool = True
+        self, suggestions: Iterable[str], *, case_sensitive: bool = True
     ) -> None:
         """Creates a suggester based off of a given iterable of possibilities.
 
--- a/src/textual/suggester.py
+++ b/src/textual/suggester.py
@@ -104,7 +104,7 @@ class SuggestFromList(Suggester):
     """
 
     def __init__(
-        self, suggestions: Iterable[str], *, case_sensitive: bool = True
+            self, suggestions: Iterable[str], *, case_sensitive: bool = True
     ) -> None:
         """Creates a suggester based off of a given iterable of possibilities.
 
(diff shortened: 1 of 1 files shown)


rodrigogiraoserrao:
Hey @lakdemon, thanks for your PR.
Do you mind giving me a bit more of context as to what your PR is trying to do, what the rationale for adding this is, etc?

Thanks!

lakdemon:
I use this class in my application and I think it might be useful to have it in textual.
The matter is that SuggestFromList can add a phrase only once at the beginning of input.
For example:
**"S"**_elect something from ..._
**"I"**_nput data into ..._
This class can offer a hint for each word as you type.
For example dictionary name
**"S"**_elect_
**"Select f"**_rom_
**"Select from t"**_able_

This is much more convenient in some cases and may come in handy later for TextArea inputs i think.
If someone is interested in this, then I can try to add tests and a description for the PR

lakdemon:
suggestion_example

rodrigogiraoserrao:
@willmcgugan I need your input on this!
Pun intended!

rodrigogiraoserrao:
> I'm afraid I don't quite grasp what this is doing over the existing suggester. I would need example(s)

See the GIF here: https://github.com/Textualize/textual/pull/2881#issuecomment-1622640222

This suggester only completes the last word whereas the suggester that is already implemented completes the whole input.

willmcgugan:
Sounds like a good idea in principle. The PR would need a bit of work to go in to the core library.

src/textual/suggester.py
```diff
@@ -136,3 +136,19 @@ async def get_suggestion(self, value: str) -> str | None:
...
+class SuggestLastWordFromList(SuggestFromList):
+    """
+        Give suggestions for completing each last word in the text based on a fixed list of options.
+    """
+
+    async def get_suggestion(self, value: str) -> str | None:
+        if not value.endswith(' '):
+            values = value.rsplit(' ', 1)
+            last_word = values[-1]
+            for idx, suggestion in enumerate(self._for_comparison):
```
willmcgugan, line 150: Probable shouldn't be using "private" data here. Suggest directly suggesting from `Suggester`.

rodrigogiraoserrao:
> The PR would need a bit of work to go in to the core library.

I'm happy to work with @lakdemon to get the PR in a good state for us to review.

src/textual/suggester.py
willmcgugan, line 150: We have a policy of no abbreviations, so `index` would be preferred over `id`

src/textual/suggester.py
```diff
@@ -136,3 +136,19 @@ async def get_suggestion(self, value: str) -> str | None:
             if suggestion.startswith(value):
                 return self._suggestions[idx]
         return None
+
+
+class SuggestLastWordFromList(SuggestFromList):
+    """
```
willmcgugan, line 142: Docstring could use more detail. See other docstrings for inspiration.

willmcgugan (changes requested):

rodrigogiraoserrao:
Hey @lakdemon, are you happy to go over the requested changes or do you want to close this PR?

willmcgugan:
Closing for now. Assumed stale.

My gut feeling is that it doesn't necessarily belong in the core library.

PR #2891, opened 2023-07-05
Ignore CancelledErrors because of Python 3.7

### LINKED ISSUE #2854: `CancelledError` when doing a `self.remove` from a timer (in Python 3.7 but not Python 3.11)
Having run into this while working on notifications, and puzzled over this with @willmcgugan when first seen, this is me isolating the problem into its own code/issue for further investigation.

For anyone looking at this, I've already uncovered what might be one really important bit of information: **THIS DEPENDS ON PYTHON VERSION**. I get the error I'm about to describe when working with Python 3.7, but *not* with Python 3.11 (I've yet to test 3.8, 3.9 or 3.10).

Given this code:

```python
from textual import on
from textual.containers import Vertical, VerticalScroll

class SelfRemovingLabel(Label):

    def on_mount(self) -> None:
        self.set_timer(2, self.remove)

class RemoveOnTimerApp(App[None]):

    CSS = """
    Label {
        margin-bottom: 1;
        width: 1fr;
        border: solid $warning;
        background: $warning 20%;
    }
    """

        with Vertical():
            yield Button("Add self-removing label")
            yield VerticalScroll()

    @on(Button.Pressed)
    def add_ephemeral_label(self):
        self.query_one(VerticalScroll).mount(SelfRemovingLabel("I will remove myself!"))

if __name__ == "__main__":
    RemoveOnTimerApp().run()
```

with Python 3.7 and the latest Textual, if you mash the button and wait for a moment for a `Label` to remove itself, you'll get the following error:

```
╭────────────────────────────────────────────────────────── Traceback (most recent call last) ───────────────────────────────────────────────────────────╮
│ /Users/davep/develop/python/textual/src/textual/await_remove.py:33 in await_prune                                                                      │
│                                                                                                                                                        │
│   30 │   def __await__(self) -> Generator[None, None, None]:                                                                                           │
│   31 │   │   async def await_prune() -> None:                                                                                                          │
│   32 │   │   │   """Wait for the prune operation to finish."""                                                                                         │
│ ❱ 33 │   │   │   await self.finished_flag.wait()                                                                                                       │
│   34 │   │                                                                                                                                             │
│   35 │   │   return await_prune().__await__()                                                                                                          │
│   36                                                                                                                                                   │
│                                                                                                                                                        │
│ ╭──────────────────────────── locals ─────────────────────────────╮                                                                                    │
│ │ self = <textual.await_remove.AwaitRemove object at 0x1020fed88> │                                                                                    │
│ ╰─────────────────────────────────────────────────────────────────╯                                                                                    │
│                                                                                                                                                        │
│ /Users/davep/.pyenv/versions/3.7.16/lib/python3.7/asyncio/locks.py:293 in wait                                                                         │
│                                                                                                                                                        │
│   290 │   │   fut = self._loop.create_future()                                                                                                         │
│   291 │   │   self._waiters.append(fut)                                                                                                                │
│   292 │   │   try:                                                                                                                                     │
│ ❱ 293 │   │   │   await fut                                                                                                                            │
│   294 │   │   │   return True                                                                                                                          │
│   295 │   │   finally:                                                                                                                                 │
│   296 │   │   │   self._waiters.remove(fut)                                                                                                            │
│                                                                                                                                                        │
│ ╭───────────────────────── locals ─────────────────────────╮                                                                                           │
│ │  fut = <Future cancelled>                                │                                                                                           │
│ │ self = <asyncio.locks.Event object at 0x1020ff1c8 [set]> │                                                                                           │
│ ╰──────────────────────────────────────────────────────────╯                                                                                           │
╰────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
CancelledError
```

If `SelfRemovingLabel` is modified to add a level of indirection to the remove, for example:

```python
class SelfRemovingLabel(Label):

    def on_mount(self) -> None:
        self.set_timer(2, self._my_remove)

    def _my_remove(self) -> None:
        self.remove()
```

it works fine.

rodrigogiraoserrao:
This fixes #2854 by ignoring the `CancelledError`.

### DIFF
--- a/src/textual/await_remove.py
+++ b/src/textual/await_remove.py
@@ -3,7 +3,7 @@
 An *optionally* awaitable object returned by methods that remove widgets.
 """
 
-from asyncio import Event, Task
+from asyncio import CancelledError, Event, Task
 from typing import Generator
 
 
@@ -30,6 +30,11 @@ class AwaitRemove:
     def __await__(self) -> Generator[None, None, None]:
         async def await_prune() -> None:
             """Wait for the prune operation to finish."""
-            await self.finished_flag.wait()
+            # Ignore CancelledErrors while 3.7 is supported.
+            # C.f. https://github.com/Textualize/textual/issues/2854
+            try:
+                await self.finished_flag.wait()
+            except CancelledError:
+                pass
 
         return await_prune().__await__()
--- a/tests/test_await_remove.py
+++ b/tests/test_await_remove.py
@@ -18,3 +18,5 @@ async def test_multiple_simultaneous_removals():
     # The app should run and finish without raising any errors.
     async with RemoveOnTimerApp().run_test() as pilot:
         await pilot.pause(0.3)
+        # Sanity check to ensure labels were removed.
+        assert len(pilot.app.query(Label)) == 0
--- a/tests/test_await_remove.py
+++ b/tests/test_await_remove.py
@@ -1,0 +1,20 @@
+from textual.app import App
+from textual.widgets import Label
+
+
+class SelfRemovingLabel(Label):
+    def on_mount(self) -> None:
+        self.set_timer(0.2, self.remove)
+
+
+class RemoveOnTimerApp(App[None]):
+    def on_mount(self):
+        for _ in range(5):
+            self.mount(SelfRemovingLabel("I will remove myself!"))
+
+
+async def test_multiple_simultaneous_removals():
+    """Regression test for https://github.com/Textualize/textual/issues/2854."""
+    # The app should run and finish without raising any errors.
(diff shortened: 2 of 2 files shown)


willmcgugan:
On consideration, this is unlikely to be the right fix. You could get in to a situation where cancelling a widget's task would get swallowed if it happens to be awaiting a remove.

I *think* what is happening is that Python3.7 raises an erroneous CancelledError when waiting a lock that has been cancelled.

The solution may be to explicitly handle `CancelledError` in `prune_widgets_task` and set the lock there.

rodrigogiraoserrao:
Closed in favour of #2895

PR #2903, opened 2023-07-08
Expose animation callback in scroll methods

jayghoshter:

- [ ] Updated documentation

Allows using callbacks after the end of the animation (or end of scrolling) in `scroll` methods in `Widget` class via the `on_complete=` keyword argument.

This is my first PR to this project, so I will need some help with any contribution protocol or additional modifications that might be required. 

Thanks for the great tool!

### DIFF
--- a/src/textual/widget.py
+++ b/src/textual/widget.py
@@ -45,6 +45,7 @@ from ._animator import DEFAULT_EASING, Animatable, BoundAnimator, EasingFunction
 from ._arrange import DockArrangeResult, arrange
 from ._asyncio import create_task
 from ._cache import FIFOCache
+from ._callback import invoke
 from ._compose import compose
 from ._context import NoActiveAppError, active_app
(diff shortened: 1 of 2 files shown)


src/textual/widget.py
```diff
@@ -1760,6 +1765,9 @@ def _scroll_to(
                 self.scroll_target_y = self.scroll_y = y
                 scrolled_y = scroll_y != self.scroll_y
 
+            if on_complete is not None:
```
willmcgugan, line 1768: I think you will need `self.call_after_refresh(on_complete)` here which will allow coroutines to be used, and will ensure the update is called after the screen has updated.

willmcgugan (commented):
Looks good, just one request.

src/textual/widget.py
jayghoshter: fixed! Thanks for going through it.

willmcgugan (approved):

PR #2904, opened 2023-07-08
linear gradient

willmcgugan:
Adds a linear gradient renderable. At some point we will expose this via CSS.

Forgive the lack of tests. I'll add tests once linear gradients are exposed to the dev.

https://github.com/Textualize/textual/assets/554369/843b9cae-9b2f-4337-86dd-b7641fdee8cc

### DIFF
--- a/src/textual/renderables/gradient.py
+++ b/src/textual/renderables/gradient.py
@@ -76,8 +76,16 @@ class LinearGradient:
         from_color = Style.from_color
 
         @lru_cache
-        def get_rich_color(color_index: int) -> RichColor:
-            return get_color(color_index / 255).rich_color
+        def get_rich_color(color_offset: int) -> RichColor:
+            """Get a Rich color in the gradient.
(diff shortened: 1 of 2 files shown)


src/textual/renderables/gradient.py
```diff
@@ -35,3 +39,90 @@ def __rich_console__(
...
+        cos_angle = cos(angle_radians)
+
+        center_x = width / 2
+        center_y = height
+
+        new_line = Segment.line()
+
+        color_gradient = Gradient(*self._stops)
+
+        _Segment = Segment
```
davep, line 74: Hangover from an iteration of the code?

davep (approved):

src/textual/renderables/gradient.py
willmcgugan: That's turning a global in to a local, which is faster. Use in L97

PR #2906, opened 2023-07-09
fuzzy matcher

willmcgugan:
Some initial work for the "command palette", which will use "fuzzy matching".

This implements the fuzzy part. Commands will be matched against the user's input. A match is recorded if all the users characters are in the command and in the correct order.

### DIFF
--- a/src/textual/_fuzzy.py
+++ b/src/textual/_fuzzy.py
@@ -27,7 +27,7 @@ class Matcher:
             input: Input string to match against.
 
         Returns:
-            Strength if match from 0 to 1.
+            Strength of the match from 0 to 1.
         """
         cached = self._cache.get(input)
         if cached is not None:
--- a/src/textual/_fuzzy.py
+++ b/src/textual/_fuzzy.py
@@ -1,0 +1,83 @@
+from re import compile, escape
+
+import rich.repr
+from rich.text import Text
+
+from ._cache import LRUCache
+
+
+@rich.repr.auto
+class Matcher:
+    """A fuzzy matcher."""
+
+    def __init__(self, query: str) -> None:
+        """
+        Args:
+            query: A query as typed in by the user.
+        """
+        self.query = query
+        self._query_regex = ".*?".join(f"({escape(character)})" for character in query)
+        self._query_regex_compiled = compile(self._query_regex)
+        self._cache: LRUCache[str, float] = LRUCache(1024 * 4)
+
+    def match(self, input: str) -> float:
+        """Match the input against the query
+
+        Args:
+            input: Input string to match against.
+
+        Returns:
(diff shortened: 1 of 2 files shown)


src/textual/_fuzzy.py
```diff
@@ -0,0 +1,82 @@
...
+        self._cache: LRUCache[str, float] = LRUCache(1024 * 4)
+
+    def match(self, input: str) -> float:
+        """Match the input against the query
+
+        Args:
+            input: Input string to match against.
+
+        Returns:
+            Strength if match from 0 to 1.
```
davep, line 30:
```suggestion
            Strength of the match from 0 to 1.
```

src/textual/_fuzzy.py
```diff
@@ -0,0 +1,82 @@
...
+            for offset in offsets:
+                if offset > last_offset + 1:
+                    group_count += 1
+                last_offset = offset
+
+            score = 1.0 - ((group_count - 1) / len(input))
+        self._cache[input] = score
+        return score
+
+    def highlight(self, input: str) -> Text:
```
davep, line 54: I wonder if it makes sense here to optionally allow feeding in a no-match and match style, so that the caller can get the matcher to style the output just as it needs for display?

src/textual/_fuzzy.py
davep, line 1: Is it intended that the following is the result?

```python
>>> x = Matcher("test")
>>> x.match("test")
1.0
>>> x.match(" test ")
1.0
>>> x.match("a test ")
1.0
>>> x.match("a t e s t test ")
0.8
```

I suspect people will be surprised that the final match scores lower than the other matches when the query exists exactly in the input, I also wonder if people might be surprised that an exact match (the first test) scores exactly the same as the subsequent tests, where the match is a substring.

src/textual/_fuzzy.py
```diff
@@ -0,0 +1,82 @@
...
+class Matcher:
+    """A fuzzy matcher."""
+
+    def __init__(self, query: str) -> None:
+        """
+        Args:
+            query: A query as typed in by the user.
+        """
+        self.query = query
+        self._query_regex = ".*?".join(f"({escape(character)})" for character in query)
```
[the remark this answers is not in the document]
willmcgugan: Debugging mostly. Might be handy to keep around.

src/textual/_fuzzy.py
willmcgugan: Yeah, nice idea. Will leave that for a future update.

src/textual/_fuzzy.py
willmcgugan: Don't think that would occur much in practice, but yeah we can probably improve the scoring function. Will save that for later.

PR #2930, opened 2023-07-12
Fix Toggle Buttons on a selection list

dmunozv04:

- [ ] Updated documentation


dmunozv04:
This PR makes the Toggle Buttons on a selection list clickable/hoverable by adding the style meta option property

davep:
Good catch. Would you mind adding a CHANGELOG entry to that effect too?

I sense we should also add an extra unit test or two to ensure that clicking on the checkbox has the desired effect (I'm happy to help with that if you're not familiar with adding tests to Textual).

dmunozv04:
As I’m not familiar with adding tests, I’d appreciate getting help on doing that

davep:
> I’ve updated the changelog

Thanks.

> As I’m not familiar with adding tests, I’d appreciate getting help on doing that

No worries; when I get a moment I'll add an appropriate test for this.

davep:
Right, that all looks like it's going through okay; I'll see about adding the extra test in the morning.

Thanks again for this @dmunozv04 -- good spot!

davep:
Tests added; as I've added code to this too I've requested review from the rest of the team.

willmcgugan (approved):

PR #2931, opened 2023-07-12
Text area

darrenburns:
Adds `TextArea`.

<details>
<summary>Themes (have since been _slightly_ updated</summary>

Textual Snapshot Test Report

</details>

<details>
<summary>Sandbox/demo app</summary>

```python

from pathlib import Path

from rich.console import RenderableType

from textual import on
from textual.document import VALID_LANGUAGES, SyntaxTheme
from textual.widgets import Footer, Static, TextArea, Pretty, OptionList, DirectoryTree, \
    Select
from textual.widgets._option_list import Option

PYTHON_SNIPPET = """\
def render_line(self, y: int) -> Strip:
    '''Render a line of the widget. y is relative to the top of the widget.'''

    row_index = y // 4  # A checkerboard square consists of 4 rows

    if row_index >= 8:  # Generate blank lines when we reach the end
        return Strip.blank(self.size.width)

    is_odd = row_index % 2  # Used to alternate the starting square on each row

    white = Style.parse("on white")  # Get a style object for a white background
    black = Style.parse("on black")  # Get a style object for a black background

    # Generate a list of segments with alternating black and white space characters
    segments = [
        Segment(" " * 8, black if (column + is_odd) % 2 else white)
        for column in range(8)
    ]
    strip = Strip(segments, 8 * 8)
    return strip
"""

TEXT = """I must not fear.
Fear is the mind-killer.
Fear is the little-death that brings total obliteration.
I will face my fear.
I will permit it to pass over me and through me.
And when it has gone past, I will turn the inner eye to see its path.
Where the fear has gone there will be nothing. Only I will remain.
"""


class NavigationOption(Option):

    def __init__(self, prompt: RenderableType, start_point: tuple[int, int],
                 end_point: tuple[int, int]):
        super().__init__(prompt)
        self.start_point = start_point
        self.end_point = end_point


class TextAreaDemo(App):
    CSS = """\
    * {
        scrollbar-size-vertical: 1;
    }
    TextArea {
        height: 1fr;
        background: $background-lighten-1;
    }
    #file-tree {
        width: 25%;
    }
    #debug {
        height: auto;
        background: $primary-darken-1 25%;
        padding: 1 2;
    }
    #options {
        dock: bottom;
        align-horizontal: right;
        height: auto;
        margin-bottom: 1;
        width: 40;
    }
    """

    BINDINGS = [
        Binding("ctrl+l", "gutter", "Toggle Gutter"),
    ]

        text_area = TextArea()

        code_path = Path(
            "/Users/darrenburns/Code/textual/src/textual/widgets/_data_table.py")

        text_area.load_text(code_path.read_text())
        # text_area.load_text("AB CD  EFG\n    HI\tJK  LM ")
        text_area.language = "python"

        with Horizontal():
            yield DirectoryTree(Path(".").resolve(), id="file-tree")
            yield text_area

        with Horizontal(id="options"):
            yield Select(
                [(lang, lang) for lang in VALID_LANGUAGES],
                prompt="Language",
                id="language-select",
            )
            yield Select(
                [(theme.name, theme.name) for theme in SyntaxTheme.available_themes()],
                prompt="Theme",
                value=SyntaxTheme.default().name,
                allow_blank=False,
                id="theme-select",
            )
        yield Footer()
        # yield Pretty(id="debug", object=text_area.selection)

    def on_mount(self):
        editor = self.query_one(TextArea)
        # self.watch(editor, "selection", self.update_debug)
        # self.watch(editor, "language", self.update_debug)

    @on(DirectoryTree.FileSelected)
    def navigate_to_definition(self, event: DirectoryTree.FileSelected) -> None:
        editor = self.query_one(TextArea)
        editor.load_text(event.path.read_text())

    @on(Select.Changed, selector="#language-select")
    def update_language(self, event: Select.Changed) -> None:
        language = event.value
        if language is None or language in VALID_LANGUAGES:
            editor = self.query_one(TextArea)
            editor.language = language

    @on(Select.Changed, selector="#theme-select")
    def update_theme(self, event: Select.Changed) -> None:
        theme = event.value
        if theme is not None:
            editor = self.query_one(TextArea)
            editor.theme = theme

    def update_debug(self):
        editor = self.query_one(TextArea)
        debug = self.query_one("#debug")
        debug.update(editor.selection)

    def action_gutter(self):
        editor = self.query_one(TextArea)
        editor.show_line_numbers = not editor.show_line_numbers


app = TextAreaDemo()
if __name__ == '__main__':
    app.run()

```

</details>

## Initial release

- Currently aiming for week of ~~21st August 2023~~ ~~28th August 2023~~. We're now awaiting release of `py-tree-sitter` binary wheels. If this doesn't happen this week we'll likely publish our own wheels.
- Syntax highlighting: powered by tree-sitter, we'll initially support a small subset of languages (python, html, markdown, sql, css, json, yaml, toml). You can plug in your own parser. Many tree-sitter parsers are available online. Adding your own theme is essentially just mapping tree-sitter tokens to rich `Style` objects.
- Keyboard support: arrow keys, word left/right, line start/end, page up/down. Hold shift + arrow key to select.
- Editing: delete left/right, delete word left/right, delete to start/end of line, kill line, delete selection, more. It's pretty simple to add your own operations using the API.
- Mouse support: click document to move cursor, click + drag to select, mouse wheel to scroll.
- API: quite a lot of utility methods for inspecting and altering the content, so adding your own operations should be straightforward.
- Line numbers, tab render width, choose tabs vs spaces.
- Can be customised using syntax highlighting themes and CSS (change fg/bg of cursor, gutter, selection, cursor line and more).

bbbfa2aa-e4d3-4b3d-8faa-9225fc498bc8

b24ea471-9b24-42ad-a57f-6bafd755a5b4

e4bd5652-a98e-4117-a8dd-0c82f9d014e1

image

text-area-matching-brackets

## Not in this release

- Read-only mode
- Wrapping
- Auto-indent (e.g. in Python files), auto-bullet point (in Markdown files) - should be easy to add in subclasses now if needed
- There is no internal clipboard
- Undo/redo - the internal machinery is there so I'm hoping this can come soon
- No utility methods for reading/writing to disk
- ~~The behaviour of cursor word left/right will be improved~~ Done!


Menschomat:
Looks great!

davep:
Assorted notes from an initial play with the PR, rather than spam the PR with lots of comments, I'll just add things to this as I play around:

- Some indication that an empty line is part of a selection would be nice.
Screenshot 2023-08-09 at 11 06 04

davep:
Following on from [the previous comment](https://github.com/Textualize/textual/pull/2931#issuecomment-1671050644), some more testing the following day so I'll start a new comment.

- `textual._languages` feels like it *shouldn't* be prefixed with an `_` as it looks like it would be useful to someone developing some sort of editor, to have access to the list of supported languages out of the box.
- Related to the above, I feel `VALID_LANGUAGES` should be a slightly richer structure that gives some handy hints for detecting those languages; especially from filename extensions anyway. Or... perhaps the module should have a `infer_language` function that takes a `Path` and does its best to figure out the language. Even a simple approach to this would be really useful.

davep:
Yes but should we build another toy editor like all those others or... 😛

davep:
@darrenburns Just tested vim: it *doesn't* go to the start/end of the line either. :-P

davep:
[Found another that doesn't](https://helix-editor.com/)!

davidbrochart:
I am trying this PR with [collaborative editing in mind](https://twitter.com/davidbrochart/status/1677407167874318336?s=20), where the text can be modified programmatically, not only by the user pressing keys. For instance, the following code emulates someone typing from another frontend, using `TextArea.insert_text_range`:
```py
import asyncio

async def insert(text_area: TextArea):
    i = 0
    while True:
        await asyncio.sleep(1)
        text_area.insert_text_range(str(i), from_location=(0, i), to_location=(0, i))
        i += 1
        if i == 10:
            break


class MyTextArea(TextArea):
    async def on_mount(self):
        asyncio.create_task(insert(self))


class MyApp(App):

        text_area = MyTextArea()
        yield text_area


if __name__ == "__main__":
    app = MyApp()
    app.run()
```
If I hit the left arrow key to move the cursor to the left, the cursor position is reset to the end of the line each time the other user inserts a character.
Do I need to handle the cursor position as well as inserting text, or am I missing something?
Anyways, this looks really promising, I can't wait to use it in [jpterm](https://github.com/davidbrochart/jpterm)!

darrenburns:
@davidbrochart Good shout, I've updated the API to include a `move_cursor` boolean which is `False` by default. Does this solve your problem?

You can do e.g. `text_area.insert_text("Hello, world!", location=(2, 5))` and it will insert text at (2, 5) without moving the cursor.

davidbrochart:
> @davidbrochart Good shout, I've updated the API to include a `move_cursor` boolean which is `False` by default. Does this solve your problem?

Not quite. Now the cursor doesn't move at all, but when working collaboratively on a document, if a character is inserted by the other user before or at my cursor position, I expect my cursor to move to the right. If a character is inserted after my cursor position, I expect my cursor to not move.

sharkusk:
This feature is looking amazing.  Will there be a read-only mode (for code browsing)?

darrenburns:
@davidbrochart I see, I misunderstood your original comment. I think you have a solid use-case, but it's possibly outside of the scope of what should be handled within the `TextArea`. These cursor movement requirements may be something to handle within a subclass - hopefully the `TextArea` API offers tools to make this relatively painless.

@Textualize/dev anyone other thoughts on this?

darrenburns:
@sharkusk Read-only is unlikely to make the initial release but I could see it coming pretty soon after as it's not too difficult and pretty high value. What do you think @willmcgugan?

davidbrochart:
> we're looking to make this work such that `move_cursor=False` means "keep the cursor on the same character", so the _location/coordinate_ will actually be adjusted, but from the user POV it will remain at the same piece of text.

It seems that the `move_cursor` parameter didn't make it in `TextArea.replace()`? With this code:
```py
import asyncio

async def insert(text_area: TextArea):
    for i in range(10):
        await asyncio.sleep(1)
        text_area.replace(str(i), start=(0, i), end=(0, i))


class MyTextArea(TextArea):
    async def on_mount(self):
        asyncio.create_task(insert(self))


class MyApp(App):

        text_area = MyTextArea()
        yield text_area


if __name__ == "__main__":
    app = MyApp()
    app.run()
```
I get the following bevahior:

https://github.com/Textualize/textual/assets/4711805/d9b62cd6-30d4-470d-96ee-3718dd72d2fe

It almost works, maybe an off-by-one error?

darrenburns:
@davidbrochart Good catch, seems like it isn't quite correct when the insertion happens at the cursor location.

Reproduced with this test case (the middle parameterisation fails):

```python
@pytest.mark.parametrize(
    "cursor_location,insert_location,cursor_destination",
    [
        ((0, 3), (0, 2), (0, 4)),  # API insert just before cursor
        ((0, 3), (0, 3), (0, 4)),  # API insert at cursor location
        ((0, 3), (0, 4), (0, 3)),  # API insert just after cursor
    ],
)
async def test_insert_character_near_cursor_maintain_selection_offset(
    cursor_location,
    insert_location,
    cursor_destination,
):
    app = TextAreaApp()
    async with app.run_test():
        text_area = app.query_one(TextArea)
        text_area.load_text("012345")
        text_area.move_cursor(cursor_location)

        selected_text_before = text_area.selected_text
        text_area.insert("X", location=insert_location)
        assert text_area.selected_text == selected_text_before
        assert text_area.selection == Selection.cursor(cursor_destination)
```

pyproject.toml
```diff
@@ -59,7 +61,9 @@ httpx = "^0.23.1"
 types-setuptools = "^67.2.0.1"
 textual-dev = "^1.1.0"
 pytest-asyncio = "*"
-pytest-textual-snapshot = "0.2.0"
+pytest-textual-snapshot = "0.2.0"  # TODO: pinned while while we resolve issue #3042
```
davep, line 64: Just to note: Once @willmcgugan deploys the latest (0.4.0) https://github.com/Textualize/pytest-textual-snapshot to PyPi, this should not be an issue.

darrenburns:
@davidbrochart That issue should hopefully be fixed now, thanks for reporting it!

src/textual/widgets/_text_area.py
```diff
@@ -0,0 +1,1266 @@
...
+class Undoable(Protocol):
+    """Protocol for actions performed in the text editor which can be done and undone.
+
+    These are typically actions which affect the document (e.g. inserting and deleting
+    text), but they can really be anything.
+
+    To perform an edit operation, pass the Edit to `TextArea.edit()`"""
+
+    def do(self, text_area: TextArea) -> object | None:
+        """Do the action."""
```
davep, line 1249: `Args` and `Returns` for the docstring.

src/textual/widgets/_text_area.py
```diff
@@ -0,0 +1,1266 @@
...
+
+    def undo(self, text_area: TextArea) -> object | None:
+        """Undo the action."""
```
davep, line 1252: `Args` and `Returns` for the docstring.

davep (commented):
Just a first pass over the PR; nothing substantial. Mainly just marking some docstring-related things that may or may not get missed.

src/textual/document/_languages.py
```diff
@@ -0,0 +1,11 @@
+VALID_LANGUAGES = [
+    "markdown",
+    "yaml",
+    "sql",
+    "css",
+    "html",
+    "json",
+    "python",
+    "regex",
+    "toml",
```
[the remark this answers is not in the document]
darrenburns: Hmm 🤔 strange!

src/textual/_fix_direction.py
```diff
@@ -0,0 +1,19 @@
+from __future__ import annotations
+
+
+def _sort_ascending(
```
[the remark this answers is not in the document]
darrenburns: Yes you're right 👍 - this utility originally did a little more and I never replaced it.

src/textual/document/_document.py
```diff
@@ -0,0 +1,366 @@
...
+
+        Args:
+            tab_width: The width to use for tab indents.
+
+        Returns:
+            The size (width, height) of the document.
+        """
+        lines = self._lines
+        cell_lengths = [cell_len(line.expandtabs(tab_width)) for line in lines]
+        max_cell_length = max(cell_lengths or [0])
```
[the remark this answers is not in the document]
darrenburns: Didn't know about this - good call.

ofek:
This is awesome! I do want to bring to your attention however that this would require every user to have a compiler for installation because `tree-sitter` does not ship binary wheels for the most part https://pypi.org/project/tree-sitter/#files

src/textual/widgets/_text_area.py
```diff
@@ -0,0 +1,1542 @@
...
+
+    def action_cursor_right(self) -> None:
+        """Move the cursor one location to the right.
+
+        If the cursor is at the end of a line, attempt to go to the start of the next line.
+        """
+        target = self.get_cursor_right_location()
+        self.move_cursor(target)
+
+    def action_cursor_right_select(self):
```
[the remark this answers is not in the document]
darrenburns: I've removed this method now and folded it into `action_cursor_right` with a `select` param.

darrenburns:
@ofek Good shout - I hope they can release soon 🙏  If not, we may need to look at publishing wheels ourselves in the interim.

ofek:
> There is no internal clipboard

Does this mean only no copying or also not being able to paste into it?

darrenburns:
@ofek For now, you can paste from your OS clipboard, but you can't copy to your OS clipboard.

src/textual/document/_document.py
```diff
@@ -0,0 +1,366 @@
...
+        """The Size of the document, taking into account the tab rendering width.
+
+        Args:
+            tab_width: The width to use for tab indents.
+
+        Returns:
+            The size (width, height) of the document.
+        """
+        lines = self._lines
+        cell_lengths = [cell_len(line.expandtabs(tab_width)) for line in lines]
```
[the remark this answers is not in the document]
darrenburns: > If lines is a list of str, then I'm afraid expandtabs won't work for us, as it is unaware of double cell characters. The Rich Text class can expand tabs in a double-cell aware manner.

Good catch - I wonder if it's worth wrapping line in a `Text` instance here or if we should extract `expandtabs` from `Text` such that I can run it against a string without constructing a new `Text` instance.

> This function could be quite inefficient. Might be significant with a large enough document. Is this called a lot, is there room for caching?

This method is top of my list for optimisation. I suspect it's where a majority of the time is spent on pressing a key at the moment, although I haven't confirmed.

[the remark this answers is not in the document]
willmcgugan: > Good catch - I wonder if it's worth wrapping line in a `Text` instance here or if we should extract `expandtabs` from `Text` such that I can run it against a string without constructing a new `Text` instance.

We could, yeah. `Text.expandtabs` has to do additional work to handle spans, which we wouldn't need here.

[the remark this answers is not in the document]
darrenburns: Some information from profiling this:

- When no syntax highlighting is used this method `Document.get_size` accounts for about 70% of the time taken when you press a key.
- When syntax highlighting is used, I can't find this method on the profile - I think it's because it's dwarfed by the time taken doing highlighting (93% of the time taken on pressing a key is related to syntax highlighting). `SyntaxAwareDocument._prepare_highlights` should probably be the first thing we target with optimisation.

tl;dr: `Document.get_size` and `SyntaxAwareDocument._prepare_highlights` are the two largest bottlenecks on keypresses.

src/textual/document/_syntax_aware_document.py
```diff
@@ -0,0 +1,362 @@
...
+        elif column == encoded_row_length:
+            return _utf8_encode(newline[0])
+        elif column == encoded_row_length + 1:
+            if newline == "\r\n":
+                return b"\n"
+
+        return b""
+
+
+@lru_cache(maxsize=128)
```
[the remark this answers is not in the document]
darrenburns: This is only used within lines, so at most it would be mapping 128 bytes objects to relatively small dictionaries.

src/textual/widgets/_text_area.py
```diff
@@ -0,0 +1,1542 @@
...
+
+        # Render the gutter and the text of this line
+        gutter_segments = self.app.console.render(gutter)
+        text_segments = self.app.console.render(
+            line,
+            self.app.console.options.update_width(virtual_width),
+        )
+
+        # Crop the line to show only the visible part (some may be scrolled out of view)
+        gutter_strip = Strip(gutter_segments)
```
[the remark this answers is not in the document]
darrenburns: Good shout!

src/textual/widgets/_text_area.py
```diff
@@ -0,0 +1,1542 @@
...
+        If we're in the middle of a row, and go down to a row with no
+        content, then we go down to another row, we want our cursor to
+        jump back to the same offset that we were originally at.
+        """
+        row, column = self.selection.end
+        column_cell_length = self.get_column_width(row, column)
+        self._last_intentional_cell_width = column_cell_length
+
+    # --- Editor operations
+    def insert(
```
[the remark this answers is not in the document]
darrenburns: What about after `location`?

`text_area.insert("hello", (2, 0))` feels clear to me without writing out the keyword arg name so I think we should keep that open, but `text_area.insert("hello", (2, 0), False)` is where it gets unclear.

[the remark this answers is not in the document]
willmcgugan: Yeah, I agree.

ofek:
tree-sitter has wheels now FYI so all good on that front: https://pypi.org/project/tree-sitter/0.20.2/#files

src/textual/document/_syntax_aware_document.py
```diff
@@ -0,0 +1,270 @@
...
+try:
+    from tree_sitter import Language, Node, Parser, Tree
+    from tree_sitter.binding import Query
+    from tree_sitter_languages import get_language, get_parser
+
+    TREE_SITTER = True
+except ImportError:
+    TREE_SITTER = False
+
+from textual.document._document import Document, EditResult, Location, _utf8_encode
```
willmcgugan, line 12: Could these be relative imports?

src/textual/document/_syntax_aware_document.py
```diff
@@ -0,0 +1,270 @@
...
+from textual.document._languages import BUILTIN_LANGUAGES
+
+
+class SyntaxAwareDocumentError(Exception):
+    """General error raised when SyntaxAwareDocument is used incorrectly."""
+
+    pass
```
willmcgugan, line 19: Superfluous `pass`

src/textual/widgets/_text_area.py
```diff
@@ -0,0 +1,1857 @@
...
+from collections import defaultdict
+from dataclasses import dataclass, field
+from functools import lru_cache
+from pathlib import Path
+from typing import TYPE_CHECKING, Any, Iterable, Optional, Tuple
+
+from rich.style import Style
+from rich.text import Text
+
+from textual._text_area_theme import TextAreaTheme
```
willmcgugan, line 13: Absolute imports again. Any reason for doing it that way? We've used relative elsewhere.

src/textual/widgets/_text_area.py
```diff
@@ -0,0 +1,1857 @@
...
+class LanguageDoesNotExist(Exception):
+    """Raised when the user tries to use a language which does not exist.
+    This means a language which is not builtin, or has not been registered.
+    """
+
+    pass
+
+
+@dataclass
+class TextAreaLanguage:
```
willmcgugan, line 74: docstring pls

willmcgugan (approved):
Just nitpicks

docs/widgets/text_area.md
```diff
@@ -0,0 +1,468 @@
...
+- The [`TextArea.get_text_range`][textual.widgets._text_area.TextArea.get_text_range] method returns the text between two locations.
+
+In all cases, when multiple lines of text are retrieved, the [document line separator](#line-separators) will be used.
+
+### Editing content inside `TextArea`
+
+The content of the `TextArea` can be updated using the [`replace`][textual.widgets._text_area.TextArea.replace] method.
+This method is the programmatic equivalent of selecting some text and then pasting.
+
+All atomic (single-cursor) edits can be represented by a `replace` operation, but for
```
davep, line 62: I wonder if this is getting a little too into the implementation detail of the widget and might mean that the casual reader misses out on the significance of the availability of `insert`, `delete` and `clear`.

src/textual/_text_area_theme.py
```diff
@@ -0,0 +1,353 @@
...
+    """The name of the theme."""
+
+    base_style: Style | None = None
+    """The background style of the text area. If `None` the parent style will be used."""
+
+    gutter_style: Style | None = None
+    """The style of the gutter. If `None`, a legible TextAreaStyle will be generated."""
+
+    cursor_style: Style | None = None
+    """The style of the cursor. If `None`, the legible TextAreaStyle will be generated."""
```
davep, line 48: Above it says "a legible", here it says "the legible", the latter sounds like it's referring to something I might find somewhere else. Just a typo or should I read more into it?

docs/widgets/text_area.md
darrenburns: Agreed, removed that initial part.

davep (approved):
/me mumbles something about no Lisp in the default languages but hits approve anyway

pyproject.toml
```diff
@@ -46,6 +46,8 @@ markdown-it-py = { extras = ["plugins", "linkify"], version = ">=2.1.0" }
 #rich = {path="../rich", develop=true}
 importlib-metadata = ">=4.11.3"
 typing-extensions = "^4.4.0"
+tree-sitter = "^0.20.1"
+tree_sitter_languages = {version = ">=1.7.0", python = "^3.8"}
```
etanol, line 50: While I'm not entitled to judge; this significantly complicates the packaging and distribution of Textual because is no longer pure Python.  Not only the TreeSitter Python bindings contain CPython extensions; but each language parser needs to be compiled.

So now, every time a new language is added to the text area, anyone repackaging or redistributing a Textual application will have to do additional work.  At least I know I will have to.

This is not really a complain and I don't want to stop progress.  However, I would like to call some attention to the fact that adding dependencies is never for free.

I do have a question, thoguh.  Why couldn't this widget have been implemented as a separate package or a Textual extension/plugin?

darrenburns: tree-sitter and tree-sitter-languages have binary wheels available. You don't need to compile anything.

ofek: For the average user yes but Isaac is talking about redistribution for example like Debian, Conda or Homebrew where everything is built from source

willmcgugan: It could still be factored out in to another package. It is something we considered. But then that is an extra hoop the end-user has to go through.

If Debian, Conda etc. are going to support the TextArea does it really matter if its part of the Textual core, or a separate package?

etanol: In my case, I have a specially built CPython distribution for internal use.  One of the peculiarities is that, while CPython extensions are built as dynamically loadable libraries (`*.so`), any other external library is statically linked (e.g. Sqlite3, zlib, GNU dbm, OpenSSL, ...).  Even packages like lxml are built in a special way.

I will probably figure out how to build _tree-sitter_ in a way that it's compatible with my distribution.  After all, we love Textual because it's **awesome**.  My point is a bit more philosophical: depending on a 3rd party package is not all dandy, there are trade-offs (like introducing source build/packaging/distribution complexity in this case).

All I'm hoping is that these decisions are not made lightly.

willmcgugan: Glad you like Textual.

I do want to keep the dependancies lean in general. There are surprisingly few (non-dev) dependencies in Textual. Syntax highlighting is one of the areas where there just doesn't exist a pure-Python solution with adequate performance.

etanol: Does this mean that TreeSitter will be favored and Pygments will be relegated to fall-back?  Will this impact rich as well?

davep: > All I'm hoping is that these decisions are not made lightly.

As someone who was in the same room when these discussions (plural) took place: I can assure that they're not.

ihabunek:
Thanks for this feature. Was looking forward to it.

One benefit of having all required dependencies being pure python is being able to distribute your app in a pyz archive and having it work across all platforms. This is not possible with compiled deps.

Would it be possible to make tree sitter an optional dependency and either falling back to pygments or simply not supporting highlighting of it's missing?

Regards,
-- Ivan

pyproject.toml
willmcgugan: > Does this mean that TreeSitter will be favored and Pygments will be relegated to fall-back? Will this impact rich as well?

No, they are quite different APIs and not readily interchangeable. No need to change anything in Rich, since it doesn't need the incremental editing ability.

darrenburns:
> Thanks for this feature. Was looking forward to it. One benefit of having all required dependencies being pure python is being able to distribute your app in a pyz archive and having it work across all platforms. This is not possible with compiled deps. Would it be possible to make tree sitter an optional dependency and either falling back to pygments or simply not supporting highlighting of it's missing? Regards, -- Ivan

The code is already written such that if tree-sitter isn't available, you simply don't get syntax highlighting. Perhaps the tree-sitter stuff required for highlighting should be moved to an optional extra.

willmcgugan:
That might be the better compromise. The tree sitter stuff could be installed via an extra. pip install textual[syntax]

PR #2939, opened 2023-07-13
fix: black formatting

tanducmai:
Default black formatting with no argument provided

- [x ] Updated documentation
- [x ] Updated CHANGELOG.md (where appropriate)


TomJGooding:
> We want to exclude tests from black formatting.

@willmcgugan FWIW, this does slightly complicate things for contributors who have their editor auto-format on save.

PR #2941, opened 2023-07-14
markdown_viewer:  jump to anchor position when clicking on an anchor link

blob42:
A little patch to handle links with anchors in `MarkdownViewer` and jumps to the anchor position after loading the new md file. 

I need help for adding tests if they are necessary.

## Rationale

I am trying to build an offline sphinx documentation that is generated into markdown output. With the current MarkdownViewer any link containing an anchor does not open the target file. This patch loads the target file and jumps to the anchor position.


blob42:
Hi @davep if you have time to make a quick review for this PR that would be great.

davep:
Sure thing, when one of us gets the time to review we will. :-)

davep:
Just to let you know: haven't forgotten about this; just been deep in some longer-running stuff, but should look at sweeping this up with the other issue soon.

blob42:
Thanks for the update @davep it's not urgent anyway. Let me know if there is something I can do.

tests/test_markdown.py
```diff
@@ -125,3 +125,18 @@ async def test_load_non_existing_file() -> None:
             await pilot.app.query_one(Markdown).load(
                 Path("---this-does-not-exist---.it.is.not.a.md")
             )
+async def test_anchor_links() -> None:
```
davep, line 128: This is looking good (I'm looking to pull this in, in part as a fix for #3094), but I don't think I understand what the test is trying to do here; especially given that it fails.

What is it aiming to test exactly?

blob42: Sorry I committed by error I didn't know the right way to test it.

davep: That's fine, I just wanted to check first. I'm happy to look at adding any appropriate testing. 🙂

davep:
I'm going to do a little more work on this; it looks like it runs into a problem that I ran into with Frogmouth too, in that loading a document needs to take into account the original location of the document.

blob42:
> @blob42 If you wish to make another change to this, could I ask a favour? Don't force push changes. I'd merged in `main` and tweaked a couple of things in anticipation of moving to merging this but the force push removed that work.

My bad sorry I didn't know. I hope you still had a local copy of the changes ! 

No changes to add on my side and I will never use force push again.

davep:
> I'm going to do a little more work on this; it looks like it runs into a problem that I ran into with Frogmouth too, in that loading a document needs to take into account the original location of the document.

Turns out it's fine, I'm just bad at writing tests.

CHANGELOG.md
```diff
@@ -10,9 +10,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 
 - Ability to enable/disable tabs via the reactive `disabled` in tab panes https://github.com/Textualize/textual/pull/3152
+- `MarkdownViewer` now correctly handles anchors in links https://github.com/Textualize/textual/pull/2941
+
 ### Fixed
 
 - Could not hide/show/disable/enable tabs in nested `TabbedContent` https://github.com/Textualize/textual/pull/3150
+- Fixed a crash in `MarkdownViewer` when clicking on a link containing an anchor https://github.com/Textualize/textual/issues/3094
```
rodrigogiraoserrao, line 18: Just a heads up that when `main` is merged into this branch, these entries could end up in the wrong place.

tests/test_markdownviewer.py
```diff
@@ -0,0 +1,25 @@
...
+            Path(__file__).with_suffix(".md")
+        )
+
+
+async def test_markdown_viewer_anchor_link() -> None:
+    """Test https://github.com/Textualize/textual/issues/3094"""
+    async with MarkdownViewerApp().run_test() as pilot:
+        # There's not really anything to test *for* here, but the lack of an
+        # exception is the win (before the fix this is testing it would have
+        # been FileNotFoundError).
```
rodrigogiraoserrao, line 24: If we make the `md` file long enough between the first header and the last one, couldn't we check that the app was scrolled to the bottom?
Is this something silly?

rodrigogiraoserrao (commented):
The changes look reasonable to me.
I thought we could have hyperlinks in markdown documents and they'd work well, i.e., I could click them and open web pages.

I just tested it on the Textual README.md and it doesn't open URLs.

Since we're handling clicks in the PR, is this something that makes sense fixing here?

tests/test_markdownviewer.py
davep: I did consider this but it's not really the purpose the the test; it was more to test that it didn't blow up as it used to.

CHANGELOG.md
davep: Aye, ta, I tend to try and make a final ChangeLog check before doing a merge, if I get to do the merge.

davep:
> I thought we could have hyperlinks in markdown documents and they'd work well, i.e., I could click them and open web pages.
> 
> I just tested it on the Textual README.md and it doesn't open URLs.
> 
> Since we're handling clicks in the PR, is this something that makes sense fixing here?

I don't recall that ever being part of the Markdown viewer (it is part of Frogmouth though). I sense it's somewhat out of scope for this PR as this is mostly about handling a genuine crash. Might be worth adding an issue to the effect of the above though.

blob42:
> The changes look reasonable to me. I thought we could have hyperlinks in markdown documents and they'd work well, i.e., I could click them and open web pages.

> Since we're handling clicks in the PR, is this something that makes sense fixing here?

Keep in mind that most modern terminal emulators handle url clicks usually ctrl+click.  They require the full link to be rendered to the terminal. 

I don't know if Textual is planning to implement some modern terminal protocol like Kitty, which would solve all these issues. 

Using Markdown from rich I can disable link rendering which allows me to use my native emulator url handler.  

Maybe the same option could be added to the widget to disable rendering urls but would ignore relative links used for jumping around. It seems out of scope for this PR though.

rodrigogiraoserrao (approved):

src/textual/widgets/_markdown.py
```diff
@@ -966,7 +966,22 @@ async def forward(self) -> None:
...
+            await self.go(message.href)
+        else:
+            link, anchor = message.href.split("#", 1)
+            await self.go(link)
+
+            def scroll_to_anchor() -> None:
+                toc = self.table_of_contents.table_of_contents
+                assert toc is not None
+                for _, name, block_id in toc:
+                    if name.lower().replace(" ", "-") == anchor:
```
willmcgugan, line 979: This is making a somewhat naive assumption about how fragments are generated from headers. It wont always be as simple as converting spaces to hyphens and lower casing. Consider characters with accents for example. 

I think it makes sense to match what GitHub markdown does, and replicate that.

src/textual/widgets/_markdown.py
```diff
@@ -966,7 +966,22 @@ async def forward(self) -> None:
 
     async def _on_markdown_link_clicked(self, message: Markdown.LinkClicked) -> None:
         message.stop()
-        await self.go(message.href)
+        if message.href.find("#") == -1:
+            await self.go(message.href)
+        else:
+            link, anchor = message.href.split("#", 1)
```
willmcgugan, line 972: Suggest using `str.partition` here, so we can avoid the additional branch.

src/textual/widgets/_markdown.py
```diff
@@ -966,7 +966,22 @@ async def forward(self) -> None:
...
+            def scroll_to_anchor() -> None:
+                toc = self.table_of_contents.table_of_contents
+                assert toc is not None
+                for _, name, block_id in toc:
+                    if name.lower().replace(" ", "-") == anchor:
+                        block = self.query_one(f"#{block_id}", MarkdownBlock)
+                        self.scroll_to_widget(block, top=True)
+                        break
+
+            self.call_later(scroll_to_anchor)
```
willmcgugan, line 984: `call_later` probably isn't the right approach here. It will execute after all other messages have been processed. This is normally very quick, but it is possible that an entirely new document has been loaded in the meantime.

I think this functionality should be built in to `go`, so we can be sure that it applies to the correct document.

src/textual/widgets/_markdown.py
```diff
@@ -966,7 +966,22 @@ async def forward(self) -> None:
...
     async def _on_markdown_link_clicked(self, message: Markdown.LinkClicked) -> None:
         message.stop()
-        await self.go(message.href)
+        if message.href.find("#") == -1:
+            await self.go(message.href)
+        else:
+            link, anchor = message.href.split("#", 1)
+            await self.go(link)
+
+            def scroll_to_anchor() -> None:
```
willmcgugan, line 975: Needs a docstring. I know it seems self explanatory, but it's how we roll.

willmcgugan (changes requested):
A good first pass, but needs work...

What happens if the very first URL we visit has a fragment?

davep:
@blob42 Do you want to make the changes based on Will's feedback?

blob42:
@davep 

I am a bit busy this week but will have time to properly work on it next week.

Feel free to do the changes if you would like.

davep:
@blob42 That's fine; not sure if I'll get back to this before next week either so let's say whoever gets back here first. :-)

davep:
Picking this up again.

davep:
https://github.com/Textualize/textual/pull/3236 is now merged into main, which should give a good basis for handling Markdown header slugs.

davep:
I think we hit this and some other stuff with #3244 so I'll close this one. Thanks again!

PR #2947, opened 2023-07-15
Remove print call from OptionList

darrenburns:


davep (approved):
Looks like this got snuck in as part of the "style review" PR the other month. Good catch.

PR #2866, opened 2023-06-30
Notifications

davep:
This PR seeks to satisfy #2846. In doing so it does the following:

- Adds a general `Notification`/`Notifications` pair of classes for holding details on an individual notification, as well as holding a list of timed notifications.
- Adds a small collection of widgets (which should sort of be considered to be "internal" at the moment, but which are named non-internal so as not to discourage styling) that show and handle the display of the above notifications via a `Toast`.
- Adds a core `App.notify` method, which takes a message and a title, as well as an optional severity level (`information`, `warning` or `error` -- default being `information`) and an optional timeout (3 seconds by default).
- Adds `App.unnotify` for removing a notification.
- Adds `App.clear_notifications` for clearing notifications.
- Adds a `Screen.notify` method, which is a thin wrapper around `App.notify`.
- Adds a `Widget.notify` method, which is a thin wrapper around `App.notify`.

The `Toast`s are always shown above the main display, scrolling up from the bottom right, and they persist between screens and modes.

By way of illustration:

https://github.com/Textualize/textual/assets/28237/1e380bff-1ffa-4b90-9509-1e5db38b7db0

was produced with this code:

```python
from pathlib import Path
from random import choice

from textual import on
from textual.app import App, ComposeResult
from textual.containers import Horizontal, Vertical, VerticalScroll
from textual.widgets import Markdown, Button, Checkbox, Input

QUOTES = [
    "Are either of you paleontologists? I'm in desperate need of a paleontologist.",
    "I should reach Defcon 1 and release my missiles in 28 hours. Would you like to see some projected kill ratios?",
    "A strange game."
    "The only winning move is not to play.",
    "How about a nice game of chess?",
    "Joshua called me.",
    "The WOPR spends all it's time thinking about World War III.",
    "Did you ever play tic-tac-toe?",
    "I loved it when you nuked Las Vegas."
]

class ToastyApp(App[None]):

    CSS = """
    Horizontal {
        align: center middle;
        height: auto;
    }
    Horizontal Button {
        margin-left: 1;
    }
    Input {
        width: 20%;
    }
    """

    def compose(self) -> ComposeResult:
        with Vertical():
            with Horizontal():
                yield Checkbox("Titled", value=True)
                yield Input(placeholder="Timeout (default 3 seconds)")
                yield Button("Information", id="information")
                yield Button("Warning", id="warning")
                yield Button("Error", id="error")
            with VerticalScroll():
                yield Markdown()

    async def on_mount(self) -> None:
[...]

PR #2962, opened 2023-07-17
Make notification titles work as documented again (notifications fix option 1)

davep:
*Note: This PR is one of a pair of either/or PRs. See also #2963 as the alternative.*

It seems that the [last moment style tweaks](https://github.com/Textualize/textual/pull/2866/commits/7fc7ef20ef52ba515033940e9f84420700cc1fe8) also changed the way that notifications work in regard to how titles are handled. As released, notifications no longer work as documented.

This PR reinstates the documented behaviour. The assumption here, based on the fact that it wasn't called out in the PR's subject or description, is that it was an unintended consequence of adding the `render` method.

### DIFF
--- a/src/textual/widgets/_toast.py
+++ b/src/textual/widgets/_toast.py
@@ -98,10 +98,15 @@ class Toast(Static, inherit_css=False):
 
     def render(self) -> RenderableType:
         notification = self._notification
-        if notification.title:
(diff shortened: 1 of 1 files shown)


davep:
The choice has been made, we'll go with #2963.

PR #2963, opened 2023-07-17
Make notification documentation and interface match changed rendering (notifications fix option 2)

davep:
*Note: This PR is one of a pair of either/or PRs. See also #2962 as the alternative.*

It seems that the [last moment style tweaks](https://github.com/Textualize/textual/pull/2866/commits/7fc7ef20ef52ba515033940e9f84420700cc1fe8) also changed the way that notifications work in regard to how titles are handled. As released, notifications no longer work as documented.

This PR changes the interface for notifications to match the new behaviour, and also updates the documentation so that it no longer misrepresents how titles work. The assumption here, based on the fact that it wasn't called out in the PR's subject or description, is that it was an intended consequence of adding the `render` method but the changes weren't carried to completion.

### DIFF
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -2795,7 +2795,7 @@ class App(Generic[ReturnType], DOMNode):
         self,
         message: str,
         *,
-        title: str | None = None,
+        title: str = "",
         severity: SeverityLevel = "information",
         timeout: float = Notification.timeout,
     ) -> Notification:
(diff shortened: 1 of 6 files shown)


willmcgugan:
I don't think the severity as a title works. Just seems redundant. So this would be my prefered option.

davep:
@willmcgugan 
indiana-jones-last-crusade

willmcgugan (approved):

PR #2967, opened 2023-07-18
web driver from textual-serve

willmcgugan:
This is a driver for Ganglion. It used to be in textual-serve but it makes more sense for it to be here.

Forgive the last of tests. It is a WIP for now.


src/textual/drivers/_byte_stream.py
davep, line 1: I suspect this wants a

```python
```

davep (approved):
Aside from the `__future__` suspicion...

Issue #2969, opened 2023-07-18
SelectionList crashes on content overflow or resize


zdovcj:
When using SelectionList and resizing the terminal so that the text cannot fit in the screen, the program will crash. I think it also works with OptionList.

Environment:
Textual version: 0.30.0
Python version: 3.10.8

Code to reproduce:
```python
import textual.widgets
import textual.app

class App(textual.app.App):

    def compose(self):
        yield textual.widgets.SelectionList (('random_text' * 10, 0), id='id')

if __name__ == '__main__':
    App().run()

```

The last error printed is:
```
                                                                                                             
│ /opt/homebrew/lib/python3.10/site-packages/textual/widgets/_option_list.py:749 in get_option_at_index       
│                                                                                                             
│   746 │   │   try:                                                                                          
│   747 │   │   │   return self._options[index]                                                               
│   748 │   │   except IndexError:                                                                            
│ ❱ 749 │   │   │   raise OptionDoesNotExist(                                                                 
│   750 │   │   │   │   f"There is no option with an index of {index}"                                        
│   751 │   │   │   ) from None                                                                               
│   752                                                                                                       
│                                                                                                             
│ ╭──────────── locals ────────────╮                                                                          
│ │ index = 1                      │                                                                          
│ │  self = SelectionList(id='id') │                                                                          
│ ╰────────────────────────────────╯                                                                          
╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────
```

I have tried to size my content upon resize event, but it doesn't solve the issue. In this case, it will even crash on increasing the terminal size. I assume that the function that sets the content length might be called before  the the code that crashes.

Code to reproduce the crash upon increasing:
```python
import textual.widgets
import textual.app
import textual.events
import textual  

class App(textual.app.App):

    @textual.on(textual.events.Resize)
    def resize_handler(self):
        width = self.size.width
        text = 'a' * (width - 4)
        widget = self.query_one(textual.widgets.SelectionList)
        widget.clear_options()
        widget.add_option((text, 0, None))

    def compose(self):
        yield textual.widgets.SelectionList (('random_text', 0), id='widget_id')

if __name__ == '__main__':
    App().run()
```

davep:
Thanks. That looks to be a duplicate of #2900 so I'll close this in favour of the earlier report.

zdovcj:
Ok. Sorry for not noticing and posting a duplicate. I will at least add the code that dynamically changes the content length to #2900.

[...]

Issue #2900, opened 2023-07-07
SelectionList with long lines throws an exception when the window is too small


valberg:
# Textual Diagnostics

## Versions

| Name    | Value  |
|---------|--------|
| Textual | 0.29.0 |
| Rich    | 13.4.2 |

## Python

| Name           | Value                                                |
|----------------|------------------------------------------------------|
| Version        | 3.11.3                                               |
| Implementation | CPython                                              |
| Compiler       | GCC 13.1.1 20230429                                  |
| Executable     | /home/valberg/.pyenv/versions/textual-dev/bin/python |

## Operating System

| Name    | Value                                                  |
|---------|--------------------------------------------------------|
| System  | Linux                                                  |
| Release | 6.4.1-arch2-1                                          |
| Version | #1 SMP PREEMPT_DYNAMIC Tue, 04 Jul 2023 08:39:40 +0000 |

## Terminal

| Name                 | Value          |
|----------------------|----------------|
| Terminal Application | *Unknown*      |
| TERM                 | xterm-256color |
| COLORTERM            | truecolor      |
| FORCE_COLOR          | *Not set*      |
| NO_COLOR             | *Not set*      |

## Rich Console options

| Name           | Value               |
|----------------|---------------------|
| size           | width=62, height=34 |
| legacy_windows | False               |
| min_width      | 1                   |
| max_width      | 62                  |
| is_terminal    | True                |
| encoding       | utf-8               |
| max_height     | 34                  |
| justify        | None                |
| overflow       | None                |
| no_wrap        | False               |
| highlight      | None                |
| markup         | None                |
| height         | None                |

# Bug report

Running the following in a smallish window, or resizing below the length of the selection length:

```python
from textual.app import App, ComposeResult
from textual.widgets import Header, Footer, SelectionList


class TestApp(App):

    def compose(self) -> ComposeResult:
        yield Header()

        yield SelectionList(
            ("xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx", 1),
        )

        yield Footer()


if __name__ == "__main__":
    app = TestApp()
    app.run()
```

results in the following exception:

```
OptionDoesNotExist: There is no option with an index of 1
```

davep:
[...]

PR #2970, opened 2023-07-19
Fix a crash caused by a prompt being wider than a `SelectionList`

davep:
This fixes the issue raised in #2900 (and also #2969).

### DIFF
--- a/src/textual/widgets/_option_list.py
+++ b/src/textual/widgets/_option_list.py
@@ -298,6 +298,7 @@ class OptionList(ScrollView, can_focus=True):
         id: str | None = None,
         classes: str | None = None,
         disabled: bool = False,
+        wrap: bool = True,
     ):
         """Initialise the option list.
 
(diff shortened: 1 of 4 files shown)


src/textual/widgets/_option_list.py
```diff
@@ -477,13 +493,18 @@ def _refresh_content_tracking(self, force: bool = False) -> None:
...
         lines_from = self.app.console.render_lines
-        options = self.app.console.options.update_width(
-            self.scrollable_content_region.width
-        )
         add_span = self._spans.append
         option_ids = self._option_ids
         add_lines = self._lines.extend
 
+        # Adjust the options for our purposes.
+        options = self.app.console.options.copy().update_width(
```
willmcgugan, line 501: `update_width` does a `copy()` internally, so you won't need the explicit copy.

willmcgugan (approved):
Just one request.

src/textual/widgets/_option_list.py
davep: Ahh hah! That explains why it wasn't there before!

Issue #2952, opened 2023-07-16
Capture  prints


willmcgugan:
At the moment we capture the output of any `print` statements, so as not to write over the Textual app. Anything printed will go to the dev tools if enabled, or dev/null if not.

This is a sensible default, but sometimes devs want to be able to capture that content and write it to a widget (for example). This can be a requirement if you working with existing code that may have printed to the terminal.

To make that possible, we should have an optional way for an app to capture data that might have otherwise been swallowed.

Please propose an API to @willmcgugan before attempting this.

PR #2971, opened 2023-07-19
capture print

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/2952

Implements a mechanism to capture prints (or writes to stdout / stderr).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## Unreleased
+
+### Added
+
+- Added App.capture_print, App.end_capture_print, Widget.capture_print, Widget.end_capture_print
+
 ## [0.30.0] - 2023-07-17
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
-- Added App.capture_print, App.end_capture_print, Widget.capture_print, Widget.end_capture_print https://github.com/Textualize/textual/issues/2952
+- Added App.begin_capture_print, App.end_capture_print, Widget.begin_capture_print, Widget.end_capture_print https://github.com/Textualize/textual/issues/2952
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
-- Added App.capture_print, App.end_capture_print, Widget.capture_print, Widget.end_capture_print
+- Added App.capture_print, App.end_capture_print, Widget.capture_print, Widget.end_capture_print https://github.com/Textualize/textual/issues/2952
 
 ## [0.30.0] - 2023-07-17
 
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -222,7 +222,7 @@ class _PrintCapture:
         self.stderr = stderr
 
     def write(self, text: str) -> None:
-        """Captured write.
+        """Called when writing to stdout or stderr.
 
         Args:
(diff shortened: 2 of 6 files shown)


src/textual/app.py
```diff
@@ -206,6 +208,37 @@ def isatty(self) -> bool:
...
+    def write(self, text: str) -> None:
+        """Captured write.
+
+        Args:
+            text: Text that was "printed".
+        """
+        self.app._print(text, stderr=self.stderr)
+
+    def flush(self) -> None:
+        """Stdout or stderr was flushed."""
```
davep, line 233: This docstring reads more like one for a boolean property than one for a method that performs an action. Might want to tweak it it?

src/textual/app.py
```diff
@@ -1026,6 +1069,54 @@ async def _press_keys(self, keys: Iterable[str]) -> None:
...
+            text: Text that has been printed.
+            stderr: True if the print was to stderr, or False for stdout.
+        """
+        if self._devtools_redirector is not None:
+            self._devtools_redirector.write(text)
+        for target, (_stdout, _stderr) in self._capture_print.items():
+            if (_stderr and stderr) or (_stdout and not stderr):
+                target.post_message(events.Print(text, stderr=stderr))
+
+    def capture_print(
```
davep, line 1095: More to note than anything else, but I feel this might be better as `begin_capture_print`; it mirrors `end_capture_print`, makes the intent more clear, and then later allows for `capture_print` to be used as a context.

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## Unreleased
+
+### Added
+
+- Added App.capture_print, App.end_capture_print, Widget.capture_print, Widget.end_capture_print https://github.com/Textualize/textual/issues/2952
```
davep, line 12: I feel we should add the same methods to `Screen` too.

davep (approved):
Couple or comments and observations, but nothing is a blocker. I do feel adding the same methods to `Screen` makes a lot of sense though.

src/textual/app.py
willmcgugan: Yeah, although there is a odd inconsistency then. There is a existing `capture_mouse` which isn't a context manager.

CHANGELOG.md
willmcgugan: The screen is a Widget, so it will inherit those methods.

davep: Damn, good point. I keep forgetting that. Amusingly screen and widget both now have a notify method. Note to self: remove the wrapper on Screen.

AntoninRousset:
I was using Textual headless mode to turn my Textual app into a CLI:

```sh
python my_app.py items list
# Amazing textual interface!
```

```sh
python my_app.py --headless items list
# Print a boring JSON and exit
```

```sh
python my_app.py items list > items.json
# Same as headless, but redirected to a file
```

I was using `print()` to write to stdout, but this no longer works. I need to hack `App.__init__`:
```python

        if kwargs.get("headless", False):
            self._capture_stdout = sys.stdout
            self._capture_stderr = sys.stderr
```

dhallas:
I have also run into problems with this change.
I have a textual app that uses the Python kubernetes-client library. But this commit causes the kubernetes client to fail detecting that the output tty is not interactive and should run the authentication scheme in a different way (this was actually fixed in the latest release, see original bug report here: https://github.com/kubernetes-client/python/issues/1986).
But this change in textual breaks that detection logic causing authentication to fail again. 
I can also confirm that the hack suggested by @AntoninRousset solves the issue

dhallas:
Seems like returning _False_ in app.py _PrintCapture::isatty also solves the problem for me. There is a comment asking if this should be made configurable. Maybe this is the use case for it?

app.py

`
    def isatty(self) -> bool:
        """Pretend we're a terminal."""
        # TODO: should this be configurable?
        return False
`

dhallas:
@willmcgugan  - would it be better to create an issue to track this instead of commenting here?

willmcgugan:
Yes please.

PR #2981, opened 2023-07-22
Fixes a bug that causes Click events to bubble up from Switch widgets.

### LINKED ISSUE #2366: `Switch` should stop the `Click` event from bubbling
At the moment `Switch` handles `Click` but then lets it bubble; there's no good reason to do that and it also stops the ability to write something like this:

```python
from textual.widgets    import Header, Footer, Label, Switch

class LabeledSwitch( Horizontal ):

    def on_click( self ) -> None:
        self.query_one(Switch).toggle()

class ClickableLabelApp( App[ None ] ):

        yield Header()
        with LabeledSwitch():
            yield Label( "Click me!" )
            yield Switch()
        yield Footer()

if __name__ == "__main__":
    ClickableLabelApp().run()
```

where the idea is to make a compound widget that lets you click on the `Label` or the `Switch` and the `Switch` will toggle -- only it doesn't work if you click on the `Switch` because it ends up double-toggling.

seifertm:

- [ ] Updated documentation

This PR stops _Click_ events from bubbling up from _Switch_ widgets.

The PR adds a corresponding test that fails, if the App containing a Switch receives a Click event when the Switch is clicked and adjusts the changelog.  

Closes #2366

### DIFF
--- a/src/textual/widgets/_switch.py
+++ b/src/textual/widgets/_switch.py
@@ -154,8 +154,9 @@ class Switch(Widget, can_focus=True):
     def get_content_height(self, container: Size, viewport: Size, width: int) -> int:
         return 1
 
-    async def _on_click(self, _: Click) -> None:
+    async def _on_click(self, event: Click) -> None:
         """Toggle the state of the switch."""
+        event.stop()
         self.toggle()
 
     def action_toggle(self) -> None:
--- a/tests/test_switch.py
+++ b/tests/test_switch.py
@@ -1,0 +1,18 @@
+from textual.app import App, ComposeResult
+from textual.widgets import Switch
+
+
+async def test_switch_click_doesnt_bubble_up():
+    """Regression test for https://github.com/Textualize/textual/issues/2366"""
+
+    class SwitchApp(App[None]):
+        def compose(self) -> ComposeResult:
+            yield Switch()
+
+        async def on_click(self) -> None:
+            raise AssertionError(
+                "The app should never receive a click event when Switch is clicked."
+            )
+
+    async with SwitchApp().run_test() as pilot:
+        await pilot.click(Switch)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 
 - Fixed a crash when a `SelectionList` had a prompt wider than itself https://github.com/Textualize/textual/issues/2900
+- Fixed a bug where `Click` events were bubbling up from `Switch` widgets https://github.com/Textualize/textual/issues/2366
 
 ## [0.30.0] - 2023-07-17
 


src/textual/widgets/_switch.py
```diff
@@ -154,8 +154,9 @@ def get_content_width(self, container: Size, viewport: Size) -> int:
     def get_content_height(self, container: Size, viewport: Size, width: int) -> int:
         return 1
 
-    async def _on_click(self, _: Click) -> None:
+    async def _on_click(self, event: Click) -> None:
         """Toggle the state of the switch."""
+        event.stop()
```
rodrigogiraoserrao, line 159: @seifertm I'm not feeling it.
You have to call this method like you mean it...

tests/test_switch.py
```diff
@@ -0,0 +1,16 @@
+from textual.app import App, ComposeResult
+from textual.widgets import Switch
+
+
+async def test_switch_click_doesnt_bubble_up():
```
rodrigogiraoserrao, line 5: Can you please add a comment with a link to the issue?
We usually say something like "Regression test for <URL>".

rodrigogiraoserrao (changes requested):

tests/test_switch.py
seifertm: Thanks. You mentioned this in the meta issue, but for some reason I thought it referred to the changelog.

I added a docstring that references the original ticket.

rodrigogiraoserrao (approved):
LGTM 🚀

tests/test_switch.py
```diff
@@ -0,0 +1,18 @@
...
+    """Regression test for https://github.com/Textualize/textual/issues/2366"""
+
+    class SwitchApp(App[None]):
+        def compose(self) -> ComposeResult:
+            yield Switch()
+
+        async def on_click(self) -> None:
+            raise AssertionError(
```
willmcgugan, line 13: Can you check this fails without the fix? I seem to recall an issue where assertion errors where swallowed when raised via pilot. (just being paranoid)

rodrigogiraoserrao: Ahhh I see. I seem to recall Darren taking care of this, but you can never be too sure.
@seifertm, let us know what you find.

seifertm: I'm equally paranoid about these things, so I wanted to see my test fail without the fix as well. I commented out `event.stop()` and got this stack trace for the test when running `pytest -k test_switch`:

```
self = SwitchApp(title='SwitchApp', classes={'-dark-mode'})

    async def on_click(self) -> None:
>       raise AssertionError(
            "The app should never receive a click event when Switch is clicked."
        )
E       AssertionError: The app should never receive a click event when Switch is clicked.

tests/test_switch.py:13: AssertionError
-----Captured stderr call -----
╭───────────────────── Traceback (most recent call last) ──────────────────────╮
│ /home/michael/oss/textual/tests/test_switch.py:13 in on_click                │
│                                                                              │
│   10 │   │   │   yield Switch()                                              │
│   11 │   │                                                                   │
│   12 │   │   async def on_click(self) -> None:                               │
│ ❱ 13 │   │   │   raise AssertionError(                                       │
│   14 │   │   │   │   "The app should never receive a click event when Switch │
│   15 │   │   │   )                                                           │
│   16                                                                         │
│                                                                              │
│ ╭────────────────────────── locals ───────────────────────────╮              │
│ │ self = SwitchApp(title='SwitchApp', classes={'-dark-mode'}) │              │
│ ╰─────────────────────────────────────────────────────────────╯              │
╰──────────────────────────────────────────────────────────────────────────────╯
AssertionError: The app should never receive a click event when Switch is 
clicked.
----- snapshot report summary -----
1 snapshot passed.
===== short test summary info =====
FAILED tests/test_switch.py::test_switch_click_doesnt_bubble_up - AssertionError: The app should never receive a click event when Switch is clicked.
====== 1 failed, 6 passed, 1518 deselected in 2.42s =====
(
```

willmcgugan: Works for me!

willmcgugan (approved):

PR #2987, opened 2023-07-22
MouseMove events bubble up and are posted by Screen

### LINKED ISSUE #2905: Send mouse move events to screen
It look like mouse move events never make it to the Screen object. I think they should.

seifertm:

- [x] Updated documentation

This PR changes the behavior of _MouseMove_ events to bubble up from Widgets. `Screen` receives the MouseMove event when there's no other Widget in the App.

Note: The relative coordinates of the MouseMove events do not change when bubbling up, i.e. they always stay relative to the Widget that received the event. This behavior is consistent with other mouse events such as _Click_.

There's some more room for refactoring as the handling of MouseMove and MouseEvent in Screen._forward_event is almost the same. The main difference seems to be that MouseEvent sets or resets the focused widget, whereas MouseMove sets or resets the mouse over.

Closes #2905


rodrigogiraoserrao (approved):
This looks good to me!

We may prefer the changelog message in the "Changed" section, though.

rodrigogiraoserrao:
@willmcgugan review away.

PR #2994, opened 2023-07-23
add css first-child and last-child pseudo selectors

### LINKED ISSUE #650: CSS pseudo selectors
We should implement more pseudo selectors in Textual.

It will be really nice in combination with queries, for instance to remove the last child in a list:

```python
self.query(".item:last-child").remove()
```

And of course for general styling.

I'm thinking of `last-child` and `first-child`, but we could implement any that make sense.

See https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-classes for a list of pseudo selectors.

csernazs:
Resolves #650

This is a "draft" MR for now, I'll do the checklist once we agree that this could be a good implementation.

- [ ] Updated documentation

### DIFF
--- a/src/textual/css/model.py
+++ b/src/textual/css/model.py
@@ -88,14 +88,26 @@ class Selector:
         """
         return self._checks[self.type](node)
 
+    def _check_first_last_child(self, node: DOMNode) -> bool:
+        if "last-child" in self.pseudo_classes and node.parent.children[-1] == node:
+            return True
+
+        if "first-child" in self.pseudo_classes and node.parent.children[0] == node:
+            return True
+
     def _check_universal(self, node: DOMNode) -> bool:
         return node.has_pseudo_class(*self.pseudo_classes)
 
     def _check_type(self, node: DOMNode) -> bool:
         if self.name not in node._css_type_names:
             return False
+
+        if self._check_first_last_child(node):
+            return True
+
         if self.pseudo_classes and not node.has_pseudo_class(*self.pseudo_classes):
             return False
+
         return True
 
     def _check_class(self, node: DOMNode) -> bool:
--- a/src/textual/css/constants.py
+++ b/src/textual/css/constants.py
@@ -69,6 +69,8 @@ VALID_PSEUDO_CLASSES: Final = {
     "focus-within",
     "focus",
     "hover",
+    "first-child",
+    "last-child",
 }
 VALID_OVERLAY: Final = {"none", "screen"}
 VALID_CONSTRAIN: Final = {"x", "y", "both", "inflect", "none"}
--- a/tests/test_query.py
+++ b/tests/test_query.py
@@ -196,6 +196,52 @@ def test_query_classes():
     assert len(app.query(".test")) == 0
 
 
+def test_query_last_child():
+    class App(Widget):
+        pass
+
+    class Child(Widget):
+        pass
+
+    CHILD_COUNT = 100
+
+    # Create a fake app to hold everything else.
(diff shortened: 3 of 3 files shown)


src/textual/css/model.py
```diff
@@ -94,6 +94,13 @@ def _check_universal(self, node: DOMNode) -> bool:
     def _check_type(self, node: DOMNode) -> bool:
         if self.name not in node._css_type_names:
             return False
+
+        if "last-child" in self.pseudo_classes and node.parent.children[-1] == node:
```
csernazs, line 98: this and the first-child can fail if the `node` has no `parent`. Is this possible in textual?
If yes, what would be the expected behavior in such case? Should the root node matches the `last-child` or `first-child` pseudo elements?

src/textual/css/model.py
```diff
@@ -94,6 +94,13 @@ def _check_universal(self, node: DOMNode) -> bool:
     def _check_type(self, node: DOMNode) -> bool:
```
csernazs, line 94: do we need to add this to `_check_id` or `_check_class` also? 

For `_check_id` we could have a match on the id, then if the element is found, there could be an extra check whether it is the first child (or the last child, depending on the selector), but I'm not sure if it would be the correct behavior.

For `_check_class` I'm thinking on the usage and the test and the following didn't work:
`app.query(Child).filter(":first-child")`

src/textual/css/model.py
rodrigogiraoserrao: All widgets in the DOM always have a parent.
So, this is fine as long as this only gets called after the widget has been mounted.

src/textual/css/model.py
rodrigogiraoserrao: I'm not sure how the semantics for the pseudo selectors work.
Maybe @willmcgugan can clarify.

src/textual/css/model.py
willmcgugan: The `app` is at the root of the DOM, and it doesn't have a parent.

src/textual/css/model.py
willmcgugan: `app.query("Child:first_child", Child)` should work.

src/textual/css/model.py
rodrigogiraoserrao: @willmcgugan I always forget that the app is a widget 🙃
In that case, what's the reply you want to give to csernazs?

rodrigogiraoserrao:
Hey @csernazs, I'm afraid there might have been some miscommunication between the two of us.
What do you need from us to finish the PR?

src/textual/css/model.py
csernazs: Maybe I miss something but `app.query("Child:first_child", Child)` does not work for me.

Is this the correct function to be called?

csernazs:
Hi @rodrigogiraoserrao,

I'm sorry for the delay.

From my point of view:

We would need to close those two threads I opened above.
So then I can make the required changes, and then we can merge this MR.

Thanks,
Zsolt

src/textual/css/model.py
rodrigogiraoserrao: > Maybe I miss something but `app.query("Child:first_child", Child)` does not work for me.

I think what Will is trying to say is that when `:first_child` is implemented properly, the query above should be made to work.
If you've implemented the pseudo-selector `:first_child` and the call above doesn't work, then something needs to be fixed.

willmcgugan:
@csernazs Thanks for your input. I think we will take it from here.

PR #2995, opened 2023-07-23
Add DigitDisplay widget

eliasdorneles:
This supersedes: https://github.com/Textualize/textual/pull/2990

- [x] Updated documentation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ unit-test:
 
 .PHONY: test-snapshot-update
 test-snapshot-update:
-	$(run) pytest --cov-report term-missing --cov=textual tests/ -vv --snapshot-update
+	$(run) pytest --cov-report term-missing --cov=textual tests/ -vv --snapshot-update $(PYTEST_ARGS)
 
 .PHONY: typecheck
 typecheck:
--- a/src/textual/widgets/__init__.py
+++ b/src/textual/widgets/__init__.py
@@ -48,6 +48,7 @@ __all__ = [
     "Checkbox",
     "ContentSwitcher",
     "DataTable",
+    "DigitDisplay",
     "DirectoryTree",
     "Footer",
     "Header",
@@ -77,7 +78,6 @@ __all__ = [
     "Tooltip",
     "Tree",
     "Welcome",
-    "DigitDisplay",
 ]
 
 _WIDGETS_LAZY_LOADING_CACHE: dict[str, type[Widget]] = {}
--- a/src/textual/widgets/__init__.py
+++ b/src/textual/widgets/__init__.py
@@ -40,6 +40,7 @@ if typing.TYPE_CHECKING:
     from ._tooltip import Tooltip
     from ._tree import Tree
     from ._welcome import Welcome
+    from ._digit_display import DigitDisplay
 
 
 __all__ = [
@@ -76,6 +77,7 @@ __all__ = [
     "Tooltip",
     "Tree",
     "Welcome",
+    "DigitDisplay",
 ]
 
 _WIDGETS_LAZY_LOADING_CACHE: dict[str, type[Widget]] = {}
--- a/src/textual/widgets/_digit_display.py
+++ b/src/textual/widgets/_digit_display.py
@@ -209,8 +209,7 @@ class _SingleDigitDisplay(Static):
     DEFAULT_CSS = """
     _SingleDigitDisplay {
       height: 3;
-      min-width: 2;
-      max-width: 3;
+      width: 3;
     }
     """
 
@@ -244,7 +243,6 @@ class _SingleDigitDisplay(Static):
             disabled=disabled,
         )
         self.digit = initial_value
-        self._content_width = 3
 
     def validate_digit(self, digit: str) -> str:
         """Sanitize and validate the digit input."""
@@ -258,11 +256,10 @@ class _SingleDigitDisplay(Static):
     def _watch_digit(self, digit: str) -> None:
         """Called when the digit attribute changes and passes validation."""
         content = _character_map[digit.upper()]
-        self._content_width = len(content.splitlines()[0])
         self.update(content)
 
     def get_content_width(self, container: Size, viewport: Size) -> int:
-        return self._content_width
+        return 3
 
     def get_content_height(self, container: Size, viewport: Size, width: int) -> int:
         return 3
--- a/src/textual/widgets/_digit_display.py
+++ b/src/textual/widgets/_digit_display.py
@@ -199,6 +199,7 @@ _character_map = {
 
 class _SingleDigitDisplay(Static):
     digit = reactive(" ", layout=True)
+    """The digit to display."""
 
     DEFAULT_CSS = """
         _SingleDigitDisplay {
@@ -212,10 +213,17 @@ class _SingleDigitDisplay(Static):
         super().__init__(**kwargs)
         self.digit = initial_value
 
-    def watch_digit(self, digit: str) -> None:
-        """Called when the digit attribute changes."""
+    def validate_digit(self, digit: str) -> str:
+        """Sanitize and validate the digit input."""
         if len(digit) > 1:
             raise ValueError(f"Expected a single character, got {len(digit)}")
+        digit = digit.upper()
+        if digit not in _character_map:
+            raise ValueError(f"Unsupported character: {digit}")
+        return digit
+
+    def _watch_digit(self, digit: str) -> None:
+        """Called when the digit attribute changes and passes validation."""
         self.update(_character_map[digit.upper()])
 
 
@@ -223,8 +231,10 @@ class DigitDisplay(Widget):
     """A widget to display digits and basic arithmetic operators using Unicode blocks."""
 
     digits = reactive("", layout=True)
+    """The digits to display."""
 
     supported_digits = frozenset(_character_map.keys())
+    """The digits and characters supported by this widget."""
 
     DEFAULT_CSS = """
     DigitDisplay {
@@ -242,27 +252,17 @@ class DigitDisplay(Widget):
         for widget in self._displays:
             yield widget
 
-    def _add_digit_widget(self, digit: str) -> None:
-        new_widget = _SingleDigitDisplay(digit)
-        self._displays.append(new_widget)
-        self.mount(new_widget)
-
-    def watch_digits(self, digits: str) -> None:
+    def _watch_digits(self, digits: str) -> None:
         """
         Called when the digits attribute changes.
         Here we update the display widgets to match the input digits.
         """
-        diff_digits_len = len(digits) - len(self._displays)
-
         # Here we add or remove widgets to match the number of digits
-        if diff_digits_len > 0:
-            start = len(self._displays)
-            for i in range(diff_digits_len):
-                self._add_digit_widget(digits[start + i])
-        elif diff_digits_len < 0:
-            for display in self._displays[diff_digits_len:]:
-                self._displays.remove(display)
-                display.remove()
+        while len(self._displays) < len(digits):
+            self._displays.append(_SingleDigitDisplay(digits[len(self._displays)]))
+            self.mount(self._displays[-1])
+        while len(self._displays) > len(digits):
+            self._displays.pop().remove()
 
         # At this point, the number of widgets matches the number of digits, and we can
         # update the contents of the widgets that might need it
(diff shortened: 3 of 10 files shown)


Makefile
```diff
@@ -10,7 +10,7 @@ unit-test:
 
 .PHONY: test-snapshot-update
 test-snapshot-update:
-	$(run) pytest --cov-report term-missing --cov=textual tests/ -vv --snapshot-update
+	$(run) pytest --cov-report term-missing --cov=textual tests/ -vv --snapshot-update $(PYTEST_ARGS)
```
eliasdorneles, line 13: Note to reviewer: I took the liberty to add this snippet, which let you do things like: `make PYTEST_ARGS="-k digit_display" test-snapshot-update` while developing.
Let me know if you prefer that I remove this.

src/textual/widgets/_digit_display.py
```diff
@@ -0,0 +1,271 @@
...
+            raise ValueError(f"Expected a single character, got {len(digit)}")
+        self.update(_character_map[digit.upper()])
+
+
+class DigitDisplay(Widget):
+    """A widget to display digits and basic arithmetic operators using Unicode blocks."""
+
+    digits = reactive("", layout=True)
+
+    supported_digits = frozenset(_character_map.keys())
```
eliasdorneles, line 227: Note to reviewer: this attribute is meant mostly to help client code to validate the input if needed, as the widget doesn't do any validation of the reactive attribute `digits` (should it?). If given unsupported digits/characters, it will crash.

I wasn't sure the best way to expose this, a class attribute w/ a frozenset seemed okay, let me know if this is fine.

eliasdorneles:
Screenshot (on Ubuntu - Terminator):

image

There are two example apps:

* Static examples: `textual run docs/examples/widgets/digit_display.py`
* Reactive example, w/ input to try things dynamically: `textual run docs/examples/widgets/digit_display_reacting.py`

@rodrigogiraoserrao @willmcgugan what do you think?

src/textual/widgets/__init__.py
rodrigogiraoserrao, line 1: Please insert the new widget in the correct place by alphabetical order! :D

src/textual/widgets/_digit_display.py
```diff
@@ -0,0 +1,271 @@
...
+_character_map[
+    "^"
+] = """
+·╻·
+▝·▘
+···
+"""
+
+
+_VIRTUAL_SPACE = "·"
```
rodrigogiraoserrao, line 192: Consider moving this up, next to the comment about stripping an auxiliary character, and fix the comment that mentions underscores when we're actually using this dot (which is a good idea, by the way).

src/textual/widgets/_digit_display.py
```diff
@@ -0,0 +1,271 @@
...
+
+class _SingleDigitDisplay(Static):
+    digit = reactive(" ", layout=True)
+
+    DEFAULT_CSS = """
+        _SingleDigitDisplay {
+          height: 3;
+          min-width: 2;
+          max-width: 3;
+        }
```
rodrigogiraoserrao, line 208:
```suggestion
    _SingleDigitDisplay {
        height: 3;
        min-width: 2;
        max-width: 3;
    }
```

src/textual/widgets/_digit_display.py
```diff
@@ -0,0 +1,271 @@
...
+    """
+
+    def __init__(self, initial_value=" ", **kwargs):
+        super().__init__(**kwargs)
```
rodrigogiraoserrao, line 212: I know it is a pain, but we typically write down explicitly the kwargs that widgets take.
I _think_ it's because it is helpful for IDEs.

src/textual/widgets/_digit_display.py
```diff
@@ -0,0 +1,271 @@
...
+        self.digit = initial_value
+
+    def watch_digit(self, digit: str) -> None:
+        """Called when the digit attribute changes."""
+        if len(digit) > 1:
+            raise ValueError(f"Expected a single character, got {len(digit)}")
```
rodrigogiraoserrao, line 218: This can be moved to the validate method `validate_digit` that gets called when the reactive attribute is about to change.

Docs: https://textual.textualize.io/guide/reactivity/#validation

src/textual/widgets/_digit_display.py
```diff
@@ -0,0 +1,271 @@
...
+          min-width: 2;
+          max-width: 3;
+        }
+    """
+
+    def __init__(self, initial_value=" ", **kwargs):
+        super().__init__(**kwargs)
+        self.digit = initial_value
+
+    def watch_digit(self, digit: str) -> None:
```
rodrigogiraoserrao, line 215:
```suggestion
    def _watch_digit(self, digit: str) -> None:
```

Consider making this private so that it doesn't show up as part of the public API (of this private widget).

src/textual/widgets/_digit_display.py
```diff
@@ -0,0 +1,271 @@
...
+    def compose(self) -> ComposeResult:
+        for widget in self._displays:
+            yield widget
+
+    def _add_digit_widget(self, digit: str) -> None:
+        new_widget = _SingleDigitDisplay(digit)
+        self._displays.append(new_widget)
+        self.mount(new_widget)
+
+    def watch_digits(self, digits: str) -> None:
```
rodrigogiraoserrao, line 250: You can make this private, it'll still work automagically.

src/textual/widgets/_digit_display.py
```diff
@@ -0,0 +1,271 @@
...
+
+        # Here we add or remove widgets to match the number of digits
+        if diff_digits_len > 0:
+            start = len(self._displays)
+            for i in range(diff_digits_len):
+                self._add_digit_widget(digits[start + i])
+        elif diff_digits_len < 0:
+            for display in self._displays[diff_digits_len:]:
+                self._displays.remove(display)
+                display.remove()
```
rodrigogiraoserrao, line 265: Tell me how you feel about this alternative in pseudocode

```py
while len(self._displays) < len(digits):
    self.add_digit_widget(digits[len(self._displays)])
while len(self._displays) > len(digits):
    self._displays.pop().remove()
```

This is much shorter and I tend to like this pattern, but I don't know how others react to it.

src/textual/widgets/_digit_display.py
```diff
@@ -0,0 +1,271 @@
...
+        diff_digits_len = len(digits) - len(self._displays)
+
+        # Here we add or remove widgets to match the number of digits
+        if diff_digits_len > 0:
+            start = len(self._displays)
+            for i in range(diff_digits_len):
+                self._add_digit_widget(digits[start + i])
+        elif diff_digits_len < 0:
+            for display in self._displays[diff_digits_len:]:
+                self._displays.remove(display)
```
rodrigogiraoserrao, line 264: It is faster to just `.pop` from the right instead of removing from the “middle” of the `_displays`.

rodrigogiraoserrao (changes requested):
Excellent work!

I've left some questions, comments, and suggestions in your code.

One general thing I'd like to ask is for docstrings in methods and etc to follow the format you can find elsewhere in the code.

Looking forward to merging this!

src/textual/widgets/_digit_display.py
eliasdorneles: Indeed, shorter and clearer, thanks!
I've done that, and inlined the `add_digit_wdiget` method, because I thought having the list manipulation here would make it more readable.

eliasdorneles:
@rodrigogiraoserrao thanks for the review! I've updated it taking into account your comments, also improved the screenshot of the reactive example by adding a few keys pressed =)

src/textual/widgets/_digit_display.py
rodrigogiraoserrao: Right.

rodrigogiraoserrao (approved):
It's looking good to me.

You left two of my comments without a response; I'm not sure if it is because you disagree (which is fine, please argue against me) or if you just missed them.

@willmcgugan will take a look and then we'll get back to you :)

eliasdorneles:
> You left two of my comments without a response; I'm not sure if it is because you disagree (which is fine, please argue against me) or if you just missed them.

Oops, i think i missed them, lemme check again!
I didn't do "resolve conversation" cause i didn't want to forget anything, but clearly i got lost :sweat_smile:

eliasdorneles:
Updated, I _think_ I got everything now, lemme know if there is still something missing.

src/textual/widgets/_digit_display.py
eliasdorneles: I did this for the `DigitDisplay` implementation only which is public facing -- I didn't do it for `_SingleDigitDisplay` as it's to be considered internal and an implementation detail subject to change.

willmcgugan: All the same, we like to have the argument spelled out. Its benefitial for IDEs...

src/textual/widgets/_digit_display.py
```diff
@@ -0,0 +1,278 @@
...
+
+    DEFAULT_CSS = """
+    _SingleDigitDisplay {
+      height: 3;
+      min-width: 2;
+      max-width: 3;
+    }
+    """
+
+    def __init__(self, initial_value=" ", **kwargs):
```
willmcgugan, line 212: Typing please.

src/textual/widgets/_digit_display.py
```diff
@@ -0,0 +1,278 @@
...
+    """The digits and characters supported by this widget."""
+
+    DEFAULT_CSS = """
+    DigitDisplay {
+        layout: horizontal;
+        height: 3;
+    }
+    """
+
+    def __init__(
```
willmcgugan, line 246: Docstring por favor

src/textual/widgets/_digit_display.py
```diff
@@ -0,0 +1,278 @@
...
+"""
+
+
+# here we strip spaces and replace virtual spaces with spaces
+_character_map = {
+    k: v.strip().replace(_VIRTUAL_SPACE, " ") for k, v in _character_map.items()
+}
+
+
+class _SingleDigitDisplay(Static):
```
willmcgugan, line 200: Docstring please (we like them for private classes also).

willmcgugan, line 200: Cat you add `get_content_width` and `get_content_height` methods. This allows "auto" dimensions pick a more optimal code path. There should be plenty of examples in other widgets...

willmcgugan (changes requested):

src/textual/widgets/_digit_display.py
eliasdorneles: Hello @willmcgugan :wave: 
Were you able to have a look at this?

willmcgugan: If you set your widget to `width: auto;` and `height: auto` it should call `get_content_width` etc.

src/textual/widgets/_digit_display.py
```diff
@@ -0,0 +1,337 @@
...
+        return digit
+
+    def _watch_digit(self, digit: str) -> None:
+        """Called when the digit attribute changes and passes validation."""
+        content = _character_map[digit.upper()]
+        self._content_width = len(content.splitlines()[0])
+        self.update(content)
+
+    def get_content_width(self, container: Size, viewport: Size) -> int:
+        return self._content_width
```
willmcgugan, line 265: Are all the digits not the same size? If so, could we just return 3 here?

eliasdorneles:
Just rebased, to be synced w/ latest code.

willmcgugan (commented):
Just a comment for now. This is looking good! :)

A little swamped, will test soon.

src/textual/widgets/_digit_display.py
eliasdorneles: I was trying to make the dot and comma to change width to 2, but in the end it's simpler to have every digit the same size indeed :+1: 
Lemme update this!

eliasdorneles: updated!

willmcgugan:
Hi @eliasdorneles 

Thanks for your work on this! We've decided to go with a slightly different implementation here (https://github.com/Textualize/textual/pull/3073).

It's not a reflection on your work (which is excellent) -- we've taken a different approach to the design. We used this PR as inspiration...

Thanks for joining in the sprint!

Will

eliasdorneles:
Hi @willmcgugan,

No worries, I understand.

I admit I do feel a bit disappointed that my work hasn't made the cut. At the same time, it's very interesting for me to see the differences between our designs and to learn from yours. 🤓

Anyway, I've got some experience working on open source projects, enough to understand your position, I know that it's important for a maintainer to know how to say no. You did that gently and kindly, thank you for that! 💜

Have a good day and keep up the great work! ☀️

Issue #2786, opened 2023-06-15
Add a method of cancelling a widget's animation


davep:
Currently we can [animate properties of a widget](https://textual.textualize.io/api/widget/#textual.widget.Widget.animate) but there is a situation where things may change just after the animation starts, such that the animation is heading towards a situation that no longer makes sense. The `Underline` of `Tabs` is one good example, where the "slide" animation of the underline might be kicked off, heading towards a tab, that is subsequently removed while the animation is still happening. This can end up with the underline highlight being the wrong size, or even highlighting a tab that's no longer there because all tabs have been removed.

Given this sort of situation could happen with other widgets (an animation is seeking an end result that might not make sense by the time it ends), it would be useful if there was a way of cancelling any current animations on a widget.

willmcgugan:
Suggest something like  `def stop_animation(attribute:str):`

davep:
After puzzling for too long over the way in here, I think it just clicked. The quick and dirty way of doing this seems to be:

```python
del self.app._animator._animations[(id(self), "the_attribute")]
```

With this as the (terrible, of course, but a clue) core concept, I think I can take it from there.

PR #3000, opened 2023-07-24
Deanimate! (aka, provide a method of stopping application and widget animations)

davep:
This PR seeks to satisfy the requirement outlined in #2786.

### DIFF
--- a/src/textual/_animator.py
+++ b/src/textual/_animator.py
@@ -374,6 +374,24 @@ class Animator:
         self._idle_event.clear()
         self._complete_event.clear()
 
+    def stop_animation(self, obj: object, attribute: str) -> None:
+        """Stop an animation on an attribute.
+
+        Args:
+            obj: The object containing the attribute.
+            attribute: The name of the attribute.
+
+        Note:
+            If there is no animation running, this is a no-op. If there is
+            an animation running the attribute will be left in the last
+            state it was in before the call to stop.
+        """
+        key = (id(obj), attribute)
+        if key in self._scheduled:
+            self._scheduled.remove(key)
+        if key in self._animations:
+            del self._animations[key]
+
     async def __call__(self) -> None:
         if not self._animations:
             self._timer.pause()
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,9 +10,6 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 
 - Added App.begin_capture_print, App.end_capture_print, Widget.begin_capture_print, Widget.end_capture_print https://github.com/Textualize/textual/issues/2952
-
-### Added
-
 - Added the ability to run async methods as thread workers https://github.com/Textualize/textual/pull/2938
 
 ### Changed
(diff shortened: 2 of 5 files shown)


src/textual/_animator.py
```diff
@@ -374,6 +374,24 @@ def _animate(
         self._idle_event.clear()
         self._complete_event.clear()
 
+    def stop_animation(self, obj: object, attribute: str) -> None:
```
willmcgugan, line 377: If we are stopping an animation, it could be left in a state before its natural end point. I think we should have an option to set the animate to its final destination.

I think we also have `on_complete` callbacks when an animation is complete. It doesn't appear that these will be called with this change.

willmcgugan (changes requested):

src/textual/_animator.py
davep: Good point. The "value never reaches conclusion" thing felt like a natural consequence of stopping early (so yeah, I guess making it an option makes sense when doing the stop). Calling any `on_complete` is something that did pass me by.

Just to be clear: do we want `on_complete` to be called no matter, or only if there's a "stop and complete"? Or should we perhaps have a `on_stopped` to handle both cases?

willmcgugan: I think just `on_complete` for now. Until we have a request / use case for a `on_stopped`.

davep: Noted. And on the matter of if it should always be called?

I'm thinking I'll add a `complete=True|False` to `stop_animation` (likely defaulting to `True`); so would we expect it to be called *only* if `on_complete` is `True`?

src/textual/_animator.py
```diff
@@ -64,6 +64,23 @@ def __call__(self, time: float) -> bool:  # pragma: no cover
...
+        """Stop the animation.
+
+        Args:
+            complete: Flag to say if the animation should be taken to completion.
+
+        Note:
+            [`on_complete`][Animation.on_complete] will be called regardless
+            of the value provided for `complete`.
+        """
+        await self.invoke_callback()
```
willmcgugan, line 82: The `complete` flag doesn't appear to be used?

willmcgugan: Could we make this abstract?

Without it, you could create a valid but subtly broken `Animation` subclass.

i.e. rather than a subclass call super, have it call `invoke_callback`...

src/textual/_animator.py
```diff
@@ -304,7 +353,10 @@ def _animate(
         ), "An Animation should have a duration OR a speed"
 
         animation_key = (id(obj), attribute)
-        self._scheduled.discard(animation_key)
+        try:
+            del self._scheduled[animation_key]
```
willmcgugan, line 357: Suggest `pop(animation_key, None)` here (fewer lines, probably more efficient).

src/textual/_animator.py
```diff
@@ -374,6 +426,65 @@ def _animate(
...
+    ) -> None:
+        """Stop a scheduled animation.
+
+        Args:
+            key: The key for the animation to stop.
+            complete: Should the animation be moved to its completed state?
+        """
+        # First off, pull the timer out of the schedule and stop it; it
+        # won't be needed.
+        schedule = self._scheduled[key]
```
willmcgugan, line 440: Danger Will Robinson. As this is async, a ton of code could have run since the caller awaited this method. Enough time for `key` to be removed in another task.

willmcgugan (changes requested):

src/textual/_animator.py
davep: A bit more detail about your thinking here would be helpful. For example, are you suggesting that it would be a good idea to handle the key not being in the dictionary any more, because the animation may have elsewhere run to completion in the moment between the call to stop and this method getting hit?

willmcgugan: Exactly, you can't assume that key will still be in that dict.

davep: Cool. Handled.

PR #3002, opened 2023-07-24
Fix a crash when setting `DirectoryTree.show_root` before DOM is ready

### LINKED ISSUE #2363: Setting `show_root` to `True` raises a `KeyError`
```py
    def compose(self):
        dir_tree = DirectoryTree("./src/routes")
        dir_tree.show_root = True
        yield Header()
        with Horizontal():
            yield Logo()
            with TabbedContent("Config", "Routes", "Server"):
                yield Config(self.config)
                yield dir_tree
                yield Label("hello")
        yield Footer()
```
setting `show_root` to `True`. raises the following error:
`KeyError: "No 'directory-tree--folder' key in COMPONENT_CLASSES"`

Diagnose output:

# Textual Diagnostics

## Versions

| Name    | Value  |
|---------|--------|
| Textual | 0.20.1 |
| Rich    | 13.3.3 |

## Python

| Name           | Value                                                                                                    |
|----------------|----------------------------------------------------------------------------------------------------------|
| Version        | 3.11.0                                                                                                   |
| Implementation | CPython                                                                                                  |
| Compiler       | Clang 14.0.0 (clang-1400.0.29.202)                                                                       |
| Executable     | /Users/felixingram/Library/Application Support/hatch/env/virtual/sanic-kit/1YPPa8fF/sanic-kit/bin/python |

## Operating System

| Name    | Value                                                                                                 |
|---------|-------------------------------------------------------------------------------------------------------|
| System  | Darwin                                                                                                |
| Release | 22.4.0                                                                                                |
| Version | Darwin Kernel Version 22.4.0: Mon Mar  6 21:00:41 PST 2023; root:xnu-8796.101.5~3/RELEASE_ARM64_T8103 |

## Terminal

| Name                 | Value                              |
|----------------------|------------------------------------|
| Terminal Application | WezTerm (20230326-111934-3666303c) |
| TERM                 | xterm-256color                     |
| COLORTERM            | truecolor                          |
| FORCE_COLOR          | *Not set*                          |
| NO_COLOR             | *Not set*                          |

## Rich Console options

| size           | width=203, height=46 |
| max_width      | 203                  |
| max_height     | 46                   |

davep:
Fixes #2363.

Because labels can sometimes be needed by the tree before the DOM is ready, this changes `DirectoryTree.render_label` so that it only attempts to *style* the labels if the DOM is up and running.


src/textual/widgets/_directory_tree.py
```diff
@@ -203,25 +203,29 @@ def render_label(
 
         if node._allow_expand:
             prefix = ("📂 " if node.is_expanded else "📁 ", base_style + TOGGLE_STYLE)
-            node_label.stylize_before(
-                self.get_component_rich_style("directory-tree--folder", partial=True)
-            )
+            if self.app.is_mounted(self):
```
willmcgugan, line 206: I have a feeling this would be better done before `render_label`.

It doesn't make much sense for `render_label` to be called prior to mount. Does the user even see the label rendered pre-mount?

Can we bailed out earlier than this?

willmcgugan (changes requested):

src/textual/widgets/_directory_tree.py
willmcgugan: The things is, setting show_root prematurely invokes this code, but it will run again post mount. The change here makes the error go away but I think the issue is that this code runs at all.

It's also an annoying precedent, in that we may get the same issue with other renderables and watches. And I don't want to establish this as a pattern.

Can you investigate a solution, which doesn't place the burden on the widget developer?

davep: Do you mean investigate a different approach to how Tree works, or how watches work?

willmcgugan: Honestly, could be either. Whatever is the most elegant solution.

davep:
Having discussed this some more in the 2023-07-25 standup, we've decided that there's a more general class of issue here with watchers (and friends) and doing things before the DOM is fully mounted. Given this I'm going to close this PR, return the motivating issue back to TODO (actually perhaps place in backlog for now), and raise a fresh issue relating to the more general problem.

rodrigogiraoserrao:
Hey Dave, given what we discussed in today's standup and the way [issue 3011](https://github.com/textualize/textual/issues/3011) was resolved, do you want to reopen this PR & change the mounted checks to make use of `Widget.is_mounted`?

I couldn't find a better place to add the `is_mounted` checks and I'd feel a bit silly to just copy all your changes and make a PR with them.

davep:
Sure. Weirdly I'd forgotten about this and the fact that there is an is_mounted kicking about anyway(?).

davep:
Or not...

image

darrenburns:
@davep I think you'll need to restore the branch before you can re-open. Hopefully there's a button to do it on the GitHub web UI for you. I think it should be next to the "deleted the branch" text here:

image

davep:
Right, that seems to have worked. @rodrigogiraoserrao did you want to snarf this and tidy it up then? (obvious thing to note: the changelog entry location will be *waaaaaaaaaaay* out of date now).

rodrigogiraoserrao:
@davep snarfed, updated, and committed.
I don't think you can review your own PR and it doesn't make sense for me to review the PR I also updated, so we'll wait for @darrenburns or @willmcgugan to review & approve.

willmcgugan:
@rodrigogiraoserrao I left a comment last year, can you please take a look.

src/textual/widgets/_directory_tree.py
```diff
@@ -297,25 +297,29 @@ def render_label(
 
         if node._allow_expand:
             prefix = ("📂 " if node.is_expanded else "📁 ", base_style + TOGGLE_STYLE)
-            node_label.stylize_before(
-                self.get_component_rich_style("directory-tree--folder", partial=True)
-            )
+            if self.is_mounted:
```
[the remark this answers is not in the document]
rodrigogiraoserrao: I think the idea was to build a label as faithful to the intended one as possible.
If you want, we can return early with a simpler version of `node_label`.
I've tested that locally and the tests still pass; do you want me to commit it?

rodrigogiraoserrao:
> @rodrigogiraoserrao I left a comment last year, can you please take a look.

Sorry, it must have slipped past me. [I replied](https://github.com/Textualize/textual/pull/3002#discussion_r1444438446).

src/textual/widgets/_directory_tree.py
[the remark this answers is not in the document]
willmcgugan: Yes please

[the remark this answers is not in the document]
rodrigogiraoserrao: Done in 8087eac.

willmcgugan (approved):

PR #3004, opened 2023-07-24
fix: scalar animation with percentages

### LINKED ISSUE #2940: CSS Transition rubber bands when %'s are used

yes

**current**: when using the `transition` css property on `width` with a percentage-based start/stop, the animation is jerky/rubber-bands. When moving from larger to smaller, it seems to move past the desired value (maybe to the default widget size?) then snap back to the specified value. When moving from smaller to larger, it appears to start moving from at or near a width of zero, hitting the current width, and then snapping out to the desired width.

**expected**: transition should look like the `textual easing` examples when applied to %-based widths.

**of note**: when using concrete values instead of percents, all seems well.

https://github.com/Textualize/textual/assets/43392618/5d5f4df1-1ab3-473b-9359-b1339e620d5d

# Textual Diagnostics

## Versions

| Name    | Value  |
|---------|--------|
| Textual | 0.28.1 |
| Rich    | 13.4.2 |

## Python

| Name           | Value                                                                        |
|----------------|------------------------------------------------------------------------------|
| Version        | 3.10.6                                                                       |
| Implementation | CPython                                                                      |
| Compiler       | GCC 11.3.0                                                                   |
| Executable     | /home/epi/.cache/pypoetry/virtualenvs/ui-overhaul-obYiRisJ-py3.10/bin/python |

## Operating System

| Release | 5.19.0-46-generic                                                |
| Version | #47~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Wed Jun 21 15:35:31 UTC 2 |

## Terminal

| Terminal Application | tmux (3.2a)    |

## Rich Console options

| size           | width=157, height=47 |
| max_width      | 157                  |
| max_height     | 47                   |

```python
from textual.widgets import Header, Input, Footer, Button


class Issue2934(App[None]):

    TITLE = "Test Issue 2934"

    CSS = """
        #my-input {
            width: 20%;
            margin: 1;
            transition: width 500ms in_out_cubic;
        }

        #my-input:focus {
            width: 80%;
        }

        #my-input:blur {
            width: 20%;
        }
    """

        yield Header()
        with Container():
            yield Input(id='my-input')
            yield Button('focusable')
        yield Footer()

    def action_toggle(self, bar: str) -> None:
        self.query_one(f"#{bar}bar").toggle_class("shown")


if __name__ == "__main__":
    Issue2934().run()
```

TomJGooding:
### Description

Fixes #2940 by resolving the scalar animation `start` and `destination` cell sizes based on the size of the container (parent) rather than the widget.

### Related Issue

- #2940

### How Has This Been Tested?

The added test sets both the initial widget width and the animation value as percentages. ~~The size is checked at specific intervals during the animation using `pilot.pause`, which might be why the test is a bit flaky?~~

This has also been tested manually, with both the `animate` method and the ([currently undocumented](https://github.com/Textualize/textual/issues/2940#issuecomment-1637523027)) CSS `transition` feature.

### Checklist:

- [ ] Updated documentation


TomJGooding:
Should a check be added here? mypy complains that `"DOMNode | None" has no attribute "size"`, but am I right in thinking that all widgets have a parent which will be either another widget or the app?

https://github.com/Textualize/textual/blob/33d775ab823f4c01d819a29e7581dfe8453c7e9a/src/textual/css/scalar_animation.py#L38

TomJGooding:
Thanks Will, that's what I thought. I guess someone might try to animating the size of the app...!

willmcgugan:
@TomJGooding Is this one still relevant?

TomJGooding:
Sorry I realise this PR has been hanging around for a while, but after some tinkering and hopefully now with a better understanding of Textual, I'm tentatively marking this as ready for review.

You'll see I've struggled with a creating a test that isn't flaky in CI (only in Windows for some reason), so I'd welcome any suggestions for improving this.

TomJGooding:
Closed as stale.

Issue #2923, opened 2023-07-11
EuroPython 2023 sprint meta issue. Start here!


rodrigogiraoserrao:
Welcome to the [EuroPython 2023 Textual sprint](https://ep2023.europython.eu/sprints#textual)!

First of all, thanks for your interest in contributing!
This meta issue lists some issues that we hand-picked for you to tackle during the sprint.
Take a look at the instructions below, pick an issue to tackle, and have fun!

Throughout the sprint, I (Rodrigo, @rodrigogiraoserrao) will be around to help you and some other maintainers may also be available to jump in when needed.
Also, if you [join the Discord server](https://discord.gg/Enf6Z3qhVr) you'll have access to many other Textual users and enthusiasts who may be able to lend you a hand.

# Before contributing

If you want to participate in this sprint, 1+ year of Python experience is recommended.
You are also encouraged to [go through the Textual tutorial](https://textual.textualize.io/tutorial/), if you haven't yet, to ensure you have at least some understanding of what Textual does.

# Setup

To develop `Textual`, you'll need to clone this repository.
Then, [make sure you have Poetry installed](https://python-poetry.org) and run `poetry install --all` to install all of the (dev) dependencies.

Make sure everything is working fine by running `textual --version`.

# Good first issues

These issues were handpicked by us for you to tackle at the EuroPython sprint.
They should be self-contained and they aren't technically _very_ demanding... but they will take time to tackle!

 - CLOSED ~~** #2916 – add a new widget to Textual; peek at how Textual actually displays things in a terminal~~
 - ** #2917 – ditto
 - CLOSED ~~** #2982 – ditto~~
 - CLOSED ~~** #2918 – build a new compound widget: a widget built out of existing widgets~~
 - #2919 – build a new compound widget: a widget built out of existing widgets (this one depends on the previous one)
 - ** #2920 – enhance a widget with more styles
 - CLOSED ~~** #2757 – enhance a widget with a new method~~
 - CLOSED ~~** #2603 – enhance a widget with a new method~~
 - ** #2921 – write an app for the Textual CLI
 - CLOSED ~~** #2922 – modify an app from the Textual CLI~~
 - CLOSED ~~** #2905 – fix a bug related to message bubbling~~
 - CLOSED ~~** #2366 – fix another bug related to message bubbling~~
 - CLOSED ~~** #1849 – fix a bug with CSS parsing~~

[...]

PR #3005, opened 2023-07-25
fixes #2991 by adjusting the y scroll after max_lines exceeded

### LINKED ISSUE #2991: TextLog auto_scroll = False doesn't work as intended if max_lines exceeded
If the amount of text in a TextLog hits max_lines, and the user scrolls back, their viewport will scroll forward as new content is added.

Here's an example where I page up, setting auto_scroll=False.

scroll

As new content is added, the point on the scrollbar stays constant, but the content changes.  With auto_scroll = False, I would expect the viewport to remain constant and the scroll bar point to continually get further back, until we reach the edge of history where that is no longer possible.

Screen grab from textual 0.30.0.

mardanst:

- [x] Updated documentation


rodrigogiraoserrao:
Hey @mardanst, thanks for the PR.

Before we take a look at it, could you make sure the tests pass locally?
You can go through the steps outlined in #2923, from the section "Guidelines" onwards.
Thanks!

mardanst:
That link was very helpful.  I was able to fix the failing test and add a new snapshot test for #2991 .  

Tests are passing locally, but failing in github on (windows-latest,3.11) on an unrelated test.

mardanst:
Resubmitted it and it passed the second time through

rodrigogiraoserrao:
Thanks for following up on it. We'll take a look and get back to you.

As you realised, some of our tests are a bit flakey. When one of those fails, no need to create a "fake" commit to trigger them again. We'll rerun them for you :)

willmcgugan:
Closing, assumed stale. I don't think the issue is still present.

PR #3012, opened 2023-07-25
feat(listview): add method to append multiple items

### LINKED ISSUE #2997: ListView method to append many
There is a method on ListView to append a single item. We could also use a method to append many items.

If we're following the same conventions as a builtin list, this would be called `ListView.extend` and would accept an Iterable of ListItems.

TomJGooding:
Closes #2997 by adding a `ListView.extend` method to append multiple items. This is essentially a cut-and-paste from the existing `append` method!

- [ ] Updated documentation


willmcgugan (approved):
LGTM

TomJGooding:
Thanks! Looks like the changelog has become a bit out of sync, do you want me to submit  new PR or will you do this housekeeping before the next release?

PR #3046, opened 2023-08-01
Rich log

willmcgugan:
- Renamed TextLog to RichLog
- Added Log widget
- Added Widget.is_vertical_scroll_end, Widget.is_horizontal_scroll_end, Widget.is_vertical_scrollbar_grabbed, Widget.is_horizontal_scrollbar_grabbed
- Added Strip.crop_extend
- Added LRUCache.discard

PR #3048, opened 2023-08-01
docs(faq): correct wording for workers

TomJGooding:
Some suggested corrections to the wording of the new worker FAQ.

Perhaps a short description (or a link) to explain what threaded means would also be useful?

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -7,7 +7,7 @@
 - [How can I select and copy text in a Textual app?](#how-can-i-select-and-copy-text-in-a-textual-app)
 - [How can I set a translucent app background?](#how-can-i-set-a-translucent-app-background)
 - [How do I center a widget in a screen?](#how-do-i-center-a-widget-in-a-screen)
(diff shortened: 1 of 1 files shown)


davep:
> Perhaps a short description (or a link) to explain what threaded means would also be useful?

Hopefully people who are running into this FAQ (which relates to a breaking change released today) [will have been using a thread worker based off the docs about them](https://textual.textualize.io/guide/workers/#thread-workers).

willmcgugan:
Good spot! I was hasty in updating that.

faq.md is built by `faqtory`. You would need to edit the appropriate `.questions.md` file and call `faqtory build` (`pip install faqtory`) to regenerate it.

TomJGooding:
Ah, thanks Will. I can't see a matching `.questions.md` in main so might not have been merged, but sure I can figure it out...

willmcgugan:
Ah, you're right. My bad. Don't worry about then, I'll sneak it in a PR.

willmcgugan:
Snuck it in https://github.com/Textualize/textual/pull/3046

PR #3050, opened 2023-08-02
fix: display same key symbol when used with modifier

### LINKED ISSUE #3035: Binding for `]` renders incorrectly in `Footer`
The binding `Binding("ctrl+right_square_bracket", "toggle_indent_width", "Cycle indent width")` renders like this:

image

It should probably render as `Ctrl+]`.

TomJGooding:
### Description

Fixes #3035, so that a key display is the same if used with a modifier.

**Before:**

before

**After:**

after

### Checklist:

- [ ] Updated documentation


TomJGooding:
Alternatively here's a quick test with spaces around the plus sign to make it less cramped:

image

Either way, I think it would be an improvement on `CTRL+RIGHT_SQUARE_BRACKET`!

TomJGooding:
I've merged main again to resolve changelog conflicts, but not added a new entry yet for this fix.

darrenburns (approved):
Sorry we missed this - looks good to me 👍 just awaiting the changelog entry.

TomJGooding:
No problem, I think just this PR fell between the cracks where I didn't to bother you by chasing but then forgot about it myself until today!

PR #3054, opened 2023-08-03
DocdocsFix width.md so the link to box_sizing.md works

dmunozv04:

- [ ] Updated documentation

### DIFF
--- a/docs/styles/width.md
+++ b/docs/styles/width.md
@@ -9,7 +9,7 @@ width: <a href="../../css_types/scalar">&lt;scalar&gt;</a>;
 --8<-- "docs/snippets/syntax_block_end.md"
 
 The style `width` needs a [`<scalar>`](../../css_types/scalar) to determine the horizontal length of the width.
-By default, it sets the width of the content area, but if [`box-sizing`](./box_sizing) is set to `border-box` it sets the width of the border area.
+By default, it sets the width of the content area, but if [`box-sizing`](./box_sizing.md) is set to `border-box` it sets the width of the border area.
 
 ## Examples
 


rodrigogiraoserrao:
Hey there, David, thanks for your PR, that's a great catch.

Sadly, I just noticed there are a bunch of issues with links like those in the docs for the styles.
If you want, you could try to hunt those down and add those fixes to this PR – but I suspect that won't be a lot of fun.

I'd like all link fixes to be made in a single go, though, so I'll wait to merge this in.
I hope you understand 🙌

dmunozv04:
I've been trying to figure a way to find all the broken links, and apparently mkdocs gives warnings whenever it finds broken links. 
[textual-docs-broken-links.txt](https://github.com/Textualize/textual/files/12290066/textual-docs-broken-links.txt)
It should be much easier to handle with this list, I'll give it a go and try to fix them all.

rodrigogiraoserrao:
> I've been trying to figure a way to find all the broken links, and apparently mkdocs gives warnings whenever it finds broken links (during the build docs process).
> [textual-docs-broken-links.txt](https://github.com/Textualize/textual/files/12290066/textual-docs-broken-links.txt)
> It should be much easier to handle with this list, I'll give it a go and try to fix them all.

This looks like a smart approach! You can probably search for lines that include link warnings, right?
What I noticed was that I mistook two styles of writing links. The broken link you fixed used a relative path, like `[something here](./some_file_here)` but I didn't provide the complete path because I forgot the file extension, which you fixed into `[something here](./some_file_here.md)`.
I suspect a search for the pattern of parenthesis with text starting with `./` _could_ also help..?

Either way, I'll wait for further updates from you!

rodrigogiraoserrao:
And thanks!

dmunozv04:
You're welcome, after doing a mix of automatic and manual tasks, I think most links are now fixed.

There are a few exceptions, which I don't know how to handle:

- Links that are inside html `a` tags. They work ok, but aren't formatted like the rest of the links (with the .md extension) [example](https://github.com/Textualize/textual/blob/5a662e68aefc695c02203c8b3720a4c90a7dce09/docs/styles/margin.md?plain=1#L8)
- Links made when showing python Docstrings, which happen in the `/api` path, which also work but also don't follow the format convention [example](https://github.com/Textualize/textual/blob/5a662e68aefc695c02203c8b3720a4c90a7dce09/src/textual/app.py#L5)

dmunozv04:
Just for reference, here is the current stripped output when building the docs:
[textual-docs-broken-links-2.txt](https://github.com/Textualize/textual/files/12290733/textual-docs-broken-links-2.txt)

docs/styles/layer.md
```diff
@@ -5,11 +5,11 @@ The `layer` style defines the layer a widget belongs to.
 ## Syntax
 
 --8<-- "docs/snippets/syntax_block_start.md"
-layer: <a href="../../css_types/name">&lt;name&gt;</a>;
+layer: <a href="../css_types/name.md">&lt;name&gt;</a>;
```
rodrigogiraoserrao, line 8: This link is now broken!

rodrigogiraoserrao (changes requested):
Wow, you fixed A LOT of links!
I found one single link that you seem to have broken by mistake – one of those links inside `<a>` tags.

rodrigogiraoserrao:
Links inside `<a>` tags need to have a path that is either absolute or relative _to the website paths_, whereas the relative links ending in `.md` that we use throughout the docs and that you used only work inside markdown.

rodrigogiraoserrao:
Links inside docstrings are links inside `python` files and it feels like they are processed in a different way, so to me it looks like relative links won't work there.
It is ok to leave them as-is.

rodrigogiraoserrao (approved):
LGTM 🚀

Once again, thanks!

PR #3056, opened 2023-08-03
Fix animation of offset with tuple and string #3028

### LINKED ISSUE #3028: Animate offset?
The docs say that "You can apply animations to styles such as offset to move widgets around the screen".

But how?

I found this discussion https://github.com/Textualize/textual/discussions/1569 from back in January, which suggests perhaps this is might be a bug?

```python
from textual.geometry import Offset
from textual.widgets import Footer, Static


class ExampleApp(App):
    BINDINGS = [("space", "animate_offset", "Animate Offet")]

    CSS = """
    Screen {
        align: center middle;
    }

    Static {
        background: red;
        width: auto;
        height: 3;
        content-align: center middle;
    }
    """

        yield Static("Animate me!")
        yield Footer()

    def action_animate_offset(self) -> None:
        static = self.query_one(Static)
        static.styles.animate(
            attribute="offset",
            value=Offset(10, 10),  # crashes
            # value="10 10",  # crashes
            # value=(10, 10),  # crashes
            duration=0.5,
        )


if __name__ == "__main__":
    app = ExampleApp()
    app.run()
```

yuvalmo:
Closes #3028 by doing to needed conversion from tuple / str into `ScalarOffset` so offset can be animated in an intuitive way.

- [ ] Updated documentation

I am not sure which docs, if any, should be updated.
The bug page states that this is the expected behavior, so maybe the documentation already matches this PR?


src/textual/css/styles.py
```diff
@@ -372,6 +372,13 @@ def __textual_animation__(
             # If destination is a number, we can convert that to a scalar
             if isinstance(value, (int, float)):
                 value = Scalar(value, Unit.CELLS, Unit.CELLS)
+            # If destination is a pair of numbers, we can convert it to a
+            # scalar offset
+            elif isinstance(value, tuple) and (
+                list(map(type, value)) == [int, int]
```
yuvalmo, line 378: This basically checks that `value` is of type `tuple[int, int]` or `tuple[float, float]`.
This is the best way I found to implement it.
If anyone has any suggestions, please let me know.

src/textual/css/scalar.py
```diff
@@ -367,6 +367,24 @@ def resolve(self, size: Size, viewport: Size) -> Offset:
...
+        """Create a scalar offset from a string.
+
+        Args:
+            token: String to parse (e.g. '1 2' or '1.0 2.0')
+
+        Returns:
+            New offset
+        """
+        x, y = token.split()
+        # fmt: off
```
yuvalmo, line 381: This comment is to prevent the pre-commit from formatting this.

I don't know if this is accepted here or not, but I thought that:
1. It easier to read this way.
2. It matches the way the rest of the class is formatted (see `ScalarOffset.resolve`).

willmcgugan: No need to disable black here, add a trailing comma...

yuvalmo: Thanks for the tip, I fixed it.

Issue #3021, opened 2023-07-27
Command Palette


willmcgugan:
A command palette where the user can enter commands that are queried via fuzzy search.

Please update with details.

PR #3058, opened 2023-08-04
Command palette

davep:
Implements #3021.

Key features of the PR:

- Adds a "command palette" (AKA minibuffer if you're an Emacs fan) that pops up over a Textual application.
- Adds a method of providing command sources, all of which will be searched.
- Runs the code associated with the command once a choice has been made.
- Hits from sources are gathered up async, so the completions can keep coming in as the user interacts with the system.

Current state of work in a simple test application, where there's a command source of lines of text, where a selected line will cause the text to be shown in a Textual notification. The command source is coded to pretend to be a very slow source, with noticeable pauses between each command been provided.

https://github.com/Textualize/textual/assets/28237/476304b6-b06b-4958-9566-5c045024c629

The matching in the command provider is done using Textual's (as yet undocumented) builtin "fuzzy matcher". The code for the command provided used here is:

```python
    async def search(self, query: str) -> CommandMatches:
        """A request to hunt for commands relevant to the given user input.

        Args:
            query: The user input to be matched.
        """
        print(f"begin search(\"{query}\")")
        try:
            # Get a Textual fuzzy matcher.
            matcher = self.matcher(query)
            # For each line in the data source...
            for candidate in self.DATA:
                # Sleep some random amount of time, just so we can pretend
                # to be a slow source.
                await sleep(random() / 10)
                # If the match is above zero...
                if matcher.match(candidate):
                    # ...create a hit object and pass it back up to the
                    # command palette for use there.
                    yield CommandSourceHit(
                        # The match score -- used for sorting.
                        matcher.match(candidate),
                        # The Rich Text object for showing the command and
                        # how it matched.
                        Text.assemble(
                            Text.from_markup("[italic green]notify('[/]"),
                            matcher.highlight(candidate),
                            Text.from_markup("[italic green]')[/]")
                        ),
                        # The code to run this if this match is picked by
                        # the user.
                        partial(self.screen.notify, candidate),
                        # The original text of the command itself.
                        candidate,
                        # Some optional help text for the command; here used
                        # to help show easy access to key information about
                        # the environment.
                        "Show the selected text as a notification\n"
                        f"I think the current screen is {self.screen!r}\n"
                        f"I think the focused widget is {self.focused!r}\n"
                        f"Match score: {matcher.match(candidate):0.5f}"
                    )
        except CancelledError:
            print(f"cancelled search(\"{query}\")")
        finally:
            print(f"end search(\"{query}\")")
```

where `self.DATA` is simply a list of lines of text.

A slightly more sensible example, this time hooked up with a command source that wraps the main Textual application actions:

https://github.com/Textualize/textual/assets/28237/2ff82c61-55ca-4472-86eb-6fc1d84fdb1d

Still to do (not an exhaustive list):

- [x] Improve and fix the fuzzy matcher.[^1]
- [x] Add support to the fuzzy matcher for being given styles for highlighting.
- [x] Add a component class to help control the styling of matched text in the hits.
- [x] Look at a good sort order for command hits.
- ~~Related to the above, consider support for `OptionList` for inserting new options in specific locations in the list.~~
- [x] Add support for using the command palette in an example app or two (the code browser looks like a prime candidate).
- [x] Finally decide how the command palette will be called upon by the user, if it should be enabled by default, and how configurable that should be by the developer.
- [x] Add global enable/disable flag on App
- [x] Add a reference command source for obvious commands on App.
- [x] Unit tests.
- [x] Snapshot tests.
- [x] Main documentation.
- ~~Likely a full entry in the guide.~~[^2]
- ~~Perhaps also a HOWTO on how to write your own command source.~~[^3]
- [x] Test on Windows; especially the choice of default key binding.
- [ ] Document the default key binding somewhere, so developers know how this is initially invoked.[^4]
- [x] Try a short delay before the loading indicator appears. Suggest 500ms, to prevent flicker.
- [x] Remove the border that separates the input from the results.
- [x] Try adding a 🔎 emoji before the input text, which aligns with the results.
- [x] Have the highlight in the command list extend as far as possible left and right.
- [x] Add appropriate entries to the ChangeLog.
- [x] Update the "Released in 0.??.0" text to the release version.
- [ ] Double-check the issues with <kbd>Ctrl</kbd>+<kbd>Space</kbd> on Windows, relating to pasting from the clipboard ([see comment way down below](https://github.com/Textualize/textual/pull/3058#issuecomment-1702591818)).

[^1]: Actually, aside from some tweaks to the interface, I've left it as-is for the moment. I do wonder if we can improve the scoring -- perhaps scoring earlier matches over later and the like -- but I think it's fine for the moment.
[^2]: The more I think on this, the more I think it makes more sense that, initially at least, this is documented as part of the API (which it is).
[^3]: I think this should come later if it's decided it's needed.
[^4]: Going to be added by @willmcgugan before going into `main`; likely a main docs and blog combo.


src/textual/command_palette.py
```diff
@@ -0,0 +1,700 @@
...
+
+        Args:
+            user_input: The user input to be matched.
+
+        Yields:
+            Instances of [`CommandSourceHit`][textual.command_palette.CommandSourceHit].
+        """
+        raise NotImplemented
+
+    def run(
```
willmcgugan, line 152: Does this need to be public? Feels like something that should be internal.

`run` seems an odd name, given it returns a callable but doesn't actually run it.

willmcgugan (changes requested):
Very nice, a few points as discussed.

- Might need to tweak the highlight colors a bit. I made a change that looks good in dark, but not great in light any more.
- Try a short delay before the loading indicator appears. Suggest 500ms, to prevent flicker.
- Try removing the border that separates the input from the results.
- Try adding a 🔎 emoji before the input text, which aligns with the results.

src/textual/command_palette.py
davep: Yes, it needs to be public as it's a helper method that's part of the public API for making a command source (see its docstring for the explanation of what it does). I'm open to a better name if you have one, but it felt reasonably declarative: "when this command is selected `run` this thing that might be sync or async", sort of thing.

src/textual/app.py
```diff
@@ -431,6 +445,14 @@ def __init__(
         The new value is always converted to string.
         """
 
+        self.use_command_palette: bool = True
```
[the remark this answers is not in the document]
davep: Done.

src/textual/_system_commands_source.py
```diff
@@ -0,0 +1,83 @@
...
+    """Help text for the command."""
+
+
+class SystemCommandSource(CommandSource):
+    """A [source][textual.command_palette.CommandSource] of command palette commands that run app-wide tasks.
+
+    Used by default in [`App.COMMAND_SOURCES`][textual.app.App.COMMAND_SOURCES].
+    """
+
+    async def hunt_for(self, user_input: str) -> CommandMatches:
```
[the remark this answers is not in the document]
davep: Done.

src/textual/command_palette.py
```diff
@@ -0,0 +1,705 @@
...
+
+    This is the text that will be placed into the `Input` field of the
+    [command palette][textual.command_palette.CommandPalette] when a
+    selection is made.
+    """
+
+    command_help: str | None = None
+    """Optional help text for the command."""
+
+    def __lt__(self, other: CommandSourceHit) -> bool:
```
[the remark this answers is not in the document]
davep: Good catch; `Command` too. Done.

davep:
> Might need to tweak the highlight colors a bit. I made a change that looks good in dark, but not great in light any more.

@willmcgugan Given mention of a `textual.Text`, which would the problem with light mode at a stroke (I think), might it not be better to just revert this the the default inverse that we had for now and revisit this once there's a `textual.Text`?

davep:
Tested on Windows 11 in the Windows Terminal and working as expected:

Screenshot 2023-08-22 at 08 33 50

src/textual/command_palette.py
davep: Now removed as per the chat this morning.

src/textual/command_palette.py
```diff
@@ -0,0 +1,754 @@
...
+        # If all the sources are pretty fast it could be that we've reached
+        # this point but the queue isn't empty yet. So here we flush the
+        # queue of anything left. Note though that rather than busy-spin the
+        # queue and just pull items and yield them, we keep using the
+        # await/wait_for so we don't block until we're done. Not doing this
+        # makes typing into the input *very* choppy when you have very fast
+        # sources.
+        while not commands.empty():
+            try:
+                yield await wait_for(commands.get(), 0.1)
```
willmcgugan, line 557: I don't think we need the `wait_for` here. If the queue is not empty, then `get` is guaranteed to return quickly.

src/textual/command_palette.py
```diff
@@ -0,0 +1,754 @@
...
+            self.get_component_rich_style("command-palette--highlight")
+        )
+
+        # Set up a queue to stream in the command hits from all the sources.
+        commands: Queue[CommandSourceHit] = Queue()
+
+        # Fire up an instance of each command source, inside a task, and
+        # have them go start looking for matches.
+        searches = [
+            create_task(
```
willmcgugan, line 516: It doesn't look like these tasks are ever cancelled when a new search comes in. This may not produce an error but it is likely that they will be doing work that is never seen, and I suspect it will generate asyncio warnings.  Might be worth try this with [asyncio debug](https://docs.python.org/3/library/asyncio-dev.html) enabled.

willmcgugan (changes requested):
Potentially some issues around concurrency here. Can chat about it if you like.

tests/command_palette/test_command_source_environment.py
```diff
@@ -0,0 +1,45 @@
...
+from textual.screen import Screen
+from textual.widget import Widget
+from textual.widgets import Input
+
+
+class SimpleSource(CommandSource):
+    environment: set[tuple[App, Screen, Widget | None]] = set()
+
+    async def search_for(self, _: str) -> CommandMatches:
+        def gndn() -> None:
```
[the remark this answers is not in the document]
davep: Goes nowhere, does nothing; an old and often-used acronym borrowed form Star Trek for placeholder functions that don't do anything.

src/textual/command_palette.py
davep: See the comment explaining this loop for why the `wait_for` is there. Long story short: it makes it all a lot more responsive when the queue has been flooded.

src/textual/_system_commands_source.py
```diff
@@ -0,0 +1,61 @@
...
+from .command_palette import CommandMatches, CommandSource, CommandSourceHit
+
+
+class SystemCommandSource(CommandSource):
+    """A [source][textual.command_palette.CommandSource] of command palette commands that run app-wide tasks.
+
+    Used by default in [`App.COMMAND_SOURCES`][textual.app.App.COMMAND_SOURCES].
+    """
+
+    async def search_for(self, user_input: str) -> CommandMatches:
```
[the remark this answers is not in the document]
davep: Done.

src/textual/command_palette.py
```diff
@@ -0,0 +1,754 @@
...
+
+    DEFAULT_CSS = """
+    SearchIcon {
+        margin-left: 1;
+        margin-top: 1;
+        width: 2;
+    }
+    """
+
+    icon: var[Text] = var(Text.from_markup(":magnifying_glass_tilted_right:"))
```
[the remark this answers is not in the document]
davep: Perfect. `rich.emoji.Emoji.replace` looks to be the thing; that's really what I wanted. I'm not a fan of hard-coding emoji into strings, preferring something more readable. That'll solve this in a nicer way.

src/textual/command_palette.py
willmcgugan: I don't see how. If `commands` is not empty, then `await commands.get()` should return instantly.

src/textual/command_palette.py
```diff
@@ -0,0 +1,754 @@
...
+                )
+            )
+            for source in self._sources
+        ]
+
+        # Set up a delay for showing that we're busy.
+        self._start_busy_countdown()
+
+        # Now, while there's some task running...
+        while any(not search.done() for search in searches):
```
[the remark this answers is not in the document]
davep: The first-encountered error is now re-raised and shows as the developer may expect.

Screenshot 2023-08-29 at 08 56 10

src/textual/command_palette.py
```diff
@@ -0,0 +1,864 @@
...
+                # A cancelled error means things are being aborted.
+                aborted = True
+            else:
+                # There was no timeout, which means that we managed to yield
+                # up that command; we're done with it so let the queue know.
+                commands.task_done()
+
+        # Check through all of the finished searches and, if any of them has
+        # an exception, bubble it up.
+        search_exception = next(
```
willmcgugan, line 566: Normally I'd like the approach of bailing out when an error occurs.

For this use case though, I think it may be better to log the exceptions rather than have the app exit.

My reasoning is that we don't want a single broken command provider to make the command palette unusable.

willmcgugan (changes requested):

src/textual/command_palette.py
davep: Done.

davep:
The choice of default key binding might be questionable after all: while it has been shown to work fine on macOS and Windows, I've just noticed that, in Windows 11 under Parallels at least, if I go to paste something into a Textual application from the clipboard it causes the command palette to appear first and then the text being pasted is pasted into that -- this would suggest that the paste operation is causing the same sequence to be passed in as that generated by <kbd>ctrl</kbd>+<kbd>space</kbd>.

willmcgugan (approved):

Issue #2982, opened 2023-07-22
Add `Rule` widget


rodrigogiraoserrao:
This issue concerns itself with adding [a new widget](https://textual.textualize.io/guide/widgets/) to Textual.
The rule widget should be similar to [HTML rules](https://www.w3schools.com/tags/tag_hr.asp).
The key difference is that HTML only has a horizontal rule and `Rule`'s orientation can be customised (`"horizontal"` or `"vertical"`).

You will want to implement a method `render` that is responsible for returning the characters that make up the rule itself.
You may also look at `_progress_bar.py::Bar.render` and `_progress_bar.py::Bar.render_indeterminate` for some inspiration on how to implement “a line” that you can render.

I suggest you start by implementing a horizontal rule.
After it works well, think about creating a vertical rule as well and make it customisable via a parameter, e.g., `Rule(orientation="horizontal")` and `Rule(orientation="vertical")`.

Take a look at #2333 to see what files need to be changed when you implement a new widget.
Then, don't be like me and also do the final step shown in #2399 to add your new widget to the [widget gallery](https://textual.textualize.io/widget_gallery/).

Good luck and have fun!

scriptogre:
Working on it! 😎

rodrigogiraoserrao:
Hey Chris, we exchanged a couple of messages in private but I'd like to reiterate my question: are you working on this at your own pace or are you dropping this?
I'm asking just so we can know whether we'll need to tackle this ourselves or not.

rodrigogiraoserrao:
This is up for grabs.
See #3060 for a pretty good PR that is almost there.

TomJGooding:
I'm not _necessarily_ volunteering, but just to clarify do you think this is better separated into `HorizontalRule` and `VerticalRule` widgets or just a single `Rule` widget with an `orientation` parameter?

willmcgugan:
I have no strong opinion on that. Surprising, coming from me. 🤷‍♂️

TomJGooding:
Ugh, apparently my fork now has snapshot tests failing on `main` even if I did want to volunteer. Any suggestions?

willmcgugan:
> How 'bout `Rule(..., orientation=...)` but then, like the `Button` variants, maybe provide `Rule.horizontal()` and `Rule.vertical()` class methods?

Sounds like the best of both worlds

willmcgugan:
[...]

PR #3060, opened 2023-08-04
Adding HorizontalRule and VerticalRule widgets

scriptogre:
_**Status**_: this PR is stale but close to being finished.

This PR introduces two new widgets, `HorizontalRule` and `VerticalRule`, into the Textual library.

I've opted for two distinct classes instead of a single `Rule` class to improve intuitiveness and readability. A base `Rule` class encapsulates common functionality, and HorizontalRule and VerticalRule extend from this.

Both widgets are defined in `_rule.py`, akin to the multi-class file `_tabbed_content.py`. To work with the widget lazy loading mechanism in `widgets/__init__.py`, I created `_horizontal_rule.py` and `_vertical_rule.py`. I've also updated `widgets/__init__.pyi` accordingly.

Despite these efforts, I'm facing import issues in `_rules.py` when trying to run `docs/examples/widgets/rules.py`. 

Any insights into the potential cause or solutions would be appreciated!

I have not forgotten about documentation and tests, but I wish to solve this issue first.

Related to #2982 

- [ ] Updated documentation


rodrigogiraoserrao:
Hey Chris, I've gone through your code (only briefly) and I think I found the issue.

If you take your PR as it stands, right now, and if you try to run your file `_rules.py` or try to import something from it, it will fail.
For example, if you run `python -m textual.widgets._rules`, it should exit silently.
Instead, it raises an error.

Another thing you can try is adding a `if __name__ == "__main__": print("hello world")` to the bottom of the file `_rules.py` and then run `python src/textual/widgets/_rules.py` and see it will fail with the exact same error.

So, the importing issue you are seeing isn't really an issue with the imports.
If you can fix the problem with the `_rules.py` file, everything should fall in place.

<details>
<summary>Spoiler for the issue</summary>

The problem is in how you defined `Rule.rule_extent` with both `@abstractmethod` and `@property`. You may want to look into `@abstractproperty` from the module `abc`.

</details>

PR #3066, opened 2023-08-07
Fix click on input with double width chars

### LINKED ISSUE #2968: Input - clicking the right half of a double width character moves the cursor to the wrong character.
こんにちは

Put that text into an `Input` and click on the right half of any character - notice that the cursor incorrectly moves to the character to the right of the one you clicked on.

yuvalmo:
Closes #2968.

- [x] Updated documentation

### DIFF
--- a/src/textual/widgets/_input.py
+++ b/src/textual/widgets/_input.py
@@ -421,10 +421,11 @@ class Input(Widget, can_focus=True):
         cell_offset = 0
         _cell_size = get_character_cell_size
         for index, char in enumerate(self.value):
-            if cell_offset >= click_x:
+            cell_width = _cell_size(char)
+            if cell_offset <= click_x < (cell_offset + cell_width):
                 self.cursor_position = index
                 break
-            cell_offset += _cell_size(char)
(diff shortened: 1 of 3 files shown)


src/textual/widgets/_input.py
```diff
@@ -421,10 +421,11 @@ async def _on_click(self, event: events.Click) -> None:
         cell_offset = 0
         _cell_size = get_character_cell_size
         for index, char in enumerate(self.value):
-            if cell_offset >= click_x:
+            cell_width = _cell_size(char)
+            if cell_offset <= click_x < (cell_offset + cell_width):
```
yuvalmo, line 425: NOTE TO REVIEWER: This line is equivalent to:
```suggestion
            if (click_x - cell_offset) < cell_width:
```
This option is shorter and does just one comparison instead of two.
However, I eventually preferred the former because it easier to understand in my opinion.

If any of the maintainers prefer the latter, please let me know.

willmcgugan (approved):
Looks good. Thanks!

PR #3070, opened 2023-08-08
Consider visible children inside invisible containers when computing focus chain

### LINKED ISSUE #3053: Visible widgets inside an invisible container are not part of the focus chain
Textual supports visible children of an invisible container -- this can be helpful with some approaches to laying things out. However, it seems that the visible children of the invisible container aren't seen as part of the focus chain; they can't be tabbed to with the keyboard.

Example code:

```python

class InvisibleContainerFocusApp(App[None]):

    CSS = """
    #invisible {
        visibility: hidden;
    }

    #invisible > * {
        visibility: visible;
    }
    """

        with Horizontal():
            with Vertical():
                yield Label("The container for this is visible")
                for _ in range(10):
                    yield Button()
            with Vertical(id="invisible"):
                yield Label("The container for this is invisible")
                for _ in range(10):
                    yield Button()

if __name__ == "__main__":
    InvisibleContainerFocusApp().run()
```

### LINKED ISSUE #3071: Change `DOMNode.visible` so it takes into account the full DOM
Following up on discussions that stemmed from #3070 and https://github.com/Textualize/textual/pull/3070#issuecomment-1669683285 in particular, we'll change `DOMNode.visible` so that it answers the question "Is the widget visible in the DOM?" instead of just comparing `styles.visibility` to `hidden`, which only checks "Is the widget not explicitly set to invisible?", which isn't a very helpful question to ask/answer to get.

This nuance exists because a node's visibility style may be in one of three states:

 1. explicitly set to `visible`
 2. explicitly set to `hidden`
 3. not set

However, the property `.visible` doesn't take into account the three states, instead only checking whether we are in case number 2. (`.visible` is `False`) or not (`.visible` is `True`).

From now on, `.visible` will check a node's explicit value and then go up the tree if its own visibility isn't set.

The setter will remain the same because the setter already (correctly) targets the visibility of the node in question.

rodrigogiraoserrao:
This fixes #3053 by not skipping invisible containers when computing the focus chain.

This fixes #3071 by going up the DOM when a node doesn't have its visibility explicitly set.

Then, in order to save some of those computations when going through the DOM to compute the focus chain, we keep a stack of visibilities that we use while computing the focus chain.


willmcgugan (commented):
I'm struggling to understand how this takes visibility in to account.

By visibility, I mean wether `visibility` is `visible` or `none`. A widget can have visibility `none` which would mean it can't be focused, but its children may have `visiblity: visible` and they should be focusable.

Call me if you want to go over this...

rodrigogiraoserrao:
`focusable_children` was the same as `DOMNode.displayed_children`, but with the extra condition that the child must be visible.

By getting rid of `focusable_children` and only going over `displayed_children`, I get all children that are displayed in the DOM.
Then, I modified `Widget.focusable` to report a widget as focusable only if it can focus, if it is not disabled, _and_ if it is visible.

This makes it so that the `while` loop inside `Screen.focus_chain` traverses through the children of invisible containers but only adds them to the focus chain if they _are_ visible.

src/textual/widget.py
```diff
@@ -1498,17 +1498,7 @@ def _self_or_ancestors_disabled(self) -> bool:
     @property
     def focusable(self) -> bool:
         """Can this widget currently be focused?"""
-        return self.can_focus and not self._self_or_ancestors_disabled
-
-    @property
-    def focusable_children(self) -> list[Widget]:
```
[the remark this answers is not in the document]
rodrigogiraoserrao: Of course. I made note of its removal in 70ff85e.

willmcgugan:
But does this take in to account that `visible` is inherited? i.e. If a widget is invisible, all its children (and ancestors) will also be invisible, unless one of them sets `visibility: visible`.

In other words a widget with `widget.visible == True` may in fact be invisible if a parent is invisible.

rodrigogiraoserrao:
> But does this take in to account that `visible` is inherited? i.e. If a widget is invisible, all its children (and ancestors) will also be invisible, unless one of them sets `visibility: visible`.
> 
> In other words a widget with `widget.visible == True` may in fact be invisible if a parent is invisible.

Oh, it doesn't.
Fixing it.

src/textual/screen.py
```diff
@@ -293,18 +294,52 @@ def focus_chain(self) -> list[Widget]:
...
+                        push_visibility(node_visibility != "hidden")
+                node_is_visible = (
+                    node_visibility != "hidden"
+                    if node_visibility
+                    else visibility_stack[-1]
+                )
+                if (  # Same as `if node.focusable`, but we cache inherited visibility.
+                    node_is_visible
+                    and node.can_focus
+                    and not node._self_or_ancestors_disabled
```
willmcgugan, line 341: The  `_self_or_ancestors_disabled` method will look up the DOM to the root, which makes it a little inefficient to call when we are recursing the entire DOM.

If a branch of the tree is disabled, then all its children will be disabled. So we could avoid traversing the remaining tree if we encounter a disabled widget.

willmcgugan (commented):
I think we can improve this algorithm somewhat.

I'm fairly sure we can do without 2 stacks. We should also be able to avoid recursing in to disable branches of the tree.

Suggest we have a chat about this tomorrow.

src/textual/screen.py
rodrigogiraoserrao: 🤦  of course!

Implemented that in ae773ed.

rodrigogiraoserrao:
Single stack implemented in f8b270d.

willmcgugan (approved):
LGTM

PR #3073, opened 2023-08-08
Digits

willmcgugan:
Adds a `Digits` widget

Screenshot 2023-08-08 at 21 17 14


docs/widgets/digits.md
```diff
@@ -0,0 +1,66 @@
+# Digits
+
+!!! tip "Added in version 0.33.0"
+
+A widget to display numerical values in tall multi-line characters.
+
+The digits 0-9 are supported, in addition to the following characters `+`, `-`, `^`, `:`, and `×`.
```
davep, line 7: I feel that `÷` (or at least `/`) should go in here too.

src/textual/renderables/digits.py
```diff
@@ -0,0 +1,119 @@
...
+
+ :
+
+""".splitlines()
+
+
+class Digits:
+    """Renders a 3X3 unicode 'font' for numerical values."""
+
+    def __init__(self, text: str, style: StyleType = "") -> None:
```
davep, line 64: Missing docstring for `Digits.__init__`.

src/textual/renderables/digits.py
```diff
@@ -0,0 +1,119 @@
...
+                row3(DIGITS3X3[position + 2].ljust(3))
+
+        new_line = Segment.line()
+        for line in digit_pieces:
+            yield Segment("".join(line), style)
+            yield new_line
+
+    @classmethod
+    def get_width(cls, text: str) -> int:
+        """Calculate the width without rendering."""
```
davep, line 105: Missing `Args` and `Returns`.

src/textual/widgets/_digits.py
```diff
@@ -0,0 +1,94 @@
...
+
+from ..geometry import Size
+from ..renderables.digits import Digits as DigitsRenderable
+from ..widget import Widget
+
+
+class Digits(Widget):
+    """A widget to display numerical values using a 3x3 grid of unicode characters.
+
+    Args:
```
davep, line 16: Correct me if I'm wrong, but I thought we'd settled on the convention that `__init__` args would be documented in the `__init__` docstring.

src/textual/widgets/_digits.py
```diff
@@ -0,0 +1,94 @@
...
+        self,
+        value: str = "",
+        *,
+        name: str | None = None,
+        id: str | None = None,
+        classes: str | None = None,
+        disabled: bool = False,
+    ) -> None:
+        if not isinstance(value, str):
+            raise TypeError("value must be a str")
```
davep, line 45: This deserves a `Raises` in the docstring.

I also wonder if it's worth actually allowing `str | float`.

src/textual/widgets/_digits.py
```diff
@@ -0,0 +1,94 @@
...
+        self.value = value
+
+    def update(self, value: str) -> None:
+        """Update the Digits with a new value.
+
+        Args:
+            value: New value to display.
+        """
+        if not isinstance(value, str):
+            raise TypeError("value must be a str")
```
davep, line 56: Needs a `Raises` in the docstring.

Also, as mentioned for `__init__`, I can see some utility in allow `str | float` for this.

src/textual/widgets/_digits.py
```diff
@@ -0,0 +1,94 @@
...
+        *,
+        name: str | None = None,
+        id: str | None = None,
+        classes: str | None = None,
+        disabled: bool = False,
+    ) -> None:
+        if not isinstance(value, str):
+            raise TypeError("value must be a str")
+        super().__init__(name=name, id=id, classes=classes, disabled=disabled)
+        self.value = value
```
davep, line 47: `self.value` should either have a docstring, or more likely (given it doesn't appear to be reactive)  prefixed with `_`.

tests/snapshot_tests/snapshot_apps/digits.py
```diff
@@ -0,0 +1,26 @@
+from textual.app import App, ComposeResult
+from textual.widgets import Digits
+
+
+class DigitApp(App):
+    CSS = """
+    #digits1 {
+    text-align: left;
```
davep, line 8:
```suggestion
        text-align: left;
```

src/textual/widgets/_digits.py
willmcgugan: I considered that, but there are more numbers than floats or ints. There's Decimal etc. We could just `str` all the things, and allow `Any`, but that didn't sit well.

By accepting `str` it doesn't limit what you can display, and it makes it clear that the dev needs to format it.

src/textual/widgets/_digits.py
willmcgugan: Changing it to a property, since setting it wouldn't do anything.

src/textual/widgets/_digits.py
willmcgugan: Did we? You may be right.

willmcgugan:
@davep comments addressed...

davep (approved):

PR #3074, opened 2023-08-08
Added: get_row_key and get_column_key from a row/column index

pslestang:
My use case is to add row to a data table with a key dynamically generated, so I need a convenient way to retrieve that key which is in my case not available in values (so not easily retrievable)

This PR add two methods in data_table widget:
* get_row_key
* get_column_key

Those methods respectively return the RowKey / ColumnKey from  the row/column identified by the index given in parameter.


pslestang:
I found that method coordinate_to_cell_key was matching my needs, so I do not need anymore this part of code. Closing the PR.

PR #3078, opened 2023-08-09
Catch exception when the markdown file cannot be found

kokotchY:

- [x] Updated documentation

Since [#2972](https://github.com/Textualize/textual/pull/2972), the example for the markdown viewer is not working anymore:
```
Unable to load PosixPath('.../demo.md')
```

In this pull request, I propose to catch the FileNotFoundException and exit the application in the same way. 
Should I update CHANGELOG.md with the fix?


willmcgugan:
Good catch! Also needed to fix the typing. https://github.com/Textualize/textual/pull/3078

Issue #3076, opened 2023-08-09
Remove docstrings from "protocol" methods


willmcgugan:
Some public methods are used internally, but never need to be called directly by the dev. If these have docstrings they appear in the API docs. I feel these are noise, and we shouldn't be showing these to the developer.

I think the docstrings on Widgets *subclasses* should be removed, for the following methods:

`get_content_width()`
`get_content_height()`
`def compose()`
`def render()`

But review what other methods may be covered by this.

willmcgugan:
@rodrigogiraoserrao Thoughts?

rodrigogiraoserrao:
> Some public methods are used internally, but never need to be called directly by the dev.

Does that mean they should/could be private, then..? 🤔 

---

For trivial one-line docstrings, I'm not _opposed_ to removing them.
But there might be cases where we use the docstrings to add more information, so I think it is preferable that we use a filter to exclude those methods from being included.

willmcgugan:
> Does that mean they should/could be private, then..? 🤔

Not really. If you subclass widget, you need to implement them.

> so I think it is preferable that we use a filter to exclude those methods from being included.

How can we filter them? Is there a mkdocs option?

rodrigogiraoserrao:
> > Does that mean they should/could be private, then..? 🤔
> 
> Not really. If you subclass widget, you need to implement them.

I don't get the causation here.
Can't something be private and still need overriding when subclassing?

> How can we filter them? Is there a mkdocs option?

[Yes](https://mkdocstrings.github.io/python/usage/configuration/members/#filters).
If you assign this to me I'll get this done.

willmcgugan:
> 
> I don't get the causation here. Can't something be private and still need overriding when subclassing?

You can in Python. Although in classic OOP, private means only for use by the one class. 

> 
> > How can we filter them? Is there a mkdocs option?
> 
> [Yes](https://mkdocstrings.github.io/python/usage/configuration/members/#filters). If you assign this to me I'll get this done.

Make it so.

rodrigogiraoserrao:
> > I don't get the causation here. Can't something be private and still need overriding when subclassing?
> 
> You can in Python. Although in classic OOP, private means only for use by the one class.

Hmmm ok, I thought the leading `_` in Python could also mean "protected", in the Java sense.

> Make it so.

👍

PR #3080, opened 2023-08-09
Hide some members from the public docs.

rodrigogiraoserrao:
See relevant issue: #3076.
Some methods need to be implemented to make the widget work but the user doesn't really care about them. For that matter, we can hide them from the public documentation.

I decided to hide these methods that some widgets inherit and document:

  - "compose"
  - "render"
  - "render_line"
  - "render_lines"
  - "get_content_width"
  - "get_content_height"
  - ~~"on_mount"~~

On a case-by-case basis, we may also want to hide specific methods from specific widgets, but that _probably_ entails manually curating such a list of excluded methods.

### DIFF
--- a/docs/api/widget.md
+++ b/docs/api/widget.md
@@ -1,1 +1,5 @@
 ::: textual.widget
+    options:
+        filters:
+          - "!^_"
+          - "^__init__$"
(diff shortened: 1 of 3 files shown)


willmcgugan:
That looks right. Although for the case of `on_mount` and other handlers, we can make those private with an underscore. Textual will still call those handlers, and don't need to worry about excluding them.

rodrigogiraoserrao:
> [...] for the case of `on_mount` [...] we can make those private with an underscore.

Of course. Updated.

willmcgugan (approved):

PR #3084, opened 2023-08-09
fix doc error in Bubble parameters for events.Focus and events.Blur

blob42:
Quick fix to reflect the bubble parameters for `Blur and Focus` in the docs. I don't remember if these events bubbled before.


willmcgugan (approved):

willmcgugan:
Thanks

PR #3115, opened 2023-08-17
Create CONTRIBUTING.md

### LINKED ISSUE #2933: Add `CONTRIBUTING.md`
Textual currently doesn't have any guidelines for contributing. 

I've been meaning to raise this for a while, but the EuroPython 2023 sprint perhaps makes this more prescient.

There's probably a lot that could just be coped from the Rich guidelines; that's where I started!

clairecharles:

Fixes #2933

- [X] Updated documentation

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,60 +4,45 @@
 
 ## 🤔 How can I contribute?
 
-### 🛠 Suggesting Enhancements
+**1.** Fix issue
 
-This section guides you through submitting an enhancement suggestion, including completely new features and minor improvements to existing functionality. Following these guidelines helps maintainers and the community understand your suggestion 📝 and find related suggestions. 🔎
+**2.** Report bug
 
-Since GitHub Issue forms we only suggest you include most information possible.
-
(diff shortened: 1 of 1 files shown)


rodrigogiraoserrao (changes requested):
Hey Claire, thanks for your PR.

While I think you nailed the welcoming tone for the document, I think the content is not what we are going for.
I think we can drop most of the generic advice, for example the style guide for commit messages or how to create a pull request.
Instead, what we can do is link to proper guides/documentation when needed.

A great alternative, that applies to pretty much everything else, is that we can point to examples of things we've done and that contributors can copy/mimic.

So, instead of focusing on these generic things that are more or less the same for every open source project, the focus of our contributing document should be on the `textual`-specific workflows.
For example,
 - how does one setup the development environment for `textual`?
 - what's the `textual`-specific checklist that we need to go through before opening a PR?
 - how does one create a snapshot test?

Some of these things have already been answered in #2923, which was linked from the issue #2933 that you are fixing.

So, please take your time to go through #2923 and we're looking forward to your revised PR.
Thanks!

willmcgugan (approved):
Great work. Thanks!

PR #3118, opened 2023-08-18
Tweak the demo

davep:
Hits a couple of issues noticed in the demo.


darrenburns:
I could be wrong but wasn't the Text.assemble used because *theoretically* console markup could appear in a path and mess up the output?

davep:
@darrenburns Good point, so really what should be happening is the `Path` should be escaped?

Issue #3068, opened 2023-08-08
Move the FAQ in to the docs


willmcgugan:
I think we should configure Faqtory to generate FAQ.md in the docs. This will make the FAQ easier to find, and look prettier.

We would need to consider where to put it in the nav. Maybe a FAQ entry in the header bar.

PR #3129, opened 2023-08-21
Move the Textual FAQ into the main docs

davep:
Implements #3068.


rodrigogiraoserrao (commented):
Don't we need to add `faqtory` to the (dev) dependencies of `Textual`, then?

davep:
Good call; I had thought it was.

willmcgugan (approved):
Nice

davep:
I just realised something: rather than have the two different link IDs floating around, I dropped faqtory's slugs from the generated FAQ and leaned into mkdocs-materials' IDs generated from the headers; but now I think about it faqtory and mkdocs-material might not slug the same.

I'll make an issue to double-check this tomorrow and I might have to add the faqtory slugs back and we'll just accept the duplication of effort in the page.

willmcgugan:
Good point!

willmcgugan:
I'm open to modifying the slug generation in FAQtory to match.

PR #3132, opened 2023-08-21
feat(loadingindicator): default min height

aaronst:
From https://github.com/Textualize/textual/discussions/3131, add `min-height: 1;` to `LoadingIndicator`'s `DEFAULT_CSS`.


willmcgugan:
Thanks

PR #3140, opened 2023-08-22
Rename CSS files to TCSS.

### LINKED ISSUE #3137: Rename CSS to TCSS
Now that we have an awesome Textual CSS syntax highlighter, we should change the extension in the code and docs.

All *.css files should be renamed to *.tcss, and code updated accordingly.

There should also be a brief explanation in the guide regarding the naming, which also mentions the vscode extension.

rodrigogiraoserrao:
Fixes #3137.


davep (approved):
Unsurprisingly there's really too much here to check line-by-line; but given tests look fine and nothing stands out obviously in the docs, I'm going with the "more eyes will find faults" approach on this one.

PR #3168, opened 2023-08-24
Fix broken f-string

rodrigogiraoserrao:
Noticed this en passant


davep:
Is token.name typed something other than str?

davep (approved):

rodrigogiraoserrao:
> Is token.name typed something other than str?

No idea, why?

PR #3169, opened 2023-08-24
Let layers be strings.

rodrigogiraoserrao:
Motivated by syntax highlighting of TCSS, I'd like us to consider allowing the values of `layer` and `layers` be strings.
I.e., allow CSS like this:

```sass
Screen {
    layers: "layer-name";
}

#widget {
    layer: "layer-name";
}
```

Nothing needs to be done for `layer` because it doesn't check its token type.
For `layers`, the change is trivial and can be seen in this PR.


src/textual/css/_styles_builder.py
```diff
@@ -722,7 +722,7 @@ def process_layer(self, name: str, tokens: list[Token]) -> None:
     def process_layers(self, name: str, tokens: list[Token]) -> None:
         layers: list[str] = []
         for token in tokens:
-            if token.name != "token":
+            if token.name not in ["token", "string"]:
```
willmcgugan, line 725: You could save a few microseconds with a set literal.

willmcgugan:
Seems reasonable. Could you add a changelog entry?

rodrigogiraoserrao:
> Seems reasonable.

Just to confirm: seems reasonable that we accept strings and push it as the "standard" or just that we have this possibility?

willmcgugan:
I think strings are only necessary if you want additional characters or spaces in your layer names. I wouldn't say it should be the standard exactly, just an alternative.

willmcgugan (approved):

PR #3183, opened 2023-08-26
feat(datatable): add cursor_type to constructor

aaronst:
I have a use-case where I need a table with `cursor_type="row"`. I'd like to be able to set this in the constructor like `yield DataTable(cursor_type="row")`.


darrenburns (approved):

willmcgugan:
Thanks @aaronst

PR #3193, opened 2023-08-28
Customisable input validation (& validation on blur events)

### LINKED ISSUE #3100: Validation on blur
Currently validation is done on input change and submit. I think we also need to do it on blur.

Consider a password check. I find it super irritating when validation shows an error before I'm done typing.

It should probably also be configrable. So the dev can decide when the validation happens.

rodrigogiraoserrao:
Fixes #3100.


src/textual/widgets/_input.py
```diff
@@ -221,6 +221,7 @@ def __init__(
         *,
         suggester: Suggester | None = None,
         validators: Validator | Iterable[Validator] | None = None,
+        prevent_validation_on: Iterable[type[Message]] | None = None,
```
willmcgugan, line 224: This parameter is negative condition. The dev is saying don't do something on this. This creates a bit of cognitive load where the developer has to reverse what they want in their mind.

Suggest we drop the `prevent_` and `have it simply `validate_on`.

Requiring an iterable of messages could also be confounding. As the typing would accept any Message.

I think this should be an iterable of literals. Something like this:

`validate_on: Iterable[ValidateCondition] | None`

Where 

`ValidateContition = Literal["changed", "submitted", "blur"]`

src/textual/widgets/_input.py
```diff
@@ -254,7 +257,26 @@ def __init__(
         elif validators is None:
             self.validators = []
         else:
-            self.validators = list(validators) or []
+            self.validators = list(validators)
+        self.prevent_validation_on: set[type[Message]] = set(
```
willmcgugan, line 261: A set literal expression may work better here.

```python
{*(prevent_validation_on or []), self.Changed, self.Submitted, Blur}
```

willmcgugan (changes requested):
A request for an alternative API.

src/textual/widgets/_input.py
rodrigogiraoserrao: Re `Literal`: consider it done.

I disagree with the negative condition & cognitive load argument.
One negation is something reasonable.
Since the default is “everything”, it is sensible to have a parameter that allows ignoring just some, just like with the context manager `prevent`.

willmcgugan: There is only three options. `validate_on = {"blur", "changed"}` tells me everything I need to know. `prevent_validation_on = {"changed"}` leaves me wondering what the default is.

rodrigogiraoserrao: See a26307278

docs/widgets/input.md
```diff
@@ -26,7 +26,8 @@ The example below shows how you might create a simple form using two `Input` wid
 
 You can supply one or more *[validators][textual.validation.Validator]* to the `Input` widget to validate the value.
 
-When the value changes or the `Input` is submitted, all the supplied validators will run.
+All the supplied validators will run when the value changes, the `Input` is submitted, or focus moves _out_ of the `Input`.
+This can be customized via the attribute [`prevent_validation_on`][textual.widgets.Input.prevent_validation_on].
```
willmcgugan, line 30: We need to inform the reader what the three possible values are, and what they mean. Also, this is referring to the old attribute name.

src/textual/widgets/_input.py
```diff
@@ -236,6 +241,8 @@ def __init__(
             suggester: [`Suggester`][textual.suggester.Suggester] associated with this
                 input instance.
             validators: An iterable of validators that the Input value will be checked against.
+            validate_on: When does input validation happen? The default is to validate
```
willmcgugan, line 244: A rhetorical question doesn't make a good docstring. Suggest something like "zero or more of the following values... which define when to perform validation"

willmcgugan (changes requested):

willmcgugan (approved):

Issue #3195, opened 2023-08-28
Screens should have a title and subtitle


MartijnBraam:
When implementing a multi-screen application that has Header widgets in the screens the title and subtitle displayed in the Header should be stored in the Screen as overrides instead of only using the global values from the application.

This would simplify displaying titles in screens properly with keeping the title and subtitle synchronized with the screen stack.

willmcgugan:
Should be easy to implement. Screen would have `title` and `subtitle`, which would override the App attributes if not None. We would need to manage titles when the active screen changes.

PR #3199, opened 2023-08-29
Add title and sub-title to screens.

rodrigogiraoserrao:
Mimicking 'App', we provide class variables TITLE and SUB_TITLE for the screen defaults and those can then be changed via the title and sub_title reactive attributes.

Related issue: #3195

### DIFF
--- a/src/textual/widgets/_header.py
+++ b/src/textual/widgets/_header.py
@@ -161,11 +161,19 @@ class Header(Widget):
         self.toggle_class("-tall")
 
     def _on_mount(self, _: Mount) -> None:
-        def set_title(title: str) -> None:
+        def set_title() -> None:
+            screen_title = self.screen.title
+            title = screen_title if screen_title is not None else self.app.title
             self.query_one(HeaderTitle).text = title
 
         def set_sub_title(sub_title: str) -> None:
+            screen_sub_title = self.screen.sub_title
+            sub_title = (
+                screen_sub_title if screen_sub_title is not None else self.app.sub_title
+            )
             self.query_one(HeaderTitle).sub_text = sub_title
 
         self.watch(self.app, "title", set_title)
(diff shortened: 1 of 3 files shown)


src/textual/widgets/_header.py
```diff
@@ -161,11 +161,19 @@ def _on_click(self):
         self.toggle_class("-tall")
 
     def _on_mount(self, _: Mount) -> None:
-        def set_title(title: str) -> None:
+        def set_title() -> None:
+            screen_title = self.screen.title
```
willmcgugan, line 165: We should have a more elegant way of exposing the actual title. So that a developer doesn't need to replicate this logic.

Perhaps a computed value called `screen_title` and `screen_subtitle`

willmcgugan (changes requested):

src/textual/widgets/_header.py
rodrigogiraoserrao: See 4ed93d45c

src/textual/widgets/_header.py
```diff
@@ -160,12 +160,28 @@ def watch_tall(self, tall: bool) -> None:
     def _on_click(self):
         self.toggle_class("-tall")
 
+    @property
+    def screen_title(self) -> str:
```
willmcgugan, line 164: Docstrings pls.

Could we make these computed reactives? That way you could watch `self.screen_title`

willmcgugan (changes requested):

src/textual/widgets/_header.py
rodrigogiraoserrao: See 5a15e9c8a for docstrings.
As per our conversation, we can't make them computed reactives because we can't trigger the compute method when the reactives on `Screen` or `App` change.

willmcgugan (approved):

PR #3202, opened 2023-08-29
Add app return codes.

### LINKED ISSUE #3189: Revise error codes
We need a way for a Textual app to generate a return code that would be used as the process return code.

- `App.return_code` should be 0 by default
- `App.exit` should accept an optional return code which will set `App.return_code`
- A fatal error should set a return code of 1. 

Textual apps should _not_ exit the process by default, but you could do something like this:

```python
my_app.run()
sys.exit(my_app.return_code)
```

rodrigogiraoserrao:
Fixes #3189.

- [x] Updated documentation

### DIFF
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -483,6 +483,8 @@ class App(Generic[ReturnType], DOMNode):
         """Internal attribute used to set the return value for the app."""
         self._return_code: int | None = None
         """Internal attribute used to set the return code for the app."""
+        self._started_running: bool = False
+        """Whether the app has ever started running or not."""
         self._exit = False
         self._disable_tooltips = False
         self._disable_notifications = False
@@ -533,14 +535,12 @@ class App(Generic[ReturnType], DOMNode):
         return self._return_value
 
     @property
-    def return_code(self) -> int:
+    def return_code(self) -> int | None:
         """The return code with which the app exited.
 
-        Non-zero codes are for errors.
+        Non-zero codes indicate errors.
         A value of 1 means the app exited with a fatal error.
-
-        Accessing this attribute before the app runs or while the app is running
-        is meaningless.
+        If the app hasn't run yet or if it is running, this will be `None`.
 
         Example:
             The return code can be used to exit the process via `sys.exit`.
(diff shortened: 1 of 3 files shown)


src/textual/app.py
```diff
@@ -529,6 +532,25 @@ def return_value(self) -> ReturnType | None:
         """
         return self._return_value
 
+    @property
+    def return_code(self) -> int:
```
willmcgugan, line 536: The `return_code` property will be 1 when no error has occurred. That would be astonishing if you inspect it before the app exits.

I know you said it was "meaningless" but I don't think we should have meaningless properties... 

I would expect either the return code to be `0` or `None` if it hasn't been set.

src/textual/app.py
```diff
@@ -649,16 +671,22 @@ def _screen_stack(self) -> list[Screen]:
...
 
         Args:
             result: Return value.
+            return_code: The return code. Use non-zero values for error codes.
             message: Optional message to display on exit.
         """
         self._exit = True
         self._return_value = result
+        self._return_code = return_code
+        print(f"Exiting; return code is {self._return_code}")
```
willmcgugan, line 689: Debug code

willmcgugan (changes requested):

src/textual/app.py
rodrigogiraoserrao: Fair point.
As of da0cd5e, the return code is `None` before/while running the app.
(So as to avoid confusion from a successful exit with return code `0`.)

src/textual/app.py
```diff
@@ -529,6 +534,25 @@ def return_value(self) -> ReturnType | None:
...
+        If the app hasn't run yet or if it is running, this will be `None`.
+
+        Example:
+            The return code can be used to exit the process via `sys.exit`.
+            ```py
+            my_app.run()
+            sys.exit(my_app.return_code)
+            ```
+        """
+        if not self._started_running or self.is_running:
```
willmcgugan, line 552: So `app.return_code` will be None, immediately after calling `exit(return_code=10)` ?

I think this is simpler than you imagine. Suggest you set the return code to `1` in `App._handle_exception`. All errors will go through there.

rodrigogiraoserrao: Done in ecd7c93.

willmcgugan (approved):

Issue #2487, opened 2023-05-04
Create base `TreeMessage` for tree messages?


rodrigogiraoserrao:
`OptionList` has a `OptionMessage` that is inherited by the two `OptionList` messages.
Is it worth doing the same for the tree messages? There are 4 messages whose code is exactly the same, the only thing that changes is the name of the class.

rodrigogiraoserrao:
> IIRC I made that change once but it got reversed. Might be worth looking back through history.

You are right, and I was the one who removed the base class 😆  https://github.com/Textualize/textual/commit/f34685b18fac829cf562513ec98797ccdae878aa

> I think the objection was to do with impact on documentation, so perhaps we're in a better place now.

I see the point. It isn't _super_ obvious what attributes the `OptionList` messages have.
Maybe if we add a sentence in the docstrings for those messages to check the base class, then we can inherit from those messages and we can, once more, have a `TreeMessage` base class.

Thoughts, @willmcgugan?

![Untitled](https://user-images.githubusercontent.com/5621605/236292880-2fda7ba2-7440-4d9f-a9cc-043509fe9f78.png)

willmcgugan:
I think with @davep 's change to the `@on` decorator it makes sense to have common base classes.

The docs issue is annoying.

willmcgugan:
Exactly. The indirection means that some people will miss the attributes in the base class.

willmcgugan:
It probably does make some sense to have a common base class. But it doesn't seem all that useful: I can't see a reason why you would want to respond to multiple tree events in a single handler.

Closing this, because it just doesn't seem worth changing, unless we have an identiable need for it.

PR #3203, opened 2023-08-29
feat(directory tree): add directory selected message

### LINKED ISSUE #3200: DirectoryTree - message when directory was selected
It doesn't seem possible to hook up an action when someone selects a directory using a `DirectoryTree` widget - let's say I want to connect it to some `Input` and then update that input when something is selected in the `DirectoryTree` widget - e.g. user selects directory where to save some file (then append some default filename to that directory).

For a file it's fine because there is a `FileSelected` message posted, but no such message of `DirectorySelected` is available. Also watching `DirectoryTree.path` doesn't seem to solve the problem as this path is not updated when a directory is selected.

TomJGooding:
### Description

Closes #3200 by adding a `DirectorySelected` message to the `DirectoryTree` widget.

### Related Issue

- #3200 

### How Has This Been Tested?

Sorry but I wasn't sure where to add tests - I couldn't find any existing test suite for `DirectoryTree`?

### Checklist:

- [ ] Updated documentation


davep (approved):
Looks good to me. Ideally I'd have made a common base class for the messages at this point (nice for the `@on` decorator if nothing else) but this is an issue that's a bit more widespread than `DirectoryTree` (see #2487).

It's also made me realise that `bubble` is `True` by default but many messages explicitly set that too. Note to self: perhaps we should tidy that up a bit?

rodrigogiraoserrao:
> Sorry but I wasn't sure where to add tests - I couldn't find any existing test suite for `DirectoryTree`?

There may be a couple of random `DirectoryTree` tests scattered throughout the `tests` directory, but most of the functionality for the `DirectoryTree` is actually inherited from `Tree` and thus most of the testing happens in `tests/tree`.
There is a `tests/tree/test_directory_tree_reload_node.py` file there; feel free to rename to `test_directory_tree.py` and add your tests there, or create a separate file to test this feature you implemented.

rodrigogiraoserrao (changes requested):
Other than the lack of tests (and Dave's comment), looks good to me.

davep:
Yup, AFAIK there are no tests for `DirectoryTree` messages (at least of this nature).

TomJGooding:
Thanks all for the feedback. I've now removed the superfluous `bubble=True`, but it sounds like a base message class might create issues in the docs.

I confess I am struggling to add tests for this. I've tried to combine the logic of the existing [test_tree_messages.py](https://github.com/Textualize/textual/blob/main/tests/tree/test_tree_messages.py) and [test_directory_tree_reload_node.py](https://github.com/Textualize/textual/blob/main/tests/tree/test_directory_tree_reload_node.py) tests,  but I must be doing something wrong as even locally tests are _very_ flaky.  

Sorry to be a pain, but could you give some guidance even just for testing just the existing `FileSelected` message?

davep:
> I've now removed the superfluous bubble=True, but it sounds like a base message class might create issues in the docs.

That's fine; it wasn't a suggestion for a change to actually make here, more a pondering for now-to-future changes (I'm also generally of the opinion that problems with documentation-generation shouldn't trump utility in the code and I do think that the win for `@on` beats a failing in the documentation generation; but that's all really for the issue I linked and issues related to that).

rodrigogiraoserrao:
> I confess I am struggling to add tests for this. I've tried to combine the logic of the existing [test_tree_messages.py](https://github.com/Textualize/textual/blob/main/tests/tree/test_tree_messages.py) and [test_directory_tree_reload_node.py](https://github.com/Textualize/textual/blob/main/tests/tree/test_directory_tree_reload_node.py) tests, but I must be doing something wrong as even locally tests are _very_ flaky.

Are you using `await pilot.pause()` or `await pilot.pause(ms)` in your tests?
I'm assuming you are doing something like clicking a row of the directory tree and then checking if a message was posted.
You may need to `await pilot.pause()` to make sure the app has time to process the click and the message that gets posted.

davep:
On the testing front, something along these lines should work:

```python

from pathlib import Path

from textual import on
from textual.widgets import DirectoryTree

class DirectoryTreeApp(App[None]):

    def __init__(self, test_path: Path) -> None:
        super().__init__()
        self.test_path = test_path
        (test_path / "0").write_text("")
        self.messages: list[str] = []


        yield DirectoryTree(self.test_path)

    @on(DirectoryTree.FileSelected)
    @on(DirectoryTree.DirectorySelected)
    def record(self, event: DirectoryTree.FileSelected | DirectoryTree.DirectorySelected) -> None:
        self.messages.append(event.__class__.__name__)


async def test_directory_tree_selected_messages(tmp_path) -> None:
    """The directory tree should generate select messages."""

    async with DirectoryTreeApp(tmp_path).run_test() as pilot:
        tree = pilot.app.query_one(DirectoryTree)
        # ... pilot.press and asserts on pilot.app.messages carry on from here.
```

and while I was testing this out I was getting unexpected results. I *think* this change might be unintentionally emitting more `DirectorySelected` messages than would be expected.

Take this bit of code for example:

```python

from pathlib import Path

from textual import on
from textual.widgets import DirectoryTree, Log

class DirTreeTest(App[None]):

        with Horizontal():
            yield DirectoryTree(Path(".."))
            yield Log()

    @on(DirectoryTree.DirectorySelected)
    @on(DirectoryTree.FileSelected)
    def log_event(self, event: DirectoryTree.DirectorySelected | DirectoryTree.FileSelected) -> None:
        self.query_one(Log).write_line(event.__class__.__name__)

if __name__ == "__main__":
    DirTreeTest().run()
```

When I first run it up I get an initial `DirectorySelected` message:

Screenshot 2023-08-31 at 11 02 45

that makes sense. However, if I cursor down once, then hit <kbd>Enter</kbd> to select that directory, I end up with *3* more `DirectorySelected` events:

Screenshot 2023-08-31 at 11 03 29

I think most people would find that surprising. We should probably figure out why that is.

davep (changes requested):
See https://github.com/Textualize/textual/pull/3203#issuecomment-1700746199 above -- we should improve that I think.

TomJGooding:
Thanks both for your help. Sorry, I've realised now I also needed to `await pilot.pause()` _before_ trying to navigate the tree to give it chance to load!

> I think this change might be unintentionally emitting more DirectorySelected messages than would be expected.

Yes sorry I realise that now too! Thanks Dave.

TomJGooding:
Sorry for all the questions, I really thought this would be a simple thing for me to add but just I seem to be causing you more hassle! :(

I obviously need to remove one of the `post_message(self.FileSelected` calls, rather than just copying the `FileSelected` messages. The problem is that if I remove it from  `_on_tree_node_selected`, if you select the directory to expand it and then select it again, it won't post a second message. But if I remove it from `_on_tree_node_expanded`, then it won't post for the root node, which Dave said makes sense.

rodrigogiraoserrao (approved):
This looks good to me.

rodrigogiraoserrao:
@davep would you mind also having a look at this PR, just so there's two sets of eyes approving this?

davep (approved):
Looks good to me also.

PR #3209, opened 2023-08-30
feat: add rule widget

### LINKED ISSUE #2982: Add `Rule` widget

TomJGooding:
### Description

Draft PR to add a `Rule` widget - closes #2982. This takes inspiration from the good work already done in #3060.

As discussed in the linked issue, this adds a single `Rule` widget rather than separate widgets for horizontal and vertical orientations.

@rodrigogiraoserrao This is still a work in progress, but I just wanted to double-check I was on the right track with this? ~~Especially with the different `line_styles` for the rule where I'm directly using the existing border styles?~~

### Related Issue

- #2982 

### How Has This Been Tested?

~~TODO!~~ Snapshot tests added for the horizontal and vertical rule examples. Also added tests that an invalid rule orientation or line style will raise an exception.

#### Horizontal rules:

image

#### Vertical rules:

image

### Checklist:

- [x] Updated documentation


src/textual/widgets/_rule.py
```diff
@@ -0,0 +1,149 @@
...
+    """The orientation of the rule."""
+
+    line_style: Reactive[LineStyle] = reactive[LineStyle]("solid")
+    """The line style of the rule."""
+
+    def __init__(
+        self,
+        orientation: RuleOrientation = "horizontal",
+        line_style: LineStyle = "solid",
+        name: str | None = None,
```
rodrigogiraoserrao, line 63:
```suggestion
        *,
        name: str | None = None,
```

src/textual/widgets/_rule.py
```diff
@@ -0,0 +1,149 @@
...
+            orientation: The orientation of the rule.
+            line_style: The line style of the rule.
+            name: The name of the widget.
+            id: The ID of the widget in the DOM.
+            classes: The CSS classes of the widget.
+            disabled: Whether the widget is disabled or not.
+        """
+        super().__init__(name=name, id=id, classes=classes, disabled=disabled)
+        self.orientation = self.validate_orientation(orientation)
+        self.line_style = self.validate_line_style(line_style)
```
rodrigogiraoserrao, line 80: I believe you don't need to call the validators explicitly here.

(Come to think of it, this may be done elsewhere in some other widgets, but I think that's just a vestige of some old reactive attributes limitation.)

src/textual/widgets/_rule.py
```diff
@@ -0,0 +1,149 @@
...
+        if self.orientation == "vertical":
+            rule_char = BORDER_CHARS[self.line_style][1][0]  # middle-left border char
+            return Text(rule_char * self.size.height)
+        elif self.orientation == "horizontal":
+            rule_char = BORDER_CHARS[self.line_style][0][1]  # top-middle border char
+            return Text(rule_char * self.size.width)
+        else:
+            raise InvalidRuleOrientation(
+                f"Valid rule orientations are {friendly_list(_VALID_RULE_ORIENTATIONS)}"
+            )
```
rodrigogiraoserrao, line 93: Given the validation, this could probably be omitted, no? Or are you playing it extra-safe?

rodrigogiraoserrao (commented):
This is looking great to me. I didn't think of using the border edge styles as styles for the rule but that makes a lot of sense to me.

Feels like you are pretty much on the right track, yes.

TomJGooding:
> I didn't think of using the border edge styles as styles for the rule but that makes a lot of sense to me.

Sorry, I've now changed my mind on this! I realised many border styles don't actually make sense in this context, so I've reduced the rule styles accordingly. See updated screenshots above.

TomJGooding:
Sorry @rodrigogiraoserrao a quick question: do I need to make the non-widget `Rule` classes available, i.e. RuleOrientation, LineStyle, and custom exceptions?

src/textual/widgets/_rule.py
rodrigogiraoserrao: Sure thing.

src/textual/widgets/_rule.py
rodrigogiraoserrao: It _shouldn't_ be necessary.

docs/widgets/rule.md
```diff
@@ -68,3 +68,8 @@ This widget sends no messages.
 ::: textual.widgets.Rule
     options:
       heading_level: 2
+
+::: textual.widgets.rule
+    options:
+      show_root_heading: true
+      show_root_toc_entry: true
```
rodrigogiraoserrao, line 75: Doesn't this make it so that the widget itself shows up twice? (I'm not sure, I'm doing some guesswork here.)

willmcgugan (approved):
LGTM. Thanks, Tom!

docs/widgets/rule.md
rodrigogiraoserrao: Ahh of course, you added the `rule.py` with the reimports. Thanks a ton for the screenshot :)

TomJGooding:
Haha, looks like I just mistimed merging main so I could update the changelog! Thanks Will.

Issue #3206, opened 2023-08-30
Flicker on tree control


willmcgugan:
When scrolling the tree control there is a little flicker where the selection doesn't update instantly.

This was working in the past. The best approach may be to `git bisect` this to find what commit introduces this flicker.

Here's `code_browser.py` to demonstrate the issue.

https://github.com/Textualize/textual/assets/554369/3efff249-730a-4dbd-ba2c-6d5190e0d33a

PR #3210, opened 2023-08-30
Fixes flicker on tree scroll

willmcgugan:
I think at some point we had a mechanism that disabled animation if you were scrolling a single line. That solved the flicker on the tree. I don't recall why we removed that, or even if it was intentional. This update solves it by explicitly disabling animation when moving the cursor up or down.

Fixes https://github.com/Textualize/textual/issues/3206


darrenburns (approved):
Wasn't expecting it to be this to be honest 😄

davep:
For reference, #2358 seems to be the last fix.

PR #3213, opened 2023-08-31
DataTable new rows can have auto height.

### LINKED ISSUE #3122: Auto expanding height
You can set a height when you add a row to the DataTable, but there is currently no way to make it expand to fit the content.

I think that `add_row` should grow an option to automatically calculate the optimal height of a row.

If we make this the default, we should be able to explicitly override it with a height. Perhaps a `height` of None means auto-detect, and a integer sets it explicitly. An explicit height will be faster, which might be significant if the user adds a lot of rows.

rodrigogiraoserrao:
Fixes #3122.

Adds `auto_height` to `Row` to try and keep `Row` and `Column` as similar as possible, as per a conversation with @darrenburns.
Temporarily sets the row height to 0 to signal the row height hasn't been computed yet (we can't set to `None` because we really need an `int` there for computations that run in the interim).

### DIFF
--- a/src/textual/widgets/_data_table.py
+++ b/src/textual/widgets/_data_table.py
@@ -1200,7 +1200,7 @@ class DataTable(ScrollView, Generic[CellType], can_focus=True):
             new_rows: The new rows that will affect the `DataTable` dimensions.
         """
         console = self.app.console
-        auto_height_rows: list[tuple[Row, list[RenderableType]]] = []
+        auto_height_rows: list[tuple[int, Row, list[RenderableType]]] = []
         for row_key in new_rows:
             row_index = self._row_locations.get(row_key)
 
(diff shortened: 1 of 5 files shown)


darrenburns (changes requested):
Could you add a snapshot test for this since it's quite a visual change? The tests at the moment just check that the `height` attribute is set correctly on the `Row`, but that says nothing about what is actually rendered and what the user actually sees.

src/textual/widgets/_data_table.py
```diff
@@ -1215,6 +1225,22 @@ def _update_dimensions(self, new_rows: Iterable[RowKey]) -> None:
                 content_width = measure(console, renderable, 1)
                 column.content_width = max(column.content_width, content_width)
 
+            if row.auto_height:
+                auto_height_rows.append((row, cells_in_row))
+
+        for row, cells_in_row in auto_height_rows:
```
willmcgugan, line 1231: Since we will need to render the cells anyway, can we get the height without rendering and throwing away the result ?

rodrigogiraoserrao: I.e., can we store the renderings?

rodrigogiraoserrao: I'm not sure how feasible it is.
I currently render the cell in a quick & dirty way just to count lines.

I looked into modifying `DataTable._render_cell`, which is responsible for rendering a cell and caching its result.
However, `_render_cell` needs a lot of information that I don't have computed, of which the base style is the more troublesome to compute.
`DataTable._render_line_in_row` spends ~90 lines on that, and I am not sure whether it's faster to do this simpler throwaway rendering here or if we should go over all that trouble at this point.

willmcgugan: I think its a clear win, and worth perservering.

rodrigogiraoserrao:
> Could you add a snapshot test for this since it's quite a visual change?

Good call. See 65aaf39.

src/textual/widgets/_data_table.py
rodrigogiraoserrao: See c9ae50d.

darrenburns:
Have you tested this with sorting and checked what happens as rows move around - does it redraw correctly with the correct heights etc?

src/textual/widgets/_data_table.py
```diff
@@ -1767,58 +1845,113 @@ def _render_cell(
...
+            is_header_cell: Is this a cell from a header?
+            is_row_label_cell: Is this the label of any given row?
+            is_fixed_style_cell: Should this cell be styled like a fixed cell?
+            hover: Does this cell have the hover pseudo class?
+            cursor: Is this cell covered by the cursor?
+            show_cursor: Do we want to show the cursor in the data table?
+            show_hover_cursor: Do we want to show the mouse hover when using the keyboard
+                to move the cursor?
+            has_css_foreground_priority: `self.cursor_foreground_priority == "css"`?
+            has_css_background_priority: `self.cursor_background_priority == "css"`?
```
rodrigogiraoserrao, line 1921: Yesterday we talked about rhetorical questions as docstrings, but given that these are just a bunch of Boolean values, this actually looks reasonable to me.

rodrigogiraoserrao:
@darrenburns @willmcgugan after much sweat and tears, I'm ready for your reviews. 🙃

src/textual/widgets/_data_table.py
```diff
@@ -1767,58 +1845,113 @@ def _render_cell(
...
-                self.app.console.options.update_dimensions(width, height),
+                options,
             )
 
             self._cell_render_cache[cell_cache_key] = lines
 
         return self._cell_render_cache[cell_cache_key]
 
+    @functools.lru_cache(
+        maxsize=512
```
willmcgugan, line 1895: This is broken I'm afraid. @lru_cache on a method will give you a *single* cache for the process. You are using `get_component` which may return different results depending on the instance. Consequently if you have 2 or more DataTables, the styles are going to get mixed.

If you need caching (suspect it is a win), have a look at `cache.py` which has some home grown cache objects you can use per-instance.

Also consider reducing the maxsize. I'm guessing some combos are used more than others.

willmcgugan (changes requested):
Tested in a sandbox, and it works nicely. However there may be an issue re caching.

src/textual/widgets/_data_table.py
rodrigogiraoserrao: While we concluded that this is not broken in the sense that you described because `lru_cache` also caches the instance, there are issues waiting to happen with hot-reloading, dark-mode toggling, etc.
Those are mostly addressed in #3313 and 2524af4.

I'd like for #3313 to be merged into main so I can merge it here to make sure this was all I needed and to add tests for it.

willmcgugan (approved):

Issue #3217, opened 2023-09-01
Windows terminal wrong size when pushing screen


dmunozv04:
In windows terminal if you push a screen after resizing the terminal window, the screen will use the old screen size, leaving parts of the screen empty. It seems to fix itself the moment you resize the terminal when inside the screen though. Here's a small video showing the issue, together with code to replicate it.
Video: https://github.com/Textualize/textual/assets/39565245/22efb96b-cc13-46b5-9c81-3c4004b3c7ac
Code: 
```Python
from textual import on
from textual.app import App, ComposeResult
from textual.screen import Screen, ModalScreen
from textual.widgets import Label, Button


class BugScreen(ModalScreen):
    CSS = """
    BugScreen {
    border: solid red;
    align: center middle;
    }
    
    """

    def compose(self) -> ComposeResult:
        yield Label("This is a screen")


class BugApp(App):
    def compose(self) -> ComposeResult:
        yield Button("Press me to open screen", name="button")

    @on(Button.Pressed)
    def open_screen(self, event: Button.Pressed):
        self.app.push_screen(BugScreen())


if __name__ == "__main__":
    app = BugApp()
    app.run()

```

# Textual Diagnostics

## Versions

| Name    | Value  |
|---------|--------|
| Textual | 0.35.1 |
| Rich    | 13.4.2 |

## Python

| Name           | Value                                                               |
|----------------|---------------------------------------------------------------------|
| Version        | 3.11.3                                                              |
| Implementation | CPython                                                             |
| Compiler       | MSC v.1934 64 bit (AMD64)                                           |
| Executable     | C:\Users\someone\somewhere\venv\Scripts\python.exe |

## Operating System

| Name    | Value      |
|---------|------------|
| System  | Windows    |
| Release | 10         |
| Version | 10.0.22621 |

## Terminal

| Name                 | Value            |
|----------------------|------------------|
| Terminal Application | Windows Terminal |
| TERM                 | *Not set*        |
| COLORTERM            | *Not set*        |
| FORCE_COLOR          | *Not set*        |
| NO_COLOR             | *Not set*        |

## Rich Console options

| Name           | Value                |
|----------------|----------------------|
| size           | width=188, height=49 |
| legacy_windows | False                |
| min_width      | 1                    |
| max_width      | 188                  |
| is_terminal    | True                 |
| encoding       | utf-8                |
| max_height     | 49                   |
| justify        | None                 |
| overflow       | None                 |
| no_wrap        | False                |
[...]

PR #3218, opened 2023-09-01
fix updating CSS on push_screen

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/3217
Fixes https://github.com/Textualize/textual/issues/3221

We recently added `CSS` to Screens, which requires a `stylesheet.update` when pushing the screen.

### DIFF
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -2052,6 +2052,7 @@ class App(Generic[ReturnType], DOMNode):
                 try:
                     try:
                         await self._dispatch_message(events.Compose())
+                        default_screen = self.screen
                         await self._dispatch_message(events.Mount())
                         self.check_idle()
                     finally:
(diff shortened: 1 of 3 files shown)


dmunozv04:
@willmcgugan This fix seems to not work correctly, since the bug also appears without using CSS.
image

src/textual/app.py
```diff
@@ -2059,7 +2061,8 @@ async def invoke_ready_callback() -> None:
                     Reactive._initialize_object(self)
 
                     self.stylesheet.update(self)
-                    self.refresh()
+                    if self.screen is not default_screen:
+                        self.stylesheet.update(default_screen)
```
rodrigogiraoserrao, line 2065: What can happen in between the time where you set `default_screen = self.screen` and this `if` statement that `self.screen` is no longer what you stored inside `default_screen`?

willmcgugan: The issue was that if you change the screen stack in App.on_mount then the default screen never has its CSS applied.

rodrigogiraoserrao: Ahhhh, I see. Thanks.

rodrigogiraoserrao (approved):

PR #3225, opened 2023-09-03
chore: remove superfluous bubble=true from messages

TomJGooding:
@davep [recently commented](https://github.com/Textualize/textual/pull/3203#pullrequestreview-1602076289) that:

> It's also made me realise that `bubble` is `True` by default but many messages explicitly set that too. Note to self: perhaps we should tidy that up a bit?

- [ ] Updated documentation


willmcgugan:
Thanks.

willmcgugan (approved):

PR #3232, opened 2023-09-04
Add see-also entries to the various blur/focus messages

davep:
This question crops up from time to time, often with people looking for "focus" or "blur" and wondering why they don't bubble and so then wondering how they can catch such events in ancestors in the DOM.

The Descendant prefix isn't obvious (I always forget what it is and need to go hunting), so having the focus and blur events link to the descendant events should help make them easier to discover.

### DIFF
--- a/docs/events/focus.md
+++ b/docs/events/focus.md
@@ -16,5 +16,5 @@ _No other attributes_
 ## See also
 
 - [Blur](blur.md)
-- [DescendantFocus](descendant_focus.md)
 - [DescendantBlur](descendant_blur.md)
+- [DescendantFocus](descendant_focus.md)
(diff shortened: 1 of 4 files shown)


docs/events/focus.md
```diff
@@ -12,3 +12,9 @@ _No other attributes_
 ## Code
 
 ::: textual.events.Focus
+
+## See also
+
+- [Blur](blur.md)
+- [DescendantFocus](descendant_focus.md)
+- [DescendantBlur](descendant_blur.md)
```
rodrigogiraoserrao, line 20: tsk tsk
```suggestion
- [DescendantBlur](descendant_blur.md)
- [DescendantFocus](descendant_focus.md)
```

rodrigogiraoserrao (approved):

docs/events/focus.md
davep: Just checking that you're reviewing properly. :-P

PR #3233, opened 2023-09-04
Modes docs

willmcgugan:
Adds docs for modes


src/textual/app.py
```diff
@@ -1535,19 +1540,19 @@ def mount_all(
...
+        """
 
         stack = self._screen_stacks.get(mode, [])
         if not stack:
             _screen = self.MODES[mode]
-            if callable(_screen):
-                screen, _ = self._get_screen(_screen())
-            else:
-                screen, _ = self._get_screen(self.MODES[mode])
+            new_screen: Screen | str = _screen() if callable(_screen) else _screen
```
willmcgugan, line 1552: Mainly to fix typing.

docs/guide/screens.md
```diff
@@ -256,3 +256,66 @@ Returning data in this way can help keep your code manageable by making it easy
...
+</div>
+
+
+To add modes to your app, define a [`MODES`][textual.app.App.MODES] class variable which should be a dictionary.
+This dictionary maps the name of the mode on to either a screen object, a callable that returns a screen, or the name of an installed screen.
+However you specify it, this screen is the base screen in an independent stack of screens.
+
+You can the switch between these screens at any time by calling [`App.switch_mode`][textual.app.App.switch_mode].
+When you switch to a new mode, it becomes the active screen stack and the topmost screen in the new stack becomes visible.
+Any calls to [`App.push_screen`][textual.app.App.push_screen] or [`App.pop_screen`][textual.app.App.pop_screen] will affect only the active mode.
```
rodrigogiraoserrao, line 292: I feel like this would go well with the paragraph that starts with “In Textual, we can manage this with *modes*.”.
As it stands, it feels like we're taking too long to tell the reader what are the methods that matter.

rodrigogiraoserrao (approved):
Looks good to me after a couple of typo fixes.

I left a suggestion on the table but I don't mind either way.

Issue #3267, opened 2023-09-09
Make notify threadsafe


willmcgugan:
Consider this issue https://github.com/Textualize/textual/discussions/3265

Calling anything UI related from a thread is generally not going to work. But maybe we should make it for for notify.

Suggest we add a `Notification` message. The `notify` method post this (non bubbling) message, and the handler would actually display it. That way you could call `self.notify` from a thread.

PR #3275, opened 2023-09-11
Make notify thread-safe

willmcgugan:
Makes `notify` thread-safe. This is a convenience when working with threaded workers.

Fixes https://github.com/Textualize/textual/issues/3267

### DIFF
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -2982,7 +2982,7 @@ class App(Generic[ReturnType], DOMNode):
         self.post_message(Notify(notification))
 
     def _on_notify(self, event: Notify) -> None:
-        """Handle notification method."""
+        """Handle notification message."""
         self._notifications.add(event.notification)
         self._refresh_notifications()
 
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -90,7 +90,7 @@ from .keys import (
     _get_unicode_name_from_key,
 )
 from .messages import CallbackType
-from .notifications import Notification, Notifications, SeverityLevel
+from .notifications import Notification, Notifications, Notify, SeverityLevel
 from .reactive import Reactive
 from .renderables.blank import Blank
 from .screen import Screen, ScreenResultCallbackType, ScreenResultType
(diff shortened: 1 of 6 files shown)


src/textual/app.py
```diff
@@ -2977,11 +2979,14 @@ def notify(
...
+        self.post_message(Notify(notification))
+
+    def _on_notify(self, event: Notify) -> None:
+        """Handle notification message."""
+        self._notifications.add(event.notification)
         self._refresh_notifications()
-        return notification
 
-    def unnotify(self, notification: Notification, refresh: bool = True) -> None:
+    def _unnotify(self, notification: Notification, refresh: bool = True) -> None:
```
darrenburns, line 2989: I've never used this API but this method seems like it's really useful if it's public?

Is this method how we discard notifications manually? It seems useful if notifications are longer-lived - to allow us to remove ones that are no longer relevant.

willmcgugan: The problem is not that `self.notify()` doesn't add the notification immediately. It posts a message, and returns `None`. That means that the user won't have a reference to the Notification object. And even if they did, it might not yet be active.

Programatically removing notifications is useful, but we will need another mechanism for that.

tests/notifications/test_app_notifications.py
```diff
@@ -17,15 +17,17 @@ async def test_app_with_notifications() -> None:
...
     async with NotificationApp().run_test() as pilot:
         pilot.app.notify("test")
+        await pilot.pause()
         assert len(pilot.app._notifications) == 1
 
 
 async def test_app_with_removing_notifications() -> None:
     """An app with notifications should have notifications in the list, which can be removed."""
     async with NotificationApp().run_test() as pilot:
         notification = pilot.app.notify("test")
```
davep, line 27:
```suggestion
        pilot.app.notify("test")
```

davep (approved):
One small tidy-up suggested in one of the tests; otherwise looks good and should be properly handy when debugging threaded workers!

src/textual/app.py
darrenburns: Not saying it's for this PR, but I think it'd be nice if `notify` still returned a `Notification` as a handle to a notification which may or may not be active. Then, if you try to `unnotify` a `Notification` that isn't active then it'd just prevent it from appearing.

PR #3276, opened 2023-09-11
docs: add datatable supplementary classes

### LINKED ISSUE #3273: `DataTable` supplementary types should be included in the documentation
It seems that the following `DataTable`-related types aren't included anywhere in the documentation:

- `CellDoesNotExist`
- `CellKey`
- `CellType`
- `Column`
- `ColumnDoesNotExist`
- `ColumnKey`
- `CursorType`
- `DuplicateKey`
- `Row`
- `RowDoesNotExist`
- `RowKey`

As such it would seem that their location, should the dev wish to import one or more of them, isn't obvious. It also means that when they're shown in the documentation as parameter and return types, etc, they're not linked to.

TomJGooding:
Closes #3273.

This also suggests slight changes to wording to make it more suitable for the documentation.

### DIFF
--- a/docs/widgets/data_table.md
+++ b/docs/widgets/data_table.md
@@ -217,3 +217,8 @@ The data table widget provides the following component classes:
 ::: textual.widgets.DataTable
     options:
       heading_level: 2
+
+::: textual.widgets.data_table
+    options:
+      show_root_heading: true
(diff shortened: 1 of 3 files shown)


davep (approved):
Thanks Tom!

Issue #872, opened 2022-10-11
Look at Windows key handling


willmcgugan:
Holding down shift creates ctrl-at keys.

sandos:
Yes, please! 

I believe this issue is prohibiting using anything with shift in windows, where, for example, ctrl+o works fine, shift+ctrl+o creates as you say ctrl-at sequences instead. With Windows Terminal also allocating a lot of shortcuts for itself it would be nice to expand the possible list of keys used. Every line after the ctrl+a was really a shift+ctrl+o: (its the same in "cmd.exe" as windows terminal)

`Key(key='ctrl+o', character='\x0f', name='ctrl_o', is_printable=False)
Key(key='ctrl+o', character='\x0f', name='ctrl_o', is_printable=False)
Key(key='ctrl+o', character='\x0f', name='ctrl_o', is_printable=False)
Key(key='ctrl+o', character='\x0f', name='ctrl_o', is_printable=False)
Key(key='ctrl+o', character='\x0f', name='ctrl_o', is_printable=False)
Key(key='ctrl+@', character='\x00', name='ctrl_@', is_printable=False)
Key(key='ctrl+o', character='\x0f', name='ctrl_o', is_printable=False)
Key(key='ctrl+o', character='\x0f', name='ctrl_o', is_printable=False)
Key(key='ctrl+o', character='\x0f', name='ctrl_o', is_printable=False)
Key(key='ctrl+o', character='\x0f', name='ctrl_o', is_printable=False)
Key(key='ctrl+o', character='\x0f', name='ctrl_o', is_printable=False)`

sandos:
Oh, I noticed another thing as well, which is maybe even worse: pressing a ctrl+alt-key combo delays / blocks combos that do work, until some class of event happens. For example the lone binding

```Binding("ctrl+o", "switch", "Hide hosts", show=True, priority=True),```

will be delayed after pressing ctrl+alt+r.

This is also easily observable in "textual keys": initially, pressing ctrl+o gets immediate responses. Pressing ctrl+alt+r once, and then ctrl+o shows nothing happening.. until for example the mouse is hovered over the console.

davep:
Dragging this into "in progress" and using this as a place to log some other issues with keys and Windows. See also [the problem with the default binding for the command palette on Windows](https://github.com/Textualize/textual/pull/3058#issuecomment-1702591818) and also #3178.

darrenburns:
[...]

PR #3280, opened 2023-09-11
Filter out any NUL keystrokes that aren't Ctrl+2/Ctrl+space

davep:
See #872.

### DIFF
--- a/src/textual/drivers/win32.py
+++ b/src/textual/drivers/win32.py
@@ -268,6 +268,13 @@ class EventMonitor(threading.Thread):
                             if (
                                 key_event.dwControlKeyState
                                 and key_event.wVirtualKeyCode == 0
+                            ) or (
+                                # Filter out any NUL characters that don't
+                                # appear to be Ctrl-2, AKA Ctrl-space.
+                                #
+                                # https://github.com/Textualize/textual/issues/872
+                                key == "\x00"
+                                and key_event.wVirtualKeyCode != 50
                             ):
                                 continue
                             append_key(key)


davep:
Note that this also solves the problem of the Command Palette popping up when you paste something into an input field in Windows.

davep:
Taking this back to draft for a moment. While it seems to work fine in Windows under Parallels on my Mac, it's not working fine on the native Windows installation. This *might* be down to Windows Terminal version or something, amongst other things, so I want to double-check all of that.

davep:
Key differences between working and non-working environments:

# Working

- Windows 11
- Under Parallels
- Under macOS
- Windows Terminal 1.17.11461.0

# Non-working

- Windows 10 Pro
- Native
- Windows Terminal 1.16.10261.0

# Conclusion

The first thing to try is to update the version of Windows Terminal so it's the same in both places.

davep:
Updated to the same version of Windows Terminal, on the native Windows 10 machine, and it made not a jot of difference. I'll throw in some low-level debugging again to see what's coming through.

The machine also has the option to be upgraded to Windows 11, but I'm not going to jump into doing that just yet -- having a Windows 10 machine to hand to test with does make a lot of sense; especially if that is the difference.

davep:
https://github.com/microsoft/terminal/issues/2865 may be important here.

davep:
Even odder. This PR, under Windows 11, under Parallels, on macOS, still supports <kbd>Ctrl</kbd>+<kbd>Space</kbd>. OTOH @rodrigogiraoserrao reports that Windows 11 native, with this PR, shows nothing for the same key combination; while at the same time without the PR shows `ctrl+@`.

TomJGooding:
I've also been doing some testing on my Windows 10 if it helps.

Running `textual keys` on main and then pressing `ctrl` on its own shows 'ctrl+@'. Pressing `ctrl+space` results in 'ctrl+@ showing twice.

With this PR, `textual keys` shows nothing both after pressing `ctrl` and `ctrl+space`.

davep:
@TomJGooding Thanks, looks like you're seeing what I'm seeing.

davep:
So, having tested native on Windows 11 here too now, and seeing the same effect, this little hack doesn't quite solve the problem at all. Or, more to the point: it *does* seem to solve the problem but at the expense of <kbd>Ctrl</kbd>+<kbd>Space</kbd>; this might be an acceptable loss if #872 is a bigger problem.

I'm going to leave this as draft while we think on this and explore other options. Meanwhile we'll move the Command Palette onto a different keyboard combination that we can be more confident about.

TomJGooding:
> AKA, for when I can handle subjecting myself to opening a Windows session again and having to try and do Python coding in the most hostile of environments.

I guess even if you had Emacs you couldn't set a mark either.... :wink:

davep:
> I guess even if you had Emacs you couldn't set a mark either.... 😉

<kbd>M</kbd>-<kbd>x</kbd> `set-mark-command` <kbd>RET</kbd> :-P

TomJGooding:
You may have found this already, but I stumbled on the [debug tap](https://github.com/microsoft/terminal/wiki/Toubleshooting-Tips#enabling-the-debug-tap) mode for Windows Terminal which might be useful?

davep:
Closing this PR as it seems the downsides outweigh how trivial it is, and hopefully there's a more fundamental solution to be had with something like #3411.

Issue #3268, opened 2023-09-09
Allow progress bars to go back to being intermediate


willmcgugan:
A Discord user https://discord.com/channels/1026214085173461072/1033754296224841768/1149742624002023594 reported that the docs say you can't go back to a intermediate state with the ProgressBar.

This seems to be an arbitrary restriction. And does actually work to some extent.

We should make sure it works, not matter how you use the interface. And update the API docs to reflect that.

PR #3286, opened 2023-09-12
Tweak progress bar docs.

rodrigogiraoserrao:
There is no good reason as to why the progress bar can't be set back to its indeterminate state (and you could actually do it with code) so this removes the docstring that says that a progress bar can't go back to its indeterminate state.

Related issue: #3268
Related Discord message: https://discord.com/channels/1026214085173461072/1033754296224841768/1149742624002023594

The user in the linked Discord message reports that `pb.update(total=None)` “didn't change the [total] value so it doesn't work” but that's how `update` works:

 - `pb.update(progress=None)` does nothing; and
 - `pb.update(advance=None)` also does nothing.

The method `update` can only affect changes for keyword arguments that are set to a value _other_ than `None`.
Setting `pb.total = None` works (it already worked) and the docstring for `total` no longer says otherwise.
I also added a test to make sure that setting `total` back to `None` works, so now this is explicitly a supported feature.


willmcgugan (changes requested):
You can also set the total with  the `update`, and it look like that prohibits setting None.

rodrigogiraoserrao:
> You can also set the total with the `update`, and it look like that prohibits setting None.

It's not that it prohibits. It's the _only_ way to design such an `update` method which can update multiple things at once. It's just like how `update` works for rich progress bars [docs](https://rich.readthedocs.io/en/stable/reference/progress.html#rich.progress.Progress.update) and it's written clearly that `update(some_key=some_value)` will only have any effect if `some_value` is not `None`.

rodrigogiraoserrao:
I can get rid of the method `update` if it's causing more confusion that doing good.

willmcgugan (approved):

PR #3289, opened 2023-09-12
Command Palette tweaks and docs

willmcgugan:
- Adds docs on the Command Palette to Guide.
- Change the module name from `command_palette` to `command`
- Changes the name of some the objects, mostly for brevity
- Adds a `post_init` function to the command sources.
- Makes textual._fuzzy public


src/textual/app.py
```diff
@@ -326,14 +326,14 @@ class MyApp(App[None]):
     """
 
     ENABLE_COMMAND_PALETTE: ClassVar[bool] = True
-    """Should the [command palette][textual.command_palette.CommandPalette] be enabled for the application?"""
+    """Should the [command palette][textual.command.CommandPalette] be enabled for the application?"""
 
-    COMMAND_SOURCES: ClassVar[set[type[CommandSource]]] = {SystemCommandSource}
-    """The [command sources](/api/command_palette/) for the application.
+    COMMAND_SOURCES: ClassVar[set[type[Source]]] = {SystemCommandSource}
+    """The [command sources](/api/command/) for the application.
```
rodrigogiraoserrao, line 332: We should use paths to textual objects because it's easier to keep URLs synced that way.
The `[some text here][textual.module.Object]` ones.

src/textual/command.py
```diff
@@ -1,9 +1,13 @@
-"""The Textual command palette."""
+"""The Textual command palette.
+
+See the guide on the [Command Palette](../guide/command_palette.md) for full details.
+
```
rodrigogiraoserrao, line 4:
```suggestion
```

src/textual/command.py
```diff
@@ -42,24 +47,19 @@
     from .app import App, ComposeResult
 
 __all__ = [
```
rodrigogiraoserrao, line 49: Alphabetical order, please? 🥹

src/textual/command.py
```diff
@@ -42,24 +47,19 @@
...
-
-
 @total_ordering
-class CommandSourceHit(NamedTuple):
+class Hit(NamedTuple):
     """Holds the details of a single command search hit."""
 
-    match_value: float
+    score: float
     """The match value of the command hit.
```
rodrigogiraoserrao, line 63:
```suggestion
    """The score of the command hit.
```

src/textual/command.py
```diff
@@ -68,43 +68,47 @@ class CommandSourceHit(NamedTuple):
...
 """Return type for the command source match searching method."""
 
 
-class CommandSource(ABC):
+class Source(ABC):
     """Base class for command palette command sources.
 
     To create a source of commands inherit from this class and implement
-    [`search`][textual.command_palette.CommandSource.search].
+    [`search`][textual.command.Source.search].
```
rodrigogiraoserrao, line 111:
```suggestion
    the method [`search`][textual.command.Source.search].
```

src/textual/command.py
```diff
@@ -136,44 +142,81 @@ def app(self) -> App[object]:
...
+        """The preferred style to use when highlighting matching portions of the [`match_display`][textual.command.Hit.match_display]."""
         return self.__match_style
 
     def matcher(self, user_input: str, case_sensitive: bool = False) -> Matcher:
-        """Create a [fuzzy matcher][textual._fuzzy.Matcher] for the given user input.
+        """Create a [fuzzy matcher][textual.fuzzy.Matcher] for the given user input.
 
         Args:
             user_input: The text that the user has input.
             case_sensitive: Should match be case sensitive?
```
rodrigogiraoserrao, line 153:
```suggestion
            case_sensitive: Should matching be case sensitive?
```

src/textual/command.py
```diff
@@ -453,10 +498,21 @@ def _on_click(self, event: Click) -> None:
             self.workers.cancel_all()
             self.dismiss()
 
-    def _on_mount(self, _: Mount) -> None:
+    def on_mount(self, _: Mount) -> None:
```
rodrigogiraoserrao, line 501: Why are we making this public?

src/textual/screen.py
```diff
@@ -157,7 +157,7 @@ class Screen(Generic[ScreenResultType], Widget):
     title: Reactive[str | None] = Reactive(None, compute=False)
     """Screen title to override [the app title][textual.app.App.title]."""
 
-    COMMAND_SOURCES: ClassVar[set[type[CommandSource]]] = set()
+    COMMAND_SOURCES: ClassVar[set[type[Source]]] = set()
     """The [command sources](/api/command_palette/) for the screen."""
```
rodrigogiraoserrao, line 161: API link should be of the type `[text here][textual.module.Object]`

tests/snapshot_tests/snapshot_apps/command_palette.py
```diff
@@ -1,25 +1,29 @@
 from textual.app import App
-from textual.command_palette import CommandSource, CommandMatches, CommandSourceHit
+from textual.command import Source, Hits, Hit
```
rodrigogiraoserrao, line 2:
```suggestion
from textual.command import Hit, Hits, Source
```

rodrigogiraoserrao:
I left a bunch of comments and suggestions, but nothing blocking.

docs/examples/guide/command_palette/command01.py
```diff
@@ -0,0 +1,65 @@
...
+                    matcher.highlight(command),  # (5)!
+                    lambda: app.open_file(path),
+                    help="Open this file in the viewer",
+                )
+
+
+class ViewerApp(App):
+    """Demonstrate a command source."""
+
+    COMMAND_SOURCES = App.COMMAND_SOURCES | {PythonFileSource}  # (6)!
```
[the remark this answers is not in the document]
davep: The decision was made a wee while ago (so as part of the original PR) that it wouldn't be the case for the command sources; with the aim to review how people use this.

src/textual/command.py
```diff
@@ -68,43 +68,47 @@ class CommandSourceHit(NamedTuple):
...
     def __eq__(self, other: object) -> bool:
-        if isinstance(other, CommandSourceHit):
-            return self.match_value == other.match_value
+        if isinstance(other, Hit):
+            return self.score == other.score
         return NotImplemented
 
+    def __post_init__(self) -> None:
+        if self.text is None:
+            self.text = str(self.match_display)
```
davep, line 100: Attributes of a `NamedTuple` are normally read-only. Is there something special about a `__post_init__` that allows this assignment?

davep (changes requested):
From what I can see, the move to make the `Input`-bound text of the `Hit` optional isn't quite right and won't work; I've pointed to where I think the issue is.

I also want to test what happens if someone uses more interesting match text without supplying the `Input`-bound text.

davep:
I did forget to add though: the docs are looking good!

src/textual/command.py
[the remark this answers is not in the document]
davep: To see this using the example in the docs, add this following to `ViewerApp`:

```python
    def on_mount(self) -> None:
        from textual.command import CommandPalette
        CommandPalette.run_on_select = False
```

When you select a file the `Input` will read `None`.

docs/examples/guide/command_palette/command01.py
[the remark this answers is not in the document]
willmcgugan: We might revisit that in the future. But I suspect the most common requirement will to override it per app, so inheritance is not that useful.

docs/guide/command_palette.md
```diff
@@ -0,0 +1,106 @@
...
+
+There are two methods you will typically override in a command source: [`post_init`][textual.command.Source.post_init] and [`search`][textual.command.Source.search].
+Both should be coroutines (`async def`).
+Let's explore those methods in detail.
+
+### post_init method
+
+The [`post_init`][textual.command.Source.post_init] method is called when the command palette is opened.
+You can use this method as way of performing work that needs to be done prior to searching.
+In the example, we use this method to get the Python (.py) files in the current working directory.
```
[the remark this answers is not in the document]
willmcgugan: The command sources are constructed each time the command palette is opened. I think `post_init` is somewhat familiar, as it has much the same meaning as dataclass `__post_init__`

src/textual/command.py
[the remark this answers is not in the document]
darrenburns: __post_init__ only works on dataclasses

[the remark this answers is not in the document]
davep: Also, to emphasise the part I mention about the intention of `match_display` vs `text`, it's worth keeping in mind that a `match_display` is a `RenderableType`. This should mean, and is done with the intention, that the display could be done within a Rich `Table` or some other more complex display. But the `str` of a `Table` would, unless I've missed something, be something like this:

```
<rich.table.Table object at 0x1022a6170>
```

If this is going to be an acceptable case of "leave off the `text` if you know your `match_display` can `str` to something readable and useable within an `Input`", we might want to tweak the documentation here to make it clear what the dev's responsibilities are regarding using the default case or not.

docs/guide/command_palette.md
```diff
@@ -0,0 +1,106 @@
...
+  3. Get a [Matcher][textual.fuzzy.Matcher] instance to compare against hits.
+  4. Use the matcher to get a score.
+  5. Highlights matching letters in the search.
+  6. Adds our custom command source and the default command sources.
+
+There are two methods you will typically override in a command source: [`post_init`][textual.command.Source.post_init] and [`search`][textual.command.Source.search].
+Both should be coroutines (`async def`).
+Let's explore those methods in detail.
+
+### post_init method
```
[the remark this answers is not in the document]
willmcgugan: I think I'd prefer to keep headers as text.

docs/guide/command_palette.md
```diff
@@ -0,0 +1,106 @@
...
+
+### search method
+
+The [`search`][textual.command.Source.search] method is responsible for finding results (or *hits*) that match the user's input.
+This method should *yield* [`Hit`][textual.command.Hit] objects for any command that matches the `query` argument.
+
+Exactly how the matching is implemented is up to the author of the command source, but we recommend using the builtin fuzzy matcher object, which you can get by calling [`matcher`][textual.command.Source.matcher].
+This object has a [`match()`][textual.fuzzy.Matcher.match] method which compares the user's search term against the potential command and returns a *score*.
+A score of zero means *no hit*, and you can discard the potential command.
+A score of above zero indicates the confidence in the result, where 1 is an exact match, and anything lower indicates a less confident match.
```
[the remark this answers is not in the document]
willmcgugan: The original reads better to me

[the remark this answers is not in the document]
davep: I'll second that; I wouldn't personally put "of" in there.

src/textual/command.py
willmcgugan: I don't think we should be too fussed about additional lines in docstrings.

src/textual/command.py
willmcgugan: "Implement the method search" ? Maybe "implement the search method", but think it reads ok as is.

src/textual/screen.py
willmcgugan: I think Dave's intent was to link to the page in the API and not a specific object. How else would you do that?

src/textual/command.py
willmcgugan: quells an error about overriding things. I don't think anyone will extend this anyway,

docs/examples/guide/command_palette/command01.py
```diff
@@ -0,0 +1,65 @@
...
+
+class PythonFileSource(Source):
+    """A command source to open a file in the current working directory."""
+
+    def read_files(self) -> list[Path]:
+        """Get a list of Python files in the current working directory."""
+        return list(Path("./").glob("*.py"))
+
+    async def post_init(self) -> None:  # (1)!
+        """Called before search."""
```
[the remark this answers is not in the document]
willmcgugan: I wouldn't have infered that. But how about "Called prior to searching" to avoid confusion?

docs/examples/guide/command_palette/command01.py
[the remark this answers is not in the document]
rodrigogiraoserrao: Ok. I would've asked on the original PR but I was waiting for the green light to review the PR and it never came :P

src/textual/screen.py
davep: Sounds about right to me: if I recall correctly the intent was that it eventually settles on somewhere where the concept of command sources, what they are, how to use them, how to build them, etc; not on a specific bit of class documentation.

src/textual/command.py
rodrigogiraoserrao: I wasn't going to make a PR to delete them, but this was one click away 🤷

docs/guide/command_palette.md
```diff
@@ -0,0 +1,112 @@
+# Command Palette
+
+Textual apps have a built-in *command palette*, which gives users a quick way to access certain functionality within your app.
+
+In this chapter we will explain what a command palette is, how to use it, and how you can add your own commands.
+
+## Launching the command palette
+
+Press ++ctrl+space++ to invoke the command palette (modal) screen, which contains of a single input widget.
```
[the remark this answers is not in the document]
willmcgugan: I don't know if we have settled on something else yet. We can change that in another PR.

src/textual/command.py
davep: Nit-pick: if going public, we can lose the `_: Mount`.

docs/guide/command_palette.md
```diff
@@ -0,0 +1,112 @@
...
+Textual will suggest commands as you type in that input.
+Press ++up++ or ++down++ to select a command from the list, and ++enter++ to invoke it.
+
+Commands are looked up via a *fuzzy* search, which means Textual will show commands that match the keys you type in the same order, but not necessarily at the start of the command.
```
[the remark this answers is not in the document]
willmcgugan: I think its fine as an opening paragraph. I don't like having to caveat everything.

docs/examples/guide/command_palette/command01.py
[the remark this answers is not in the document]
davep: Perfect, I think that's nice and unambiguous!

docs/guide/command_palette.md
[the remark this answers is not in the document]
rodrigogiraoserrao: It looks familiar but that familiarity is misleading: the `__post_init__` in dataclasses runs _once_ after instantiation.
Is the command source _instantiated_ every single time we open it?

If it is, then I retract my comment about `post_init` being a bad name and instead request that we make it _very_, _very_ clear that the command source is instantiated every time the palette is opened.

src/textual/command.py
willmcgugan: Strange. I made `text` allow None, and assumed I would have to populate it. But this code will never even run, so I guess thats already taken care of elsewhere.

src/textual/screen.py
rodrigogiraoserrao: `/api/xxx` pages can be linked to with `[textual.xxx]`, e.g.:

```suggestion
    """The [command sources][textual.command] for the screen."""
```

If you don't want to switch to that link style, then at least the link requires fixing:

```suggestion
    """The [command sources](/api/command/) for the screen."""
```

docs/guide/command_palette.md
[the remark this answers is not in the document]
davep: Cool, I can go with that; just as long as we're aware of it and are happy with it.

docs/guide/command_palette.md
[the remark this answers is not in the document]
davep: Sure thing. But before release, yeah?

src/textual/command.py
davep: As mentioned elsewhere; the upshot of this is that the `text` that's intended for the `Input` ends up being `"None"` -- this will be even more important if/when we do the thing we were talking about a while back, where a command can complete into the `Input` and then require further parameters. This is why the distinction between the `match_display` (which goes into the dropdown, and can be any renderable type: highlighted text, a Rich Table, probably a Rich Pixels Pokemon image, etc...) and the `text` is important here.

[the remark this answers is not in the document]
willmcgugan: Ok, some misunderstandings on my part. I've made thing work the way I assumed they were! The `text` is populated automatically now, if `match_display` is a str or Text, which will be the 99% case.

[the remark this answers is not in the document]
davep: Perfect! Rich Pixels-based Pokédex command source is safe once again!

docs/guide/command_palette.md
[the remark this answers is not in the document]
willmcgugan: yup.

davep (approved):
Given my main concern here was the display text vs the input text... I'm all on board with this PR now.

docs/guide/command_palette.md
```diff
@@ -0,0 +1,118 @@
...
+
+    Unlike most other places in Textual, errors in command sources will not *exit* the app.
+    This is a deliberate design decision taken to prevent a single broken `Source` class from making the command palette unusable.
+    Errors in command sources will be logged to the [console](./devtools.md).
+
+## Screen commands
+
+You can also associate commands with a screen by adding a `COMMAND_SOURCES` class var to your Screen class.
+
+This is useful for commands that only make sense when a give screen is active.
```
davep, line 106:
```suggestion
This is useful for commands that only make sense when a given screen is active.
```

Issue #3178, opened 2023-08-26
app hangs when pasting emoji from clipboard to an Input widget


HubertReX:
Although Input widgets display emojis correctly when a value is set in code (I mean actual emoji like 😊not :smiling_face_with_smiling_eyes:), the whole app hangs when trying to paste emoji from the clipboard on Windows 10.

As a sample code, one can use even this first example:
[Input - Textual](https://textual.textualize.io/widgets/input/#__tabbed_1_2)
Then copy an emoji to the clipboard and then try to paste it (either Ctr+V or Ctrl+Shift+V)
For me, the main app hangs and the respective python process needs to be terminated. The textual console continues to work. 

Here is the output from the textual console:
```
───────────── Client '127.0.0.1' connected ────────────────────────────────────────────────────
[09:56:30] PRINT                                                                                                                                                                                      app.py:1102
Exception in thread Thread-1:
[09:56:30] PRINT                                                                                                                                                                                      app.py:1102
Traceback (most recent call last):
[09:56:30] PRINT                                                                                                                                                                                      app.py:1102
  File "c:\Users\user\AppData\Local\Programs\Python\Python311\Lib\threading.py", line 1038, in _bootstrap_inner
[09:56:30] PRINT                                                                                                                                                                                      app.py:1102
    self.run()
[09:56:30] PRINT                                                                                                                                                                                      app.py:1102
  File "C:\Projekty\RPG\.py311_env\Lib\site-packages\textual\drivers\_writer_thread.py", line 60, in run
[09:56:30] PRINT                                                                                                                                                                                      app.py:1102
    write(text)
[09:56:30] PRINT                                                                                                                                                                                      app.py:1102
UnicodeEncodeError: 'utf-8' codec can't encode characters in position 423-424: surrogates not allowed


```
I could reproduce this on Win10 in Windows Terminal 1.17 using PowerShell 7. Works fine on Ubuntu 22.
When run using PowerShell, but not inside Windows Terminal, nothing is pasted, but at least the app continues to work. Still, this is useless since emojis do not work in PS without WT.

# Textual Diagnostics

## Versions

| Name    | Value  |
|---------|--------|
| Textual | 0.33.0 |
| Rich    | 13.5.2 |

## Python

| Name           | Value                                         |
|----------------|-----------------------------------------------|
| Version        | 3.11.4                                        |
| Implementation | CPython                                       |
| Compiler       | MSC v.1934 64 bit (AMD64)                     |
[...]

PR #3290, opened 2023-09-13
Recode incoming text on Windows before input event processing

davep:
This is a tentative fix for #3178. I'm keeping this as draft for the moment as this affects the Windows input at a very low level, and I'd like the input of @darrenburns and @willmcgugan on if this makes sense. What I do know is that this fixes the problem reported in #3178, and so I suspect there's the core of the solution in here.

So... eyes and experience with handling emoji and the like on this would be welcome.


davep:
Moving out of draft for full review [as per](https://github.com/Textualize/textual/issues/3178#issuecomment-1733209574).

davep:
Merged in `main` as it's a couple of days since it was put into the review queue.

willmcgugan (approved):

willmcgugan:
Would be awesome if you could find some reference in the win32 docs as to why this is required on Windows, and not elsewhere. But if this is the fix, lets merge.

darrenburns:
I spotted a few references to data on the clipboard being encoded as UTF-16 on the MS developer docs, but nothing that stands out to explain exactly why the approach in this PR approach works.

e.g. https://learn.microsoft.com/en-us/dotnet/api/system.windows.clipboard.settext?view=windowsdesktop-7.0

PR #3294, opened 2023-09-13
Ensure that `print` reports the correct location in console

### LINKED ISSUE #3237: Textual console showing wrong path
The `textual console` command is showing the wrong path.

IIRC there is an offset in the stack which it uses to find the caller. We may have introduced another level in the stack.

Image

davep:
**DO NOT MERGE***

Fixes #3237 

See https://github.com/Textualize/textual-dev/pull/19 for the background to this. Once that side of things is sorted in `textual-dev` and the package has been published, then this can be merged.

Marking for review but as draft; it can be looked over and reviewed, but it should not be merged yet.

### DIFF
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -1161,7 +1161,10 @@ class App(Generic[ReturnType], DOMNode):
             stderr: True if the print was to stderr, or False for stdout.
         """
         if self._devtools_redirector is not None:
-            self._devtools_redirector.write(text)
+            current_frame = inspect.currentframe()
+            self._devtools_redirector.write(
+                text, current_frame.f_back if current_frame is not None else None
+            )
         for target, (_stdout, _stderr) in self._capture_print.items():
             if (_stderr and stderr) or (_stdout and not stderr):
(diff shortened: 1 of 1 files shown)


src/textual/app.py
```diff
@@ -1156,7 +1156,10 @@ def _print(self, text: str, stderr: bool = False) -> None:
             stderr: True if the print was to stderr, or False for stdout.
         """
         if self._devtools_redirector is not None:
-            self._devtools_redirector.write(text)
+            current_frame = inspect.currentframe()
+            self._devtools_redirector.write(
+                text, current_frame.f_back if current_frame is not None else None
```
rodrigogiraoserrao, line 1161: So, the previous frame is the one where `print` was called?
Can you briefly explain how a call to `print` ends up here?

davep: > Can you briefly explain how a call to print ends up here?

Not off the top of my head; I don't recall going near that.

rodrigogiraoserrao (approved):

PR #3305, opened 2023-09-14
Update Collapsible

willmcgugan:
- Fixed / updated docs for Collapsible
- Simplified implementation a bit


willmcgugan:
@rodrigogiraoserrao you are most familiar with this...

docs/widgets/collapsible.md
```diff
@@ -1,34 +1,37 @@
 # Collapsible
 
-!!! tip "Added in version 0.36"
+!!! tip "Added in version 0.37"
 
-Widget that wraps its contents in a collapsible container.
+A container with a title than can be used to show (expand) or hide (collapse) content, either by clicking or focusing and pressing ++enter++.
```
rodrigogiraoserrao, line 5:
```suggestion
A container with a title that can be used to show (expand) or hide (collapse) content, either by clicking or focusing and pressing ++enter++.
```

docs/widgets/collapsible.md
```diff
@@ -1,34 +1,37 @@
...
+
+Here is an example of using the constructor to add content:
 
 ```python
     yield Collapsible(Label("Hello, world."))
 ```
 
-Alternatively, you can compose other widgets under the context manager:
+Here's how the to use it with the context manager.
```
rodrigogiraoserrao, line 22: For consistency with the example above.

```suggestion
Here's how the to use it with the context manager:
```

src/textual/widgets/_collapsible.py
```diff
@@ -1,87 +1,93 @@
...
+    def action_toggle(self) -> None:
+        """Toggle the state of the parent collapsible."""
+        self.post_message(self.Toggle())
+
+    def render(self) -> RenderableType:
+        """Compose right/down arrow and label."""
+        if self.collapsed:
+            return Text.assemble(self.collapsed_symbol, " ", self.label)
+        else:
+            return Text.assemble(self.expanded_symbol, " ", self.label)
```
rodrigogiraoserrao, line 73: Doesn't this get rerun every single time we need to show the title, even if the collapsed state didn't change?
Couldn't we cache this if we inherited from `Static`?

src/textual/widgets/_collapsible.py
```diff
@@ -1,87 +1,93 @@
...
-        Title {
-            width: 100%;
-            height: auto;
-        }
+    BINDINGS = [Binding("enter", "toggle", "Toggle collapsible", show=False)]
 
-        Title:hover {
-            background: grey;
-        }
+    collapsed = reactive(False)
```
rodrigogiraoserrao, line 40: I find it a bit odd that `Collapsible.collapsed` starts off as `True` and this one starts off as `False`!

rodrigogiraoserrao (changes requested):
I think I spotted something small that needs changing.

src/textual/widgets/_collapsible.py
willmcgugan: Is a bit. Changed to match the default.

src/textual/widgets/_collapsible.py
willmcgugan: Only when it is repainted. It's such a trivial function that caching the renderable may not even be a win.

But I have noticed, that it is doing a little more work that required!

PR #3307, opened 2023-09-14
docs: correct grid gutter type

TomJGooding:
Correct the docs for `grid-gutter` (as discussed in #3249), which currently says this can be a scalar where it only accepts an integer.


willmcgugan:
Thanks, Tom!

PR #3314, opened 2023-09-15
color command example

willmcgugan:
- Added command palette examples to examples
- Tweaked the highlight color in the CP


willmcgugan:
#executivedecision

PR #3321, opened 2023-09-16
Fix command palette `TimeoutError` error

### LINKED ISSUE #3320: `TimeoutError` when finding nothing in the Command Palette (Python <3.11)
To test this, make a fresh installation of Textual, then run:

```sh
$ python -m textual
```

To run up the demo. Press <kbd>Ctrl</kbd>+<kbd>\\</kbd> to pull up the command palette, then type something that won't be found, such as "hello". Within a moment or so you should get a `TimeoutError`.

Note that this seems to *only* be an issue with Pythons less than 3.11.

davep:
Fixes #3320 

It looks like eaa749665f9b8271eff45be8e5e1e72ac8729b9e unintentionally removed the import of `TimeoutError` from `asyncio`, causing the command palette to only work correctly on Python 3.11 or later.

Tested on Python 3.7 and 3.11 with this change and all working again in both Pythonic extremes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## Unreleased
+
+### Fixed
+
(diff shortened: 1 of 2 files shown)


willmcgugan (approved):

willmcgugan:
I'll do a point release today.

PR #3322, opened 2023-09-16
Stop command palette `Input` message leakage

davep:
It was possible for `Input` events to leak from the `CommandPalette` up to the developer's application, which could result in unintended side-effects if the developer had broad event handlers on their application (not an unreasonable situation if writing a simple all-in-the-app tool).

This does raise the question of what to do about all possible messages that could be posted from something like the command palette, which the developer may not be expecting and which could bubble up to their application. I'll raise an issue for considering that (perhaps there should be, or is there, a way to say "no message shall pass!" on a screen).

Edit to add: raised the wider issue as #3323

### DIFF
--- a/src/textual/command.py
+++ b/src/textual/command.py
@@ -876,6 +876,7 @@ class CommandPalette(ModalScreen[CallbackType], inherit_css=False):
         Args:
             event: The input event.
         """
+        event.stop()
         self.workers.cancel_all()
         search_value = event.value.strip()
         if search_value:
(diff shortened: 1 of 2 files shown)


src/textual/command.py
```diff
@@ -906,10 +907,14 @@ def _select_command(self, event: OptionList.OptionSelected) -> None:
 
     @on(Input.Submitted)
     @on(Button.Pressed)
-    def _select_or_command(self) -> None:
+    def _select_or_command(
+        self, event: Input.Submitted | Button.Pressed | None = None
```
willmcgugan, line 911: If this is a message handlers, then I don't think it will ever be None.

davep: It doubles as a message handler and an internally-called method; the message itself, until now, was throwaway. It's called elsewhere in the code with no message.

davep: (to add: this is one of the things I love about `@on` -- it removes the need for thin wrappers for messages to fire code that normally doesn't care one jot about the message itself)

willmcgugan: Oh I see. I would probably favor factoring out the common code rather than doing it that way. Not a blocker though.

willmcgugan (approved):

src/textual/command.py
davep: Yeah, I think I'm on the `@on`-is-awesome-lean-into-it-hard side of the fence (at least for my own code), but if this method had started out with an `event` parameter, I probably would have gone that way.

PR #3329, opened 2023-09-16
Testing guide

willmcgugan:
Adds Testing to the guide.

This covers a lot. In the future I might add a few more examples.

Still do do is the snapshot framework, which I'll leave for @darrenburns


docs/guide/testing.md
```diff
@@ -0,0 +1,168 @@
...
+
+!!! tip "Learn more about testing"
+
+    We recommend [Python Testing with pytest](https://pythontest.com/pytest-book/) for a comprehensive guide to writing tests.
+
+## Do you need to write tests?
+
+The short answer is "no", you don't *need* to write tests.
+
+In practice however, it is almost always a good idea to write tests.
```
brokenshield, line 19:
```suggestion
1. In practice however, it is almost always a good idea to write tests.
```

docs/guide/testing.md
```diff
@@ -0,0 +1,168 @@
...
+Writing code that is completely bug free is virtually impossible, even for experienced developers.
```
brokenshield, line 20:
```suggestion
1. Writing code that is completely bug free is virtually impossible, even for experienced developers.
```

docs/guide/testing.md
```diff
@@ -0,0 +1,168 @@
...
+If you want to have confidence that your application will run as you intended it to, then you should write tests.
```
brokenshield, line 21:
```suggestion
1. If you want to have confidence that your application will run as you intended it to, then you should write tests.
```

docs/guide/testing.md
```diff
@@ -0,0 +1,168 @@
...
+Your test code will help you find bugs early, and alert you if you accidentally break something in the future.
```
brokenshield, line 22:
```suggestion
1. Your test code will help you find bugs early, and alert you if you accidentally break something in the future.
```

docs/guide/testing.md
```diff
@@ -0,0 +1,168 @@
...
+The `run_test()` method is an *async context manager* which returns a [`Pilot`][textual.pilot.Pilot] object.
+You can use this object to interact with the app as if you were operating it with a keyboard and mouse.
+
+Let's look at the tests for the example above:
+
+```python title="test_rgb.py"
+--8<-- "docs/examples/guide/testing/test_rgb.py"
+```
+
+1. The `run_test()` method requires that it run in a coroutine, so test must use the `async` keyword.
```
brokenshield, line 65:
```suggestion
1. The `run_test()` method requires that it run in a coroutine, so tests must use the `async` keyword.
```

docs/guide/testing.md
```diff
@@ -0,0 +1,168 @@
...
+2. This runs the app and returns a Pilot instance we can use to interact with it.
+3. Simulates pressing the ++r++ key.
+4. This checks that pressing the ++r++ key has resulted in the background color changing.
+5. Simulates clicking on the widget with an `id` of `red` (the button labelled "Red").
+
+There are two tests defined in `test_rgb.py`.
+One to test keys and one to test button clicks.
```
brokenshield, line 72:
```suggestion
The first to test keys and the second to test button clicks.
```

docs/guide/testing.md
```diff
@@ -0,0 +1,168 @@
...
+Both tests first construct an instance of the app and call `run_test()` to get a Pilot object.
```
brokenshield, line 73:
```suggestion
Both tests first construct an instance of the app and then call `run_test()` to get a Pilot object.
```

docs/guide/testing.md
```diff
@@ -0,0 +1,168 @@
...
+```
+
+Each string creates a single keypress.
+You can also use the name for non-printable keys (such as "enter") and the "ctrl+" modifier.
+These are the same identifiers as used for key events, which you can experiment with by running `textual keys`.
+
+## Simulating clicks
+
+You can simulate mouse clicks in a similar way with [`Pilot.click`][textual.pilot.Pilot.click].
+If you supply a css selector Textual will simulate clicking on the matching widget.
```
brokenshield, line 101:
```suggestion
If you supply a CSS selector Textual will simulate clicking on the matching widget.
```

docs/guide/testing.md
```diff
@@ -0,0 +1,168 @@
...
+
+```python
+async with app.run_test(size=(100, 50)) as pilot:
+    ...
+```
+
+## Pausing the pilot
+
+Some actions in a Textual app won't change the state immediately.
+For instance messages may take a moment to bubble from the widget that sent them.
```
brokenshield, line 158:
```suggestion
For instance, messages may take a moment to bubble from the widget that sent them.
```

docs/guide/testing.md
willmcgugan: Not keen on a list here. Wasn't intended to be list like.

willmcgugan:
Thanks @brokenshield

PR #3348, opened 2023-09-19
Mark the command palette as existing

davep:
I've not checked off "Command menu" as that sounds like something slightly different from the command palette.

### DIFF
--- a/docs/roadmap.md
+++ b/docs/roadmap.md
@@ -19,9 +19,9 @@ High-level features we plan on implementing.
     * [x] Monochrome mode
     * [ ] High contrast theme
     * [ ] Color-blind themes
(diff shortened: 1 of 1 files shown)


TomJGooding:
Mark as existing... or M-x for short?

willmcgugan:
That's it. We're going to have an emacs jar. Whenever somebody mentions anything re emacs they have to deposit £1 in the jar.

Pretty soon we won't need VC money!

willmcgugan:
I think we could probably replace that section with just "Command Palette" and check it off. Don't recall what I had in mind with "command menu"

davep:
Roadmap tweaked.

willmcgugan (approved):

PR #3357, opened 2023-09-20
Snapshot testing guide

darrenburns:


docs/guide/testing.md
```diff
@@ -162,7 +162,150 @@ You can generally solve this by calling [`pause()`][textual.pilot.Pilot.pause] w
...
+This offers a simple, automated way of checking our application displays like we expect.
+
+### pytest-textual-snapshot
+
+You can use [`pytest-textual-snapshot`](https://github.com/Textualize/pytest-textual-snapshot) to snapshot test your Textual app.
+This is a plugin for pytest which adds support for snapshot testing Textual apps, and it's maintained by the developers of Textual.
+
+A test using this package saves a snapshot (in this case, an SVG screenshot) of a running Textual app to disk.
+The next time the test runs, it takes another snapshot and compares it to the previously saved one.
+If the snapshots differ, the test fails, and you can view side-by-side diff showing the visual change.
```
rodrigogiraoserrao, line 188:
```suggestion
If the snapshots differ, the test fails, and you can view a side-by-side diff showing the visual change.
```
```suggestion
If the snapshots differ, the test fails, and you can view the side-by-side diff showing the visual change.
```

docs/guide/testing.md
```diff
@@ -162,7 +162,150 @@ You can generally solve this by calling [`pause()`][textual.pilot.Pilot.pause] w
...
+
+```
+pytest --snapshot-update
+```
+
+!!! warning
+
+    Only ever run pytest with `--snapshot-update` if you're happy with how the output looks
+    on the left hand side of the snapshot report. When using `--snapshot-update`, you're saying "I'm happy with this
+    screenshot, and it will now represent the ground truth which all future runs of this test should be compared against".
```
rodrigogiraoserrao, line 247: I think this is a key warning. Also, it may be worth putting extra emphasis on the fact that this will happen for _all_ snapshot tests that are already in place, so if any of those was wrong, it'll be updated in the wrong way.

docs/guide/testing.md
```diff
@@ -162,7 +162,150 @@ You can generally solve this by calling [`pause()`][textual.pilot.Pilot.pause] w
...
+You can simulate pressing keys before the snapshot is captured using the `press` parameter.
+
+```python
+def test_calculator_pressing_numbers(snap_compare):
+    assert snap_compare("path/to/calculator.py", press=["1", "2", "3"])
+```
+
+#### Changing the terminal size
+
+To capture the snapshot with a different terminal size, pass a tuple `(height, width)` as the `terminal_size` parameter.
```
rodrigogiraoserrao, line 289:
```suggestion
To capture the snapshot with a different terminal size, pass a tuple `(width, height)` as the `terminal_size` parameter.
```

rodrigogiraoserrao (changes requested):
Looking good! Just a tiny request.

docs/guide/testing.md
```diff
@@ -162,7 +162,150 @@ You can generally solve this by calling [`pause()`][textual.pilot.Pilot.pause] w
 You can also supply a `delay` parameter, which will insert a delay prior to waiting for pending messages.
 
 
-## Textual's test
+## Textual's tests
 
 Textual itself has a large battery of tests.
 If you are interested in how we write tests, see the [tests/](https://github.com/Textualize/textual/tree/main/tests) directory in the Textual repository.
+
+## Snapshot testing
```
willmcgugan, line 170: I dislike a heading followed by another heading. Could we remove the heading on L172 ?

willmcgugan (approved):
Just one request

docs/guide/testing.md
darrenburns: Agreed - done

rodrigogiraoserrao (approved):

docs/guide/testing.md
darrenburns: Good point - I've made that clearer

PR #3358, opened 2023-09-20
Scoped css

willmcgugan:
- Implements scoped CSS by default
- Adds :dark and :light psuedo classes


davep (approved):
Looking good to me. One typo spotted (and change added).

PR #3360, opened 2023-09-20
Pilot.click/hover methods restricted to the visible area

### LINKED ISSUE #3349: Restrict Pilot.click to visible area
I suspect that `Pilot.click` will allow clicking of widgets that may not be in the visible area of the screen.

Since we are simulating a real user operating the mouse, we should make that not work. Probably by throwing an exception.

It also looks like it is possible to supply an `offset` that would click outside of the visible area. This should probably also result in an exception.

rodrigogiraoserrao:
This fixes #3349.

The two methods now raise an `OutOfBounds` error if the click/hover offset target is outside of the visible screen and it returns a Boolean that determines whether the click/hover hit the correct widget or not.


src/textual/pilot.py
```diff
@@ -104,21 +112,41 @@ async def click(
...
+                f"Target size is {target_widget.size}, click offset is {offset}."
+            )
+
         message_arguments = _get_mouse_message_arguments(
             target_widget, offset, button=1, shift=shift, meta=meta, control=control
         )
+
+        click_offset = Offset(message_arguments["x"], message_arguments["y"])
+        visible_screen_region = screen.region + screen.scroll_offset
+        if not visible_screen_region.contains(*click_offset):
```
willmcgugan, line 126: I think this works:

```suggestion
        if click_offset not in visible_screen_region:
```

src/textual/pilot.py
```diff
@@ -104,21 +112,41 @@ async def click(
...
+            raise OutOfBounds(
+                "Target offset is outside of currently-visible"
+                + f"screen region {visible_screen_region}."
```
willmcgugan, line 129: Don't think we need the region repr in there. Can't see it being helpful.

src/textual/pilot.py
```diff
@@ -104,21 +112,41 @@ async def click(
         else:
             target_widget = screen
 
+        if not target_widget.size.contains(*offset):
```
willmcgugan, line 115: I don't understand this. The offset should be added to the click coordinate. 

It is not an error to have an offset that puts the click position outside of the selected widget. It is only out of bounds if the click is outside of the screen.

src/textual/pilot.py
rodrigogiraoserrao: Changed.

src/textual/pilot.py
rodrigogiraoserrao: Removed.

src/textual/pilot.py
rodrigogiraoserrao: Refactored after the back-and-forth in the original issue.

src/textual/pilot.py
```diff
@@ -107,27 +123,52 @@ async def click(
         message_arguments = _get_mouse_message_arguments(
             target_widget, offset, button=1, shift=shift, meta=meta, control=control
         )
+
+        click_offset = Offset(message_arguments["x"], message_arguments["y"])
+        visible_screen_region = screen.region + screen.scroll_offset
```
willmcgugan, line 128: This *feels* wrong... `click_offset` is relative to the Screen. i.e. (0, 0) is the top left of the screen regardless of the scroll position. No matter how much I scroll (0, 0) will never be out of bounds.

Can you confirm this?

rodrigogiraoserrao: The `click_offset` is an offset with respect to the top-left corner of the _virtual_ screen.

So, both `pilot.click(None, (0, 0))` and `pilot.click(Screen, (0, 0))` can be outside of the visible screen if you've scrolled down.

Do you think `pilot.click(None, offset)` should give an offset relative to the visible screen instead?

willmcgugan: If I do `pilot.click(offset=(0, 0))` I would expect that to mean the top left, regardless of the scroll position. Mouse coordinates come in that way.

rodrigogiraoserrao: Ok, but that's _not_ how it works currently.

Do you want me to change it, then?

rodrigogiraoserrao: >> Ok, but that's _not_ how it works currently.
> I would be very surprised if that is the case.

I guarantee it is not how it works currently.
<details>
<summary> Test it with this app in `main`. </summary>

```py
import asyncio

from textual.app import App


class MyApp(App):
    clicked = False

    def compose(self):
        yield Button()
        for idx in range(200):
            yield Label(str(idx), id=f"lbl{idx}")

    def on_button_pressed(self):
        self.clicked = True


async def main():
    app = MyApp()
    async with app.run_test(size=(80, 24)) as pilot:
        app.query_one("#lbl199").scroll_visible(animate=False)
        await pilot.pause()
        await pilot.click(offset=(0, 0))
        await pilot.pause()

    assert not app.clicked


if __name__ == "__main__":
    asyncio.run(main())

```

</details>

> Clicking (0, 0) should always be possible. Clicking (0, -1) should always be out of bounds.

I'll make it happen, then. `offset` with `selector=None` will be relative to visible screen. If any selector is provided, then the offset is relative to the top-left corner of the selected widget.

rodrigogiraoserrao: Something wrong is not right. I think the app above proves your point, not mine.
I'll figure out what I'm missing and I'll find out why the tests didn't catch it.

src/textual/pilot.py
```diff
@@ -139,10 +179,21 @@ async def hover(
         message_arguments = _get_mouse_message_arguments(
             target_widget, offset, button=0
         )
+
+        hover_offset = Offset(message_arguments["x"], message_arguments["y"])
+        visible_screen_region = screen.region + screen.scroll_offset
```
willmcgugan, line 184: I don't think this is correct. Which suggests the tests aren't doing their job.

rodrigogiraoserrao: Of course this isn't correct. It should match `Pilot.click`. See 7d11422.

src/textual/pilot.py
rodrigogiraoserrao: For the record, what I was missing was the correct interpretation of the screen coordinate system.
The `Screen` doesn't move and its top-left corner always matches the top-left corner of the terminal.

PR #3373, opened 2023-09-21
Added option to strip spaces with the Length validator

SkwalExe:

- [x] Updated documentation

### DIFF
--- a/src/textual/validation.py
+++ b/src/textual/validation.py
@@ -384,12 +384,15 @@ class Length(Validator):
         minimum: int | None = None,
         maximum: int | None = None,
         failure_description: str | None = None,
+        strip_spaces: bool = False,
     ) -> None:
         super().__init__(failure_description=failure_description)
         self.minimum = minimum
         """The inclusive minimum length of the value, or None if unbounded."""
         self.maximum = maximum
         """The inclusive maximum length of the value, or None if unbounded."""
+        self.strip_spaces = strip_spaces
+        """If true, then check the length while ignoring trailing and leading spaces"""
 
     class Incorrect(Failure):
         """Indicates a failure due to the length of the value being outside the range."""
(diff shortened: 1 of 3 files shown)


src/textual/validation.py
```diff
@@ -419,12 +423,19 @@ def describe_failure(self, failure: Failure) -> str | None:
...
-                return f"Must be shorter than {self.maximum} characters."
+                desc += f"Must be shorter than {self.maximum} characters."
             elif self.minimum is not None and self.maximum is None:
-                return f"Must be longer than {self.minimum} characters."
+                desc += f"Must be longer than {self.minimum} characters."
             else:
-                return f"Must be between {self.minimum} and {self.maximum} characters."
+                desc += f"Must be between {self.minimum} and {self.maximum} characters."
+
+            desc += (
```
willmcgugan, line 434: I think we could probably do without this messages. For non-engineers, you wouldn't typically count spaces anyway.

src/textual/validation.py
```diff
@@ -384,12 +384,15 @@ def __init__(
         minimum: int | None = None,
         maximum: int | None = None,
         failure_description: str | None = None,
+        strip_spaces: bool = False,
```
willmcgugan, line 387: I'm actually thinking this should default to `True`. I can't imagine why you would want to allow spaces.

willmcgugan (changes requested):

src/textual/validation.py
```diff
@@ -384,12 +384,15 @@ def __init__(
         minimum: int | None = None,
         maximum: int | None = None,
         failure_description: str | None = None,
+        strip_spaces: bool = True,
```
willmcgugan, line 387: Calling `strip` can potentially strip more than spaces. can we rename this to `strip_whitespace`?

willmcgugan (changes requested):

PR #3399, opened 2023-09-25
Add countdown to showing command palette `no matches` message

### LINKED ISSUE #3278: Command palette flicker when there are no results
When there are no results, each keypress produces a flicker where the "No matches found" message is shown and hidden. I think we should be able to avoid this, so the "No matches found" message persists.

https://github.com/Textualize/textual/assets/554369/98e93373-e7e1-4f95-84e7-0f193684cc7c

guy-av:
The countdown protects from spamming the message and helping the palette feel smoother when there are no matches.

- [x] Change `test_no_results` to wait for the countdown

Resolves #3278

### DIFF
--- a/src/textual/command.py
+++ b/src/textual/command.py
@@ -458,6 +458,8 @@ class CommandPalette(ModalScreen[CallbackType], inherit_css=False):
         """The command that was selected by the user."""
         self._busy_timer: Timer | None = None
         """Keeps track of if there's a busy indication timer in effect."""
+        self._no_matches_timer: Timer | None = None
+        """Keeps track of if there are 'No matches found' message waiting to be displayed."""
         self._providers: list[Provider] = []
         """List of Provider instances involved in searches."""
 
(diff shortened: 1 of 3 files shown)


guy-av:
This PR tries to follow suggested changes noted in [this comment](https://github.com/Textualize/textual/issues/3278#issuecomment-1713875625).

src/textual/command.py
```diff
@@ -559,6 +561,38 @@ def _become_busy() -> None:
 
         self._busy_timer = self.set_timer(self._BUSY_COUNTDOWN, _become_busy)
 
+    def _stop_no_matches_countdown(self) -> None:
```
davep, line 564: Could we get a docstring for this please?

src/textual/command.py
```diff
@@ -559,6 +561,38 @@ def _become_busy() -> None:
...
+        if self._no_matches_timer is not None:
+            self._no_matches_timer.stop()
+            self._no_matches_timer = None
+
+    _NO_MATCHES_COUNTDOWN: Final[float] = 0.5
+    """How many seconds to wait before showing 'No matches found'"""
```
davep, line 570:
```suggestion
    """How many seconds to wait before showing 'No matches found'."""
```

src/textual/command.py
```diff
@@ -559,6 +561,38 @@ def _become_busy() -> None:
...
+
+    def _start_no_matches_countdown(self, command_list: CommandList) -> None:
```
davep, line 572: What's the thinking about having the `CommandList` instance be an argument here? The `CommandList` that is part of the command palette should be queryable.

davep (commented):
Looking good; thanks. Couple of very small tweaks and one question.

Also, very small general observation (which really isn't a problem, *very* much appreciate the contribution, but worth a mention in regards to FOSS contributions in general): this issue was already assigned so it would be nice to check with the assignee first if it's okay to jump in.

guy-av:
> Looking good; thanks. Couple of very small tweaks and one question.
> 
> 
> 
> Also, very small general observation (which really isn't a problem, *very* much appreciate the contribution, but worth a mention in regards to FOSS contributions in general): this issue was already assigned so it would be nice to check with the assignee first if it's okay to jump in.

Sure thing!

src/textual/command.py
guy-av: You're right!
Nice catch

guy-av: The argument has been removed and the `CommandList` is queried instead

davep (approved):
Looks good to me! Thanks again.

I've also added an entry to the CHANGELOG.

PR #3406, opened 2023-09-26
Minor refactoring

rndusr:
This doesn't actually fix or change anything (maybe some unnoticable performance improvements), but it makes `XTermParser._sequence_to_key_events()` easier to reason about.


rndusr:
Included in https://github.com/Textualize/textual/pull/3406.

PR #3409, opened 2023-09-27
fix Tree(disabled=True) breaking app

### LINKED ISSUE #3407: Tree `disabled=True` breaks when setting at initialization
Using the `Tree` example from the docs, if you pass in `disabled=True` during initialization the app break with 2 exceptions.

**Thrown Exceptions:**
1. `AttributeError: 'Tree' object has no attribute '_line_cache'`
2. `ScreenStackError: No screens on stack`

I can see in the source that the call to `super().__init__()` on the base `Widget` object is called before setting the `_line_cache` so I'm not sure if that is the problem or not.

```python
from textual.widgets import Tree


class TreeApp(App):
        tree: Tree[dict] = Tree("Dune", disabled=True)
        tree.root.expand()
        characters = tree.root.add("Characters", expand=True)
        characters.add_leaf("Paul")
        characters.add_leaf("Jessica")
        characters.add_leaf("Chani")
        yield tree


if __name__ == "__main__":
    app = TreeApp()
    app.run()
```

Enabling and disabling the Tree after the app is composed works just fine as can be seen using this example (using the 'D' keybinding).

```python
from textual.widgets import Footer, Header, Tree


class TreeApp(App):
    BINDINGS = [
        ("d", "disable_tree", "Toggle Tree Availability"),
    ]

        tree: Tree[dict] = Tree("Dune")
        tree.root.expand()
        characters = tree.root.add("Characters", expand=True)
        characters.add_leaf("Paul")
        characters.add_leaf("Jessica")
        characters.add_leaf("Chani")
        yield Header()
        yield Footer()
        yield tree

    def action_disable_tree(self) -> None:
        tree = self.query_one(Tree)
        tree.disabled = not tree.disabled


if __name__ == "__main__":
    app = TreeApp()
    app.run()
```

joshbduncan:
Fixes #3407 where `Tree` widget initialized/mounted with `disabled=True` would break its parent app

- [x] Updated documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 
 - `Pilot.click`/`Pilot.hover` can't use `Screen` as a selector https://github.com/Textualize/textual/issues/3395
+- App exception when a `Tree` is initialized/mounted with `disabled=True` https://github.com/Textualize/textual/issues/3407
 
 ### Added
 
--- a/src/textual/widgets/_tree.py
+++ b/src/textual/widgets/_tree.py
@@ -597,8 +597,6 @@ class Tree(Generic[TreeDataType], ScrollView, can_focus=True):
             disabled: Whether the tree is disabled or not.
         """
 
-        super().__init__(name=name, id=id, classes=classes, disabled=disabled)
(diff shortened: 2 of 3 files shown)


davep (approved):
Looks good; thanks for reporting and fixing!

PR #3410, opened 2023-09-27
Add Document get_index_from_location / get_location_from_index

davidbrochart:

- [ ] Updated documentation

### DIFF
--- a/src/textual/document/_document.py
+++ b/src/textual/document/_document.py
@@ -91,6 +91,28 @@ class DocumentBase(ABC):
         """Return the line separator used in the document."""
 
     @abstractmethod
+    def get_index_from_location(self, location: Location) -> int:
+        """Given a location, returns the index from the document's text.
+
+        Args:
+            location: The location in the document.
+
+        Returns:
+            The index in the document's text.
+        """
+
+    @abstractmethod
+    def get_location_from_index(self, index: int) -> Location:
+        """Given an index in the document's text, returns the corresponding location.
+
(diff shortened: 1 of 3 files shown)


davidbrochart:
I think these methods could be useful when some document changes come from a system which is not `Location`-aware, but only reports changes pointing to raw indexes in the document content, like in collaborative editing.

willmcgugan:
@darrenburns Could you have a look at this one pls.

src/textual/document/_document.py
```diff
@@ -320,6 +342,40 @@ def line_count(self) -> int:
         """Returns the number of lines in the document."""
         return len(self._lines)
 
+    def get_index_from_location(self, location: Location) -> int:
```
darrenburns, line 345: I can see why both of these would be useful. Would you mind writing a few tests for them?

Also, we have a policy of avoiding shorthand variable names where possible, so if you wouldn't mind renaming `col` -> `column`, `i` -> `line_index`, etc. that'd be appreciated too.

darrenburns (changes requested):

davidbrochart:
Thanks for the review @darrenburns, I made the changes, rebased and added some tests.

darrenburns:
Thanks @davidbrochart, could you also update the changelog?

davidbrochart:
Done.

darrenburns (approved):

darrenburns:
Thanks!

Issue #3304, opened 2023-09-14
Re-render options in OptionList


willmcgugan:
The OptionList will render all the options, and then apply styles to them post render to indicate a highlight.

The problem with this approach is that alpha / dim etc are not taken in to account. The result is that text with alpha, or the dim style, don't apply to the new highlight background.

I think we can re-render highlighted items so that the alpha applies.

willmcgugan:
This will require a fairly substantial refactor.

Essentially the approach of pre-rendering content, and then applying styles to get modifiers like select / hover, isn't going to take alpha in to account.

We need to re-render options with different component classes.

This may be one for @davep since he built the original.

rodrigogiraoserrao:
I can do it, no problem.
But if everyone thinks it's better, I'll yield to Dave.

rodrigogiraoserrao:
TL;DR: This will _likely_ be fixed if [Textual#3342](https://github.com/Textualize/textual/issues/3342) is fixed first.

willmcgugan:
Recently fixed

Issue #3342, opened 2023-09-18
(Text) Opacity isn't read from component classes styles.


rodrigogiraoserrao:
(TL;DR: This is _likely_ the cause of [Textual#3304](https://github.com/Textualize/textual/issues/3304) and this will _likely_ be fixed if [Textual#3413](https://github.com/Textualize/textual/issues/3413) is fixed first.)

Following #3304, an investigation showed that there are widgets for which the (text-)opacity in component classes is ignored.

Here are said widgets:
    - `Checkbox` (`ToggleButton`'s fault)
    - `DirectoryTree`
    - `Input`
    - `OptionList` (as first reported in [Textual#3304](https://github.com/Textualize/textual/issues/3304))
    - `RadioButton` (`ToggleButton`'s fault)
    - `Tree`

Below you can find an app that showcases these:

<details>
<summary>myapp.py</summary>

```py
from textual.app import App, ComposeResult
from textual.widgets import (
    Checkbox,
    RadioButton,
    DirectoryTree,
    Input,
    Switch,
    Tree,
)


class CCApp(App[None]):
    CSS_PATH = "myapp.tcss"

    def compose(self) -> ComposeResult:
        yield Checkbox("checkbox")
        yield DirectoryTree(".")
        yield Input(placeholder="placeholder text")
        yield RadioButton("radio button")
        yield Switch()
        tree = Tree("Dune", id="tree")
        tree.root.expand()
        characters = tree.root.add("Characters", expand=True)
        characters.add_leaf("Paul")
        characters.add_leaf("Jessica")
        characters.add_leaf("Chani")
        yield tree


if __name__ == "__main__":
    CCApp().run()
```

</details>

<details>
<summary>myapp.tcss</summary>

```tcss
Checkbox > .toggle--label {
    opacity: 0%;
    text-opacity: 0%;
    color: #ffffff00;  # alpha also ignored here.
}

DirectoryTree > .directory-tree--file {
    text-opacity: 0%;
    opacity: 0%;
    color: #ffffff00;  # alpha ignored here.
}

Input > .input--placeholder {
    text-opacity: 0%;
    opacity: 0%;  # opacity of background is respected.
    background: red;  # invisible because opacity is 0%.
    # color: #ffffff00;  # alpha in color is respected.
}

RadioButton > .toggle--label {
    text-opacity: 0%;
    opacity: 0%;
    background: #00000000;  # alpha also ignored here.
    color: #ffffff00;  # alpha also ignored here.
}

[...]

PR #3415, opened 2023-09-27
Fix issues around component classes / (text) opacity

### LINKED ISSUE #3413: Rich styles computation for dom nodes doesn't take text-opacity into account
`DOMNode.rich_style` does not take into account the text opacity when computing the foreground colour.

This will _likely_ fix [Textual#3342](https://github.com/Textualize/textual/issues/3342), which in turn will _likely_ fix [Textual#3304](https://github.com/Textualize/textual/issues/3304).

rodrigogiraoserrao:
This PR will fix #3413, #3342, and #3304.

This started out as a PR that intended to fix an issue with the option list and kept growing in size as I kept uncovering more issues with component classes & styles, (partial) rich styles, and interactions with (text-)opacity styles.

I've stopped digging because there's some things with component classes we'd like to iron out and those will probably help with what's not handled in this PR yet.

Some things that were left to be done/fixed/considered:
 - #3421 
 - #3467


rodrigogiraoserrao:
de6a479 introduces a snapshot test app that you can run with `textual run tests/snapshot_tests/snapshot_apps/component_classes_opacity.py`.

The “empty boxes” are a good thing because I'm using component classes to set colours to transparent.
The screenshot shows that the only widgets that aren't fixed are `DirectoryTree`, `Tree`, `Checkbox`, and `RadioButton` (i.e., we need to fix `Tree` and `ToggleButton`).

Screenshot 2023-09-27 at 17 43 13

The screenshot shows that the previous commits fix #3304.

rodrigogiraoserrao:
The reason the (directory) tree and toggle button labels are still white is because they use partial rich styles to enable markup to be used in the labels, which don't currently take into account text opacity.

To know how to resolve this, we need to know what should happen if a widget has mark up and conflicting TCSS.
For example, what should the checkbox `Checkbox("[red bold]This is just[/] some text.")` look like if the following TCSS is specified:

```sass
Checkbox > .toggle--label {
    color: blue;
    text-opacity: 50%;
}
```

1. The red/bold remains, but faded 50% and the “some text” is blue 50%
2. The red/bold remains and not faded, “some text” is blue 50%
3. The bold remains, the whole text is blue 50%
4. The TCSS overrides the markup completely and everything is just blue 50%
5. The existence of markup overrides the TCSS completely and the TCSS is essentially ignored
6. Some other combination I'm not considering 

After I have the answer to this question, I can get to fixing the code so that it complies with what we decide.

src/textual/widgets/_tabs.py
```diff
@@ -107,7 +107,7 @@ class Tab(Static):
     }
     Tab:disabled {
         color: $text-disabled;
-        text-opacity: 50%;
+        text-opacity: 70%;
```
rodrigogiraoserrao, line 110: This accounts for the now-correct calculation of the final text color when opacity is present.

willmcgugan:
I have an alternative solution in mind for `_redirect_component_styles`. Solution is good, but I want to make this change as part of a larger update.

PR #3432, opened 2023-09-29
fix: allow integer for fractional styles

### LINKED ISSUE #3414: Fractional styles raise exception when set to integers
Although technically correct (the best kind of correct), it's quite awkward that you can't do something like

```py
some_widget.styles.text_opacity = 0
```

This raises an exception because `0` is not a float.
Text opacity – and the other related styles – should be made to work with integers, too.

TomJGooding:
Fixes #3414.

- [ ] Updated documentation


willmcgugan (approved):
THanks, Tom.

TomJGooding:
@willmcgugan I've resolved the changelog conflicts again, if you're happy to merge this?

willmcgugan:
There are some tests that are particularly sensitive to timing. In the CI the tests run much slower than a regular app, and it tends to highlight these issues. We're going to revisit that soon, because it is a pain.

PR #3438, opened 2023-10-02
Detect non widgets

holdenweb:

- [x] Updated documentation

I've been dogged by an `AttributeError` that occurs when trying to mount non-widgets (either directly, or because I've omitted the call to `Widget__init__()`), that points out a particular item has no `id`, so I added a message that should catch both cases. You may find better wording for the message.

Impact on speed should be zero, as the code only executes of a fatal error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
-- Widgets now complain on attempts to mount a non-widget.
+- Attribute errors are suppressed when mounting widgets to improve reporting 
 - `Pilot.click`/`Pilot.hover` now raises `OutOfBounds` when clicking outside visible screen https://github.com/Textualize/textual/pull/3360
 - `Pilot.click`/`Pilot.hover` now return a Boolean indicating whether the click/hover landed on the widget that matches the selector https://github.com/Textualize/textual/pull/3360
 - Added a delay to when the `No Matches` message appears in the command palette, thus removing a flicker https://github.com/Textualize/textual/pull/3399
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
+- Widgets now complain on attempts to mount a non-widget.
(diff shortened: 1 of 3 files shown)


davep:
There is a later check for the object being a widget, which will raise an `AppError`. For example, this:

```python

@dataclass
class NotAWidget:
    id: str = ""

class MountNonWidgetApp(App[None]):

        yield NotAWidget()

if __name__ == "__main__":
    MountNonWidgetApp().run()
```

results in this:

```
AppError: Can't register NotAWidget(id=''); expected a Widget instance
```

I sense the error raised here should match. Moreover, the existing check that the object is a widget could be used anyway, and so potentially the change could be:

```python
        try:
            ids_to_mount = [widget.id for widget in widgets if widget.id is not None]
        except AttributeError:
            ids_to_mount = []
```

In other words the lack of an ID would be a pass but then the "not a `Widget` instance" logic would catch this anyway.

However, neither of these approaches would seem to clarify the situation of failing to call `super().__init__()` in an actual `Widget` instance as the `id` property would be missing while at the same time it could be an instance of a widget that failed to have it.

Perhaps the ultimate improvement here would be in `App._register`, along with the check that the widget is a `Widget` instance, it also then goes on to check if the `Widget` has an `id` property and raises an `AppError` suggesting `__init__()` may have been overridden while `super().__init__()` might not have been called.

holdenweb:
I've persisted with updating the PR, but it sounds as though your greater intimacy with the source has come up with more appropriate actions. Would you like me to take a look to see what the necessary changes to `App._register` might look like?

holdenweb:
Hmmm ... it seems that Widget subclasses _are_ assigned an `id` somewhere. Perhaps the "no id" error was caused by passing a class instead of an instance of the class, probably a less frequent occurrence. Because subclasses aren't instances this case is already caught by the existing `_register()` code and so the only change required would seem to be the suppression of the `AttributeError` as you suggested above. The message could, perhaps, be improved, and perhaps the code could raise a `WidgetError` or `TypeError` rather than `AppError`, but the essential change is done.

willmcgugan:
Thanks @holdenweb This was a good opportunity to improve DX.

Closing in favour of https://github.com/Textualize/textual/pull/3780

PR #3441, opened 2023-10-02
Prevent stdout/stderr capture on headless mode

darrenburns:
Debugging tests got much harder since recent stdout/stderr capturing changes, since we can no longer print inside them. The prints get captured by Textual, meaning they don't make it as far as the pytest output.

I think when running apps headless, we want to allow prints to make it through, but I may be missing broader context on the print capturing changes that went in.

I've confirmed allows the `print` output to make it to the pytest output as expected.

@willmcgugan thoughts on this?

Edit - a test is failing but that's expected. I won't look to resolve that test until we agree on direction for this.


willmcgugan:
I think not capturing in headless mode probably should be the default. But we do have a mechaniam now where the dev can capture prints and decide what to do with them, so I feel there should also be an option to do a headless run with capturing enabled.

darrenburns:
Should adding a mechanism to toggle capturing in headless mode be done in this PR?

willmcgugan:
May as well. Feels like a small change.

PR #3443, opened 2023-10-02
Data table cell padding

### LINKED ISSUE #3435: Add ability to remove cell padding
I haven't been able to modify the style of a DataTable so that there is zero space between the columns. This in contrast to the rows, which are stacked with zero space between the cells. 

I've tried setting `margin`, `grid-gutter` and `padding` to zero in the TCSS file, with no success: 

```
DataTable {
	grid-gutter: 0;
	padding: 0;
	margin: 0;
} 
```

rodrigogiraoserrao:

- [x] Updated documentation

This fixes #3435.


src/textual/widgets/_data_table.py
```diff
@@ -170,14 +171,18 @@ class Column:
...
 
-    @property
-    def render_width(self) -> int:
-        """Width in cells, required to render a column."""
-        # +2 is to account for space padding either side of the cell
-        if self.auto_width:
-            return self.content_width + CELL_X_PADDING
-        else:
-            return self.width + CELL_X_PADDING
+    def render_width(self, data_table: DataTable[Any]) -> int:
```
willmcgugan, line 174: Since this is no longer a property, it should be named "<VERB>_<NOUN>".  Suggest "get_render_width".

src/textual/widgets/_data_table.py
```diff
@@ -309,6 +314,8 @@ class DataTable(ScrollView, Generic[CellType], can_focus=True):
     show_cursor = Reactive(True)
     cursor_type: Reactive[CursorType] = Reactive[CursorType]("cell")
     """The type of the cursor of the `DataTable`."""
+    cell_padding = Reactive(_DEFAULT_CELL_X_PADDING)
```
willmcgugan, line 317: Will this re-render the table if set after the first render?

rodrigogiraoserrao: It does.

src/textual/widgets/_data_table.py
rodrigogiraoserrao: That's reasonable. See b84334c.

src/textual/widgets/_data_table.py
willmcgugan: Are you absolutely certain of this? Like if you were John McClane and Hans Gruber was pointing a gun to your head, and a wrong answer would lead to Nakatomi Plaza blowing up and killing all the hostages. Would this still be your answer?

rodrigogiraoserrao: I don't want to blow Nakatomi Plaza up. I had tested assigning to `cell_padding` while an app was running and the data table refreshed correctly.
However, I tried hooking up an `Input` to the cell padding and if I make more than one modification, it wasn't.

This was tested and fixed in bbe5c85.

rodrigogiraoserrao: Also, I don't get Die Hard references.

Teletubbies references are more likely the way to go.

giphy

rodrigogiraoserrao: Thanks for your input! See 1936f10.

willmcgugan (approved):

Issue #3141, opened 2023-08-22
Revisit the removal of "protocol" methods from the docs


davep:
While on the whole #3076 made perfect sense, there are some situations where we would want to keep the "protocol" methods in the docs. An example would be `textual.app.App.compose`; the idea here being that when such a method is introduced, it should go into the documentation; only the overrides in child classes need to be excluded.

Not keeping the implementation that introduces the method means the method doesn't get documented at all, and so then results in things like this in the docs:

<img width="833" alt="Screenshot 2023-08-22 at 14 19 37" src="https://github.com/Textualize/textual/assets/28237/4d4e4b85-a129-4221-acc9-79e1eb516721">

(note how `compose()` [isn't a link any more](https://textual.textualize.io/guide/app/#composing))

More generally: we seem to have acquired a lot more warnings and similar things during docs generation; we should perhaps look to them to see if they're related to this change.

rodrigogiraoserrao:
The increased number of things in the log are a bunch of INFO logs that we get as of a more recent mkdocs version (or a more recent version of one of the plugins) where we get warned about absolute links that the logs recommend be made relative, so it's unrelated to the “protocol” changes.

PR #3445, opened 2023-10-02
Fix styling in app.md

joehakimrahme:
Backquotes within the hyperlink syntax seems to break something in https://textual.textualize.io/guide/app/#composing

### DIFF
--- a/docs/blog/posts/textual-plotext.md
+++ b/docs/blog/posts/textual-plotext.md
@@ -102,7 +102,7 @@ it](https://github.com/Textualize/textual-plotext/blob/main/examples/textual_tow
 ![The Textual Towers weather history app](/blog/images/textual-plotext/weather.png)
 
 If you are an existing Plotext user who wants to turn your plots into full
-terminal applications, we think this will be a very familiar and accessible.
+terminal applications, we think this will be very familiar and accessible.
 If you're a Textual user who wants to add plots to your application, we
 think Plotext is a great library for this.
 


TomJGooding:
[Did you preview the docs after these changes?](https://github.com/Textualize/textual/blob/main/CONTRIBUTING.md#updating-and-building-the-documentation) I don't think the backticks are the issue - see #3141.

willmcgugan:
Yeah, I think its that the import path is wrong there.

rodrigogiraoserrao:
Hey @joehakimrahme , thanks for the PR.
Like Tom said, the backticks aren't the problem and they should stay!

Interestingly enough, both links inside the square brackets are correct.
The reason the one with `compose` isn't working is because we configured the documentation to ignore the method `compose` everywhere:

https://github.com/Textualize/textual/blob/03e3a6956182e441b6c3db1ca6f50679ae508163/mkdocs-common.yml#L85

However, `App.compose` should show in the documentation.

To do this, take a look at the `filters` that is linked above and see the `docs/widget.md` to see how we re-define `filters` so that [the `Widget` API page](https://textual.textualize.io/api/widget/) contains methods like `compose` and `render_line`.

After you make the needed change(s), make sure you build the docs locally (see Tom's link)!

Thanks 🤗

joehakimrahme:
Apologies, I close this PR by mistake. I opened #3461 to follow.

@rodrigogiraoserrao thanks so much for this, I don't think I'd ever been able to figure this one out on my own.

Thanks also Will and Tom for the pointers :)

PR #3446, opened 2023-10-03
Remove an unnecessary a

davep:
Saw this in passing.


rodrigogiraoserrao (approved):
I think the implementation is suboptimal but I'll let it slide this time. 🙄

davep:
True, I'm very much sure there there are at least a couple or so more "unnecessary words" I could have remove from the prose to make it much less verbose.

PR #3447, opened 2023-10-03
Improve typing for queries.

rodrigogiraoserrao:
This PR makes it so that things like `reveal_type(app.query(Button).first())` correctly report `Button`, instead of `Widget`.

### DIFF
--- a/src/textual/css/query.py
+++ b/src/textual/css/query.py
@@ -48,7 +48,10 @@ class WrongType(QueryError):
     """Query result was not of the correct type."""
 
 
-QueryType = TypeVar("QueryType", bound="Widget")
+QueryType = TypeVar("QueryType", bound=Widget)
+"""Type variable used to type generic queries."""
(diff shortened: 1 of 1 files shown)


davep (commented):
Might it be worth adding a test or two to ensure this all works exactly as intended, and no future changes mess this up?

davep:
If we're cool that that's all it needs, that's fine with me; just wanted to be sure there wasn't any scope here for a test or two where there might not have been one before.

rodrigogiraoserrao:
> If we're cool that that's all it needs, that's fine with me; just wanted to be sure there wasn't any scope here for a test or two where there might not have been one before.

As of #3450, `query.py` is at 100% code coverage, so the functions whose typing I changed are being exercised and the typechecker was happy with my changes, so I am happy with this.
But it may be that you are thinking of something I'm not, so if there's something like that, please let me know!

willmcgugan (approved):
Nice 👍

PR #3459, opened 2023-10-04
`OptionList` safe duplicate check

### LINKED ISSUE #3455: `OptionList` duplicate ID detection not working as expected
Consider the following code:

```python
from textual.widgets import OptionList, Log, Footer
from textual.widgets.option_list import Option

class OptionListAddBug(App[None]):

    BINDINGS = [
        (str(n), f"add({n})", f"Add {n}") for n in range(10)
    ]

    CSS = """
    OptionList, Log {
        height: 1fr;
        border: round cornflowerblue;
    }
    """

        yield OptionList()
        yield Log()
        yield Footer()

    def action_add(self, value: int) -> None:
        try:
                Option(f"Added {value} just fine", id=f"id-{value}")
            )
        except Exception as error:
            self.query_one(Log).write_line(f"{error!r}")


if __name__ == "__main__":
    OptionListAddBug().run()
```

If you run it and press keys 0-9, you'll find that corresponding options are added to the `OptionList`.

Screenshot 2023-10-04 at 15 26 43

However, if you start the app afresh, then press (for example) <kbd>2</kbd> twice, resulting in a logged `DuplicateID` exception (expected and desired), after than you can't add any more options to the list, and the attempted added option is always reported as having the original duplicate ID:

Screenshot 2023-10-04 at 15 28 53

In the above I pressed <kbd>2</kbd> twice and then a number of other number keys.

It would appear that `OptionList` is getting every confused in its `id` tracking and/or duplicate detection.

davep:
Simply put: the duplicate check inside `OptionList` was happening *after* a new option was added to the internal structures. This was almost fine back when there was only a method for adding a single option, and even then it was only fine if a `DuplicateID` exception was treated as a "just give up using this `OptionList` signal". Or put another way: attempting to add a duplicate ID had unintended side-effects, making every subsequent attempt to add an option appear to be the creation of a duplicate.

This PR moves to a preflight check for duplicate IDs.

Fixes #3455.

### DIFF
--- a/src/textual/widgets/_option_list.py
+++ b/src/textual/widgets/_option_list.py
@@ -7,6 +7,7 @@ forms of bounce-bar menu.
 
 from __future__ import annotations
 
+from collections import Counter
 from typing import ClassVar, Iterable, NamedTuple
 
 from rich.console import RenderableType
@@ -540,10 +541,6 @@ class OptionList(ScrollView, can_focus=True):
                 if content.id is not None:
                     # The option has an ID set, create a mapping from that
                     # ID to the option so we can use it later.
-                    if content.id in option_ids:
-                        raise DuplicateID(
-                            f"The option list already has an option with id '{content.id}'"
-                        )
                     option_ids[content.id] = option
                 option += 1
             else:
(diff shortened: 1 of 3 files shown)


src/textual/widgets/_option_list.py
```diff
@@ -558,6 +555,38 @@ def _refresh_content_tracking(self, force: bool = False) -> None:
...
+            item_id for item_id in new_option_ids if item_id in current_ids
+        }.union(
+            item_id
+            for item_id, id_count in Counter(new_option_ids).items()
+            if id_count > 1
+        )
+        # If there are any duplicates...
+        if duplicates:
+            # ...complain.
+            raise DuplicateID(
```
TomJGooding, line 586: Sorry to nit, but I think the code is self-evident without the comments?

src/textual/widgets/_option_list.py
```diff
@@ -558,6 +555,38 @@ def _refresh_content_tracking(self, force: bool = False) -> None:
...
+                f"Operation would result in duplicate ids {', '.join(duplicates)}"
```
TomJGooding, line 587: "Operation would result in" perhaps could be worded to be more user-friendly?

davep: Go for it.

src/textual/widgets/_option_list.py
davep: Eh, I don't much care for commented code where the comments fizzle out. Feel free to PR to the PR to taste though.

TomJGooding: Perhaps I should just stay in my lane and not comment on PR's in future. 

I'm just a minor contributor so obviously you can take or leave my suggestions.

davep: Nah, most welcome. I actually tried to add you to the reviewer list on this one when I opened the PR but apparently GitHub wasn't having it. 🤷🏻‍♀️

src/textual/widgets/_option_list.py
```diff
@@ -558,6 +554,30 @@ def _refresh_content_tracking(self, force: bool = False) -> None:
...
+            DuplicateID: If there is an attempt to use a duplicate ID.
+        """
+        # We're only interested in options, and only those that have IDs.
+        new_options = [
+            item
+            for item in candidate_items
+            if isinstance(item, Option) and item.id is not None
+        ]
+        # Get the set of new IDs that we're being given.
+        new_option_ids = set(option.id for option in new_options)
```
willmcgugan, line 573: Could use a set exception here.

willmcgugan (approved):
Just a recomendation.

PR #3475, opened 2023-10-07
Link to the repo rather than the pypi page

davep:
See https://github.com/Textualize/textual-plotext/issues/1


willmcgugan (approved):
Cheers

PR #3476, opened 2023-10-08
fix doc: change `rule:` to `text-align:` in css_types

laixintao:
I think that's an mistake in the doc, because the `rule` actually will throw an Exception. `Invalid CSS property 'rule'`.

- [x] Updated documentation


willmcgugan:
Thanks.

PR #3477, opened 2023-10-08
Add wait_for_dismiss to push_screen

willmcgugan:
An addition to the screens API to avoid callbacks. Adds a `wait_for_dismiss` parameter which waits for the new screen to be dismissed.

It works like this:

```python
result = await self.push_screen(MyScreen(), wait_for_dismiss=True)
```

This only works from within a worker. If not run within a worker it will raise a `NoActiveWorker` error.

Without a worker, waiting on the result would cause a deadlock. But at least we can identify that and raise a helpful error.

This allows for more natural code than callbacks. For example, here's how you might use a screen to ask the user if they want to quit:

```python
@work
async def action_close(self) -> None:
    if await self.push_screen(QuitScreen(), wait_for_dismiss=True):
        self.app.quit()
```

The `push_screen` has overloaded signatures, so the type checker knows that there will only be a return value when `wait_for_dismiss` is True, and it knows the type of the return -- which is nice!

### DIFF
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -247,25 +247,6 @@ class _PrintCapture:
         return -1
 
 
-class ScreenAwaitable(Generic[ScreenResultType]):
-    """An optional awaitable to get the result of a screen."""
-
-    def __init__(
-        self, await_mount: AwaitMount, future: asyncio.Future[ScreenResultType]
-    ) -> None:
-        self._await_mount = await_mount
-        self._future = future
-
-    def __await__(self) -> Generator[None, None, ScreenResultType]:
-        async def await_screen() -> ScreenResultType:
-            """Await the mount and the future."""
-            await self._await_mount
-            await self._future
-            return self._future.result()
-
-        return await_screen().__await__()
-
-
 @rich.repr.auto
 class App(Generic[ReturnType], DOMNode):
     """The base class for Textual Applications."""
@@ -1810,7 +1791,7 @@ class App(Generic[ReturnType], DOMNode):
         screen: Screen[ScreenResultType] | str,
         callback: ScreenResultCallbackType[ScreenResultType] | None = None,
         wait_for_dismiss: Literal[True] = True,
-    ) -> ScreenAwaitable[ScreenResultType]:
+    ) -> asyncio.Future[ScreenResultType]:
         ...
 
     def push_screen(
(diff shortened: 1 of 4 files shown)


src/textual/app.py
```diff
@@ -1778,25 +1776,58 @@ def _replace_screen(self, screen: Screen) -> Screen:
...
+        ...
+
+    @overload
+    def push_screen(
+        self,
+        screen: Screen[ScreenResultType] | str,
+        callback: ScreenResultCallbackType[ScreenResultType] | None = None,
+        wait_for_dismiss: Literal[True] = True,
+    ) -> asyncio.Future[ScreenResultType]:
+        ...
```
rodrigogiraoserrao, line 1795: Nice.

tests/test_screens.py
```diff
@@ -407,4 +409,71 @@ def on_mount(self):
...
+
+    app = ScreensApp()
+    # Press X to exit, then Y to dismiss, expect True result
+    async with app.run_test() as pilot:
+        await pilot.press("x", "y")
+    assert results == [True]
+
+    results.clear()
+    app = ScreensApp()
+    # Press X to exit, then Y to dismiss, expect False result
```
rodrigogiraoserrao, line 448:
```suggestion
    # Press X to exit, then N to dismiss, expect False result
```

src/textual/app.py
```diff
@@ -1805,13 +1836,23 @@ def push_screen(
...
+        next_screen._push_result_callback(message_pump, callback, future)
         self._load_screen_css(next_screen)
         self._screen_stack.append(next_screen)
         self.stylesheet.update(next_screen)
         next_screen.post_message(events.ScreenResume())
         self.log.system(f"{self.screen} is current (PUSHED)")
-        return await_mount
+        if wait_for_dismiss:
+            try:
+                get_current_worker()
```
rodrigogiraoserrao, line 1848: The only point of this is to make sure we're running from within a worker, right?

rodrigogiraoserrao (approved):

src/textual/app.py
willmcgugan: Correct

mistotebe:
Isn't it better (definitely cleaner) to just add a new method, so you don't have to add `wait_for_dismiss`? Also no need to keep `callback` in that case...

willmcgugan:
@mistotebe Nothing prevents us for doing both. But awaiting the screen is more verbose.

```python
screen = QuitScreen()
await self.push_screen(screen)
result = await screen.wait_for_dismiss()
```

Compared to:

```python
result = await self.push_screen(QuitScreen(), wait_for_dismiss=True)
```

And callbacks still have their uses.

mistotebe:
> @mistotebe Nothing prevents us for doing both. But awaiting the screen is more verbose.
> 
> ```python
> screen = QuitScreen()
> await self.push_screen(screen)
> result = await screen.wait_for_dismiss()
> ```

If such a method existed, yes, but it appears to be missing from the PR? And sounds race-prone unless implemented correctly.

My instinct would have been to add something like this instead:

```python

async def push_modal_screen+bikeshed(self, screen: ModalScreen[ScreenResultType]) -> ScreenResultType:
    ...
```

PR #3488, opened 2023-10-09
Allow scrollbar size of zero (scrollable containers without scrollbars)

darrenburns:
Scrollbar size can be set to zero for a scrollable container with no visible scrollbars.


rodrigogiraoserrao (commented):
The code changes look good but I'd expect a snapshot test and a test that makes sure scrolling is actually possible.
(These two could be the same.)

willmcgugan (approved):

darrenburns:
@rodrigogiraoserrao gonna leave the additional tests for now and merge this to get it into the release we're doing now.

Issue #3484, opened 2023-10-09
Revisit flaky tests


willmcgugan:
We have some flaky tests. It seems to be a handful that are particularly sensitive to timing. I suspect that the issue lies in the tests themselves, and there is some genuine issue there. Let's look at those tests and see if we can make them stable.

rodrigogiraoserrao:
Tests I often find flaky:
 - a test related to the `textual borders` preview (possibly related to #3437)
 - a test that checks tabs messages

darrenburns:
@rodrigogiraoserrao Thanks for doing this summary - it's really helpful!

Issue #2914, opened 2023-07-11
Review the order in which a `Mount` event is received and `render` is called


davep:
Following on from #2912, we should look to see if we can guarantee that any freshly-mounted widget receives its `Mount` event *before* any call on `render` is made.

MorningLightMountain713:
Have been caught out by this. Watching. Would be good to guarantee `on_mount` is called before any calls to `render` Cheers!

darrenburns:
I've changed the label to bug, because it's basically considered idiomatic Textual at this point to initialise state inside your `on_mount`. If you use that state inside a `render` method, Textual should not crash.

This currently causes sporadic crashes in the `LoadingIndicator` widget.

rodrigogiraoserrao:
FYI I have a way to repro the issue of calling `render` before calling `on_mount`.
Just run the app below and click the button twice in under one second:

```py
import asyncio


from textual.app import App
from textual.widget import Widget
from textual.widgets import Button


class W(Widget):
    DEFAULT_CSS = "W { height: 1; }"
    def render(self):
        return self.renderable

    async def on_mount(self):
        await asyncio.sleep(1)
        self.renderable = "1234"
        print("on mount")


class MyApp(App[None]):
    def compose(self):
        yield Button()

    async def on_button_pressed(self):
        self.mount(W())


if __name__ == "__main__":
    MyApp().run()
```

rodrigogiraoserrao:
Progress being made. Widgets that haven't been mounted yet are being considered as "visible widgets" by the compositor too soon. E.g., changing `Compositor.visible_widgets` to only consider widgets for which the `_mounted_event` has been set seems to fix the problem.

Next up I'll figure out why unmounted widgets are making it to the compositor map so soon.

rodrigogiraoserrao:
It seems that the issue is in the way the compositor builds the map, which does not take into account which widgets have been mounted yet.

Consider this tentative fix:

```py
class Compositor:
    def _arrange_root(...) -> tuple[CompositorMap, set[Widget]]:
        # ...

        def add_widget(...) -> None:
            if not widget._mounted_event.is_set():  # <-- new
                return                              # <-- new
```

This passes the full test suite and it takes care of the problem by preventing unmounted events from being considered when building the map.

Another possible alternative would be to register widgets only after they're mounted but that doesn't sound correct.

Another possibly possible alternative is to check if the “registering” process can be split into two in a way that makes sense:

 - some “registering” happens before mounting
[...]

PR #3498, opened 2023-10-10
Test flakiness investigation and attempted fixes ❄

### LINKED ISSUE #3523: `Tabs` still has active tab after cleared
The docs say that `Tabs.Cleared` is sent "when there are no active tabs", but this raises an assertion error

```python
from textual.widgets import Footer, Tab, Tabs


class TabsApp(App):
    BINDINGS = [("space", "clear_tabs", "Clear tabs")]

        yield Tabs(
            Tab("First tab", id="one"),
            Tab("Second tab", id="two"),
        )
        yield Footer()

    def on_tabs_cleared(self, event: Tabs.Cleared) -> None:
        assert not event.tabs.active, f"{event.tabs.active} should not be!"

    def action_clear_tabs(self) -> None:
        self.query_one(Tabs).clear()


if __name__ == "__main__":
    app = TabsApp()
    app.run()
```

[Edited to correct code example: I always mix up active and active_tab!]

darrenburns:
Here are my theories and the corresponding changes for each of the flaky tests identified by Rodrigo here: https://github.com/Textualize/textual/issues/3484#issuecomment-1752739081.

Work-in-progress, but feel free to comment/discuss the theories.

✅ = I think I've fixed it

### `test_schedule_reverse_animations` ✅
This test essentially was starting two timers one line after the other, and assuming the order in which the timers finished was guaranteed. This isn't the case. 

**Solution**: I've modified the test to make it almost certain that the reverse animation begins to run while the forward animation is in progress.

### `test_scheduling_animation` ✅
I think that the `pause` was actually detrimental here - the delay `0.1` and `0.9 * 0.1` are too close to each other, and so sometimes the execution didn't continue after the `pause` until the animation was already complete.

**Solution**: I opted to remove the `pause` here rather than reduce the `0.9` value, as I don't think the pause adds much.

### `test_inactive_stack_is_alive` ✅
`switch_mode` leads to important asynchronous (queue-based) work being done, but it cannot be awaited.

When you perform any action after switching modes, you cannot be sure what state the application is in - you don't know, for example, if the root screen associated with the mode has been mounted yet.

**Solution**: I've made `switch_mode` return an `AwaitMount`, and updated the test to `await` it.

**Edit:** 
Turns out the problem was simpler - the test was querying for a `Label` in the DOM every 0.1 seconds. Since the app was switching screens, there were small windows where there was no `Label` in the DOM, and this error wasn't accounted for.

I opted to remove this test because I think the approach it took was generally flawed and I don't think it gave much value.

### `test_remove_tabs`

The problem is that the `AwaitRemove` returned by `remove_tab` only waits for the removal of the `Tab` from the DOM. The addition of the `.-active` class happens some time in the future. This ultimately means that we cannot safely use `Tabs.active_tab` in an application that supports tab removal.

**Solution**: I've added a more general object for optional awaits called `AwaitComplete`. `remove_tab` now returns an `AwaitComplete` which waits for both the DOM node being removed, **_and_** for the corresponding internal state updates.

### `test_remove_tabs_messages`

This looks like it was closely related to the issue above. `remove_tab` was being called in a loop, which was ultimately resulting in many messages being dispatched **after refresh**.

**Solution**: Same as above. Also, the messages are no longer dispatched after refresh. I tested an example to see if this reintroduced an animation issue with the underline bar, but couldn't see any problems.

---
There's still some flakiness here I can't work out because it's rare:

It looks like _on_mount can, very occasionally, still be called before its children can be queried. I can’t see any other explanation after tracing through code as much as possible. I got 2 failures in a row within Tabs where the child Underline wasn’t being returned from a query inside Tabs._on_mount.

---

### `test_textual_dev_border_preview` ✅

Looks like this may have been due to the button press animation.

The test code uses `wait_for_scheduled_animations`, but the `Button` widget doesn't use animations for the active/click/press effect.

### `test_command_palette` ✅

This failure happened during a period where snapshot reports weren't being produced in CI, so it's not possible to see why this failure occurred.

There were 2 issues here:

1. The `Input` widget cursor blinking in the `CommandPalette` - all other snapshot tests which use an `Input` set `cursor_blink = False`.
2.  The `cursor_blink` reactive didn't do anything when changed at runtime. It could only be set before the widget was actually mounted.

**Solution:** I've switched off the blinking cursor for this test. I've fixed the `Input` widget such that it handles toggling the `cursor_blink` value at runtime.

### `test_input_suggestions` ✅

This failed during investigating the above issue - very similar to the above failure.

**Solution:** I've switched off the blinking cursor for this test. I also added a new watcher to `Input` which ensures that as soon as you set the `cursor_blink` to `False`, the cursor becomes visible again.

### `test_directory_tree_reload_other_node` ✅

This is a problem that affects many `DirectoryTree` tests.

Loading and reloading of directories are done inside a worker. The test doesn't know when this worker is complete - yet it's not safe to use the `DirectoryTree` unless you know the worker is complete.

**Solution:** I've made the processing of the queue which performs the "reload" awaitable.

### `test_app_with_notifications_that_expire` ✅

This test used `time.sleep` instead of `asyncio.sleep` which blocked the event loop and likely contributed to flakiness - giving less time for notification timers to expire the notifications correctly.

**Solution:** I've reworked the timeouts used in the test so the notifications will time out much faster, and we'll wait for a long time (relative to their timeout) for them to be expired. Also switched to `asyncio.sleep`.

### `test_loading` ✅

There's a [known issue](https://github.com/Textualize/textual/issues/2914) where `render` can fire before `on_mount`. This means if you're initialising state inside `on_mount` instead of `__init__`, and your `render` method uses that state, there may be a crash.

This caused a crash in the test for `LoadingIndicator`.

**Solution:** The `LoadingIndicator` now initialises the state inside the constructor instead of in `on_mount`. This doesn't solve the root problem, but avoids a crash which can occur in this test (and presumably also in any real applications which use `LoadingIndictor`).

# Other issues

- `Tabs.active` wasn't being re-assigned when tabs get removed or cleared (fixes #3523).
- The `cursor_blink` reactive on `Input` didn't do anything when changed at runtime.

	
Fixes https://github.com/Textualize/textual/issues/3484


src/textual/await_complete.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def __init__(self, *coroutine: Coroutine) -> None:
+        """Create an AwaitComplete.
+
+        Args:
+            coroutine: One or more coroutines to execute.
+        """
+        self.coroutine = coroutine
+        AwaitComplete._instances.append(self)
+        self._future: Future = gather(*[coroutine for coroutine in self.coroutine])
```
willmcgugan, line 21: I'm surprised `gather` returns a `Future`. It doesn't appear to be documented.

src/textual/await_complete.py
```diff
@@ -0,0 +1,60 @@
...
+
+    @property
+    def exception(self) -> BaseException | None:
+        """An exception if it occurred in any of the coroutines."""
+        if self._future and self._future.done():
+            return self._future.exception()
+        return None
+
+    @classmethod
+    async def wait_all(cls):
```
willmcgugan, line 47: Is this for tests? Maybe it should be private, and the `timeout` parameter exposed.

willmcgugan (commented):
Pre-review review. I think you're on the right track. These changes seem to be worthwhile, even if we didn't have the flaky tests issue.

src/textual/await_complete.py
```diff
@@ -0,0 +1,60 @@
+from __future__ import annotations
+
+from asyncio import Future, gather, wait
+from typing import Coroutine
+
+
+class AwaitComplete:
+    """An 'optionally-awaitable' object."""
+
+    _instances: list["AwaitComplete"] = []
```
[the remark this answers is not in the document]
darrenburns: Will do 👍

src/textual/await_complete.py
```diff
@@ -0,0 +1,60 @@
+from __future__ import annotations
+
+from asyncio import Future, gather, wait
+from typing import Coroutine
+
+
+class AwaitComplete:
```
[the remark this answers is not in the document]
darrenburns: I think this is a good idea - I'll update it.

src/textual/await_complete.py
darrenburns: Yeah, it returns a "GatheringFuture" instance 😄

src/textual/await_complete.py
darrenburns: I was thinking that since this class offers a way of scheduling coroutines, it should also offer a cleanup mechanism. It's not used at the moment, but I'll likely be using it in tests. I can make it private.

src/textual/widgets/_tabs.py
```diff
@@ -467,50 +474,54 @@ def clear(self) -> AwaitRemove:
...
 
         remove_await = remove_tab.remove()
 
         async def do_remove() -> None:
             """Perform the remove after refresh so the underline bar gets new positions."""
             await remove_await
-            if removing_active_tab:
-                if next_tab is not None:
+
+            removal_complete = asyncio.Event()
```
darrenburns, line 509: @willmcgugan Any thoughts on this - you mentioned the threat of deadlocks when using events. Does this usage look alright?

We need to wait until after the refresh before handing control back the awaiting user, since the `-active` class cannot be set until after after refresh.

willmcgugan: Yeah, I suspect this could deadlock.

The `call_after_refresh` will post a message to self. If you have `await self.remove_tab(...)` in a message handler then it will get stuck waiting for the refresh, which won't ever be processed.

At least I think that's what would happen. Worth testing.

I think this may suggest that we need a different implementation of `call_after_refresh`. Or maybe we can do without it in this instance.

darrenburns: I'll test it out. Unfortunately I think the `call_after_refresh` is necessary here, a quick test without it caused issues. Unless there's a smarter implementation we can use here that can calculate the new tab position in advance, before the refresh happens.

darrenburns: Fixed the deadlock here by waiting for the event in another coroutine, scheduling it, then including the Future in the returned AwaitComplete (via `gather`).

rodrigogiraoserrao (changes requested):
I only found a couple of small things that could be improved.

There were also a couple of Return docstrings which I requested you indent. I recall doing this for longer docstrings and I think there was a reason we did it, so I suggested changes in line with that.

src/textual/widgets/_data_table.py
```diff
@@ -1696,9 +1690,9 @@ async def _on_idle(self, _: events.Idle) -> None:
         if self._updated_cells:
             # Cell contents have already been updated at this point.
             # Now we only need to worry about measuring column widths.
-            updated_cells = self._updated_cells.copy()
+            updated_columns = self._updated_cells.copy()
             self._updated_cells.clear()
-            self._update_column_widths(updated_cells)
+            self._update_column_widths(updated_columns)
```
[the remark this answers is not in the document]
darrenburns: Yeah, not sure what I was thinking here - reverted it.

src/textual/widgets/_directory_tree.py
```diff
@@ -213,7 +218,7 @@ def reset_node(
         node.data = data
         return self
 
-    def reload_node(self, node: TreeNode[DirEntry]) -> None:
+    def reload_node(self, node: TreeNode[DirEntry]) -> AwaitComplete:
```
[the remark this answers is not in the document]
darrenburns: Yeah - I'm not sure if it's a problem in practice. Maybe we should push a sentinel node on to the queue and then finish awaiting after that sentinel is discovered by the consuming loop.

src/textual/widgets/_directory_tree.py
```diff
@@ -163,7 +165,9 @@ def _add_to_load_queue(self, node: TreeNode[DirEntry]) -> None:
             node.data.loaded = True
             self._load_queue.put_nowait(node)
 
-    def reload(self) -> None:
+        return AwaitComplete(self._load_queue.join())
+
+    def reload(self) -> AwaitComplete:
```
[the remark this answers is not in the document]
darrenburns: Is this required? `mypy` doesn't complain about it for me. I was hoping it'd be optional in the case of `AwaitComplete[None]`.

src/textual/widgets/_directory_tree.py
```diff
@@ -421,16 +426,17 @@ async def _loader(self) -> None:
...
                 if content:
                     self._populate_node(node, content)
-            # Mark this iteration as done.
-            self._load_queue.task_done()
+            finally:
+                # Mark this iteration as done.
+                self._load_queue.task_done()
 
-    def _on_tree_node_expanded(self, event: Tree.NodeExpanded) -> None:
+    async def _on_tree_node_expanded(self, event: Tree.NodeExpanded):
```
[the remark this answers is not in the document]
darrenburns: Fixed

src/textual/widgets/_directory_tree.py
```diff
@@ -152,7 +154,7 @@ def __init__(
         )
         self.path = path
 
-    def _add_to_load_queue(self, node: TreeNode[DirEntry]) -> None:
+    def _add_to_load_queue(self, node: TreeNode[DirEntry]) -> AwaitComplete:
```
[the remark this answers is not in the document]
darrenburns: You might be on to something - I thought `join()` returned a Future. I'll investigate this more.

src/textual/widgets/_input.py
```diff
@@ -330,6 +334,15 @@ def _watch_cursor_position(self) -> None:
...
         self.app.cursor_position = self.cursor_screen_offset
 
+    def _watch_cursor_blink(self, blink: bool) -> None:
+        """Ensure we handle updating the cursor blink at runtime."""
+        if self._blink_timer is not None:
+            if blink:
+                self._blink_timer.resume()
+            else:
+                self._cursor_visible = True
+                self._blink_timer.pause()
```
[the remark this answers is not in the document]
darrenburns: The only reason the `if self._blink_timer is not None` guard exists is because this watcher can run before the `Input` is mounted, and the `_blink_timer` isn't initialised until `on_mount`.

I don't think we have to worry about the timer disappearing or anything.

src/textual/widgets/_tabs.py
```diff
@@ -467,50 +475,50 @@ def clear(self) -> AwaitRemove:
...
-        removing_active_tab = remove_tab.has_class("-active")
+                return AwaitComplete(self.app._remove_nodes([], None)())
 
+        removing_active_tab = remove_tab.has_class("-active")
         next_tab = self._next_active
-        result_message: Tabs.Cleared | Tabs.TabActivated | None = None
-        if removing_active_tab and next_tab is not None:
-            result_message = self.TabActivated(self, next_tab)
-        elif self.tab_count == 1:
-            result_message = self.Cleared(self)
```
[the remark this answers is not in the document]
darrenburns: Yes - the messages were also posted from a watcher.

tests/test_screen_modes.py
```diff
@@ -172,47 +172,6 @@ async def test_screen_stack_preserved(ModesApp: Type[App]):
             await pilot.press("o")
 
 
-async def test_inactive_stack_is_alive():
-    """This tests that timers in screens outside the active stack keep going."""
```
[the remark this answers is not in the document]
darrenburns: I wasn't really sure how to rewrite this in a way that wouldn't be flaky - open to suggestions.

darrenburns:
@willmcgugan Ready whenever.

src/textual/widgets/_input.py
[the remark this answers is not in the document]
rodrigogiraoserrao: Exactly. So, what if the user sets cursor visibility before mounting the widget?
Then, it'll expect cursor visibility to have been set but it won't have been because we skipped the set altogether.

E.g., in a situation like this:

```py
def compose(self):
    inp = Input()
    inp.cursor_blink = False
    yield inp
```

src/textual/widgets/_directory_tree.py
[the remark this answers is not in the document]
rodrigogiraoserrao: I'm not _too_ fussed if the await waits for everything but if that's the case, we need to be clear about it in the docstrings.

CHANGELOG.md
```diff
@@ -19,6 +22,19 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
+- Breaking change: `Button.ACTIVE_EFFECT_DURATION` classvar converted to `Button.active_effect_duration` attribute https://github.com/Textualize/textual/pull/3498
+- Breaking change: `Input.blink_timer` made private (renamed to `Input._blink_timer`) https://github.com/Textualize/textual/pull/3498
+- Breaking change: `Input.cursor_blink` reactive updated to not run on mount (now `init=False`) https://github.com/Textualize/textual/pull/3498
+- Breaking change: `AwaitTabbedContent` class removed https://github.com/Textualize/textual/pull/3498
+- Breaking change: `TabbedContent.add_pane` now returns an `AwaitComplete` instead of an `AwaitTabbedContent` https://github.com/Textualize/textual/pull/3498
+- Breaking change: `TabbedContent.remove_pane` now returns an `AwaitComplete` instead of an `AwaitTabbedContent` https://github.com/Textualize/textual/pull/3498
+- Breaking change: `TabbedContent.clear_pane` now returns an `AwaitComplete` instead of an `AwaitTabbedContent` https://github.com/Textualize/textual/pull/3498
```
willmcgugan, line 31: Are these really breaking? Presumably you await them both. Is there anything that the caller would need to change in their code?

src/textual/await_complete.py
```diff
@@ -0,0 +1,66 @@
+from __future__ import annotations
+
+from asyncio import Future, gather, wait
+from typing import Any, Coroutine, Generator, Generic, TypeVar
+
+ReturnType = TypeVar("ReturnType")
+
+
+class AwaitComplete(Generic[ReturnType]):
```
willmcgugan, line 9: Can we have a repr for this?

src/textual/await_complete.py
```diff
@@ -0,0 +1,66 @@
...
+    """An 'optionally-awaitable' object."""
+
+    _instances: set["AwaitComplete"] = set()
```
willmcgugan, line 12: This makes me a little nervous in they are at the class level. If you have multiple apps, then they could clash.

I know its an exotic scenario at the moment, but I can see a use for apps in apps.

Do we need this at all? I see its used in `_wait_all`, but I don't see where that is used. Unless GitHub has hidden it for me.

src/textual/widgets/_data_table.py
```diff
@@ -1116,10 +1116,9 @@ def move_cursor(
             cursor_row = row
         if column is not None:
             cursor_column = column
-
         destination = Coordinate(cursor_row, cursor_column)
         self.cursor_coordinate = destination
-        self.call_after_refresh(self._scroll_cursor_into_view, animate=animate)
+        self._scroll_cursor_into_view(animate=animate)
```
willmcgugan, line 1121: It's good to get rid of the `call_after_refresh`, but I do wonder what it was solving originally.

In hindsight I wish we had added a comment there. Might be a good policy going forward every time we use `call_after_refresh` as a workaround.

src/textual/await_complete.py
```diff
@@ -0,0 +1,66 @@
...
+    """Track all active instances of AwaitComplete."""
+
+    def __init__(self, *coroutine: Coroutine) -> None:
```
willmcgugan, line 15: Could you type the return type of `Coroutine` here. I think that it will allow the checker to infer the return type of the awaitable.

src/textual/widgets/_tabbed_content.py
```diff
@@ -280,7 +260,7 @@ def add_pane(
         *,
         before: TabPane | str | None = None,
         after: TabPane | str | None = None,
-    ) -> AwaitTabbedContent:
+    ) -> AwaitComplete:
```
willmcgugan, line 263: Do we know the return type here?

willmcgugan (commented):
Nice work. Some very satisfying changes here (removing all those pauses). Some questions and suggestions.

src/textual/widgets/_data_table.py
darrenburns: I actually added the `call_after_refresh` here last week because it was solving the problem of the DataTable not scrolling if you added a bunch of rows and then immediately moved the cursor.

The `call_after_refresh` is required to ensure the DataTable only scrolls after the virtual height settles inside on_idle.

The removal here was a merge conflict issue - I was getting this confused with another piece of code where I actually did remove the `call_after_refresh` - good spot!

src/textual/widgets/_input.py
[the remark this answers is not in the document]
darrenburns: If they set the blink before the mount, then the blink timer will still be initialised in `on_mount`, but it will start as paused:

```python
    def _on_mount(self, _: Mount) -> None:
        self._blink_timer = self.set_interval(
            0.5,
            self._toggle_cursor,
            pause=not (self.cursor_blink and self.has_focus),
        )
```

src/textual/widgets/_tabbed_content.py
darrenburns: Actually, the plan to use generics with this is a bit flawed as it stands because the `AwaitComplete` can take multiple coroutines, and they may have different return types. We may not want to pass the return values from all of the coroutines up.

I think to add support for generics would require extra API work for specifying the return value. For now it might be better to just not support the generic since it's never used, and if we have need for it in the future, add it?

willmcgugan: I think that's reasonable. If all we are doing it awaiting until it is ready, then no return type needed.

src/textual/widgets/_directory_tree.py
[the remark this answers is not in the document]
darrenburns: `AwaitComplete` is no longer generic, so I'll mark this as resolved since it's no longer relevant.

src/textual/await_complete.py
```diff
@@ -0,0 +1,48 @@
...
+
+from asyncio import Future, gather
+from typing import Any, Coroutine, Generator, Generic, Iterator, TypeVar
+
+import rich.repr
+
+ReturnType = TypeVar("ReturnType")
+
+
+@rich.repr.auto(angular=True)
```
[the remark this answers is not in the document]
darrenburns: Wouldn't it already include the coroutines, since the parameter is named the same as the attribute?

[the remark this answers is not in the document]
willmcgugan: Ah, you're right. What a nice feature. I must thank the dev who implemented that.

Curious what the repr will look like.

rodrigogiraoserrao (approved):

PR #3499, opened 2023-10-10
Be less strict about typing timer callbacks.

### LINKED ISSUE #3434: linter complaint about using a worker as a timer callback

yes

```
from textual import work
from textual.app import App
class testy(App):
    @work
    async def worky(self) -> None:
        pass
    def compose(self):
        self.set_timer(1, self.worky)
```

```
$ pyright test.py
/home/guy/gits/ttuna/test.py
  /home/guy/gits/ttuna/test.py:8:27 - error: Argument of type "() -> Worker[None]" cannot be assigned to parameter "callback" of type "TimerCallback | None" in function "set_timer"
    Type "() -> Worker[None]" cannot be assigned to type "TimerCallback | None"
      Type "() -> Worker[None]" cannot be assigned to type "() -> Awaitable[None]"
        Function return type "Worker[None]" is incompatible with type "Awaitable[None]"
          "Worker[None]" is incompatible with protocol "Awaitable[None]"
            "__await__" is not present
      Type "() -> Worker[None]" cannot be assigned to type "() -> None"
        Function return type "Worker[None]" is incompatible with type "None"
          Type cannot be assigned to type "None"
    ... (reportGeneralTypeIssues)
1 error, 0 warnings, 0 informations 
```

The same error is reported for `set_interval` as well

rodrigogiraoserrao:
Fixes #3434.

Timer callbacks can now return `Any` instead of `None`.

This change did not introduce any new typing errors and it fixed two typing errors in our codebase.

### DIFF
--- a/src/textual/_work_decorator.py
+++ b/src/textual/_work_decorator.py
@@ -87,7 +87,7 @@ def work(
     exclusive: bool = False,
     description: str | None = None,
     thread: bool = False,
-) -> Callable[FactoryParamSpec, Worker[ReturnType]] | Decorator:
+) -> Callable[FactoryParamSpec, Worker[ReturnType]] | Decorator[..., ReturnType]:
     """A decorator used to create [workers](/guide/workers).
 
     Args:
--- a/src/textual/_work_decorator.py
+++ b/src/textual/_work_decorator.py
@@ -1,5 +1,4 @@
 """
-
 A decorator used to create [workers](/guide/workers).
 """
 
@@ -87,7 +86,7 @@ def work(
     exclusive: bool = False,
     description: str | None = None,
     thread: bool = False,
-) -> Callable[FactoryParamSpec, Worker[ReturnType]] | Decorator[..., ReturnType]:
+) -> Callable[FactoryParamSpec, Worker[ReturnType]] | Decorator:
     """A decorator used to create [workers](/guide/workers).
 
     Args:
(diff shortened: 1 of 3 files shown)


src/textual/_work_decorator.py
```diff
@@ -87,7 +87,7 @@ def work(
     exclusive: bool = False,
     description: str | None = None,
     thread: bool = False,
-) -> Callable[FactoryParamSpec, Worker[ReturnType]] | Decorator:
+) -> Callable[FactoryParamSpec, Worker[ReturnType]] | Decorator[..., ReturnType]:
```
willmcgugan, line 90: This confuses me. Decorator appears to be a type alias and not a generic.

rodrigogiraoserrao: `Decorator` is a generic type alias, see the (type variable) `ReturnType` in it:

https://github.com/Textualize/textual/blob/d0f0d76f789a7ed4542b85d196d09875d1fbf86e/src/textual/_work_decorator.py#L23-L31

And my change is to align this usage of `Decorator` in the overload with the previous definition:

https://github.com/Textualize/textual/blob/d0f0d76f789a7ed4542b85d196d09875d1fbf86e/src/textual/_work_decorator.py#L66-L75

willmcgugan: I don't understand that syntax. It doesn't appear to do anything. And now Mypy is complaining. Are you sure that `Decorator` wouldn't suffice in both places?

rodrigogiraoserrao: No, I'm not sure.

With mypy 1.4.1, my change doesn't introduce any new complaints.
mypy complains about line 75, which I didn't touch.

If I change line 75 to use `Decorator` only, mypy stops complaining.

If I use `Decorator` only, pylance complains (in both locations).

What I can do is undo my `Decorator` change and leave it as a follow-up issue.

rodrigogiraoserrao: Rolled back this specific change and opened #3510.

davep (approved):

PR #3503, opened 2023-10-10
Add will-change to excalidraw diagrams.

### LINKED ISSUE #3497: Excalidraw diagrams don't look good in dark mode on Safari
I think this is due to the CSS that modifies the SVGs when in dark mode.

The CSS in question uses css `"filter:"`. I wonder duplicating those rules with `"-webkit-filter"` would fix it.

rodrigogiraoserrao:
Adding will-change: filter seems to fix the issue with the Excalidraw diagrams on dark mode and recent Safari versions. This change was also tested on Firefox and Chrome and it didn't break the diagrams in those browsers.

This fixes #3497.

Docs on macbook:

Screenshot 2023-10-10 at 16 57 58

Docs on mobile:

IMG_1883

### DIFF
--- a/docs/stylesheets/custom.css
+++ b/docs/stylesheets/custom.css
@@ -14,6 +14,7 @@ h3 .doc-heading code {
 }
 
 body[data-md-color-primary="black"] .excalidraw  svg  {
+  will-change: filter;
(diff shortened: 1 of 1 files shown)


willmcgugan:
Awesome. Will deploy this now.

PR #3509, opened 2023-10-11
Add loading reactive

willmcgugan:
Adds a loading reactive to widgets.

To put anything in to a loading state, you can do this:

```python
data_table.loading = True
```

This will overlay the widget with a LoadingIndicator.

### DIFF
--- a/docs/guide/widgets.md
+++ b/docs/guide/widgets.md
@@ -317,7 +317,7 @@ Let's look at an example of this.
     ```
 
 
-In this example we have four [DataTable](../widgets/data_table.md) widgets, which we put in to a loading state by setting the widget's `loading` property to `True`.
+In this example we have four [DataTable](../widgets/data_table.md) widgets, which we put into a loading state by setting the widget's `loading` property to `True`.
 This will temporarily replace the widget with a loading indicator animation.
 When the (simulated) data has been retrieved, we reset the `loading` property to show the new data.
 
@@ -564,7 +564,7 @@ Here's a sketch of what the app should ultimately look like:
 --8<-- "docs/images/byte01.excalidraw.svg"
 </div>
 
-There are three types of built-in widget in the sketch, namely ([Input](../widgets/input.md), [Label](../widgets/label.md), and [Switch](../widgets/switch.md)). Rather than manage these as a single collection of widgets, we can arrange them in to logical groups with compound widgets. This will make our app easier to work with.
+There are three types of built-in widget in the sketch, namely ([Input](../widgets/input.md), [Label](../widgets/label.md), and [Switch](../widgets/switch.md)). Rather than manage these as a single collection of widgets, we can arrange them into logical groups with compound widgets. This will make our app easier to work with.
 
 ??? textualize "Try in Textual-web"
 
@@ -605,7 +605,7 @@ Note the `compose()` methods of each of the widgets.
 
 - The `ByteInput` yields 8 `BitSwitch` widgets and arranges them horizontally. It also adds a `focus-within` style in its CSS to draw an accent border when any of the switches are focused.
 
-- The `ByteEditor` yields a `ByteInput` and an `Input` control. The default CSS stacks the two controls on top of each other to divide the screen in to two parts.
+- The `ByteEditor` yields a `ByteInput` and an `Input` control. The default CSS stacks the two controls on top of each other to divide the screen into two parts.
 
 With these three widgets, the [DOM](CSS.md#the-dom) for our app will look like this:
 
--- a/docs/guide/widgets.md
+++ b/docs/guide/widgets.md
@@ -294,6 +294,37 @@ Add a rule to your CSS that targets `Tooltip`. Here's an example:
     ```{.textual path="docs/examples/guide/widgets/tooltip02.py" hover="Button"}
     ```
 
+## Loading indicator
+
+Widgets have a [`loading`][textual.widget.Widget.loading] reactive which when set to `True` will temporarily replace your widget with a [LoadingIndicator](../widgets/loading_indicator.md).
+
+You can use this to indicate to the user that the app is currently working on getting data, and there will be content when that data is available.
+Let's look at an example of this.
+
+=== "loading01.py"
+
+    ```python title="loading01.py"
+    --8<-- "docs/examples/guide/widgets/loading01.py"
+    ```
+
+    1. Shows the loading indicator in place of the data table.
+    2. Insert a random sleep to simulate a network request.
+    3. Show the new data.
+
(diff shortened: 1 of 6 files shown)


docs/guide/widgets.md
```diff
@@ -294,6 +294,37 @@ Add a rule to your CSS that targets `Tooltip`. Here's an example:
     ```{.textual path="docs/examples/guide/widgets/tooltip02.py" hover="Button"}
     ```
 
+## Loading indicator
+
+Widgets have a [`loading`][textual.widget.Widget.loading] reactive which when set to `True` will temporarily replace your widget with a [LoadingIndicator](../widgets/loading_indicator.md).
```
rodrigogiraoserrao, line 299:
```suggestion
Widgets have a [`loading`][textual.widget.Widget.loading] reactive which when set to `True` will temporarily replace your widget with a [`LoadingIndicator`](../widgets/loading_indicator.md).
```

docs/guide/widgets.md
```diff
@@ -294,6 +294,37 @@ Add a rule to your CSS that targets `Tooltip`. Here's an example:
...
+    2. Insert a random sleep to simulate a network request.
+    3. Show the new data.
+
+=== "Output"
+
+    ```{.textual path="docs/examples/guide/widgets/loading01.py"}
+    ```
+
+
+In this example we have four [DataTable](../widgets/data_table.md) widgets, which we put in to a loading state by setting the widget's `loading` property to `True`.
```
rodrigogiraoserrao, line 320:
```suggestion
In this example we have four [`DataTable`](../widgets/data_table.md) widgets, which we put into a loading state by setting the widget's `loading` reactive to `True`.
```

docs/guide/widgets.md
```diff
@@ -294,6 +294,37 @@ Add a rule to your CSS that targets `Tooltip`. Here's an example:
...
+This will temporarily replace the widget with a loading indicator animation.
+When the (simulated) data has been retrieved, we reset the `loading` property to show the new data.
```
rodrigogiraoserrao, line 322:
```suggestion
When the (simulated) data has been retrieved, we reset the `loading` reactive to show the new data.
```

src/textual/widget.py
```diff
@@ -497,6 +500,29 @@ def __exit__(
         else:
             self.app._composed[-1].append(composed)
 
+    def set_loading(self, loading: bool) -> Awaitable:
+        """Set or reset the loading state of this widget.
+
+        A widget in a loading state will display a LoadingIndicator that obscures the widget.
+
+        Args:
+            loading: `True` to put the widget in to a loading state, or `False` to reset the loading state.
```
rodrigogiraoserrao, line 509:
```suggestion
            loading: `True` to put the widget into a loading state, or `False` to reset the loading state.
```

src/textual/widgets/_loading_indicator.py
```diff
@@ -22,8 +24,49 @@ class LoadingIndicator(Widget):
         content-align: center middle;
         color: $accent;
     }
+    LoadingIndicator.-overlay {
+        overlay: screen;
+        background: $boost;
+    }
     """
 
+    def apply(self, widget: Widget) -> AwaitMount:
```
rodrigogiraoserrao, line 33: I actually expected `apply` to also be a `classmethod`.
Is there any reason why we prefer it not to be a `classmethod`?
Is it so that we can have greater control over which loading indicator gets overlayed on the given widget?

tests/test_widget.py
```diff
@@ -355,3 +355,23 @@ def test_get_set_tooltip():
...
+        assert label.loading == False
+        assert len(label.query(LoadingIndicator)) == 0
+        label.loading = True
+        await pilot.pause()
+        assert len(label.query(LoadingIndicator)) == 1
+        label.loading = True  # Setting to same value is a null-op
+        label.loading = False
+        await pilot.pause()
+        assert len(label.query(LoadingIndicator)) == 0
+        label.loading = False  # Setting to same value is a null-op
```
rodrigogiraoserrao, line 377: If setting to the same value is a no-op, we should be able to assert that the state didn't change, right?
This makes a bit more sense to me:

```suggestion
        assert len(label.query(LoadingIndicator)) == 0

        label.loading = True
        await pilot.pause()
        assert len(label.query(LoadingIndicator)) == 1

        label.loading = True  # Setting to same value is a null-op
        await pilot.pause()
        assert len(label.query(LoadingIndicator)) == 1

        label.loading = False
        await pilot.pause()
        assert len(label.query(LoadingIndicator)) == 0

        label.loading = False  # Setting to same value is a null-op
        await pilot.pause()
        assert len(label.query(LoadingIndicator)) == 0
```

Otherwise, we can just remove the no-op assignments:

```suggestion
        assert len(label.query(LoadingIndicator)) == 0
        label.loading = True
        await pilot.pause()
        assert len(label.query(LoadingIndicator)) == 1
        label.loading = False
        await pilot.pause()
        assert len(label.query(LoadingIndicator)) == 0
```

rodrigogiraoserrao (approved):
The implementation was surprisingly straightforward.

I left a couple of suggested changes and I think we should have a snapshot test, maybe the example `loading01.py`, but nothing blocking.

docs/guide/widgets.md
willmcgugan: It's a reactive property. The term `reactive` could be confused with the descriptor object.

src/textual/widgets/_loading_indicator.py
willmcgugan: It's so the caller can construct the LoadingIndicator.

src/textual/widgets/_loading_indicator.py
```diff
@@ -22,8 +24,49 @@ class LoadingIndicator(Widget):
...
+
+        Returns:
+            AwaitMount: An awaitable for mounting the indicator.
+        """
+        self.add_class("-overlay")
+        await_mount = widget.mount(self, before=0)
+        return await_mount
+
+    @classmethod
+    def clear(cls, widget: Widget) -> Awaitable:
```
davep, line 49: Just to note, and perhaps it's something to handle or not (or at least document): turns out that `get_child_by_type` is really `get_first_child_by_type`; so if by some bizarre situation there are multiple loading indicators here, it'll remove the first one it finds (which is fine).

I do wonder though if this should perhaps remove *all* the loading indicators it finds?

davep (approved):

tests/test_widget.py
willmcgugan: It was more to test that setting the same value didn't crash. But it couldn't hurt to check it again.

PR #3514, opened 2023-10-11
Add a test and remove dead code

rodrigogiraoserrao:


src/textual/screen.py
```diff
@@ -976,17 +976,6 @@ def _forward_event(self, event: events.Event) -> None:
                 else:
                     widget._forward_event(event._apply_offset(-region.x, -region.y))
 
-        elif isinstance(event, (events.MouseScrollDown, events.MouseScrollUp)):
```
rodrigogiraoserrao, line 979: `MouseScrollUp` and `MouseScrollDown` inherit from `MouseEvent` so this branch is never taken.
(See the branch immediately before.)

willmcgugan (approved):
good catch

PR #3528, opened 2023-10-13
Fix missing Tree.NodeHighlighted message

robinvd:
When the tree has no selection the cursor_line is -1, then clicking on the last node in the tree will check if the selection is different from the last selection, and do a check nodes[previous_cursor_line] != nodes[new_cursor_line] but if previous_cursor_line=-1 and new_cursor_line=len(nodes)-1 then this will fail and the message will not be emitted.

- [ ] Updated documentation

### DIFF
--- a/src/textual/widgets/_tree.py
+++ b/src/textual/widgets/_tree.py
@@ -825,6 +825,8 @@ class Tree(Generic[TreeDataType], ScrollView, can_focus=True):
         return NodeID(id)
 
     def _get_node(self, line: int) -> TreeNode[TreeDataType] | None:
+        if line < 0:
+            return None
         try:
             tree_line = self._tree_lines[line]
         except IndexError:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## Unreleased
+### Fixed
+- Clicking on the last node of an new Tree would not emit NodeHighlighted message https://github.com/Textualize/textual/pull/3528
+### Added
+### Changed
+
 ## [0.40.0] - 2023-10-11
 
 - Added `loading` reactive property to widgets https://github.com/Textualize/textual/pull/3509


TomJGooding:
Any chance of a [regression test](https://github.com/Textualize/textual/blob/main/CONTRIBUTING.md#guidelines)? I'm not a maintainer of Textual, but I'm sure you'll be asked this when someone gets to reviewing this. Thanks for you PR!

robinvd:
any change to review this @davep (or other maintainers)

davep:
@robinvd We've got a lot going on so sometimes it can take us a wee while to get through PRs. Meanwhile though what Tom said is good advice: if you've identified a bug and this is a PR to fix it then a regression test would be very welcome.

robinvd:
@davep ofc I understand!On 31 Oct 2023, at 09:22, Dave Pearson ***@***.***> wrote:﻿
@robinvd We've got a lot going on so sometimes it can take us a wee while to get through PRs. Meanwhile though what Tom said is good advice: if you've identified a bug and this is a PR to fix it then a regression test would be very welcome.

davep:
Just to add too: having enabled the running of tests in CI: it looks like your change might be causing some existing tests to fail; you will want to take that into account.

willmcgugan:
@robinvd Could you open an issue for this please. With a small example that reproduces the problem. Just to help me understand what this is fixing.

willmcgugan:
Closing. Assumed stale.

robinvd:
@willmcgugan repro example

```python
from textual.app import App
from textual.widgets import Tree


class BugApp(App):
    def compose(self):
        t = Tree("test")
        t.show_root = False
        t.root.add_leaf("node1")
        t.root.add_leaf("node2")
        t.root.add_leaf("node3")
        yield t

    def on_tree_node_highlighted(self, event):
        self.notify(f"hightlight! {event}")


BugApp().run()
```

if you now click on the last node, no notification will popup

src/textual/widgets/_tree.py
```diff
@@ -825,6 +825,8 @@ def _new_id(self) -> NodeID:
         return NodeID(id)
 
     def _get_node(self, line: int) -> TreeNode[TreeDataType] | None:
+        if line < 0:
```
TomJGooding, line 828: This would at least pass existing tests and seems to work when running your MRE:

```suggestion
        if line < 0 and not self.show_root:
```

TomJGooding (commented):
I might be wrong but this looks like an issue only when `show_root=False` is set in `compose`?

Currently this change causes multiple tests to fail as @davep already mentioned. I wonder perhaps this just needs a check if `show_root` is False.

src/textual/widgets/_tree.py
robinvd: It indeed only happens when show_root=False as otherwise the root is always selected

PR #3430, opened 2023-09-29
feat(input): add clear method

TomJGooding:
Closes #3428 by adding a `Input.clear` method.

Example usage:

```python
from textual.app import App, ComposeResult
from textual.widgets import Input


class InputApp(App):
    def compose(self) -> ComposeResult:
        yield Input(placeholder="Enter some text and press enter")

    def on_input_submitted(self, event: Input.Submitted) -> None:
        self.notify(f"You entered '{event.value}'")
        event.input.clear()


if __name__ == "__main__":
    app = InputApp()
    app.run()
```

- [x] Docstrings on all new or modified functions / classes 
- [ ] Updated documentation
- [x] Updated CHANGELOG.md (where appropriate)

PR #3529, opened 2023-10-13
clear option from select

stefgmz:
Adding a `clear` method for the `Select` similar to https://github.com/Textualize/textual/pull/3430.

Example:
```
from textual.widgets import Select


class SelectApp(App[None]):

        yield Select[int]([(str(n), n) for n in range(10)])

    def on_select_changed(self, event: Select.Changed) -> None:
        self.notify(f"You have selected '{event.value}'")
        event.select.clear()


if __name__ == "__main__":
    SelectApp().run()
```

- [ ]  Updated documentation

### DIFF
--- a/src/textual/widgets/_select.py
+++ b/src/textual/widgets/_select.py
@@ -385,3 +385,8 @@ class Select(Generic[SelectType], Vertical, can_focus=True):
         select_current = self.query_one(SelectCurrent)
         select_current.has_value = True
         self.expanded = True
+        
+    def clear(self) -> None:
+        """Clear the selected option."""
+        self.value = None
+        self.query_one(SelectCurrent).update(None)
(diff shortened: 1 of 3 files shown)


src/textual/widgets/_select.py
```diff
@@ -385,3 +385,8 @@ def action_show_overlay(self) -> None:
         select_current = self.query_one(SelectCurrent)
         select_current.has_value = True
         self.expanded = True
+        
+    def clear(self) -> None:
+        """Clear the selected option."""
+        self.value = None
+        self.query_one(SelectCurrent).update(None)
```
rodrigogiraoserrao, line 392: If you look at `_watch_value`, you will see that the watcher method will already update the `SelectCurrent`, so `self.value = None` is sufficient.

Also, make sure to include an empty newline at the end of the file.
(If you install the pre commit hooks, they'll help you.)

rodrigogiraoserrao (changes requested):
Thanks for the PR!

I've reviewed your code and suggested a change.

~~The method `clear` should also post a message `Select.Changed` but #3611 will likely fix it for you.~~

rodrigogiraoserrao:
Hey @stefgmz, thanks for your work on this.
My work on #3614 supersedes this PR as I ended up having to refactor a couple of things.
I did include a method `clear`, though 😄

PR #3539, opened 2023-10-15
button tweak

willmcgugan:
Makes buttons a little more flexible. They now have auto height, and a padding will be applied to all lines.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## Unreleased
+
+### Changed
+
+- Buttons will now display multiple lines, and have auto height https://github.com/Textualize/textual/pull/3539
+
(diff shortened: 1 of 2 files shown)


davep (commented):
I sense this needs a nice big juicy "breaking change" entry in the CHANGELOG given it'll have knock-on effects for layouts in applications. With this code:

```python

class ButtonsApp(App[None]):

        for n in range(5):
            yield Button(f"Button {n}\n" * (n+1))

if __name__ == "__main__":
    ButtonsApp().run()
```

going from this:

Screenshot 2023-10-15 at 17 49 15

to this:

Screenshot 2023-10-15 at 17 49 36

willmcgugan:
Why would anyone have buttons with new lines that don't display?

davep:
Accidentally doing it would be the most obvious cause; or having observed only the first line shows and unfortunately relying on that. Given this would cause visible layout changes it seems prudent to me to give everyone a heads up in the document intended for that.

willmcgugan:
I've added a changelog, but I don't see that as a breaking change. A visual change, certainly, but we've not marked such changes as breaking in the past.

davep (approved):

PR #3540, opened 2023-10-15
Tweak the docstrings of the ancestors properties

davep:
Both look to have been a little bit mangled at some point.

### DIFF
--- a/src/textual/dom.py
+++ b/src/textual/dom.py
@@ -896,7 +896,7 @@ class DOMNode(MessagePump):
 
     @property
     def ancestors_with_self(self) -> list[DOMNode]:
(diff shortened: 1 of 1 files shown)


darrenburns (approved):

src/textual/dom.py
```diff
@@ -914,7 +914,7 @@ def ancestors_with_self(self) -> list[DOMNode]:
 
     @property
     def ancestors(self) -> list[DOMNode]:
-        """A list of ancestor nodes Nodes by tracing ancestors all the way back to App.
+        """A list of ancestor Nodes found by tracing a path all the way back to App.
```
willmcgugan, line 917: I think `Nodes` should not be capitalized.

PR #3545, opened 2023-10-16
Setter for TextArea load_text

davetapley:
As discussed https://github.com/Textualize/textual/discussions/3525#discussioncomment-7277111

- [ ] Updated documentation

### DIFF
--- a/src/textual/widgets/_text_area.py
+++ b/src/textual/widgets/_text_area.py
@@ -929,6 +929,15 @@ TextArea {
         """The entire text content of the document."""
         return self.document.text
 
+    @text.setter
+    def text(self, value: str) -> None:
+        """Replace the text currently in the TextArea. This is an alias of `load_text`.
+
+        Args:
+            value: The text to load into the TextArea.
+        """
+        self.load_text(value)
+
     @property
     def selected_text(self) -> str:
         """The text between the start and end points of the current selection."""
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 
 - Add Document `get_index_from_location` / `get_location_from_index` https://github.com/Textualize/textual/pull/3410
+- Add setter for `TextArea.text` https://github.com/Textualize/textual/discussions/3525
 
 ### Changed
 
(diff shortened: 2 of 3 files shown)


davetapley:
Thoughts on tests / doc changes? 🤔

src/textual/widgets/_text_area.py
```diff
@@ -929,6 +929,16 @@ def text(self) -> str:
         """The entire text content of the document."""
         return self.document.text
 
+    @text.setter
+    def text(self, value):
```
TomJGooding, line 933:
```suggestion
    def text(self, value: str) -> None:
```

src/textual/widgets/_text_area.py
```diff
@@ -929,6 +929,16 @@ def text(self) -> str:
...
+        """Load text into the TextArea.
+
+        This will replace the text currently in the TextArea.
+
+        This is an alias of `load_text`.
```
TomJGooding, line 938:
```suggestion
        """Replace the text currently in the TextArea. This is an alias of `load_text`.
        
        Args:
            value: The text to load into the TextArea.
```

TomJGooding (commented):
I'm not a maintainer of Textual, but your code is missing type annotations and I've suggested tweaks to the docstring.

Do you think you could also add a simple test?

davetapley:
Thanks @TomJGooding, I wasn't sure about the argument doc string for a setter, but I guess there's no harm.

And r.e. tests it seems needlessly duplicitous when we already have `load_text` covered.
I suppose I could test that the setter calls that with a mock, but that's not really testing anything 🤪

TomJGooding:
Sorry you might be right with this only being a setter. I think I was just on auto-pilot (and my change upset black).

I will mind my own business and leave this for the maintainers!

darrenburns:
A simple test is still valuable as a guard against future regressions - it's a simple setter for now but could be modified in the future, so we want to capture it's expected behaviour. Just something like this would suffice:

```python
new_text = "hello\nworld\n"
text_area.text = new_text
assert text_area.text = new_text
```

darrenburns:
I wrote this test here for it, but I think your fork is locked as I can't push to this PR:

```python
async def test_text_setter():
    app = TextAreaApp()
    async with app.run_test():
        text_area = app.query_one(TextArea)
        new_text = "hello\nworld\n"
        text_area.text = new_text
        assert text_area.text == new_text
```

Also, if you could add a CHANGELOG.md entry too when you get a moment, that'd be great.

darrenburns (approved):
Looks good, thanks!

PR #3566, opened 2023-10-21
CSS initial special value

willmcgugan:
- Makes empty CSS rules an error
- Adds `initial` to all rules, which restores default
- Fixed issues with CSS error reporting duplicate errors
- Set the max height of the DataTable to 100vh by default. This is because the previous value of 100% broke with horizontal layouts.
- Added log message for startup time
- Added xfail for CSS inheritance. Need to come back to this one.

### DIFF
--- a/docs/guide/CSS.md
+++ b/docs/guide/CSS.md
@@ -456,7 +456,7 @@ Button {
 }
 ```
 
-If we want a specific button to use the default color, we can set the value to `initial`.
+If we want a specific button (or buttons) to use the default color, we can set the value to `initial`.
 For instance, if we have a widget with a (CSS) class called `dialog`, we could reset the background color of all buttons inside the dialog with the following CSS:
 
 ```sass
--- a/docs/guide/CSS.md
+++ b/docs/guide/CSS.md
@@ -442,3 +442,28 @@ Variables can refer to other variables.
 Let's say we define a variable `$success: lime;`.
 Our `$border` variable could then be updated to `$border: wide $success;`, which will
 be translated to `$border: wide lime;`.
+
+## Initial value
+
+All CSS rules support a special value called `initial`, which will reset a value back to its default.
+
+Let's look at an example.
+The following will set the background of a button to green:
+
+```sass
+Button {
+  background: green;
+}
+```
+
+If we want a specific button to use the default color, we can set the value to `initial`.
+For instance, if we have a widget with a (CSS) class called `dialog`, we could reset the background color of all buttons inside the dialog with the following CSS:
+
+```sass
+.dialog Button {
+  background: initial;
+}
+```
+
+Note that `initial` will set the value back to the value defined in any [default css](./widgets.md#default-css).
+If you use `initial` within default css, it will treat the rule as completely unstyled.
--- a/src/textual/_layout.py
+++ b/src/textual/_layout.py
@@ -150,7 +150,7 @@ class Layout(ABC):
             width = 0
         else:
             # Use a size of 0, 0 to ignore relative sizes, since those are flexible anyway
-            arrangement = widget._arrange(Size(container.width, 0))
+            arrangement = widget._arrange(Size(0, 0))
             return arrangement.total_region.right
         return width
 
--- a/src/textual/_layout.py
+++ b/src/textual/_layout.py
@@ -150,7 +150,7 @@ class Layout(ABC):
             width = 0
         else:
             # Use a size of 0, 0 to ignore relative sizes, since those are flexible anyway
-            arrangement = widget._arrange(Size(0, 0))
+            arrangement = widget._arrange(Size(container.width, 0))
             return arrangement.total_region.right
         return width
 
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -2231,6 +2231,7 @@ class App(Generic[ReturnType], DOMNode):
             self._handle_exception(error)
 
     async def _pre_process(self) -> bool:
+        """Special case for the app, which doesn't need the functionality in MessagePump."""
         return True
 
     async def _ready(self) -> None:
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -375,7 +375,6 @@ class App(Generic[ReturnType], DOMNode):
         super().__init__()
         self.features: frozenset[FeatureFlag] = parse_features(os.getenv("TEXTUAL", ""))
 
-        self._panicked = False
         self._filters: list[LineFilter] = []
         environ = dict(os.environ)
         no_color = environ.pop("NO_COLOR", None)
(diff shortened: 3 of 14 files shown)


src/textual/app.py
```diff
@@ -3023,10 +3029,14 @@ def _end_update(self) -> None:
     def _refresh_notifications(self) -> None:
         """Refresh the notifications on the current screen, if one is available."""
         # If we've got a screen to hand...
-        if self.screen is not None:
```
willmcgugan, line 3026: I don't know why this didn't raise a typic error, as `self.screen` was not typed as returning None

src/textual/css/stylesheet.py
```diff
@@ -336,6 +337,7 @@ def parse(self) -> None:
                 raise
             if any(rule.errors for rule in css_rules):
                 error_renderable = StylesheetErrors(css_rules)
+                self._invalid_css.add(css)
```
willmcgugan, line 340: This was the root cause of CSS errors being repeating. Failures here were never recorded.

davep (approved):

src/textual/css/stylesheet.py
```diff
@@ -409,6 +415,12 @@ def apply(
         # Collect the rules defined in the stylesheet
         node._has_hover_style = False
         node._has_focus_within = False
+
+        # Rules that may have an initial value
```
rodrigogiraoserrao, line 419: This comment isn't clear because the English word “initial” and the special CSS value `initial` clash.
I've had to reread these comments and the code below multiple times and each time my understanding of what we mean changes.

I suggest a slight rewording for the future person reading this code:

```suggestion
        # Rules that may be set to the special value `initial`
```

src/textual/css/stylesheet.py
```diff
@@ -409,6 +415,12 @@ def apply(
...
+        initial: set[str] = set()
+        # Rules in DEFAULT_CSS with an initial value
```
rodrigogiraoserrao, line 421: See my comment above

```suggestion
        # Rules in DEFAULT_CSS set to the special value `initial`
```

src/textual/css/stylesheet.py
```diff
@@ -433,6 +451,40 @@ def apply(
...
+        # Set initial values
+        for initial_rule_name in initial:
+            # Rules with a value of None should be set to the default value
+            if node_rules[initial_rule_name] is None:  # type: ignore[literal-required]
+                # Exclude non default values
+                # rule[0] is the specificity, rule[0][1] is 0 for default
+                default_rules = [
+                    rule
+                    for rule in rule_attributes[initial_rule_name]
+                    if not rule[0][0]
```
rodrigogiraoserrao, line 464: The comment talks about `rule[0][1]` but the code uses `rule[0][0]`, which lost me.

src/textual/css/stylesheet.py
```diff
@@ -433,6 +451,40 @@ def apply(
...
+                default_rules = [
+                    rule for rule in rule_attributes[initial_rule_name] if rule[0][0]
+                ]
+                if default_rules:
+                    # There is a default value
+                    new_value = max(default_rules, key=get_first_item)[1]
+                    node_rules[initial_rule_name] = new_value  # type: ignore[literal-required]
+                else:
+                    initial_value = getattr(_DEFAULT_STYLES, initial_rule_name)
+                    node_rules[initial_rule_name] = initial_value  # type: ignore[literal-required]
```
rodrigogiraoserrao, line 486: Assigning outside of the `if: ... else: ...` simplifies the code a bit and we only get one typing ignore.

```suggestion
                if default_rules:
                    # There is a default value
                    rule_value = max(default_rules, key=get_first_item)[1]
                else:
                    rule_value = getattr(_DEFAULT_STYLES, initial_rule_name)
                node_rules[initial_rule_name] = rule_value  # type: ignore[literal-required]
```

src/textual/message_pump.py
```diff
@@ -486,20 +488,28 @@ async def _process_messages(self) -> None:
             for timer in list(self._timers):
                 timer.stop()
 
-    async def _pre_process(self) -> None:
-        """Procedure to run before processing messages."""
+    async def _pre_process(self) -> bool:
+        """Procedure to run before processing messages.
+
+        Returns:
+            `True` if successful, or `False` if any exception occurred
```
rodrigogiraoserrao, line 495:
```suggestion
            `True` if successful, or `False` if any exception occurred.
```

src/textual/css/stylesheet.py
```diff
@@ -433,6 +451,40 @@ def apply(
                 for name, specificity_rules in rule_attributes.items()
             },
         )
+
+        # Set initial values
```
rodrigogiraoserrao, line 455: What “initial” is this? English word or `initial` value?
Are we replacing special values `initial` with an actual CSS value?

Could we rephrase the comment as something like what you can see below?

```suggestion
        # Replace the values `initial` with the appropriate default
```

rodrigogiraoserrao (commented):
Looks good to me.
I _really_ wish you accept my recommendation of rewording a couple of comments where the English word “initial” and the new special value `initial` make reading the code hard.
But I didn't find anything worth blocking the PR for.

docs/guide/CSS.md
```diff
@@ -442,3 +442,28 @@ Variables can refer to other variables.
...
+For instance, if we have a widget with a (CSS) class called `dialog`, we could reset the background color of all buttons inside the dialog with the following CSS:
+
+```sass
+.dialog Button {
+  background: initial;
+}
+```
+
+Note that `initial` will set the value back to the value defined in any [default css](./widgets.md#default-css).
+If you use `initial` within default css, it will treat the rule as completely unstyled.
```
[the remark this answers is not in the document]
willmcgugan: The value will be whatever rule wins with CSS specificity. If there is no default css, it will be the default as if no CSS was applied. The properties in Styles all have their own defaults.

src/textual/css/stylesheet.py
willmcgugan: I'm not sure how they clash. They pretty much mean the same thing. Just one is a token. But I don't mind rewording.

src/textual/css/stylesheet.py
willmcgugan: Fixed the comment

src/textual/_layout.py
```diff
@@ -150,7 +150,7 @@ def get_content_width(self, widget: Widget, container: Size, viewport: Size) ->
             width = 0
         else:
             # Use a size of 0, 0 to ignore relative sizes, since those are flexible anyway
-            arrangement = widget._arrange(Size(0, 0))
+            arrangement = widget._arrange(Size(container.width, 0))
```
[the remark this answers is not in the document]
willmcgugan: Restored it to previous code. It was an experiement.

Issue #3570, opened 2023-10-22
Button rendering fails with non-standard text alignment


holdenweb:
Have you checked closed issues? https://github.com/Textualize/textual/issues?q=is%3Aissue+is%3Aclosed

- Up to a year ago I can't find anything but one issue referring to https://textual.textualize.io/FAQ/#why-doesnt-textual-look-good-on-macos, which doesn't appear to address this issue.

Please give a brief but clear explanation of the issue. If you can, include a complete working example that demonstrates the bug. **Check it can run without modifications.**

- In the attached program the buttons are styled as `text-align: left;` and they render incorrectly, with misalignment of the right-hand borders, and hovering or clicking makes things worse. Removing the styling, so the button text is centred by default, removes the rendering error.
- With `text-align: left;`: 
<img width="959" alt="image" src="https://github.com/Textualize/textual/assets/105312/046d6dae-42bc-4f81-ac89-09abb7ea0db9">

- Without: 
<img width="957" alt="image" src="https://github.com/Textualize/textual/assets/105312/19506477-b2bb-45f1-b949-0373021f11fe">

Buggy program (as .txt to allow uploading): [button_bug.txt](https://github.com/Textualize/textual/files/13063400/button_bug.txt)

It will be helpful if you run the following command and paste the results:

# Textual Diagnostics

## Versions

| Name    | Value  |
|---------|--------|
| Textual | 0.40.0 |
| Rich    | 13.6.0 |

## Python

| Name           | Value                                                                                   |
|----------------|-----------------------------------------------------------------------------------------|
| Version        | 3.10.13                                                                                 |
| Implementation | CPython                                                                                 |
| Compiler       | Clang 14.0.3 (clang-1403.0.22.14.1)                                                     |
| Executable     | /Users/sholden/Library/Caches/pypoetry/virtualenvs/textutils-sKIyU2bn-py3.10/bin/python |

## Operating System

| Name    | Value                                                                                                |
|---------|------------------------------------------------------------------------------------------------------|
| System  | Darwin                                                                                               |
| Release | 22.1.0                                                                                               |
| Version | Darwin Kernel Version 22.1.0: Sun Oct  9 20:14:30 PDT 2022; root:xnu-8792.41.9~2/RELEASE_ARM64_T8103 |

## Terminal

| Name                 | Value              |
|----------------------|--------------------|
| Terminal Application | iTerm.app (3.4.21) |
| TERM                 | xterm-256color     |
| COLORTERM            | truecolor          |
[...]

PR #3571, opened 2023-10-22
button render

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/3570 and an issue with highlights on buttons larger than one line. Note the chunky highlights that aren't center aligned.

Screenshot 2023-10-22 at 18 22 16

Adds pad.py which contains HorizontalPad. This is similar to Rich's Padding, but simpler and won't stretch the height. This new renderable is public, but I'm going to hold off adding it to the docs for now. Just until I'm certain this is stable.


davep (commented):
Do you think it would be a good idea to add either a snapshot test for this style of button and its highlight, or a unit test for `HorizontalPad`, or both?

willmcgugan:
> Do you think it would be a good idea to add either a snapshot test for this style of button and its highlight, or a unit test for `HorizontalPad`, or both?

Yeah, I'll add a test.

PR #3578, opened 2023-10-24
Stop `OptionList.OptionHighlighted` leaking to the `App`

davep:
If `App` was catching `OptionList.OptionHighlighted` without any other checks, moving through the options in the command palette could cause false positives.


darrenburns (approved):
Good catch - do we need to worry about any of the other CommandPalette related events, e.g. the ones from the Input?

willmcgugan (approved):

davep:
@darrenburns Aye, `Input.Changed` and `Input.Submitted` are already consumed and `stop`ped.

PR #3580, opened 2023-10-24
Change some log groups

darrenburns:
We usually use the SYSTEM log group when logging from the App, but there were a few cases of using INFO - I'm not sure I understand the difference and how to choose which group to use.

I'm not sure if this PR is correct, but I opened it partly just to try to understand how logging works and how end users are expected to use it.

Have we defined what these groups mean and when we should use each of them? I can't see anything in the docs.

With our current system, it seems like there's no easy way to distinguish between Textual's logs and application logs. e.g. if there's a warning, how do I know if it's a Textual warning or an application warning?

### DIFF
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -2242,7 +2242,7 @@ class App(Generic[ReturnType], DOMNode):
         """
 
         ready_time = (perf_counter() - self._start_time) * 1000
-        self.log.info(f"ready in {ready_time:0.0f} milliseconds")
+        self.log.system(f"ready in {ready_time:0.0f} milliseconds")
 
         async def take_screenshot() -> None:
             """Take a screenshot and exit."""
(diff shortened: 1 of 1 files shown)


rodrigogiraoserrao:
Could it be that `system` _is_ for Textual logs and then application logs use the other levels?

davep:
The idea behind this makes sense to me: `system` should be for (Textual) system-level logs.

davep (approved):

darrenburns:
It feels like there should be a way to differentiate between a Textual WARNING log and a Textual INFO or DEBUG log? 

How can I, as a developer, filter out DEBUG logs printed by Textual (I can't, afaik, because they're all SYSTEM... if I filter out SYSTEM, I will also filter out Textual warnings) defined as?

Almost like "system" vs "application" is a totally different attribute of a log message (like a `source` or something), but right now it's lumped in with log groups.

There's also a `WORKER` group - so how do we know if the log from a worker is an ERROR or a DEBUG message?

davep:
In other words: it might make sense to have a split between "group" and "level"? (I'd be in favour)

PR #3582, opened 2023-10-24
Improve css error reporting

### LINKED ISSUE #3569: CSS errors have a broken path
CSS errors will create an invalid path to the CSS, if the CSS is in code. This is because it adds the widget name to the path.

Here's an example:

Image

If you click the heading it should take you to the file containing the error, but it doesn't because the path is invalid.

It would be nice to display the widget within the title, but it should probably be a separate field.

rodrigogiraoserrao:
This PR fixes #3569.

This also improves error importing by providing the widget name/class variable from where CSS was read in case the CSS is inline CSS.
This is still a draft because I still need to figure out if we can link to a specific line/column for when the CSS comes from an external file.

### DIFF
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -1753,20 +1753,19 @@ class App(Generic[ReturnType], DOMNode):
 
         update = False
         for path in screen.css_path:
-            if not self.stylesheet.has_source(path):
+            if not self.stylesheet.has_source((str(path), "")):
                 self.stylesheet.read(path)
                 update = True
         if screen.CSS:
             try:
-                screen_css_path = (
-                    f"{inspect.getfile(screen.__class__)}:{screen.__class__.__name__}"
-                )
+                screen_path = inspect.getfile(screen.__class__)
             except (TypeError, OSError):
-                screen_css_path = f"{screen.__class__.__name__}"
-            if not self.stylesheet.has_source(screen_css_path):
+                screen_path = ""
+            read_from = (screen_path, f"{screen.__class__.__name__}.CSS")
+            if not self.stylesheet.has_source(read_from):
                 self.stylesheet.add_source(
                     screen.CSS,
-                    path=screen_css_path,
+                    read_from=read_from,
                     is_default_css=False,
                     scope=screen._css_type_name if screen.SCOPED_CSS else "",
                 )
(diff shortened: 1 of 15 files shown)


src/textual/css/_styles_builder.py
```diff
@@ -65,19 +65,6 @@
...
-    Args:
-        tokens: Tokens to join
-        joiner: String to join on.
-
-    Returns:
-        The tokens, joined together to form a string.
-    """
-    return joiner.join(token.value for token in tokens)
-
-
```
rodrigogiraoserrao, line 80: This was dead code.

src/textual/dom.py
```diff
@@ -416,33 +416,41 @@ def __rich_repr__(self) -> rich.repr.Result:
...
             try:
-                return f"{getfile(base)}:{base.__name__}"
+                return (getfile(base), f"{base.__name__}.DEFAULT_CSS")
             except (TypeError, OSError):
-                return f"{base.__name__}"
+                return ("", f"{base.__name__}.DEFAULT_CSS")
 
         for tie_breaker, base in enumerate(self._node_bases):
-            css: str = base.__dict__.get("DEFAULT_CSS", "").strip()
+            css: str = base.__dict__.get("DEFAULT_CSS", "")
```
rodrigogiraoserrao, line 448: This messed up the line/column reporting of inline CSS errors.

src/textual/app.py
```diff
@@ -1753,20 +1753,19 @@ def _load_screen_css(self, screen: Screen):
 
         update = False
         for path in screen.css_path:
-            if not self.stylesheet.has_source(path):
+            if not self.stylesheet.has_source((str(path), "")):
```
willmcgugan, line 1756: I'd like to veto the tuple here. It doesn't read well. Could the second value be a keyword arg?

src/textual/css/stylesheet.py
```diff
@@ -68,27 +68,35 @@ def __rich_console__(
         for token, message in errors:
             error_count += 1
 
-            if token.path:
-                path = Path(token.path)
-                filename = path.name
+            if token.read_from:
```
willmcgugan, line 71: Could this be an explicit `is not None` check.

willmcgugan (changes requested):
Looks good. Couple of requests.

Can you add a screenshot of how these changes look in CSS errors?

src/textual/app.py
rodrigogiraoserrao: Yeah, that sounds good. See 6c6eecf.

src/textual/css/stylesheet.py
rodrigogiraoserrao: Reworked this a bit better so that the checks make more sense.

davep:
Giving this spin with this code:

```python

class CSSErrorApp(App[None]):

    CSS = """
    :
    """

        yield Label()

if __name__ == "__main__":
    CSSErrorApp().run()
```

I get this error:

```
 Error in stylesheet:
 /Users/davep/develop/python/textual-upstream/sandbox/foo.py, CSSErrorApp.CSS:1:4
│   1 │                                                                                                                                                                                                                      │
│ ❱ 2 │   :                                                                                                                                                                                                                  │
│   3                                                                                                                                                                                                                        │
 • Expected one of 'comment line', 'comment start', 'selector start', 'selector start class', 'selector start id', 'selector start universal', 'variable name', or 'whitespace'.                                              
```

This *seems* to be saying that the error is on line 1, but also on line 2. Am I reading that correctly?

rodrigogiraoserrao:
> This _seems_ to be saying that the error is on line 1, but also on line 2. Am I reading that correctly?

You're reading that correctly, yes, but that also seems to be an issue in `master`.
Should I open an issue to log this so that this PR is independent of that bug?

davep:
Cool, if it's a pre-existing thing unrelated this makes sense it's handled in a different fix.

davep (approved):

PR #3595, opened 2023-10-26
Add more pre-commit hooks.

rodrigogiraoserrao:
Went through the list of Rich pre-commit hooks and added the ones that look useful.
Prompted by the fact that I commit unused imports way too many times.

Took the opportunity to fix imports in the code and make sure one of the examples is compatible with Python 3.7 out of the box, which we still support.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,18 +4,18 @@ repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v4.3.0
     hooks:
-      - id: check-ast
-      - id: check-builtin-literals
-      - id: check-case-conflict
-      - id: check-merge-conflict
-      - id: check-json
-      - id: check-toml
-      - id: check-yaml
-        args: [ '--unsafe' ]
-      - id: check-shebang-scripts-are-executable
-      - id: check-vcs-permalinks
-      - id: end-of-file-fixer
-      - id: mixed-line-ending
+      - id: check-ast  # simply checks whether the files parse as valid python
+      - id: check-builtin-literals  # requires literal syntax when initializing empty or zero python builtin types
+      - id: check-case-conflict  # checks for files that would conflict in case-insensitive filesystems
+      - id: check-merge-conflict  # checks for files that contain merge conflict strings
+      - id: check-json  # checks json files for parseable syntax
+      - id: check-toml  # checks toml files for parseable syntax
+      - id: check-yaml  # checks yaml files for parseable syntax
+        args: [ '--unsafe' ]  # Instead of loading the files, simply parse them for syntax.
+      - id: check-shebang-scripts-are-executable  # ensures that (non-binary) files with a shebang are executable
+      - id: check-vcs-permalinks  # ensures that links to vcs websites are permalinks
+      - id: end-of-file-fixer  # ensures that a file is either empty, or ends with one newline
+      - id: mixed-line-ending  # replaces or checks mixed line ending
   - repo: https://github.com/pycqa/isort
     rev: 5.12.0
     hooks:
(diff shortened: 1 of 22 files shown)


darrenburns:
Did you test all of these out first? `pre-commit run --all` iirc.

.pre-commit-config.yaml
davep, line 1: I feel it might be useful to get some information on what problem each of these solve, and also what files they work on, etc. For example, `trailing-whitespace` seems like a good one (I've got Emacs set to solve that issue on save, but I see it in our code from time to time from other environments) but can be very problematic depending on which files it affects and if it understands context (you *want* some trailing space in markdown files, for example).

rodrigogiraoserrao:
@darrenburns I didn't, that will retroactively apply all of them.
Not even the current hooks have been applied to the whole codebase.

.pre-commit-config.yaml
rodrigogiraoserrao: So, you're asking for a comment next to each one, is that it?
I can add those.

As for the trailing whitespace, when/where do you want trailing whitespace in markdown files?
I can't see what situations you are talking about.

darrenburns: The trailing whitespace one isn't worth the hassle IMO - it'll cause problems in Markdown and in Python multiline strings - it doesn't understand that sometimes in multiline strings you want trailing whitespace.

darrenburns:
@rodrigogiraoserrao Ah okay, if the existing ones haven't been applied to all files then it's fine - I think it's just files that Will has made that nobody else has edited, so I'm assuming he doesn't use the hooks.

We can just let them run as we update files.

.pre-commit-config.yaml
darrenburns: Actually I'm surprised that one is used in Rich - I thought we removed it.

rodrigogiraoserrao: @davep I've added a comment next to the hooks so that we can see at a glance what they do.

davep: @rodrigogiraoserrao Trailing whitespace [is a way of forcing line breaks in Markdown](https://spec.commonmark.org/0.30/#hard-line-breaks); for one.

rodrigogiraoserrao: Thanks. I've removed it already.

src/textual/_types.py
```diff
@@ -1,12 +1,6 @@
 from typing import TYPE_CHECKING, Any, Awaitable, Callable, List, Union
 
-from typing_extensions import (
-    Literal,
-    Protocol,
-    SupportsIndex,
-    get_args,
-    runtime_checkable,
-)
+from typing_extensions import Protocol
```
davep, line 3: I have this very vague recollection (probably from a conversation I wasn't fully involved in) that there was a reason why these were imported here, like this, and then imported from here elsewhere in Textual; but I don't recall the detail.

As I look at the code I can't see why this PR *wouldn't* make this change -- it makes sense to be -- but it might be worth double-checking what the original motivation was.

davep (approved):

src/textual/_types.py
rodrigogiraoserrao: Thanks for the heads-up.
I'll merge after I make sure all is good.

rodrigogiraoserrao: After talking to Darren, it looks like these are remains of an old pattern we used to follow in our code base where we had plenty of Python version checks to figure out whether to import things from `typing` or `typing_extensions`, and so we wanted to import all those types in a single `_types.py` file so that we could then import from there without version checks.

Given that removing them and “fixing” the imports elsewhere to import directly from `typing_extensions` didn't break anything, we can go ahead with this PR.

PR #3444, opened 2023-10-02
fix(tabbed content): handle pane containing tabs

TomJGooding:
Fixes #3412. There might be a better way of discerning Tabs messages, so creating only as a draft for now.

I would still need to add regression tests, but this seems to work from some quick manual testing with the app below:

```python
from textual.app import App, ComposeResult
from textual.widgets import TabbedContent, TabPane, Tabs


class TabsApp(App):
    BINDINGS = [("space", "clear_inner_tabs")]
    CSS = """
    TabPane > Tabs {
        height: auto;
    }
    """

    def compose(self) -> ComposeResult:
        with TabbedContent():
            with TabPane("outer tab #1"):
                yield Tabs(
                    "inner tab#1",
                    "inner tab#2",
                    "inner tab#3",
                    id="inner-tabs-one",
                )
            with TabPane("outer tab #2"):
                yield Tabs(
                    "inner tab#4",
                    "inner tab#5",
                    "inner tab#6",
                    id="inner-tabs-two",
                )

    def action_clear_inner_tabs(self) -> None:
        self.query_one("#inner-tabs-one", Tabs).clear()
        self.query_one("#inner-tabs-two", Tabs).clear()


if __name__ == "__main__":
    app = TabsApp()
    app.run()
```

**Please review the following checklist.**

- [ ] Docstrings on all new or modified functions / classes 
- [ ] Updated documentation
- [ ] Updated CHANGELOG.md (where appropriate)

PR #3602, opened 2023-10-27
Handle other Tabs widgets as DOM descendants of a TabbedContent

darrenburns:


TomJGooding:
Thanks Darren, this looks like a better way of handling this. I still don't quite understand how `Tabs` would still have an active tab after a `Cleared` message though which is why tests were failing on my attempt #3444.

I think the docstring is a bit misleading for `_on_tabs_cleared`, as I think this message is posted when there are no active tabs, not when all tabs are removed.

darrenburns:
@TomJGooding I think on your attempt, the `tab_count != 0` early return inside the tabs cleared wasn't correct, because a `Cleared` message is sent when all tabs are _hidden_ (which feels wrong, but that's a different story).

So, when that watcher was called, the tab_count was actually 2, triggering the early return. Since this early return prevented `TabbedContent.active` from being updated, it remained stuck on its last value, and the test failed.

I don't think I've explained great, but the whole flow is very confusing with lots of messages, reactive updates, and side effects happening - it's really not easy to trace execution here.

TomJGooding:
Sorry just to confirm I meant my change to the early return in https://github.com/Textualize/textual/pull/3444#issuecomment-1747382544. I tried both `active` and `active_tab` (which I always mix up!) but still had the failing tests, suggesting it still has an active tab? I agree the flow is a pretty confusing, so maybe I'm just misunderstanding something here.

darrenburns:
I _think_ that's a race-condition. `active_tab` is updated after refresh (via `call_after_refresh`), so when you check it it may not have been updated yet. `active` is updated immediately, so maybe using `active` instead of `active_tab` would work there... but I'm not sure.

darrenburns:
Actually - this doesn't work as the `Tabs` isn't a direct child of the `TabbedContent`. 😞

src/textual/widgets/_tabbed_content.py
```diff
@@ -356,35 +386,51 @@ def compose_add_child(self, widget: Widget) -> None:
...
+            self.post_message(
+                TabbedContent.TabActivated(
+                    tabbed_content=self,
+                    tab=event.tab,
+                )
             )
-        )
 
     def _on_tabs_cleared(self, event: Tabs.Cleared) -> None:
         """All tabs were removed."""
```
TomJGooding, line 403:
```suggestion
        """There are no active tabs."""
```

TomJGooding (commented):
Thanks Darren, this obviously was more involved than I expected from by attempt!

My only suggestion is could we also change this misleading docstring please?

src/textual/widgets/_tabbed_content.py
darrenburns: Good shout - updated.

rodrigogiraoserrao (approved):

PR #3614, opened 2023-10-30
Validate `Select.value` and minor `Select` improvements.

### LINKED ISSUE #3612: Select.value isn't validated
As it stands, the reactive `Select.value` can be in an invalid state if you assign to it an option that the select doesn't know of.
Visually, the widget looks ok but the reactive will retain its bad value.
This could be fixed with a validator method, for example.

rodrigogiraoserrao:
This PR will now validate changes made to `value`.
Fixes #3612.


davep (commented):
As I'm testing this, it seems that the PR is designed to set the `value` to `None` if an unknown value is given. Are we sure this is the correct approach, especially given the value is sort of akin to an `id` and there's other parts of Textual where using an unknown ID would result in an exception.

Also, this would seem to allow `value` to be set to `None`, via an invalid value, if `allow_blank=False`. For example, if I have this code:

```python
from textual import on
from textual.widgets import Button, Label, Select

class BadSelectApp(App[None]):

        yield Select[int]((
            ("Filthy", 0),
            ("Rich", 1),
            ("Catflap", 2),
        ), allow_blank=False, value=1)
        yield Label()
        yield Button("Make bad")

    @on(Select.Changed)
    def show_value(self) -> None:
        self.query_one(Label).update(f"{self.query_one(Select).value!r}")

    @on(Button.Pressed)
    def make_bad(self) -> None:
        self.query_one(Select).value = "nope"
        self.show_value()

if __name__ == "__main__":
    BadSelectApp().run()
```

which is designed so that it can never be blank, the bad value lets me sneak a blank in there. Is that intentional?

rodrigogiraoserrao:
With this PR I only wanted to iron out _some_ wrinkles on the `Select` widget.
After all, the issues you're pointing out were already existing issues.

However, if you'd like, I can grow this PR further if you have preferences with regard to how you'd like to see the matters discussed below addressed.

> It seems that the PR is designed to set the `value` to `None` if an unknown value is given.

Currently, setting `Select.value` to an unknown value goes through, so setting to `None` seems better in that it is a value that we can expect the `Select.value` to have.
If you'd prefer that setting to an invalid value leaves `Select.value` unchanged, I don't oppose.
I don't see any other options, though.

> Also, this would seem to allow `value` to be set to `None`, via an invalid value, if `allow_blank=False`. [...] Is that intentional?

Not really, it just follows from the work done above.
If you create an instance of `Select` with `allow_blank=False`, it already starts off with the value `None` either way, so this doesn't seem like a bad consequence.
However, if you think that makes sense, I'd be happy to tweak this so that `Select(..., allow_blank=False)` has the first option selected by default and so that setting unknown values will leave `Select.value` unchanged.

davep:
My sense for this widget is that:

- Setting `value` to an invalid value should probably raise an exception.
- If it isn't raising an exception then setting it to `None` makes sense if `allow_blank=True`.
- If it isn't raising an exception and `allow_blank=False` and there is no `value=` on construction, setting it to `None` does seem like the only way to go, albeit a bit confusing (although I take your point about it being the default if `value=` isn't provided).
- If it isn't raising an exception and `allow_blank=False` and there *is* a `value=` on construction (and it is valid, of course), then setting `value` to that would seem to make sense if the value provided is invalid.

I think though, my sense for this, is that setting `value` to an invalid value should be an error, resulting in an exception.

darrenburns:
I agree that if it gets set to an invalid value then an exception should be raised.

davep (approved):

PR #3624, opened 2023-11-01
Document CellKey being a RowKey, ColumnKey tuple

pawamoy:

- [x] Updated documentation

---

Following https://github.com/Textualize/textual/discussions/3623.

An alternative is to add docstrings to the `row_key` and `column_key` attributes of the `CellKey` class. (UPDATE: also added)

### DIFF
--- a/src/textual/widgets/_data_table.py
+++ b/src/textual/widgets/_data_table.py
@@ -138,7 +138,9 @@ class CellKey(NamedTuple):
     can still be used to retrieve it, regardless of where it currently is."""
 
     row_key: RowKey
+    """The key of this cell's row."""
     column_key: ColumnKey
+    """The key of this cell's column."""
 
     def __rich_repr__(self):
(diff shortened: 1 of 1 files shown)


src/textual/widgets/_data_table.py
```diff
@@ -131,6 +131,8 @@ class ColumnKey(StringKey):
 class CellKey(NamedTuple):
     """A unique identifier for a cell in the DataTable.
 
+    A cell key is a `(row_key, column_key)` tuple.
```
davep, line 134: As per your main comment, I think if it were me I'd be tempted to do both: make a note to this effect here, but also add docstrings to the two properties as well.

pawamoy: Done, WDYT?

davep: I think it's perfect. :-)

davep (approved):

Issue #2369, opened 2023-04-25
Remove ScrollableContainer


willmcgugan:
I don't think we need a base class here (sorry dave).

The Vertical and Horizontal scrollable containers should have the key bindings, without inheritance.

davep:
In a subsequent conversation, reviewing why we introduced this last week and the possible use cases for it, we've decided that it will remain. However, before closing this issue, I'm going to have one last wander over where it's used and how it's documented and be sure it's all fine.

PR #3632, opened 2023-11-02
fix(scrollable container): receive focus

TomJGooding:
Fixes #3629.

- [ ] Updated documentation


davep:
IIRC `ScrollableContainer` is intended more as a base class for the likes of `HorizontalScroll` and `VerticalScroll`, originally taking those bindings out of `Widget` (I think) and moving them down a level to solve a few other issues. I do seem to remember that it was deliberate that it didn't receive focus by default, but I can't recall the reason now.

Might be worth a wee dig to check what the reasoning was; and also consider if this change makes sense (I suspect it does), or if a `BothScroll` (not that name really, but you get the idea) container should be introduced too.

(none of the above is an objection to this change; just a while-eating-lunch "this should be considered")

TomJGooding:
Thanks Dave. I found #2369 where it was decided `ScrollableContainer` would remain. The changes to the docstring in linked PR #2377 suggests to me this is intended as a 'public' widget? It sounds like conversations about this were offline though...

davep:
@TomJGooding Looks like you're right there; in which case that would suggest that `ScrollableContainer` *is* `BothScroll` so this PR makes sense. And if so, we can lose the `can_focus=True` on `VerticalScroll` and `HorizontalScroll` (and really we can possibly drop the `layout: vertical;` on `VerticalScroll` too).

TomJGooding:
Thanks for confirming. I'm not sure how you're notified by PR changes, but I've now removed the `can_focus` where now inherited. I left the layout CSS alone as I wasn't sure about removing it.

willmcgugan (approved):
Thanks

PR #3636, opened 2023-11-03
Fix the Focusable flag in the Markdown docs

davep:
It said it was focusable; it isn't.

### DIFF
--- a/docs/widgets/markdown.md
+++ b/docs/widgets/markdown.md
@@ -4,7 +4,7 @@
 
 A widget to display a Markdown document.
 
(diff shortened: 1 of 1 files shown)


rodrigogiraoserrao (approved):
I like how you approached the solution to this problem.
I think I would've done it differently, so it's always good to see other ways of looking at problems.

darrenburns:
@rodrigogiraoserrao Agreed

rodrigogiraoserrao:
> @rodrigogiraoserrao Agreed

Great minds think alike.

Issue #3637, opened 2023-11-03
Tree.refresh_line can fail to re-render.


paul-ollis:
This can be demonstrated using this program.

```python
from rich.style import Style
from rich.text import Text
from textual.app import App, ComposeResult
from textual.widgets import Button, Tree
from textual.widgets.tree import TreeNode


class MyTree(Tree):
    """Subclass of Tree."""

    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.test_counter = 0
        self.render_history: set[str] = set()

    def render_label(
            self, node: TreeNode, base_style: Style, style: Style
        ) -> Text:
        """Render nodes to include the test_counter."""
        node_label = node._label.copy()
        node_label.stylize(style)
        prefix = (f'{self.test_counter} ', base_style)
        text = Text.assemble(prefix, node_label)
        self.render_history.add(text.plain)
        return text


class TreeApp(App[None]):
    """Test tree app."""

    def compose(self) -> ComposeResult:
        yield MyTree("Test")
        yield Button('Click me')

    def on_mount(self) -> None:
        tree = self.query_one(Tree)
        tree.root.expand()
        node = tree.root.add_leaf('Leaf')

    def on_button_pressed(self, event: Button.Pressed) -> None:
        tree = self.query_one(Tree)
        tree.test_counter += 1
        tree.refresh_line(tree.root.children[0].line)


if __name__ == '__main__':
    app = TreeApp()
    app.run()
```

Pressing the button has no effect on the Leaf node. It stays showing as '0
Leaf'.  A fix for this is to change refresh_line to increment he node's
`_updates` attribute.

```python
    def refresh_line(self, line: int) -> None:
        """Refresh (repaint) a given line in the tree.

        Args:
            line: Line number.
        """
        node = self.get_node_at_line(line)
        if node:
            node._updates += 1
        region = Region(0, line - self.scroll_offset.y, self.size.width, 1)
        self.refresh(region)
```

Then the tree node's representation updates when the button is clicked.

# Textual Diagnostics

## Versions

| Name    | Value  |
|---------|--------|
| Textual | 0.41.0 |
| Rich    | 13.6.0 |

## Python

| Name           | Value                                                                        |
|----------------|------------------------------------------------------------------------------|
| Version        | 3.11.3                                                                       |
| Implementation | CPython                                                                      |
| Compiler       | GCC 11.3.0                                                                   |
[...]

PR #3639, opened 2023-11-04
Fix for issue #3637.

paul-ollis:

- [ ] Updated documentation


rodrigogiraoserrao (changes requested):
Thanks for the PR!

Before I approve the PR, do you think you could simplify the test so that it is focused on making sure that `refresh_line` doesn't use cached results instead of having such a large tree subclass and app?

Also, please make sure that the regression test(s) link to the original issue: #3637.

If you're not interested in making these changes yourself, then just let us know & please make sure that you tick the box that lets maintainers make changes to your PR or we might have to make a new PR and close yours :/

Thanks a lot, Rodrigo

paul-ollis:
I'll look making the changes suggested.

paul-ollis:
Hi Rodrigo,

I have added references to issue 3637. However, I have struggled to find a way
to make things simpler. 

I may be missing something and am more that happy for you to give me a steer in
the right direction. My basic problem is that I cannot figure out how to
determine that a given line has actually been re-rendered, using just a plain
Tree instance. (I have spent some time looking at the other Tree tests and the
Tree source code.)

I discovered this issue as a result of subclassing Tree, which may have pushed
me towards a subclassing based test. I am totally in favour of keeping tests as
focused and simple as possible, so I am more than willing to do the work on
this.

I believe you have edit access, so if it is easier for you to make the changes
instead of guiding me, please do.

Cheers,
Paul

willmcgugan:
@davep Would you mind looking at this one.

davep:
Sure thing.

davep:
I think I might be struggling to understand the motivation behind this PR. Looking at the linked issue, and testing out the code in it against `main` and also this PR, I can see the difference, but I'm struggling to understand why it's a difference someone would want. The example code seems to be subverting the whole idea of having a `label` property on the nodes, and even when it does refresh, it's still causing a refresh in a way that's surprising (only one node gets refreshed, but all nodes will end up with different labels when a whole-widget refresh happens via some other means, giving the impression that the tree is generally out of sync anyway).

Or, put another way, if `render_label` is going to be overridden in a way that a change to a variable is going to imply that *all* labels are invalidated, does it make sense to concentrate on just a single line?

As an illustration of my confusion about this PR, it feels to me like everything is taken care of with a simple call to `Tree._invalidate` when the prefix used by `render_label` gets updated.

```python
from rich.style import Style
from rich.text import Text

from textual import on
from textual.widgets import Button, Tree
from textual.widgets.tree import TreeNode

class PrefixedTree(Tree):

    counter: var[int] = var(0, init=False)

    def render_label(self, node: TreeNode, base_style: Style, style: Style) -> Text:
        return Text.assemble(f"{self.counter} ", super().render_label(node, base_style, style))

    def watch_counter(self) -> None:
        self._invalidate()

class PrefixedTreeApp(App[None]):

    prefix: var[int] = var(0)

        yield PrefixedTree("Root")
        yield Button("Increment prefix")

    def on_mount(self) -> None:
        tree = self.query_one(Tree)
        tree.root.expand()
        for n in range(10):
            node = tree.root.add(f"Child {n}")
            for m in range(10):
                node.add_leaf(f"Grandchild {m}")

    @on(Button.Pressed)
    def bump(self) -> None:
        self.query_one(PrefixedTree).counter += 1

if __name__ == "__main__":
    PrefixedTreeApp().run()
```

To be clear: none of what I say here is implying that the PR doesn't make sense, it's there to document my confusion over what the motivating problem is and how this change would solve it.

paul-ollis:
@davep A reasonable question.

My specific use case that triggered discovery of this issue is as follows.

I have an application that presents a tree of pytest test IDs. The user is
allowed to select individual tests (tree leaves) or who groups of tests
(branches) to be executed. 

The rendering of the label is used to indicate the 'selectedness' of the leaf
or part of the tree. When the user makes achnage to the selection, one or more
nodes' labels need to be 're-rendered' and the refresh_line() method appears to
the natural way to make this happen.

mytree

davep:
Thanks @paul-ollis -- that makes a ton more sense now. 

Out of curiosity, do you feel that refreshing the line makes more sense than refreshing the affected nodes?

paul-ollis:
I am fairly sure that I could achieve the same results by setting the label property in response to user selection changes, but using render_label to reflect state changes feels more natural to me.

davep:
Oh, no, I don't mean *not* overriding `render_label`, I mean telling the tree it should refresh a node (the thing that needs refreshing) vs telling it to refresh a line (the thing that happens to contain the thing that needs to refresh).

I'm just musing over if the API for `Tree` is unhelpful here, perhaps.

paul-ollis:
@davep Re refreshing nodes. Do you mean should there be a refresh_node method()? If so, then I would say yes. I found having to work from the node to a line a bit awkward.

paul-ollis:
@davep. Sorry for the confusion over my last but one post. Just bad timing. I was not trying to answer the 'refresh_node' question.

davep:
> Do you mean should there be a refresh_node method()? If so, then I would say yes.

Right, yeah, exactly that. So I think you could achieve that right now, in a roundabout way, by assigning a a node's label to its label. Eg, if I take my example above:

```python
from rich.style import Style
from rich.text import Text

from textual import on
from textual.widgets import Button, Tree
from textual.widgets.tree import TreeNode

class PrefixedTree(Tree):

    counter: var[int] = var(0)

    def render_label(self, node: TreeNode, base_style: Style, style: Style) -> Text:
        return Text.assemble(f"{self.counter} ", super().render_label(node, base_style, style))

class PrefixedTreeApp(App[None]):

        yield PrefixedTree("Root")
        yield Button("Increment prefix")

    def on_mount(self) -> None:
        tree = self.query_one(Tree)
        for n in range(10):
            tree.root.add_leaf(f"Child {n}")
        tree.root.expand_all()

    @on(Button.Pressed)
    def bump(self) -> None:
        tree = self.query_one(PrefixedTree)
        tree.counter += 1
        tree.root.children[0].label = tree.root.children[0].label

if __name__ == "__main__":
    PrefixedTreeApp().run()
```

(where I'm forcing a refresh of the first child node).

paul-ollis:
As it happens, "assigning a a node's label to its label" is exactly the work-around I currently have in place.

davep:
Got to love it when a conversation comes full circle! ;-)

So... yeah, how do we feel about taking that approach then (well, not the re-assign label approach, but a node refresh approach)? I'm still wondering if `refresh_line` should even be part of the public API? On the surface it feels to me more like it makes sense to have a `refresh` method on `TreeNode`.

paul-ollis:
I agree on all points.

1. TreeNode should have a refresh method.

2. Which make refresh_line not useful as part of the public API. Removal from
   the API seems sensible to me.

I am happy to make the changes if that would be helpful.

--

Now. About how to simplify the test as requested by Rodrigo ...

paul-ollis:
I took a punt and implemented my understanding of the changes because it was a quick and easy change. I will happily undo them if required.

davep:
Thanks @paul-ollis, that's cool. I'll have a look over it; I'll also see if I can help wrangle the tests.

davep:
> [the PR should have a test that] is focused on making sure that refresh_line doesn't use cached results

The more I look at this, the more I think there's no simple way of doing this, and the more I feel it's an incorrect focus on some internal detail of `Tree`, when this is about adding a public interface. I can see any number of ways of usefully checking that `TreeNode.refresh` (as this has moved towards) gets called *at least for the refreshed node*, but I fail to see any clean and future-proof way of checking that the cache wasn't hit.

Moreover, no other aspect of the testing for `Tree` tests this level of internal implementation detail.

I'll doodle on this a wee bit more, but I'm not seeing a clean approach to this, and neither do I see any real benefit in it. If the internal cache of `Tree` needs unit testing, I sense that should be better exposed and should have a set of tests of its own.

davep:
During standup yesterday (2023-11-21) evening, it was decided that we'd take a fairly simple and pragmatic approach to adding a test for this; moreover, this PR has turned into a new feature PR, adding `TreeNode.refresh`.

At this point I think we're good to go on this.

davep:
@paul-ollis When you get a moment could you give this one last glance over and confirm it's a change that lets you do what it is you need, and then I think we can put it back up for review and look to merging.

paul-ollis:
To @davep and the rest of the team.
Thanks for the care and effort put into this. 
I can confirm that this works perfectly in my application and the resulting code is cleaner and easier to understand.
--
Paul

davep:
@paul-ollis Cool, and thanks for kicking this off!

willmcgugan (changes requested):
Can we have a CHANGELOG entry please, for the new method, and also a breaking change by making `refresh_line` private.

davep:
> Can we have a CHANGELOG entry please, for the new method, and also a breaking change by making refresh_line private.

Done.

willmcgugan (approved):

PR #3640, opened 2023-11-06
Set pycln pre-commit hook language_version to 3.8

darrenburns:
`pycln` doesn't work on 3.7 unless a Rust compiler is available.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -31,6 +31,6 @@ repos:
     rev: v2.3.0
     hooks:
       - id: pycln
-        language_version: "3.8"
+        language_version: "3.11"
         args: [--all]
 exclude: ^tests/snapshot_tests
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -31,5 +31,6 @@ repos:
     rev: v2.3.0
     hooks:
       - id: pycln
+        language_version: "3.8"
```
rodrigogiraoserrao, line 34: Do we want to specifically say 3.8 or something like >3.8?
If we're pinning a specific version, why not the latest?

darrenburns: You can't use that kind of syntax in this context.

Updated to 3.11 to match the Black hook. I won't use 3.12 because I think it's too new and there are still some projects missing new binary wheels for it.

PR #3641, opened 2023-11-06
Command palette query isolation

### LINKED ISSUE #3633: Background queries on `App` when command palette is shown causes an exception
Technically this isn't an issue with the command palette, it would be an issue with *any* screen with a background process that causes a query to be handled by the app, but via the command palette is the way that a dev would most likely encounter it and it's where they'd find it most surprising.

Consider this code:

```python
from textual.widgets import Log

class CommandPaletteIsolation(App[None]):

        yield Log()

    def tick(self) -> None:
        self.query_one(Log).write_line("Tick")

    def on_mount(self) -> None:
        self.set_interval(1, self.tick)

if __name__ == "__main__":
    CommandPaletteIsolation().run()
```

Let it run for a moment, then summon the command palette and keep it up. The next time `tick` runs a `NoMatches` will result.

This version, where the default screen isn't being used, isn't a problem:

```python
from textual.screen import Screen
from textual.widgets import Log

class MyScreen(Screen):

        yield Log()

    def tick(self) -> None:
        self.query_one(Log).write_line("Tick")

    def on_mount(self) -> None:
        self.set_interval(1, self.tick)

class CommandPaletteIsolation(App[None]):

    def on_mount(self) -> None:
        self.push_screen(MyScreen())

if __name__ == "__main__":
    CommandPaletteIsolation().run()
```

davep:
If a query was made against the app while the command palette was active it would at best result in an unexpected result, and at worst cause a crash; this could apply if there was some sort of background process constantly performing such a query. This PR isolates the command palette from such queries by always returning the calling screen from `App.children` if the command palette is active.

Fixes #3633.


davep:
Putting this up for review despite CI failing; as the fails seem to have nothing to do with this code, and I've tried to diagnose the issue but I'm struggling to figure out what's going on and I don't want to lose any more time on that.

willmcgugan:
You might have to explain this one to me when I'm back in the office tomorrow.

davep:
Sure, although there is a very simple illustration of the problem in the issue (#3633) this addresses.

davep:
Converting back to draft: while we only have the one "system screen" at the moment (the command palette, obviously), there's some scope for introducing the concept of "we have system screens and the dev should not be thrown off by them", and this PR should introduce this.

So I'll rework this a wee bit to introduce this, rather than special-case the command palette (in the end it *will* special case the command palette, it'll just be special-casing with extra steps).

davep:
Wee reminder that this has been back out of draft for a while now and looking for review.

src/textual/screen.py
```diff
@@ -1067,3 +1067,15 @@ def __init__(
     ) -> None:
         super().__init__(name=name, id=id, classes=classes)
         self._modal = True
+
+
+class _SystemModalScreen(ModalScreen[ScreenResultType], inherit_css=False):
```
rodrigogiraoserrao, line 1072: Shouldn't this be more of a mixin kind of thing, `_SystemScreen` or something like that?

It'd let `CommandPalette` inherit cleanly from `ModalScreen` and it'd let the coming system screens inherit from whatever type of screen they want (regular one or modal).

tests/command_palette/test_declare_sources.py
```diff
@@ -38,10 +39,8 @@ class AppWithSources(AppWithActiveCommandPalette):
 async def test_app_command_sources() -> None:
     """Command sources declared on an app should be in the command palette."""
     async with AppWithSources().run_test() as pilot:
-        assert (
-            pilot.app.query_one(CommandPalette)._provider_classes
-            == AppWithSources.COMMANDS
-        )
```
rodrigogiraoserrao, line 44: Does this change mean we can no longer query for the `CommandPalette`?
Don't we find this awkward?

rodrigogiraoserrao (approved):

src/textual/screen.py
```diff
@@ -1067,3 +1067,15 @@ def __init__(
...
+    """A variant of `ModalScreen` for internal use.
+
+    This version of `ModalScreen` allows us to build system-level screens;
+    the type being used to indicate that the screen should be isolated from
+    the main application.
+
+    Note:
+        This screen is set to *not* inherit CSS.
```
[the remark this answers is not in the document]
davep: When a dev styles their app `Screen ...` we don't really want it obliterating the styling of (for example) the command palette.

src/textual/command.py
```diff
@@ -519,7 +519,7 @@ def _on_click(self, event: Click) -> None:
             self.dismiss()
 
     def on_mount(self, _: Mount) -> None:
```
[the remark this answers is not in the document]
davep: Sure. GitHub will let us know.

src/textual/screen.py
davep: > Shouldn't this be more of a mixin

I'll let *you* have that conversation with Will. :-)

tests/command_palette/test_declare_sources.py
davep: I don't think there's ever been a reasonable moment or reason for anyone to need to query for the `CommandPalette`. Probably the only time you'd want to look for it is to check if it's open, and [there's a method for that](https://textual.textualize.io/api/command/#textual.command.CommandPalette.is_open).

Issue #3439, opened 2023-10-02
`blur` is listed as a valid pseudo class but doesn't seem to do anything


davep:
Currently, `blur` is [listed as a valid pseudo class](https://github.com/Textualize/textual/blob/89991025e9c21fc293ee3233fde5c0b4f679061b/src/textual/css/constants.py#L64-L74), and as such is listed as an option if you make a mistake when specifying a pseudo class in your stylesheet. However, using it doesn't seem to do anything, and it's not handled [with the other pseudo classes](https://github.com/Textualize/textual/blob/89991025e9c21fc293ee3233fde5c0b4f679061b/src/textual/widget.py#L2739-L2771). This has caused at least one person to be confused that it's shown as a valid pseudo class while apparently doing nothing when used.

I suspect we either need to enable it as a valid pseudo class, or perhaps remove it from the list of valid pseudo classes?

davep:
In in-person conversation, the view seems to be that it would make sense to enable `:blur` for use.

willmcgugan:
https://github.com/Textualize/textual/pull/3645

PR #3645, opened 2023-11-07
Blur

willmcgugan:
Adds `:blur` pseudo class 

Fixes https://github.com/Textualize/textual/issues/3439


darrenburns (commented):
I think this should have a simple pilot test which checks that when something is focused it does not have the `blur` pseudoclass, and when the focus is removed it gains the `blur` pseudoclass. The existing tests don't seem to cover this.

davep (approved):

willmcgugan:
@darrenburns Added test

darrenburns (approved):

PR #3646, opened 2023-11-07
optimization

willmcgugan:
A small optimization to calculate applicable rules a little faster.

### DIFF
--- a/src/textual/css/stylesheet.py
+++ b/src/textual/css/stylesheet.py
@@ -440,7 +440,7 @@ class Stylesheet:
         css_path_nodes = node.css_path_nodes
 
         rules: Iterable[RuleSet]
-        if limit_rules:
+        if limit_rules is not None:
             rules = [rule for rule in reversed(self.rules) if rule in limit_rules]
         else:
             rules = reversed(self.rules)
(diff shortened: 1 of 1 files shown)


src/textual/css/stylesheet.py
```diff
@@ -636,18 +636,28 @@ def update_nodes(self, nodes: Iterable[DOMNode], animate: bool = False) -> None:
...
-                for name in node._selector_names
-                if name in rules_map
+                for name in rules_map.keys() & node._selector_names
                 for rule in rules_map[name]
             }
-            apply(node, limit_rules=rules, animate=animate)
+
+        for node in nodes:
+            rules = get_rules(node)
+            if rules:
```
darrenburns, line 649: Possibly silly question:

If `rules` is an empty set, would this prevent the update from working in the case where we go from some rules to no rules on the node? Or can rules not be empty?

darrenburns (approved):
I left one question - but happy for you to merge if that's not an issue.

src/textual/css/stylesheet.py
willmcgugan: If `rules` is empty, then it means that none of the selectors could possibly match the node. i.e something completely unstyled.

PR #3647, opened 2023-11-07
Fix erroneous ButtonDown mouse event reporting.

EricWF:
After v0.38.0, I was seeing Input fields lose focus whenever the mouse left the input field area (but without having clicked anywhere). This behavior started after the release v0.38.0 release when Mouse ButtonDown began to trigger loss of focus (This change seemed intentional). However, I wasn't pressing the mouse button down.

The erronious ButtonDown events were actually MouseMove events with no keys held down. According to the docs for xterm escape sequences, when SGR (1006) mode is enabled, the encoded button value isn't always incremented by 32 for move movements events. [1]

I believe the correct fix to this issue is to detect mouse movement events with no button down by checking if the "pressed button" is 0. Which appears to indicate that no mouse button is pressed (in this specific case). Whereas, the button value will be set to 1 when left click is pressed for example.

I'm not sure if this change is fully correct for terminals which don't support SGR mode.

[1] https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Extended-coordinates

- [x] Updated documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -252,6 +252,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Fixed CSS watcher crash if file becomes unreadable (even temporarily) https://github.com/Textualize/textual/pull/4079
 - Fixed display of keys when used in conjunction with other keys https://github.com/Textualize/textual/pull/3050
 - Fixed double detection of <kbd>Escape</kbd> on Windows https://github.com/Textualize/textual/issues/4038
+- Fixed erroneous mouse 'ButtonDown' reporting for mouse movement when any-event mode is enabled in xterm. https://github.com/Textualize/textual/pull/3647
+
 
 ## [0.47.1] - 2024-01-05
 
--- a/src/textual/_xterm_parser.py
+++ b/src/textual/_xterm_parser.py
@@ -73,12 +73,13 @@ class XTermParser(Parser[events.Event]):
                 )
                 button = 0
             else:
-                if buttons & 32:
+                button = (buttons + 1) & 3
+                # XTerm events for mouse movement can look like mouse button down events. But if there is no key pressed,
+                # it's a mouse move event.
+                if buttons & 32 or button == 0:
                     event_class = events.MouseMove
                 else:
                     event_class = events.MouseDown if state == "M" else events.MouseUp
-
-                button = (buttons + 1) & 3
 
             event = event_class(
                 x,
--- a/tests/test_xterm_parser.py
+++ b/tests/test_xterm_parser.py
@@ -218,6 +218,7 @@ def test_mouse_click(parser, sequence, event_type, shift, meta):
         ("\x1b[<35;15;38M", False, False, 0),  # Basic cursor movement
         ("\x1b[<39;15;38M", True, False, 0),  # Shift held down
         ("\x1b[<43;15;38M", False, True, 0),  # Meta held down
+        ("\x1b[<3;15;38M", False, False, 0),
     ],
 )
 def test_mouse_move(parser, sequence, shift, meta, button):


davep:
I'm not seeing the effect you mention in my usual environment (macOS, kitty); so we can recreate and then test out this PR, what environment are you seeing this in?

EricWF:
Oh, good question. After some testing this appears to only reproduce inside Intellij's builtin terminal implementation.
It does not seem to reproduce with gnome's terminal implementation. 
I'm not sure exactly what to provide, so let me know if you would like something in particular. Here are some relevant environment variables:

```
JEDITERM_SOURCE_ARGS=
FIG_JETBRAINS_SHELL_INTEGRATION=1
TERM=xterm-256color
LANG=en_US.UTF-8
XMODIFIERS=@im=ibus
XDG_SESSION_TYPE=x11
TERMINAL_EMULATOR=JetBrains-JediTerm
XDG_VTNR=2
VTE_VERSION=7400
SHELL=/usr/bin/fish
COLORTERM=truecolor
PLATFORM_TYPE=Linux
```

I could dig more into whether the bug is in Intellij's terminal. But the behavior seems potentially inline with the very hard to understand documentation about terminal modes.

Testing with my fix applied causes mouse events in TextArea an Input to match across both terminal implementations, and brings the behavior with the Intellij terminal into line with Gnome terminals existing behavior.

EricWF:
Ah, sorry I just saw your commit. Here's some of the information you need...

You can find a video of the bug & fix [here](http://efcs.us/public/term-vid.html)

```
<!-- This is valid Markdown, please paste the following directly in to a GitHub issue -->
# Textual Diagnostics

## Versions

| Name    | Value  |
|---------|--------|
| Textual | 0.40.0 |
| Rich    | 13.6.0 |

## Python

| Name           | Value                               |
|----------------|-------------------------------------|
| Version        | 3.11.6                              |
| Implementation | CPython                             |
| Compiler       | GCC 13.2.0                          |
| Executable     | /home/eric/.pyenvs/idea/bin/python3 |

## Operating System

| Name    | Value                                              |
|---------|----------------------------------------------------|
| System  | Linux                                              |
| Release | 6.5.0-2-amd64                                      |
| Version | #1 SMP PREEMPT_DYNAMIC Debian 6.5.6-1 (2023-10-07) |

## Terminal

| Terminal Application | PyCharm        |

## Rich Console options

| size           | width=112, height=35 |
| max_width      | 112                  |
| max_height     | 35                   |
```
Additionally, the output captured from _xterm_parser.py can be found below.
[mousecodes.log](https://github.com/Textualize/textual/files/13324144/mousecodes.log)

EricWF:
@davep Let me know if I can provide more.

davep:
I suspect everything necessary is here; just needs one of us to find the time to recreate the problem and review (personally I'm on holiday at the moment). 

Might take a wee while (lots of holiday all around heading into this time of year, plus lots going on around Textual), but we'll get to it.

EricWF:
@davep Sounds good. Enjoy your time away. 

I would love to land this before the next release. Because I use textual everywhere, having to patch it is :-( 

PS. I fixed the link for the video.

EricWF:
Hey friends, Happy Holidays! 

I see there have been a couple of releases since this PR has been created. Unfortunately none of them contain a fix for the issues described here.

How can I help this land?

willmcgugan:
We have a backlog. But I will be looking at this soon...

EricWF:
@willmcgugan Please let me know if I can help move things along.

willmcgugan:
Would you mind creating an issue for this. Or linking to it if it already exists.

willmcgugan:
I haven't manage to reproduce this, but it doesn't break anything. Thanks for the PR, and sorry it took so long to go in.

PR #3648, opened 2023-11-08
More `Makefile` tweaks

davep:
Ensure that the `syntax` extras gets installed as part of `make setup`, and also add `make repl`[^1].

[^1]: Sure, I'm probably the only one who'll ever use it, but it's a handy thing I have within every Python project and my muscle memory fails me every time I'm in the Textual development directory and every time I keep thinking "I should add that" so here I am adding that.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -85,6 +85,7 @@ clean: clean-screenshot-cache clean-offline-docs
 .PHONY: setup
 setup:
 	poetry install
+	poetry install --extras syntax
 
 .PHONY: update
 update:
--- a/Makefile
+++ b/Makefile
@@ -98,3 +98,7 @@ install-pre-commit:
 .PHONY: demo
 demo:
 	$(run) python -m textual
+
+.PHONY: repl
+repl:
+	$(run) python


Makefile
```diff
@@ -85,6 +85,7 @@ clean: clean-screenshot-cache clean-offline-docs
 .PHONY: setup
 setup:
 	poetry install
+	poetry install --extras syntax
```
darrenburns, line 88: You only need `poetry install --extras syntax`, which does both the normal dependencies plus the extras.

davep: Will it leave the normal dependencies intact of the extras fail?

darrenburns: That I don't know

darrenburns (approved):

Makefile
davep: Quickly testing, fresh 3.7 environment, doing the single-line version, stuff installs but you seem to end up with a non-working venv. On the other hand the two-step approach does leave you with a working venv, which is what I was aiming for (you'll still get errors, but at least the venv will be usable).

PR #3654, opened 2023-11-08
Fix live-reloading of screen CSS.

### LINKED ISSUE #3454: Live editing doesn't work if editing screen's CSS
Hi! Seems like the live editing doesn't work if you try to edit CSS for a screen (not the app itself).

I've created a repo so you could quickly reproduce the issue: https://github.com/Klavionik/textual-live-editing-issue

# Textual Diagnostics

## Versions

| Name    | Value  |
|---------|--------|
| Textual | 0.38.1 |
| Rich    | 13.6.0 |

## Python

| Name           | Value                                          |
|----------------|------------------------------------------------|
| Version        | 3.10.2                                         |
| Implementation | CPython                                        |
| Compiler       | GCC 11.3.0                                     |
| Executable     | /home/jediroman/Desktop/textual-live-editing-issue/venv/bin/python |

## Operating System

| Name    | Value                                       |
|---------|---------------------------------------------|
| System  | Linux                                       |
| Release | 5.15.0-83-generic                           |
| Version | #92-Ubuntu SMP Mon Aug 14 09:30:42 UTC 2023 |

## Terminal

| COLORTERM            | *Not set*      |

## Rich Console options

| size           | width=305, height=11 |
| max_width      | 305                  |
| max_height     | 11                   |

rodrigogiraoserrao:
Fix #3454 so that live-reloading also works on screen CSS.
The screen CSS was already being added to the file monitor but the file monitor callback sourced the files from the app and not from the monitor, so it was reloading only a subset of the files that it should.

### DIFF
--- a/tests/snapshot_tests/snapshot_apps/hot_reloading_app_with_screen_css.tcss
+++ b/tests/snapshot_tests/snapshot_apps/hot_reloading_app_with_screen_css.tcss
@@ -1,0 +1,1 @@
+/* This file is purposefully empty. */
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Duplicate CSS errors when parsing CSS from a screen https://github.com/Textualize/textual/issues/3581
 - Added missing `blur` pseudo class https://github.com/Textualize/textual/issues/3439
 - Fixed visual glitched characters on Windows due to Python limitation https://github.com/Textualize/textual/issues/2548
+- Fixed live-reloading of screen CSS https://github.com/Textualize/textual/issues/3454
 
 ### Changed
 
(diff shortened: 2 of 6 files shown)


tests/snapshot_tests/snapshot_apps/hot_reloading_app_with_screen_css.tcss
```diff
@@ -0,0 +1 @@
+/* This file is purposefully empty. */
```
davep, line 1: Point of order: it isn't empty.

davep (approved):

tests/snapshot_tests/snapshot_apps/hot_reloading_app_with_screen_css.tcss
rodrigogiraoserrao: How about "This file is purposefully devoid of TCSS rules."?

PR #3656, opened 2023-11-09
Fix for CSS outline issues

### LINKED ISSUE #3628: `outline` causes layout issues for `Button`
It might be that this is an issue elsewhere too, but it was first noticed with `Button` (#3570). Given this code:

```python

class ButtonIssue(App[None]):

    CSS = """
    Button {
        outline: white;
    }
    """

        yield Button("Test")

if __name__ == "__main__":
    ButtonIssue().run()
```

the resulting display looks like this:

Screenshot 2023-11-02 at 11 00 12

darrenburns:
`align_lines` had two problems:

- If the content was the same width as the available space, empty segments were produced.
    - This was true for both `center` and `right` align.
- It always produced one extra line of segments due to a miscalculation. I'm surprised this didn't cause issues elsewhere.
    - This appeared to be the case of all types of vertical alignment.

Fixes #3628 

## Before this PR
image

## After this PR
image


darrenburns:
@willmcgugan Probably, yeah. I can fix that too if you want (it's because `make_blank` is being called with `width=0`).

I think this function should probably still be kept to work when an empty `Segment` is in the list though rather than producing an incorrect return value.

willmcgugan:
TBH I think the fix is in preventing the empty segments.

I suspect there is a lot of code that makes the assumption there will be no empty segments. Handling that everywhere is probably too much of a burden.

src/textual/_segment_tools.py
```diff
@@ -129,19 +129,39 @@ def line_trim(segments: list[Segment], start: bool, end: bool) -> list[Segment]:
...
     """
     segments = segments.copy()
-    if segments and start:
-        _, first_segment = segments[0].split_cells(1)
-        if first_segment.text:
-            segments[0] = first_segment
-        else:
+
+    # Empty segments should be ignored as they don't contribute to cell width
+    while segments and start:
```
willmcgugan, line 134: Now that the root cause of empty segments is fixed, I'd be tempted to roll this back. We should treat "no empty segments" as an invariant.

Since it is easy to do accidentally, maybe we should put in a check wrapped with `if app.debug` that logs a warning if there are empty segments.

src/textual/_segment_tools.py
```diff
@@ -211,7 +230,7 @@ def blank_lines(count: int) -> list[list[Segment]]:
         bottom_blank_lines = vertical_excess_space
     elif vertical == "middle":
         top_blank_lines = vertical_excess_space // 2
-        bottom_blank_lines = height - top_blank_lines
+        bottom_blank_lines = vertical_excess_space - top_blank_lines
```
willmcgugan, line 233: Good catch

willmcgugan (approved):

src/textual/_segment_tools.py
darrenburns: Do you think we should put this check in this function or should it be an invariant in broader Textual that "empty, non-control Segments don't make sense"?

I've found a bunch of places that I think could produce empty Segments if we're not careful, so if you think that there's lots of code that wouldn't handle that correctly then maybe it should be a broader check?

willmcgugan: Yeah, not here. Perhaps `_styles_cache.py`.

darrenburns: Ok, I've added something to the styles cache, let me know what you think

src/textual/_segment_tools.py
```diff
@@ -121,9 +121,9 @@ def line_trim(segments: list[Segment], start: bool, end: bool) -> list[Segment]:
     """Optionally remove a cell from the start and / or end of a list of segments.
 
     Args:
-        segments: A line (list of Segments)
-        start: Remove cell from start.
-        end: Remove cell from end.
+        segments (list[Segment]): A line (list of Segments)
```
willmcgugan, line 124: I think these types get pulled in from the signature in the docs.

darrenburns: Whoops, reverted too far :)

PR #3659, opened 2023-11-09
Be consistent with line/col numbers.

### LINKED ISSUE #3625: CSS error reporting sometimes off by one
See https://github.com/Textualize/textual/pull/3582#issuecomment-1787507687.

Running the app at the bottom produces the error below, where the error reporting is off by one.
See the line above the panel and the code lines in the code snippet printed.

```
 Error in stylesheet:
 /Users/davep/develop/python/textual-upstream/sandbox/foo.py, CSSErrorApp.CSS:1:4
│   1 │                                                                                                                                                                                                                      │
│ ❱ 2 │   :                                                                                                                                                                                                                  │
│   3                                                                                                                                                                                                                        │
 • Expected one of 'comment line', 'comment start', 'selector start', 'selector start class', 'selector start id', 'selector start universal', 'variable name', or 'whitespace'.                                              
```

```py

class CSSErrorApp(App[None]):

    CSS = """
    :
    """

        yield Label()

if __name__ == "__main__":
    CSSErrorApp().run()
```

rodrigogiraoserrao:
I added some comments so that we know what line/column numbers are 0-based and which ones are 1-based.

Fixes #3625.


tests/css/test_parse.py
```diff
@@ -1248,9 +1248,27 @@ def test_parse_bad_psuedo_selector():
...
+}
+"""
+
+    stylesheet = Stylesheet()
+    stylesheet.add_source(bad_selector, None)
+
+    with pytest.raises(TokenError) as error:
+        stylesheet.parse()
+
+    assert error.value.start == (2, 7)
```
rodrigogiraoserrao, line 1274: This checks the “same thing” as the test above but it's a different code path.

darrenburns:
My initial thought on reading this is that we should be instead ensuring that everything is 0-indexed at the code level, and only converting to 1-indexed where absolutely necessary (e.g. converting the index to a line/col number for display to the user).

src/textual/css/tokenizer.py
```diff
@@ -34,9 +34,9 @@ def __init__(
         Args:
             read_from: The location where the CSS was read from.
             code: The code being parsed.
-            start: Line number of the error.
+            start: Line and column number of the error (1-indexed).
             message: A message associated with the error.
-            end: End location of token, or None if not known.
+            end: End location of token (1-indexed), or None if not known.
```
darrenburns, line 39: It seems like this was 0-indexed before, but now you've converted it to 1 indexed by ensuring we subtract 1 from all the places it was used. What do you think about keeping it zero indexed? Am I missing some context?

rodrigogiraoserrao: I don't mind keeping everything 0-indexed and converting only when about to print, but since the errors are created to be user-facing, I thought it made sense to make them 1-indexed from the get-go.

willmcgugan: As a rule of thumb, the best choice is the one that results in less `-1` and `+1` in your code.

darrenburns: I think this doesn't really matter either way, just wanted to be sure we're being consistent

darrenburns (approved):

PR #3681, opened 2023-11-15
Add missing return type and return detail for `OptionList.get_option_index`

davep:
https://github.com/Textualize/textual/commit/a3590ac19253b43bb2f92d1264ea1fd4ab981426 added this but it looks like the type hint for the return value, and the return value explanation were left off in the PR.


darrenburns (approved):

willmcgugan:
Does `option_id` not need a type?

davep:
Yes, yes it does.

PR #3683, opened 2023-11-15
Change loading indicator layer and effects.

### LINKED ISSUE #3677: Setting `loading` to `True` causes widgets to overwrite notifications
Given this code:

```python

class LoadingOverlayApp(App[None]):

    CSS = """
    Label {
        width: 90%;
        height: 1fr;
        text-align: center;
        border: solid red;
    }
    """

        yield Label("This is a really big label")

    def on_mount(self) -> None:
        self.notify(
            "This is a really big notification. Loading should not overwrite it.",
            timeout=9_999_999
        )
        self.query_one(Label).loading = True

if __name__ == "__main__":
    LoadingOverlayApp().run()
```

you get this result:

Screenshot 2023-11-14 at 15 32 15

a widget in `loading=True` state shouldn't appear to overwrite notifications.

rodrigogiraoserrao:
The loading indicator will now disable the underlying widget.
It is also no longer on top of notifications.

Fixes #3677.


willmcgugan (approved):
Good solution

PR #3692, opened 2023-11-16
Hide command palette mount and unmount handlers from the docs

davep:
CommandPalette.on_mount and CommandPalette.on_unmount were in the docs and indexed because they're not private and they have docstrings documenting what they do for the command palette. This has the unfortunate side-effect of CommandPalette.on_mount (for example) turning up as the first it if someone searches the docs for `on_mount` (which, if they've not gone through the guide and have read up on events and messages, they're likely to do and then end up confusing themselves due to not reading the docs in a sensible order):

Screenshot 2023-11-16 at 14 32 17

While it won't help someone understand what `on_mount` does, it will hide an implementation of it that won't help them.


rodrigogiraoserrao (approved):
LGTM 🐍 🚀

Issue #3705, opened 2023-11-20
Wrong Type Signature of ListView.watch_index


AulonSal:
The type signature of parameters `old_index` and `new_index` in `ListView.watch_index` should be `int | None`.
https://github.com/Textualize/textual/blob/78d9c4c59cbcacb27c0fa80935fdcbd4bb516ce3/src/textual/widgets/_list_view.py#L157C25-L157C25
```python
def watch_index(self, old_index: int, new_index: int) -> None:
```
-- should be -->
```python
def watch_index(self, old_index: int | None, new_index: int | None) -> None:
```
## Reason
`ListView.index` has type `reactive[Optional[int]]`

davep:
Good catch; thanks.

PR #3706, opened 2023-11-20
Correct the type hints for ListView.watch_index

davep:
The old and new index can be None, as highlighted in #3705.

### DIFF
--- a/src/textual/widgets/_list_view.py
+++ b/src/textual/widgets/_list_view.py
@@ -154,15 +154,17 @@ class ListView(VerticalScroll, can_focus=True, can_focus_children=False):
             return False
         return 0 <= index < len(self._nodes)
 
-    def watch_index(self, old_index: int, new_index: int) -> None:
+    def watch_index(self, old_index: int | None, new_index: int | None) -> None:
         """Updates the highlighting when the index changes."""
         if self._is_valid_index(old_index):
+            assert old_index is not None
             old_child = self._nodes[old_index]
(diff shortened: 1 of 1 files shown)


src/textual/widgets/_list_view.py
```diff
@@ -154,15 +154,17 @@ def _is_valid_index(self, index: int | None) -> bool:
...
         """Updates the highlighting when the index changes."""
         if self._is_valid_index(old_index):
+            assert old_index is not None
             old_child = self._nodes[old_index]
             assert isinstance(old_child, ListItem)
             old_child.highlighted = False
 
         new_child: Widget | None
         if self._is_valid_index(new_index):
+            assert new_index is not None
```
rodrigogiraoserrao, line 167: Why don't we make `_is_valid_index` a `TypeGuard` so that we can get rid of the `assert x_index is not None`?

davep: Good call; curiously I've never seen it used in Textual code.

rodrigogiraoserrao: Me neither. Will you do it?

willmcgugan (approved):

src/textual/widgets/_list_view.py
davep: Sorry, just saw this. I did it.

rodrigogiraoserrao (approved):

Issue #3011, opened 2023-07-25
Look into options for handling reactive-related methods firing before the DOM is mounted


davep:
There is a class of problem, that can be worked around by someone developing with Textual, where there is a `var` or `reactive` that has watch or other similar methods tied to it, that in turn try and do work that needs that the DOM is fully-mounted; but then it's possible for the developer to modify the value of the `var`/`reactive` *before* mounting has happened.

This inevitably results in DOM-related errors.

This can be best seen in #2363 and the PR that sought to fix it (#3002).

In conversation we've decided that it would be a good idea to look for a more general way of solving this problem. While `DirectoryTree`, or even `Tree`, could be modified to not have this be an issue, the fact remains that anyone building their own widgets will run into the same problem.

Current initial thinking is that perhaps `var` and `reactive` should grow some sort of `post_mount` keyword flag, which when `True` would tell Textual to delay any watch or other similar calls until *after* the DOM is mounted.

darrenburns:
We've opted against `post_mount`, but I'm out of ideas for a generic approach that is clean.

Accessing the DOM or calling `get_component_styles` inside a watcher will currently result in a crash, if a user sets the corresponding reactive before mounting it.

Closely related to https://github.com/Textualize/textual/issues/2363 which is caused by this.

@willmcgugan I'm blocked on this at the moment.

rodrigogiraoserrao:
We don't want to skip or delay calling watchers until the `Mount` event because that would probably be confusing: the user sets a reactive, expects the remainder of the widget state to update (as per the logic they might implement in the watcher) but then the watcher doesn't run because the widget hasn't been mounted yet.

The “solution” seems to be avoiding making use of the DOM in watchers and/or checking if the widget has been mounted with `is_mounted`.

PR #3709, opened 2023-11-20
feat(widget): add is_mounted property

TomJGooding:
Credit to @darrenburns for this suggestion.

This helps avoid DOM-related errors by checking if the widget is mounted before trying to do work relating to the DOM (querying, updating, etc.) For example:

```python
    def watch_some_reactive(self) -> None:
        if not self.is_mounted:
            return
        table = self.query_one(DataTable)  # Avoids NoMatches error
        ...
```

### Related Issue

- #3011 

### Checklist:

- [ ] Updated documentation

### DIFF
--- a/src/textual/widget.py
+++ b/src/textual/widget.py
@@ -394,6 +394,15 @@ class Widget(DOMNode):
     """A title to show in the bottom border (if there is one)."""
 
     @property
+    def is_mounted(self) -> bool:
+        """Check if this widget is mounted.
+
+        Returns:
+            True if this widget is mounted.
+        """
+        return self._mounted_event.is_set()
+
+    @property
     def siblings(self) -> list[Widget]:
         """Get the widget's siblings (self is removed from the return list).
 
--- a/src/textual/widget.py
+++ b/src/textual/widget.py
@@ -395,11 +395,7 @@ class Widget(DOMNode):
 
     @property
     def is_mounted(self) -> bool:
-        """Check if this widget is mounted.
-
-        Returns:
-            True if this widget is mounted.
-        """
+        """Check if this widget is mounted."""
         return self._mounted_event.is_set()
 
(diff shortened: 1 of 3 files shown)


src/textual/widget.py
```diff
@@ -393,6 +393,15 @@ def __init__(
     border_subtitle: str | Text | None = _BorderTitle()  # type: ignore
     """A title to show in the bottom border (if there is one)."""
 
+    @property
+    def is_mounted(self) -> bool:
```
davep, line 397: I see it's draft, but a wee FYI... we have the convention[^1] of: if it's a `@property`, document it as if its a property, so no `Returns` etc.

[^1]: Yes, yes, I know, right next to where you put this there's a couple of properties that violate that guideline; they're not all cleaned up from the beforetimes. :-P

TomJGooding: Haha, I actually almost didn't include this but second-guessed myself!

TomJGooding:
Something I am a little confused by is that the reactive docstring mentions watchers being called **post mount**.

https://textual.textualize.io/api/reactive/

If that was the case, obviously this workaround is unnecessary! Am I misunderstanding what is meant here?

willmcgugan:
The `init` option will call all the watchers once on mount. You can use this when a watcher needs to run when the widget is mounted.

darrenburns:
`init=True` means _Textual_ will automatically call your watchers for you once, post mount.

It doesn't affect what happens when you set a reactive, e.g. `widget.value = "hello"` will still result in `watch_value` running as soon as you set it, regardless of `init=True/False`.

TomJGooding:
Thanks both for the clarification, I think I confused myself by misreading what was meant by "on initialize" here.

willmcgugan (approved):

Issue #3687, opened 2023-11-15
Vertical scrollbar loss in horizontal layout


davep:
This is a bit of a tricky one to narrow down; it started out with a missing scrollbar in the right-most column of my [current pet project](https://github.com/davep/tinboard), and ended up with a bit of a MRE-making rabbit hole; but I think I got there. Take this code:

```python
from textual.app import App, ComposeResult
from textual.containers import Horizontal
from textual.widgets import OptionList

class MissingScrollbarApp(App[None]):

    CSS = """
    OptionList {
        height: 1fr;
    }

    #left {
        min-width: 25;
    }

    #middle {
        width: 5fr;
    }

    #right {
        min-width: 30;
    }
    """

    def compose(self) -> ComposeResult:
        options = [str(n) for n in range(200)]
        with Horizontal():
            yield OptionList(*options, id="left")
            yield OptionList(*options, id="middle")
            yield OptionList(*options, id="right")

if __name__ == "__main__":
    MissingScrollbarApp().run()
```

The numbers are *mostly* important, and derive from the layout I had in my application, and are whittled down to what I hope is the bare minimum to reproduce. Also of note is in my app it's two `OptionList`s and a `VerticalScroll` in the third column, but the above keeps it simple and has the same effect.

So, given the above, you'd expect a vertical scrollbar on each of the "panels". But, what you actually get is the third one gets cut off:

![Screenshot 2023-11-15 at 21 25 01](https://github.com/Textualize/textual/assets/28237/e0880ff6-54bc-44b8-8df0-10f6c75efd62)

If I make the terminal sufficiently wide enough, it *eventually* appears:

![Screenshot 2023-11-15 at 21 25 48](https://github.com/Textualize/textual/assets/28237/aea747da-1829-4640-836c-d562307000fb)

I had to drag this out to about 220 characters wide.

There is also an intermediate step, between no scrollbar, and scrollbar, which shows half a scrollbar:

![Screenshot 2023-11-15 at 21 27 20](https://github.com/Textualize/textual/assets/28237/393e82fa-a3dc-41c0-9ffb-9958bd32824f)

I originally tried to recreate this using just three `VerticalScroll` but couldn't get the no-scrollbar effect (but *did* get the half-scrollbar); but as soon as I threw the `OptionList` into the mix (so perhaps a `ScrollView` thing?).

Tagging @willmcgugan  for triage.

williballenthin:
[...]

PR #3712, opened 2023-11-20
fix min width

willmcgugan:
Fix min-width and friends. 

Fixes https://github.com/Textualize/textual/issues/3687


davep (approved):
Looks good, and nicely fixes where I initially noticed this.

Screenshot 2023-11-21 at 09 36 50

PR #3713, opened 2023-11-20
docs(contributing): update setup guide

TomJGooding:
Currently the contributing guidelines advise in the setup steps to:

```md
Run `poetry install` to install all dependencies
```

However this doesn't install all dependencies required for development. Running `make test` will have failing tests due to missing dependencies in the `syntax` extras.

- [ ] Updated documentation

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,9 +20,11 @@ You can follow these steps:
  1. Make sure you have Poetry installed ([see instructions here](https://python-poetry.org))
  2. Clone the Textual repository
  3. Run `poetry shell` to create a virtual environment for the dependencies
- 4. Run `make setup` to install all dependencies ([read this](#makefile-commands) if the command `make` doesn't work for you.)
+ 4. Run `make setup` to install all dependencies
  5. Make sure the latest version of Textual was installed by running the command `textual --version`
  6. Install the pre-commit hooks with the command `pre-commit install`
+
+([Read this](#makefile-commands) if the command `make` doesn't work for you.)
 
 ## Demo
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,7 +20,7 @@ You can follow these steps:
  1. Make sure you have Poetry installed ([see instructions here](https://python-poetry.org))
  2. Clone the Textual repository
  3. Run `poetry shell` to create a virtual environment for the dependencies
- 4. Run `poetry install` to install all dependencies
+ 4. Run `make setup` to install all dependencies ([read this](#makefile-commands) if the command `make` doesn't work for you.)
  5. Make sure the latest version of Textual was installed by running the command `textual --version`
  6. Install the pre-commit hooks with the command `pre-commit install`
 


CONTRIBUTING.md
```diff
@@ -20,10 +20,12 @@ You can follow these steps:
  1. Make sure you have Poetry installed ([see instructions here](https://python-poetry.org))
  2. Clone the Textual repository
  3. Run `poetry shell` to create a virtual environment for the dependencies
```
davep, line 22: While not a review of this change (which I think makes a lot of sense), while we're in here... I don't think step 3 make sense here, and the reasoning given also makes no sense. I'm no poetry expert but doesn't step 4 take care of putting the dependencies in the virtual environment anyway? That's been my experience.

I sense we should swap step 3 and 4 here, and reword the reasoning behind doing `poetry shell`

CONTRIBUTING.md
```diff
@@ -20,10 +20,12 @@ You can follow these steps:
...
- 4. Run `poetry install` to install all dependencies
+ 4. Run `make setup` to install all dependencies
  5. Make sure the latest version of Textual was installed by running the command `textual --version`
  6. Install the pre-commit hooks with the command `pre-commit install`
```
davep, line 25: An alternative would be to `make install-pre-commit`, and that `make` target handles ensuring the venv is used for that.

davep (commented):
Good catch. Noticed a couple of other things around here too so I'm thinking perhaps we should tidy up this text some more while here?

CONTRIBUTING.md
rodrigogiraoserrao: 4 will create the venv. The point of 3 was very poorly phrased. Running `poetry shell` starts a shell where the venv is activated (and maybe some other things?) so that we can run commands without having to prefix them with `poetry run` (like the makefile does).

davep: Sure, but the ordering above suggests that the venv is created as a side-effect and *then* the venv content is installed. I'm suggesting it makes more sense to two those two steps the other way round:

- Create/setup the venv
- Open a shell within the venv

At the moment, while it works, it seems the steps are the other way round, and that feels odd.

rodrigogiraoserrao: I don't mind swapping them around. I was just dropping that info so that you consider _not_ deleting that step altogether.

(Running `poetry shell` _will_ create a`.venv` if it's not there yet.)

TomJGooding: If the `shell` command is the equivilent of `source .venv/bin/activate`, the current order makes sense to me.

davep: If it was the equivalent it would fail, wouldn't it, as at that point there is no venv available to activate?

I'll stress again, of course, I'm no active poetry user (I tend to use pipenv move), but it was my understanding that poetry was similar so that if you run it, and there is no venv, it'll make it, on any invocation of it.

I agree that it's likely mostly a wording thing, but I was trying to stress that `poetry shell` wasn't at all necessary for `poetry install` to work, as `poetry install` will make (and subsequently install into) the venv anyway.

Activating the venv is, of course, necessary before step 5 and 6 (which is why I was saying that it would suggest a more correct flow of dependency if 3 and 4 were swapped).

Or.... try this right now:

1. Clone a fresh textual
2. `poetry install`

and note it builds the venv just fine.

TomJGooding: I meant `shell` is equivilent in the sense it activates the venv with the sub-shell. I do realise `install` will also create the venv, it just  _feels_ like activates the venv makes sense before installing the dependencies (like you would if not using Poetry).

If you think it makes more sense for the order to be swapped though, go for it!

willmcgugan:
@rodrigogiraoserrao @davep Is this good to go?

davep:
I don't see a blocker with it (had some wider feedback based on it, that was it). Honestly I thought it was approved already.

davep (approved):

PR #3718, opened 2023-11-21
Compositor ignores non-mounted widgets.

rodrigogiraoserrao:
By making sure that the compositor ignores widgets that haven't been mounted we can be sure that our widgets are only rendered after being mounted.

An exception to this seems to be scrollbars and scrollbar corners which are created on the fly in the `Widget` properties `horizontal_scrollbar`, `vertical_scrollbar`, and `scrollbar_corner`, because they are handled in a slightly different way (they aren't in the DOM).


willmcgugan (approved):
Looks good! Nice and simple.

I assume you've checked the regression tests fails on main?

PR #3724, opened 2023-11-22
Document DOMNode.watch

rodrigogiraoserrao:


willmcgugan (changes requested):
Can we have a fully working example. There are a few exceptions in the docs, but in general we like to have apps that can standalone.

Suggest also defining a new reactive attribute, rather than relying on an existing.

rodrigogiraoserrao:
Good call.
Here's the screenshot of the code & output in the docs:

Screenshot 2023-11-22 at 14 10 31
Screenshot 2023-11-22 at 14 10 39

willmcgugan (approved):

PR #3725, opened 2023-11-22
Fixing click targeting/tab expansion interaction

darrenburns:
Extracts logic for an operation that's required in a couple of different places.

This could be used in `Input` and navigation within the `TextArea`.

Breaks out some of the logic into a separate function which is required for wrapping in the TextArea - we need to know, on a tab by tab basis, the width they are converted to on tab expansion so that we can correctly implement the click offset -> selected character location function.

Also fixes an issue where tab width wasn't correctly being counted when you click the cursor on a line, which, to the left of the click location, contains a tab character which was not fully expanded to the tab stop width.

### DIFF
--- a/src/textual/_cells.py
+++ b/src/textual/_cells.py
@@ -1,6 +1,6 @@
 from typing import Callable
 
-from textual.expand_tabs import expand_tabs_inline
+from textual.expand_tabs import get_tab_widths
 
 __all__ = ["cell_len", "cell_width_to_column_index"]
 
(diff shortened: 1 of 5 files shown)


src/textual/_cells.py
```diff
@@ -1,10 +1,44 @@
...
 
 
 cell_len: Callable[[str], int]
 try:
     from rich.cells import cached_cell_len as cell_len
 except ImportError:
     from rich.cells import cell_len
+
+
+def cell_width_to_column_index(line: str, cell_width: int, tab_width: int) -> int:
```
willmcgugan, line 15: Could a small `lru_cache` work here? i.e. will this be called frequently with the same values?

darrenburns: An LRU cache would mean that if you move the cursor up and down over the same character on a line, without editing that line, we'd hit the cache.

Also if you hit click the same offset on the same line multiple times, it'd be faster.

Could be wrong as I'm just going off a feeling here, but I don't feel like we'd get many cache hits on a small cache here, and a big cache would have a relatively large memory cost (relatively speaking) since it's caching full lines.

willmcgugan: Fair enough. If its not called frequently then there is probably little benefit. It just jumped out as a "pure function".

src/textual/expand_tabs.py
```diff
@@ -7,37 +7,59 @@
 _TABS_SPLITTER_RE = re.compile(r"(.*?\t|.+?$)")
 
 
-def expand_tabs_inline(line: str, tab_size: int = 4) -> str:
-    """Expands tabs, taking into account double cell characters.
+def get_tab_widths(line: str, tab_size: int = 4) -> list[tuple[str, int]]:
+    """Splits a string line into tuples (str, int).
```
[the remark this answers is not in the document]
rodrigogiraoserrao: How about 

 > Splits a line into sections before tab characters and the width left for tab alignment.

I'm sure you can do better :P

tests/test_expand_tabs.py
rodrigogiraoserrao, line 1: I approve the 💩s !

willmcgugan (approved):
Hmm. Yes. This looks like it is doing the right kind of things.

PR #3730, opened 2023-11-22
Improve documentation.

rodrigogiraoserrao:
Multiple small fixes to documentation and docstrings.

### DIFF
--- a/src/textual/types.py
+++ b/src/textual/types.py
@@ -14,8 +14,15 @@ from ._types import (
 )
 from .actions import ActionParseResult
 from .css.styles import RenderStyles
-from .widgets._data_table import CursorType
+from .widgets._directory_tree import DirEntry
 from .widgets._input import InputValidationOn
+from .widgets._option_list import (
+    DuplicateID,
+    NewOptionListContent,
+    OptionDoesNotExist,
+    OptionListContent,
+)
+from .widgets._placeholder import PlaceholderVariant
 from .widgets._select import NoSelection, SelectType
 
 __all__ = [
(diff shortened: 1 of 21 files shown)


src/textual/widgets/_directory_tree.py
```diff
@@ -22,7 +22,7 @@
 
 @dataclass
 class DirEntry:
-    """Attaches directory information to a node."""
+    """Attaches directory information to a [DirectoryTree][textual.widgets.DirectoryTree] node."""
```
davep, line 25:
```suggestion
    """Attaches directory information to a [`DirectoryTree`][textual.widgets.DirectoryTree] node."""
```

src/textual/types.py
davep, line 1: More of a general question for anyone reading: why do some things go into `types.py` but others not? I don't think I've ever seen a rationale for what goes where.

src/textual/widgets/_markdown.py
```diff
@@ -1005,12 +1012,12 @@ def __init__(
 
     @property
     def document(self) -> Markdown:
-        """The Markdown document object."""
+        """The [Markdown][textual.widgets.Markdown] document widget."""
```
davep, line 1015:
```suggestion
        """The [`Markdown`][textual.widgets.Markdown] document widget."""
```

davep (approved):

src/textual/types.py
rodrigogiraoserrao: Honestly, at this point I'm not sure.
Maybe we should make sure everything comes from its own original file?

I think in the beginning there were a couple of types that didn't really have a place they could call home, and so I exported them here.
Maybe we could clean this up a bit...
Do you think that's important now?

rodrigogiraoserrao: I'll open a follow-up issue if you think that's a good idea.
(Thanks for the other two fixes, good catches!)

davep: Just to confirm: no, not important now, and not anything that made sense to do in this PR; just a question for clarification for anyone who happened to be looking at this given the context, and possibly a catalyst for spinning off into another issue to consider this if there wasn't an obvious answer.

PR #3743, opened 2023-11-23
Add Select.from_values class method for initializing with iterator

### LINKED ISSUE #3691: Alternative method to initialise `Select`
I propose we extend `Select.__init__` or create a classmethod `Select.from_values` to accept an iterable of `SelectType` and automatically build the labels by converting them to strings.

I find that often we just want labels that are string representations of the values we want to select:

```py
Select.from_values(data_source)
# vs
Select([(str(datum), datum) for datum in data_source])
```

`data_source` is often just a list of strings or paths and the comprehension to create the tuple looks really awkward.

For example, in the `Select` docs we do this:

```py
# ...
yield Select((line, line) for line in LINES)
```

I personally don't like it at all.

azinneck0485:
Added class method that can be used to create a Select control using an iterator.  Method takes the same inputs as the constructor and has been tested to work with and without specifying parameters.

Snapshot tests have been updated for this method, and all other tests passed.

Fixes #3691. 

- [x] Updated documentation


CHANGELOG.md
```diff
@@ -17,6 +17,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - CSS error reporting will no longer provide links to the files in question https://github.com/Textualize/textual/pull/3582
 - inline CSS error reporting will report widget/class variable where the CSS was read from https://github.com/Textualize/textual/pull/3582
 
+### Added
+- Added `Select.from_values` class method that can be used to initialize a Select control with an iterator https://github.com/Textualize/textual/issues/3691
```
rodrigogiraoserrao, line 21: This needs to be moved under the correct place once you pull in the updated changelog.

src/textual/widgets/_select.py
```diff
@@ -278,6 +278,44 @@ def __init__(
         self._value: SelectType | None = value
         self._options = options
 
+    @classmethod
+    def from_values(
+        cls,
+        opts: [SelectType],
```
rodrigogiraoserrao, line 284: I don't think I've ever seen this syntax `opts: [SelectType]` to declare a list of `SelectType` objects, but we don't want to restrict ourselves to lists as inputs!

Please consider changing the typing to accept a sequence of `SelectType`, instead of only a list.

src/textual/widgets/_select.py
```diff
@@ -278,6 +278,44 @@ def __init__(
...
+        value: SelectType | None = None,
+        name: str | None = None,
+        id: str | None = None,
+        classes: str | None = None,
+        disabled: bool = False,
+    ):
+        """Initialize the Select control with options specified by an iterator
+
+        Args:
+         options: Options to select from.
```
rodrigogiraoserrao, line 297: I think users could benefit from a sentence indicating that we specify the values (like you said) and that the string representations are generated automatically by passing each option to the built-in `str`.

In other words, I want it to be absolutely clear for users that this works with options other than strings.

src/textual/widgets/_select.py
```diff
@@ -278,6 +278,44 @@ def __init__(
...
+         value: Initial value (should be one of the values in `options`).
+         name: The name of the select control.
+         id: The ID of the control the DOM.
+         classes: The CSS classes of the control.
+         disabled: Whether the control is disabled or not.
+        """
+        opts_iterator = [(str(o), o) for o in opts]
+
+        return cls(
+            opts_iterator,
```
rodrigogiraoserrao, line 309: We avoid shortening names as much as possible, so feel free to type out `options_iterator`, `options`, and `option`, or rename to something else if you prefer.

rodrigogiraoserrao (changes requested):
I hate to be that guy, but I've left some nitpick comments in your PR.

Overall, the PR is going absolutely in the right direction!
I'll have to ask you to address the review comments & to pull in `main` and merge it to update your PR, as there are currently some conflicts.

Thanks!

src/textual/widgets/_select.py
azinneck0485: I've fixed this in my latest set of commits (clarified that the iterator can be of any type).  Let me know if the changes are OK or if they need to be tweaked any more

src/textual/widgets/_select.py
azinneck0485: I changed to Iterator[SelectType], which I think is more general than a list?  Let me know if this is correct or if it should be something else.

docs/widgets/select.md
```diff
@@ -57,6 +57,32 @@ The following example presents a `Select` with a number of options.
     --8<-- "docs/examples/widgets/select.tcss"
     ```
 
+## Example
```
rodrigogiraoserrao, line 60: We don't want to have two H2 headings with the exact same name.
Instead, consider changing the original `## Example` to `## Examples` and then separate each example with an H3 heading, e.g., `### Basic Example` and `### Class method`.

src/textual/widgets/_select.py
```diff
@@ -305,6 +305,46 @@ def __init__(
...
+        allow_blank: bool = True,
+        value: SelectType | NoSelection = BLANK,
+        name: str | None = None,
+        id: str | None = None,
+        classes: str | None = None,
+        disabled: bool = False,
+    ):
+        """Initialize the Select control with options specified by an iterable of
+        of any type (string, integer, etc.).  The string representation in the
+        control is generated by this method using the built-in str().
```
rodrigogiraoserrao, line 323:
```suggestion
        """Initialize the Select control with values specified by an arbitrary iterable.
        
        The options shown in the control are computed by calling the built-in `str` on each value.
```

src/textual/widgets/_select.py
```diff
@@ -305,6 +305,46 @@ def __init__(
...
+
+        Args:
+         options: Options to select from.
+         prompt: Text to show in the control when no option is select.
+         allow_blank: Allow the selection of a blank option.
+         value: Initial value (should be one of the values in `options`).
+         name: The name of the select control.
+         id: The ID of the control the DOM.
+         classes: The CSS classes of the control.
+         disabled: Whether the control is disabled or not.
```
rodrigogiraoserrao, line 333: These lines should be indented 4 spaces instead of just 1 or 2.

src/textual/widgets/_select.py
```diff
@@ -305,6 +305,46 @@ def __init__(
...
+        """
+        options_iterator = [(str(o), o) for o in options]
```
rodrigogiraoserrao, line 335: You only replaced part of abbreviations here :)

```suggestion
        options_iterator = [(str(option), option) for option in options]
```

src/textual/widgets/_select.py
```diff
@@ -305,6 +305,46 @@ def __init__(
         self._value = value
         self._setup_variables_for_options(options)
 
+    @classmethod
+    def from_values(
+        cls,
+        options: Iterable[SelectType],
```
rodrigogiraoserrao, line 311: Sorry to only catch this now, but if the method is called `from_values`, doesn't it make sense that the parameter is called `values`?

Also, looking at what is done, we _are_ passing in the values, so the class method name is correct and the parameter name should be updated.

rodrigogiraoserrao (changes requested):
Thanks! The choice of `Iterable` is the most correct one, I think.

A couple of other updates are needed, if you don't midn.

src/textual/widgets/_select.py
rodrigogiraoserrao: Your new docstring was much better, thanks.

src/textual/widgets/_select.py
rodrigogiraoserrao: `Iterable[SelectType]` was the correct choice, I think. Thanks.

src/textual/widgets/_select.py
azinneck0485: I've fixed/changed this and addressed the other comments in the 2 most recent commits.

tests/snapshot_tests/test_snapshots.py
```diff
@@ -333,6 +333,10 @@ def test_select(snap_compare):
     assert snap_compare(WIDGET_EXAMPLES_DIR / "select_widget.py")
 
 
+def test_select_from_values(snap_compare):
```
darrenburns, line 336: I don't think we need quite so many snapshot tests here. These tests are for higher level things, so we shouldn't have multiple snapshots for a single method on a single widget.

src/textual/widgets/_select.py
```diff
@@ -305,6 +305,51 @@ def __init__(
...
+        values: Iterable[SelectType],
+        *,
+        prompt: str = "Select",
+        allow_blank: bool = True,
+        value: SelectType | NoSelection = BLANK,
+        name: str | None = None,
+        id: str | None = None,
+        classes: str | None = None,
+        disabled: bool = False,
+    ):
```
davep, line 320: This seems to be missing a return type hint.

tests/snapshot_tests/test_snapshots.py
rodrigogiraoserrao: Do you suggest we keep just `test_select_from_values_expanded` and delete the other 4?

darrenburns: I think so, yeah.

davep (changes requested):

rodrigogiraoserrao:
@azinneck0485 my colleagues left two pertinent comments:
https://github.com/Textualize/textual/pull/3743#discussion_r1410516396
https://github.com/Textualize/textual/pull/3743#discussion_r1410518972

Those are the only two things standing between you and me clicking “merge”!
Looking forward to merging your PR.

tests/snapshot_tests/test_snapshots.py
azinneck0485: i've updated the snapshot tests and snapshots accordingly

src/textual/widgets/_select.py
azinneck0485: I've made a change in the latest commit, but I'm not entirely sure it is correct.  I've not used type hints with class methods before, so please let me know if what I did was wrong.

src/textual/widgets/_select.py
```diff
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import sys
```
TomJGooding, line 3:
```suggestion
```

src/textual/widgets/_select.py
```diff
@@ -305,6 +311,51 @@ def __init__(
...
+        values: Iterable[SelectType],
+        *,
+        prompt: str = "Select",
+        allow_blank: bool = True,
+        value: SelectType | NoSelection = BLANK,
+        name: str | None = None,
+        id: str | None = None,
+        classes: str | None = None,
+        disabled: bool = False,
+    ) -> Self:
```
TomJGooding, line 326:
```suggestion
    ) -> Select:
```

src/textual/widgets/_select.py
```diff
@@ -19,6 +20,11 @@
 
     from ..app import ComposeResult
 
+if sys.version_info.minor >= 11:
+    from typing import Self
+else:
+    from typing_extensions import Self
+
```
TomJGooding, line 28:
```suggestion

```

TomJGooding (commented):
You just need to add `Select` as the return type.

src/textual/widgets/_select.py
azinneck0485: Thanks.  I've resolve this in the latest commit

src/textual/widgets/_select.py
```diff
@@ -305,6 +305,51 @@ def __init__(
...
+                with no selection made, in which case its value is set to the constant
+                [`Select.BLANK`][textual.widgets.Select.BLANK].
+            value: Initial value selected. Should be one of the values in `values`.
+                If no initial value is set and `allow_blank` is `False`, the widget
+                will auto-select the first available value.
+            name: The name of the select control.
+            id: The ID of the control in the DOM.
+            classes: The CSS classes of the control.
+            disabled: Whether the control is disabled or not.
+        """
```
davep, line 339: One more very minor issue that I just noticed: The docstring should have a `Returns:` section too.

azinneck0485: updated in latest commit

davep:
This is looking good to me, albeit for the failing snapshot test. I've run it locally and am getting the same fail, and there's zero visual difference so I'm guessing it's just a need to either a) update the snapshots and b) perhaps merge `main` once more?

rodrigogiraoserrao:
> This is looking good to me, albeit for the failing snapshot test. I've run it locally and am getting the same fail, and there's zero visual difference so I'm guessing it's just a need to either a) update the snapshots and b) perhaps merge `main` once more?

I also got the fail locally and I even checked, but the test passes on `main`...
I've merged and updated the snapshots.
I'll merge once the tests pass.

rodrigogiraoserrao:
🥳  @azinneck0485 thanks for your contribution!

azinneck0485:
Thank you all for the feedback (and being so patient with fixing my little mistakes :-) ).  I also noticed that that snapshot failed for me after I changed anything in my branch (i.e. I would update the snapshot, then after making changes it would fail on teh next text run), so hopefully merging main fixes things.

Issue #3754, opened 2023-11-27
Make SystemCommand lazy


willmcgugan:
In app we import `SystemCommands` which import `command.py`.

By necessity this imports a number of widgets, which means we always load  those widgets. I think we could make this lazy somehow, to improve startup time.

Couple of options:

- Delay work until the user summons the command palette.
- Delay work until after the first render.

Discuss with @willmcgugan once you have a plan for this.

willmcgugan:
https://github.com/Textualize/textual/pull/3756

PR #3756, opened 2023-11-27
Lazy command provider

willmcgugan:
Avoids importing the system commands until the user opens the command palette.

This is a significant improvement to startup time, for simpler apps at least.

Fixes https://github.com/Textualize/textual/issues/3754


darrenburns (approved):
Looks good - do you know how much faster it is to start up? I'm surprised that you said it has "significant" impact 😄

Issue #2975, opened 2023-07-19
`max-height: 100%` with a `height: auto` container causes child widget to disappear


davep:
Given this code:

```python
from textual.app import App, ComposeResult
from textual.widgets import DataTable, Static

class HappyDataTableFunApp(App[None]):

    CSS = """
    DataTable {
        max-height: 100%;
    }
    """

    def populate(self, table: DataTable) -> DataTable:
        for n in range(20):
            table.add_column(f"Column {n}")
        for row in range(1000):
            table.add_row(
                *[str(row * n) for n in range(20)]
            )
        return table

    def compose(self) -> ComposeResult:
        with Static():
            yield self.populate(DataTable())

if __name__ == "__main__":
    HappyDataTableFunApp().run()
```

the `DataTable` doesn't show. This seems to be the combination of a child widget that is `height: auto;` and `max-height: 100%;` with a parent container that is `height: auto`.

More generally the issue seems to be that if a child widget is `height: auto` and `max-height: 100%` and the parent container is `height: auto` there's no way to work out a height.

Perhaps there's some scope here to settle on a less astonishing outcome?

davep:
Somewhat related, because I see folk using `Static` as a container quite often, perhaps this wording:

> Can be used for Rich renderables and can also be **the base for other types of widgets**.

in the [the docs for `Static`](https://textual.textualize.io/widgets/static/) could do with a wee revamp, as I suspect it has caused people to think it should be used as a container.

Also, the tutorial uses `Static` as the base class for a compound widget (which now I think about it might account for why I see people do that so often).

TomJGooding:
Yup, I can confirm this is what I was doing too when I was new to Textual based on the docs!

The specific mention of `Static` in the widgets guide to "create custom widgets of your own" might also be compounding this confusion: https://textual.textualize.io/guide/widgets/#static-widget

PR #3758, opened 2023-11-27
error if no children allowed

willmcgugan:
Adds a `ALLOW_CHILDREN` flag to widgets.

If set to False, and the dev attempts to add a child, it will raise a `NotAContainer` exception. This should prevent a potential gotcha.

Screenshot 2023-11-27 at 15 57 59

Fixes https://github.com/Textualize/textual/issues/2975

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -48,8 +48,7 @@ jobs:
       #          make typecheck
       - name: Test with pytest
         run: |
-          source $VENV
-          pytest tests -v --cov=./src/textual --cov-report=xml:./coverage.xml --cov-report term-missing
+          poetry run pytest tests -v --cov=./src/textual --cov-report=xml:./coverage.xml --cov-report term-missing
       - name: Upload snapshot report
         if: always()
         uses: actions/upload-artifact@v3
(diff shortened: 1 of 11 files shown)


rodrigogiraoserrao (approved):
You didn't add `ALLOW_CHILDREN = False` exhaustively to all widgets where it would make sense, right?

darrenburns (approved):

willmcgugan:
> You didn't add `ALLOW_CHILDREN = False` exhaustively to all widgets where it would make sense, right?

There may be more.

davep:
So I'm 100% clear on this, and so can answer any questions about the change with confidence, the original example code that was the catalyst for this PR, if written to comport with the tutorial, would be more akin to:

```python
from textual.widgets import DataTable, Static

class Table(Static):

    def populate(self, table: DataTable) -> DataTable:
        for n in range(20):
            table.add_column(f"Column {n}")
        for row in range(1000):
            table.add_row(
                *[str(row * n) for n in range(20)]
            )
        return table

        yield self.populate(DataTable())


class HappyDataTableFunApp(App[None]):

    CSS = """
    DataTable {
        max-height: 100%;
    }
    """

        yield Table()

if __name__ == "__main__":
    HappyDataTableFunApp().run()
```

This still runs fine. That's by design?

PR #3765, opened 2023-11-28
Allow setting a toggle button's label

davep:
Until now `CheckBox` and `RadioButton` labels have been fixed after the widgets have been created. Prompted by #3764 and seeing no reasonable reason to not allow updating the labels later on, this adds that ability.

### DIFF
--- a/src/textual/widgets/_toggle_button.py
+++ b/src/textual/widgets/_toggle_button.py
@@ -161,6 +161,11 @@ class ToggleButton(Static, can_focus=True):
         """The label associated with the button."""
         return self._label
 
+    @label.setter
+    def label(self, label: TextType) -> None:
+        self._label = Text.from_markup(label) if isinstance(label, str) else label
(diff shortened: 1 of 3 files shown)


src/textual/widgets/_toggle_button.py
```diff
@@ -161,6 +161,11 @@ def label(self) -> Text:
         """The label associated with the button."""
         return self._label
 
+    @label.setter
+    def label(self, label: TextType) -> None:
+        self._label = Text.from_markup(label) if isinstance(label, str) else label
```
willmcgugan, line 166: The constructor also takes just the first line. Should we not do that here?

davep: Excellent catch!

davep: Done.

willmcgugan (approved):

Issue #3741, opened 2023-11-23
Dumb down some modified keypad keys in rxvt


davep:
In testing during #3440 it's been found that rxvt seems to stand apart from (almost?) all other terminals tested in that modified keys on the numeric keypad have their own escape sequences; so <kbd>shift</kbd>+<kbd>-</kbd> (on the numeric keypad, not top row `-`) results in an escape sequence, whereas in most other terminals it results in `minus` being emitted.

In this particular case, translate the sequences into the lowest-common-form of those keys.

PR #3737, opened 2023-11-23
Add support for ctrl+Fn keys under rxvt

davep:
Found during #3440, this adds support for <kbd>ctrl</kbd>+<kbd>Fn</kbd> and <kbd>ctrl</kbd>+<kbd>sh</kbd>+<kbd>Fn</kbd> keys within rxvt. Mainly discovered and tested on macOS/XQuartz, but I think this will enable this under GNU/Linux too (Parallels keeps forgetting my GNU/Linux VMs for reasons I can't fathom -- I should probably just set up [UTM](https://mac.getutm.app/) for GNU/Linux like I have on my own machine[^1]).

See #3735 (and by extension #3440).

[^1]: I've since done that and tested it and sure enough I'm seeing the same thing on GNU/Linux too (urxvt v9.30); the unknown sequences are all the ones covered here.

PR #3769, opened 2023-11-28
Simplify some sequences that seem unique to rxvt

davep:
With (almost(?)) all other terminals I've tested, shifted numeric keypad keys just yield the unshifted version of that key. rxvt seems to have sequences for those keys. Much like #3737 and #3739 we *could* add support for these sequences to the ANSI sequence list, but this is a bit different. Whereas in those other PRs we've been adding the alternate sequences that rxvt seems to have for well known and well-supported keys, here it would be the other way round: it would be adding bindable keys that are only available on a small subset of environments.

Ideally, at least for the moment, we want to encourage people to bind keys that are well-supported; so here we turn the shifted keypad keys into their unshifted counterparts, thus matching many other environments.

See #3741

The approach taken here is to slightly extend `ANSI_SEQUENCES_KEYS` os that, rather than being a sequence to `tuple[Keys, ...]` mapping, it can optionally map to a `str`, which is treated as a redirect for `_sequence_to_key_events`.


davep:
Also tagging in @TomJGooding as someone with daily rxvt experience.

willmcgugan (approved):
LGTM Will wait for Tom's reply before merging.

davep:
Good plan; I've tested it with urxvt on macOS (XQuartz); but confirmation that it makes sense in his environment would be welcome.

TomJGooding:
I honestly never use the numpad! Sorry just to confirm, are we talking about when num lock is on or off? It might just be my particular environment, but I *think* I'm seeing something different to what you're describing.

davep:
I'm testing physically on Mac, so no numlock involved. So this would be in effect numlock on (shift plus the numerical keys as numbers). I'd double check but on this machine at home it's a TKL keyboard.

If you're seeing something different and have a moment, I'd very much appreciate a report on what you have. 

The keys this relates to are:

- Shift +
- Shift -
- Shift *
- Shift /
- Shift 1-9

(Now I'm doubting I checked 0)

TomJGooding:
Ah, sorry I've used macOS even less than the numpad, so honestly had no idea there wasn't a concept of numlock!

Funnily enough 0 was probably the first key I tried!

Here's a screenshot from `textual keys` with my numlock on. This is running through the numlock keys first without then with <kbd>Shift</kbd>

image

TomJGooding:
Here's also some data from `tkrec`. Just to confirm, `showkey -a` doesn't return anything for <kbd>Shift</kbd>+<kbd>5</kbd> either.

```json
    "unknown": [
        {
            "key": "\\x1b[6$",
            "notes": "shift+3"
        },
        {
            "key": "\\x1b[5$",
            "notes": "shift+9"
        },
        {
            "key": "\\x1b[2$",
            "notes": "shift+0"
        }
    ],
```

davep:
Thanks Tom. When you say:

> Here's a screenshot from textual keys with my numlock on. This is running through the numlock keys first without then with Shift

that's *with* this PR?

davep:
The results you're getting [with `tkrec`](https://github.com/Textualize/textual/pull/3769#issuecomment-1832798017) are curious, in that you seem to be seeing something very different from what I'm seeing. With urxvt v9.3 (2021-11-27), under both macOS/XQuartz and in Debian GNU/Linux (running on macOS inside UTM) when I press (for example) shift+3 on the numpad, I get `^[Os`.

If I'm reading this correctly it looks like in your environment you get a fully different sequence coming through.

This is starting to look like some other environmental difference? As such I sense it doesn't impact on this PR then, it just means there might be a whole other set of sequences kicking about out there.

TomJGooding:
I'm curious about this too but not managed to find any answers yet.

The man page for `xkeyboard-config` does show different configuration options for the numpad:

       │ numpad:pc                    Default numeric keypad keys                                                                                        │
       │ numpad:mac                   Numeric keypad always enters digits (as in macOS)                                                                  │
       │ numpad:microsoft             Num Lock on: digits; Shift for arrows. Num Lock off: arrows (as in Windows)                                        │
       │ numpad:shift3                Shift does not cancel Num Lock, chooses 3rd level instead                                                          │


But as far as I can tell, I don't have any `XKbOptions` set in my config (I never use the numpad so it wouldn't be something I would have configured anyway!)

```bash
$ setxkbmap -print -verbose 10
```

```
Setting verbose level to 10
locale is C
Trying to load rules file ./rules/evdev...
Trying to load rules file /usr/share/X11/xkb/rules/evdev...
Success.
Applied rules from evdev:
rules:      evdev
model:      pc105
layout:     gb
Trying to build keymap using the following components:
keycodes:   evdev+aliases(qwerty)
types:      complete
compat:     complete
symbols:    pc+gb+inet(evdev)
geometry:   pc(pc105)
xkb_keymap {
	xkb_keycodes  { include "evdev+aliases(qwerty)"	};
	xkb_types     { include "complete"	};
	xkb_compat    { include "complete"	};
	xkb_symbols   { include "pc+gb+inet(evdev)"	};
	xkb_geometry  { include "pc(pc105)"	};
};
```

PR #3773, opened 2023-11-28
_box_drawing.py: Fix 3 of the characters

encukou:
Hello, and thank you for Textual!

This is a quick one-file commit -- I had an autogenerated dict of box-drawing characters handy for a quick check. Could you please add a changelog entry if appropriate?

If you'd like to explain an algorithm for entries that don't quite match, like `(3, 1, 3, 3): "╉",` or `(3, 3, 2, 2): "╬",` I can check those as well :)

- [n/a] Updated documentation


src/textual/_box_drawing.py
```diff
@@ -45,7 +45,7 @@
     #
     (0, 1, 0, 0): "╶",
     (0, 1, 0, 1): "─",
-    (0, 1, 0, 2): "╼",
+    (0, 1, 0, 2): '╾'
```
encukou, line 48:
```suggestion
    (0, 1, 0, 2): '╾',
```

src/textual/_box_drawing.py
```diff
@@ -45,7 +45,7 @@
     #
     (0, 1, 0, 0): "╶",
     (0, 1, 0, 1): "─",
-    (0, 1, 0, 2): "╼",
+    (0, 1, 0, 2): '╾',
```
encukou, line 48:
```suggestion
    (0, 1, 0, 2): "╾",
```

src/textual/_box_drawing.py
```diff
@@ -64,7 +64,7 @@
     (0, 1, 3, 3): "╥",
     #
     (0, 2, 0, 0): "╺",
-    (0, 2, 0, 1): "╾",
+    (0, 2, 0, 1): '╼',
```
encukou, line 67:
```suggestion
    (0, 2, 0, 1): "╼",
```

src/textual/_box_drawing.py
```diff
@@ -208,7 +208,7 @@
     (2, 1, 0, 2): "┹",
     (2, 1, 0, 3): "┹",
     #
-    (2, 1, 1, 0): "╀",
+    (2, 1, 1, 0): '┞',
```
encukou, line 211:
```suggestion
    (2, 1, 1, 0): "┞",
```

encukou:
Guess I should've just opened an issue if I don't have the full dev environment set up. Sorry for the noisy fixups!

willmcgugan:
The double lines are problematic since they don't cover all combinations. My rule of thumb was that if it contained a double, then it would promote the other lines to double.

Issue #3777, opened 2023-11-29
`NoWidget: No widget under screen coordinate (x, y)` crash when clicking on the terminal after new screen was recently pushed


mzebrak:
Textual version: 0.43.0 (observable since 0.42.0, I haven't noticed it on 0.41.0)

<details>
<summary>Minimal working example to reproduce</summary>

```python
from textual import on
from textual.app import App, ComposeResult
from textual.screen import Screen
from textual.widgets import Label, Footer, Button


class SomeScreen(Screen):
    def compose(self) -> ComposeResult:
        for i in range(50):
            yield Label(f"Label {i}")

        yield Footer()


class NoWidgetCrashApp(App):
    def compose(self) -> ComposeResult:
        yield Label("Hello, world!")
        yield Button("Click me!")
        yield Footer()

    @on(Button.Pressed)
    def push_some_screen(self) -> None:
        self.push_screen(SomeScreen())


NoWidgetCrashApp().run()
```
</details>

Steps:
Just launch the app, then start spamming LMB on the "Click me!" Button. When you click in the right time - right after new screen was pushed - you'll get a crash.

<details>
<summary>Video</summary>

[Screencast from 11-29-2023 08:41:29 AM.webm](https://github.com/Textualize/textual/assets/73518558/bbf07d3f-f308-493e-835d-b86dde7f79c4)

</details>

willmcgugan:
Please paste the output of `textual diagnose`. A copy of the traceback would also be helpful.

mzebrak:
> Please paste the output of `textual diagnose`. A copy of the traceback would also be helpful.

Here you go:

<details>
<summary> Traceback </summary>

```python
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ /home/mzebrak/.pyenv/versions/textual-tests/lib/python3.10/site-packages/textual/drivers/linux_d │
│ river.py:242 in _run_input_thread                                                                │
│                                                                                                  │
│   239 │   │   an exception                                                                       │
│   240 │   │   """                                                                                │
│   241 │   │   try:                                                                               │
│ ❱ 242 │   │   │   self.run_input_thread()                                                        │
│   243 │   │   except BaseException as error:                                                     │
│   244 │   │   │   import rich.traceback                                                          │
│   245                                                                                            │
│                                                                                                  │
│ /home/mzebrak/.pyenv/versions/textual-tests/lib/python3.10/site-packages/textual/drivers/linux_d │
[...]

PR #3779, opened 2023-11-29
fix threading issue

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/3777

The driver was calling `get_widget_at` which is not thread-safe. If called while the compositor is building the map it could fail.

The solution was to move this logic from the driver to the app, which is ultimately a better place for it.


willmcgugan:
ignore review request. Fixing tests...

willmcgugan:
Ok, knock yourself out.

darrenburns (approved):

PR #3780, opened 2023-11-29
mount error

willmcgugan:
Catch attempts to mount invalid objects, and provide a helpful error.

Catches both yielding a non-widget, and forgetting to call `super()` in a widget constructor

Screenshot 2023-11-29 at 14 39 27

Supersedes https://github.com/Textualize/textual/pull/3438


davep (approved):
Can see this being helpful.

rodrigogiraoserrao (approved):

PR #3786, opened 2023-11-30
Force the call to scroll to the region of a node's label

davep:
Re: https://discord.com/channels/1026214085173461072/1033754296224841768/1179803173498658966

It is possible that the developer may be making a call to scroll to a node, while the scrollbar isn't quite in play yet; without a force to scroll the scroll won't happen.

To illustrate the issue, take this code:

```python
from textual import on
from textual.widgets import Tree, Button

class ScrollToNode(App[None]):

        yield Button("Expand and scroll to last")
        yield Tree("Root")

    def on_mount(self) -> None:
        tree = self.query_one(Tree)
        for n in range(10):
            outer = tree.root.add(str(n))
            for m in range(10):
                outer.add_leaf(str(m))

    @on(Button.Pressed)
    def show_problem(self) -> None:
        tree = self.query_one(Tree)
        tree.root.expand_all()
        self.call_after_refresh(
            tree.scroll_to_node, tree.root.children[-1].children[-1]
        )

if __name__ == "__main__":
    ScrollToNode().run()
```

When the button is pressed, the nodes are expanded but the scroll doesn't happen. With this PR the scroll does happen. It seems to be a reasonable expectation that the scroll *does* happen.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
  - `link-hover-background` renamed to `link-background-hover`
  - `link-hover-color` renamed to `link-color-hover`
  - `link-hover-style` renamed to `link-style-hover`
+- `Tree` now forces a scroll when `scroll_to_node` is called https://github.com/Textualize/textual/pull/3786
 
 ### Added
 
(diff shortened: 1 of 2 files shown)


rodrigogiraoserrao (approved):
I also feel it's reasonable to have the scroll happen.

I'm assuming there are no other implications stemming from the forced scroll?

davep:
None that I could find, test for or otherwise ascertain.

PR #3798, opened 2023-12-02
optimize cuts

willmcgugan:
A modest optimization.

### DIFF
--- a/src/textual/_compositor.py
+++ b/src/textual/_compositor.py
@@ -877,15 +877,13 @@ class Compositor:
             return self._cuts
 
         width, height = self.size
-        screen_region = self.size.region
         cuts = [[0, width] for _ in range(height)]
 
         intersection = Region.intersection
         extend = list.extend
 
         for region, clip in self.visible_widgets.values():
-            region = intersection(region, clip)
-            if region and (region in screen_region):
+            if region := intersection(region, clip):
                 x, y, region_width, region_height = region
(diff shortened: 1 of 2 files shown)


src/textual/_compositor.py
```diff
@@ -877,15 +877,13 @@ def cuts(self) -> list[list[int]]:
...
-        screen_region = self.size.region
         cuts = [[0, width] for _ in range(height)]
 
         intersection = Region.intersection
         extend = list.extend
 
         for region, clip in self.visible_widgets.values():
-            region = intersection(region, clip)
-            if region and (region in screen_region):
+            if region := intersection(region, clip):
```
willmcgugan, line 886: `visible_widgets` is already limited to the visible area, so no need for another `region in screen_region` check.

src/textual/_compositor.py
```diff
@@ -877,16 +877,14 @@ def cuts(self) -> list[list[int]]:
...
         cuts = [[0, width] for _ in range(height)]
 
         intersection = Region.intersection
         extend = list.extend
 
         for region, clip in self.visible_widgets.values():
-            region = intersection(region, clip)
-            if region and (region in screen_region):
-                x, y, region_width, region_height = region
+            x, y, region_width, region_height = intersection(region, clip)
```
willmcgugan, line 886: We don't need the `region in screen_region` because `self.visible_widgets` has already been limited to visible widgets.

PR #3800, opened 2023-12-04
Improve the method via which unsupported sequences are ignored

davep:
This PR extends the current support for ignoring unsupported sequences. Until now a sequence could map to `Keys.Ignore` but this would result in an actual, bindable, `<ignore>` key event being raised. Here we make the ignoring of the sequence silent, but also ensure that it gets logged to `keys.log` if the key sequence parser logging is enabled.

Additionally, the PR makes a start on working through the [known unhandled sequences list](https://github.com/Textualize/textual/wiki/Key-names-and-escape-sequences#unknown-sequences), either adding more aliasing for keys, or making use of the above change to ignore certain sequences altogether. In conjunction with this I've extended the code that produces the unknown sequences table so that it shows which sequences have been taken care of, and which PR took care of them. This makes it easier to keep track of what is and isn't done, and also provides a paper trail of how and why a sequence was ignored.


src/textual/_xterm_parser.py
```diff
@@ -244,10 +258,20 @@ def _sequence_to_key_events(
         """
         keys = ANSI_SEQUENCES_KEYS.get(sequence)
         if isinstance(keys, tuple):
-            # If the sequence mapped to a tuple, then it's values from the
-            # `Keys` enum. Raise key events from what we find in the tuple.
-            for key in keys:
-                yield events.Key(key.value, sequence if len(sequence) == 1 else None)
+            # If we're being asked to ignore the key...
+            if keys == (Keys.Ignore,):
```
willmcgugan, line 262: Do we need this branch? The data structure implies that there could be `Keys.Ignore` mixed with other keys. If that can't happen then a `None` sentinal would be more canonical.

How about we perform the check for ignores keys inside the for loop on L271

davep: This approach is following the [pre-established convention](https://github.com/Textualize/textual/blob/3ab7b57d3b7bcc63bd68a259b0bce60b65c0772e/src/textual/_ansi_sequences.py#L176-L177) of marking an ignored sequence as a tuple of a single `Keys.Ignore`; by that convention `Keys.Ignore` will only be used this way. Then following on from that an event is created that contains the sequence information so it can be logged.

Sure, this could be done inside the loop, but given that `Keys.Ignore` should never appear in a context other than when used as `(Keys.Ignore,)`, it didn't seem to make sense to repeat the comparison within the loop: made more sense to be to check once and be done.

willmcgugan: Using a sequence like this is always going to imply that you can mix ignored keys with actual keys. I fear future readers of the keys file may make that assumption.

If that will never occur, then I suggest we go with a `None` sentinal. Or we roll it in to the loop, so its no longer a special case.

davep: Cool, works for me: was just going with what we discussed yesterday (keeping with the inherited convention). I'll switch it up.

src/textual/_ansi_sequences.py
```diff
@@ -1,11 +1,16 @@
...
 from .keys import Keys
 
+
+class IgnoredSequence:
+    """Class used to mark that a sequence should be ignored."""
+
+
 # Mapping of vt100 escape codes to Keys.
-ANSI_SEQUENCES_KEYS: Mapping[str, Tuple[Keys, ...] | str] = {
+ANSI_SEQUENCES_KEYS: Mapping[str, Tuple[Keys, ...] | str | Type[IgnoredSequence]] = {
```
willmcgugan, line 13: Using a type as a sentinal is unconventional. I would find it a little astonishing reading this.

Suggest we use `None` as a sentinal, or stick with `Keys.Ignore` in the tuple. 

I objected to this:

```python
if keys == (Keys.Ignore,):
```

because I would find it astonishing to treat that as a sentinal, when the tuple is otherwise a sequence of things. But if check for the `Ignore` in the loop itself, there is no special case needed.

willmcgugan (changes requested):

willmcgugan (approved):

PR #3813, opened 2023-12-05
`Widget.loading` scroll-related glitch fixes

### LINKED ISSUE #3733: `scrollbar-gutter` causes `loading` overlay to not fully overlay
This has turned up with Textual 0.42. If you have a widget that is styled `scrollbar-gutter: stable`, the `loading` overlay seems to "leak" whatever's underneath on the edge of the indicator. Using this code (using an `OptionList` as that's the widget I discovered it with):

```python

class LoadingOverlayRedux(App[None]):

    CSS = """
    OptionList {
        scrollbar-gutter: stable;
        width: 1fr;
        height: 1fr;
    }
    """

    BINDINGS = [
        ("space", "toggle")
    ]

        yield OptionList(*[
            ("X" * 500) for _ in range(1_000)
        ])

    def action_toggle(self) -> None:
        self.query_one(OptionList).loading = not self.query_one(OptionList).loading

if __name__ == "__main__":
    LoadingOverlayRedux().run()
```

it initially looks like this:

Screenshot 2023-11-22 at 19 55 03

and then if I toggle `loading`:

Screenshot 2023-11-22 at 19 55 13

davep:
This PR addresses a couple of issues when using `Widget.loading` to show a loading indicator on a widget:

- If the scrollbar gutter was fixed, it would cause some of the background widget to show through.
- If the underlying widget was scrollable, and was scrolled down some way, the loading indicator would either be partially visible or not visible at all.

Fixes #3733.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Fixed `DataTable.update_cell` not raising an error with an invalid column key https://github.com/Textualize/textual/issues/3335
 - Fixed `Input` showing suggestions when not focused https://github.com/Textualize/textual/pull/3808
-- Fixed `Widget.loading` overlay being affected by scroll position and scrollbar gutter
+- Fixed `Widget.loading` overlay being affected by scroll position and scrollbar gutter https://github.com/Textualize/textual/pull/3813
 
 ### Removed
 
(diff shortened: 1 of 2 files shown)


davep:
As per in-person discussion, handing this off to @willmcgugan for a wee play to consider some alternative methods of doing this type of overlay.

davep:
Closing in favour of #3816.

Issue #3695, opened 2023-11-17
Querying TabPane by ID returns multiple nodes


liancheng:
When assigning an ID to a `TabPane`, it seems that there is always a corresponding `ContentTab` created with the same ID, and therefore failing `query_one(id)` calls.

This [discussion thread](https://github.com/Textualize/textual/discussions/3666) is relevant, and I can use the `TabPane#foo` trick to workaround this issue. But this still seems to be a bug to me.

Repro:

```python
from textual.app import App, ComposeResult
from textual.widgets import TabbedContent, TabPane, Button


class ExampleApp(App):
    def compose(self) -> ComposeResult:
        with TabbedContent():
            yield TabPane("A", id="tab-a")

        yield Button("Repro")

    def on_button_pressed(self) -> None:
        for node in self.query("#tab-a"):
            self.log(f"### {node} ###")
        self.query_one("#tab-a")


if __name__ == "__main__":
    ExampleApp().run()
```

Clicking the button produces the following error:

```
TooManyMatches: Call to only_one resulted in more than one matched node
```

Relevant log lines in Textual console:

```
[01:05:37] INFO                                         tabbed_content_bug.py:14
### ContentTab(id='tab-a') ###
[01:05:37] INFO                                         tabbed_content_bug.py:14
### TabPane(id='tab-a') ###
```

# Textual Diagnostics

## Versions

| Name    | Value  |
|---------|--------|
| Textual | 0.41.0 |
| Rich    | 13.5.2 |

## Python

| Name           | Value                                               |
|----------------|-----------------------------------------------------|
| Version        | 3.11.2                                              |
| Implementation | CPython                                             |
| Compiler       | Clang 15.0.7                                        |
| Executable     | /Users/lian/local/src/textual/main/.venv/bin/python |

## Operating System

| Name    | Value                                                                                                 |
|---------|-------------------------------------------------------------------------------------------------------|
| System  | Darwin                                                                                                |
| Release | 22.6.0                                                                                                |
| Version | Darwin Kernel Version 22.6.0: Wed Jul  5 22:22:05 PDT 2023; root:xnu-8796.141.3~6/RELEASE_ARM64_T6000 |

## Terminal

| Name                 | Value           |
|----------------------|-----------------|
| Terminal Application | tmux (3.3a)     |
| TERM                 | screen-256color |
| COLORTERM            | truecolor       |
| FORCE_COLOR          | *Not set*       |
| NO_COLOR             | *Not set*       |

## Rich Console options

[...]

PR #3815, opened 2023-12-05
Namespace IDs for `ContentTab` in `TabbedContent`

davep:
To satisfy #3695 this PR adds an "internal" prefix to the tabs that go to make up a `TabbedContent`, thus ensuring that an ID for a `TabPane` doesn't end up being a duplicate ID amongst the descendants of the `TabbedContent`[^1]. All `ContentTab` instances now have and ID of `--content-tab-` followed by the ID of the `TabPane` they pertain to.

In doing this the following changes have been included:

- `TabbedContent.TabActivated` has grown a `pane` property, which is a reference to the `TabPane` that has been activated.
- `TabbedContent.TabActivated` has been added to `TabbedContent.TabActivated.ALLOW_SELECTOR_MATCH`, allowing it to be used with the `@on` decorator.
- `TabbedContent.TabActivated.tab` has been removed from `TabbedContent.TabActivated.ALLOW_SELECTOR_MATCH`, an intentional breaking change to encourage anyone using it to switch to `pane` (the reason being that the ID of the `tab` will be the "internally-prefixed" one so will be less easy to reason about).
- `ContentTab` has grown a couple of methods for handling the addition and removal of the prefix to IDs.
- `ContentTabs` has grown wrappers around `disable`, `enable`, `show` and `hide` to handle the transition between IDs.
- `TabbedContent` has grown `get_tab` and `get_pane` methods, partially in service of this PR, but also as a handy utility for the developer using `TabbedContent`.
- Added some extra information to the docs for `TabbedContent` that informs the developer how they can go about styling the tabs themselves. *(Note: this is as-yet unchecked with local docs serving as both myself and @rodrigogiraoserrao are unable to get the docs to build right now)*

[^1]: The initial complaint being that a `query_one("#some-tab-pane")` failing was surprising.

### DIFF
--- a/src/textual/widgets/_tabbed_content.py
+++ b/src/textual/widgets/_tabbed_content.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from asyncio import gather
 from dataclasses import dataclass
 from itertools import zip_longest
+from typing import cast
 
 from rich.repr import Result
 from rich.text import Text, TextType
@@ -242,7 +243,7 @@ class TabbedContent(Widget):
     class TabActivated(Message):
         """Posted when the active tab changes."""
 
-        ALLOW_SELECTOR_MATCH = {"tab"}
+        ALLOW_SELECTOR_MATCH = {"pane"}
         """Additional message attributes that can be used with the [`on` decorator][textual.on]."""
 
         def __init__(self, tabbed_content: TabbedContent, tab: ContentTab) -> None:
@@ -256,6 +257,11 @@ class TabbedContent(Widget):
             """The `TabbedContent` widget that contains the tab activated."""
             self.tab = tab
             """The `Tab` widget that was selected (contains the tab label)."""
+            self.pane: TabPane = cast(
+                TabPane,
+                tabbed_content.get_child_by_type(ContentSwitcher).visible_content,
+            )
+            """The `TabPane` widget that was activated by selecting the tab."""
             super().__init__()
 
         @property
@@ -270,6 +276,7 @@ class TabbedContent(Widget):
         def __rich_repr__(self) -> Result:
             yield self.tabbed_content
             yield self.tab
+            yield self.pane
 
     class Cleared(Message):
         """Posted when there are no more tab panes."""
--- a/src/textual/widgets/_tabbed_content.py
+++ b/src/textual/widgets/_tabbed_content.py
@@ -51,7 +51,11 @@ class ContentTab(Tab):
         Returns:
             The ID with the prefix removed.
         """
-        return content_id[len(cls._PREFIX):] if content_id.startswith(cls._PREFIX) else content_id
+        return (
+            content_id[len(cls._PREFIX) :]
+            if content_id.startswith(cls._PREFIX)
+            else content_id
+        )
 
     def __init__(self, label: Text, content_id: str, disabled: bool = False) -> None:
         """Initialize a ContentTab.
@@ -80,7 +84,9 @@ class ContentTabs(Tabs):
             active: ID of the tab which should be active on start.
             tabbed_content: The associated TabbedContent instance.
         """
-        super().__init__(*tabs, active=active if active is None else ContentTab.add_prefix(active))
+        super().__init__(
+            *tabs, active=active if active is None else ContentTab.add_prefix(active)
+        )
         self.tabbed_content = tabbed_content
 
     def get_content_tab(self, tab_id: str) -> ContentTab:
@@ -149,6 +155,7 @@ class ContentTabs(Tabs):
             TabError: If there are any issues with the request.
         """
         return super().show(ContentTab.add_prefix(tab_id))
+
 
 class TabPane(Widget):
     """A container for switchable content, with additional title.
@@ -413,7 +420,7 @@ class TabbedContent(Widget):
             tabs.add_tab(
                 ContentTab(pane._title, pane.id),
                 before=before if before is None else ContentTab.add_prefix(before),
-                after=after if after is None else ContentTab.add_prefix(after)
+                after=after if after is None else ContentTab.add_prefix(after),
             ),
             self.get_child_by_type(ContentSwitcher).mount(pane),
         )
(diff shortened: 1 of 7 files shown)


src/textual/widgets/_tabbed_content.py
```diff
@@ -26,15 +27,46 @@
...
+            content_id: The ID to remove the prefix from.
+
+        Returns:
+            The ID with the prefix removed.
+        """
+        return (
+            content_id[len(cls._PREFIX) :]
+            if content_id.startswith(cls._PREFIX)
+            else content_id
+        )
```
rodrigogiraoserrao, line 59: Let's drop support for 3.8 so this can be `return content_id.removeprefix(cls._PREFIX)` 😆

src/textual/widgets/_tabbed_content.py
```diff
@@ -422,38 +534,77 @@ def _is_associated_tabs(self, tabs: Tabs) -> bool:
         """
         return isinstance(tabs, ContentTabs) and tabs.tabbed_content is self
 
-    def _watch_active(self, active: str) -> None:
+    def _watch_active(self, old_active: str, active: str) -> None:
```
rodrigogiraoserrao, line 537: You're not really using this `old_active`, are you?

src/textual/widgets/_tabbed_content.py
```diff
@@ -422,38 +534,77 @@ def _is_associated_tabs(self, tabs: Tabs) -> bool:
...
+    def get_tab(self, pane_id: str | TabPane) -> Tab:
+        """Get the `Tab` associated with the given ID or `TabPane`.
+
+        Args:
+            pane_id: The ID of the pane, or the pane itself.
+
+        Returns:
+            The Tab associated with the ID.
+        """
+        if target_id := pane_id if isinstance(pane_id, str) else pane_id.id:
```
rodrigogiraoserrao, line 557: I'd suggest parenthesising the conditional expression to make it easier to read inside the `if` statement.

```suggestion
        if target_id := (pane_id if isinstance(pane_id, str) else pane_id.id):
```

src/textual/widgets/_tabbed_content.py
```diff
@@ -422,38 +534,77 @@ def _is_associated_tabs(self, tabs: Tabs) -> bool:
...
 
+    def get_tab(self, pane_id: str | TabPane) -> Tab:
+        """Get the `Tab` associated with the given ID or `TabPane`.
+
+        Args:
+            pane_id: The ID of the pane, or the pane itself.
+
+        Returns:
+            The Tab associated with the ID.
+        """
```
rodrigogiraoserrao, line 556: You seem to be missing the `Raises` section of this docstring.

tests/test_tabbed_content.py
```diff
@@ -209,7 +210,10 @@ def compose(self) -> ComposeResult:
         await tabbed_content.add_pane(TabPane("Added", id="new-1"), before="initial-1")
         assert tabbed_content.tab_count == 2
         assert tabbed_content.active == "initial-1"
-        assert [tab.id for tab in tabbed_content.query(Tab).results(Tab)] == [
+        assert [
+            ContentTab.sans_prefix(tab.id)
+            for tab in tabbed_content.query(Tab).results(Tab)
```
rodrigogiraoserrao, line 215: I think we _used_ to need the `.results(Tab)` so that the type checker was happy but I think this is no longer the case.
I.e., I think you can remove `.results(Tab)` from here and the other lines below.

rodrigogiraoserrao (changes requested):
A couple of small changes requested.

How easy/convenient is it to use the helper methods `add_prefix`/`sans_prefix`?
While I agree they were defined in the more logical place from the point of view of the organisation of the code, I wonder if we chould/should allow the user to get to those methods in a more direct manner, as we usually don't import `ContentTabs`.

davep:
They're intentionally a hidden detail.

src/textual/widgets/_tabbed_content.py
davep: If only...

src/textual/widgets/_tabbed_content.py
davep: Doesn't seem so; I would suspect it being vestigial of prior WiP.

src/textual/widgets/_tabbed_content.py
davep: Good spot.

tests/test_tabbed_content.py
davep: Good catch; old habit.

src/textual/widgets/_tabbed_content.py
```diff
@@ -26,15 +27,46 @@
 class ContentTab(Tab):
     """A Tab with an associated content id."""
 
-    def __init__(self, label: Text, content_id: str, disabled: bool = False):
+    _PREFIX: Final[str] = "--content-tab-"
```
rodrigogiraoserrao, line 30: Ah, I was just reminded of one last thing!

The users need to know what this prefix is if they want to write their own CSS that styles tabs, right?
That's why this whole conversation started.

Wouldn't it be a good idea if this were actually a public attribute of the class, or maybe even a public global?
This way, it's easier to document.

I'm not sure this is a good idea, so I'll just leave this as a comment and not as a requested change.

rodrigogiraoserrao (approved):
Looks good to me!

src/textual/widgets/_tabbed_content.py
davep: I did consider this but the problem I see is that 99.9% of the time, when they want to use it, it'll be in CSS; where some public constant would be utterly useless.

rodrigogiraoserrao: I think your statistic looks accurate. I only wondered that because of documentation. With a leading `_`, it won't show in the documentation.

davep: I would hope it doesn't.

rodrigogiraoserrao: 😆 then one of us misunderstood the other.
If the prefix isn't documented, how can the user know what to write in CSS?

davep: I'm not in a position to check properly right now, but I have a recollection of documenting it (hence the thing about not being able to build the docs last week).

davep: image

rodrigogiraoserrao: I saw that.
If/when we change the value of `_PREFIX`, that will need updating and it's likely we'll forget to do so.
If instead it linked to `ContentTabs.PREFIX` in the documentation, then updating the code would be enough.
That was my only concern.

davep: My view is it hides the detail in the wrong location; and it would need (or at least deserve) the example code anyway, leaving exactly the same problem.

PR #3816, opened 2023-12-05
simplify loading

willmcgugan:
Changes how `loading` is implemented.

Rather than save and restore state (which has the potential to be brittle), this adds a `loading` state.

Fixes https://github.com/Textualize/textual/issues/3733


rodrigogiraoserrao (approved):
Makes sense to me.

Although I don't get why the loading indicator must have a min-width of 1 if its height is already set to 100%.

PR #3820, opened 2023-12-06
Opacity typing

willmcgugan:
Fix for a typing issue. Also a very minor performance win.

### DIFF
--- a/src/textual/_opacity.py
+++ b/src/textual/_opacity.py
@@ -1,4 +1,4 @@
-from typing import Iterable
+from typing import Iterable, cast
 
 from rich.segment import Segment
 from rich.style import Style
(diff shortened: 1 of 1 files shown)


src/textual/_opacity.py
```diff
@@ -25,8 +25,8 @@ def _apply_opacity(
     from_rich_color = Color.from_rich_color
     from_color = Style.from_color
     blend = base_background.blend
-    for segment in segments:
-        text, style, _ = segment
+    styled_segments = cast("Iterable[tuple[str, Style, object]]", segments)
```
willmcgugan, line 28: The second value in the tuple is typed as optional, but we know from context that it never is None.

Doing the cast outside of the loop for efficiency.

willmcgugan:
#executivedecision It's pretty trivial...

PR #3825, opened 2023-12-07
Add Collapsible.Toggled.

### LINKED ISSUE #3824: Add message `Collapsible.Toggled`
What it says on the tin.

rodrigogiraoserrao:
Fix #3824

### DIFF
--- a/src/textual/widgets/_collapsible.py
+++ b/src/textual/widgets/_collapsible.py
@@ -99,6 +99,26 @@ class Collapsible(Widget):
     }
     """
 
+    class Toggled(Message):
+        """Event sent when the `Collapsible` is toggled.
+
+        Can be handled using `on_collapsible_toggled` in a subclass of
+        [`Collapsible`][textual.widgets.Collapsible] or in a parent widget in the DOM.
+        """
+
+        def __init__(self, collapsible: Collapsible) -> None:
+            self.collapsible: Collapsible = collapsible
+            """The collapsible that was pressed."""
+            super().__init__()
+
+        @property
+        def control(self) -> Collapsible:
+            """An alias for [Toggled.collapsible][textual.widgets.Collapsible.Toggled.collapsible].
+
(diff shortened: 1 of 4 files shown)


src/textual/widgets/_collapsible.py
```diff
@@ -99,6 +99,26 @@ class Collapsible(Widget):
...
+            self.collapsible: Collapsible = collapsible
+            """The collapsible that was pressed."""
+            super().__init__()
+
+        @property
+        def control(self) -> Collapsible:
+            """An alias for [Toggled.collapsible][textual.widgets.Collapsible.Toggled.collapsible].
+
+            This will be the same value as [Toggled.collapsible][textual.widgets.Collapsible.Toggled.collapsible].
+            """
```
rodrigogiraoserrao, line 119: I copied the docstring from `Button.Pressed`.
Doesn't this feel a bit redundant, though?

src/textual/widgets/_collapsible.py
```diff
@@ -99,6 +99,23 @@ class Collapsible(Widget):
     }
     """
 
+    class Toggled(Message):
```
davep, line 102: I'd observe that leaning into a single message rather than one for expanded and one for collapsed makes it just a touch harder to write more declarative code with the `@on` decorator; but I'm not so bothered about it as to heavily advocate for it.

rodrigogiraoserrao: How about if `Collapsed` and `Expanded` inherit from `Toggled`, so that we can actually get the best of both worlds?

davep: Exactly what I was thinking. :-)

rodrigogiraoserrao: Do you mind taking another look just to make sure my change ba9c00e was in the right direction?

davep: Done; double approved.

PR #3830, opened 2023-12-07
feat(collapsible): make title a reactive attribute

### LINKED ISSUE #3829: Make Title of a Collapsible Widget Editable
Can not update title property of the Collapsible widget , looking at CollapsibleTitle class it seems to only set at init
Looks like this was an intentional choice ?
But I think it would be helpful to make the title property editable ,
Eg: changing content in the collapsed state and don't want to toggle it open to get a quick status update, so write something in .title... 

~~~
from textual.containers import Container, Horizontal, VerticalScroll
from textual.widgets import Header, Footer, Collapsible, Placeholder

class AnApp(App):

    TITLE = "Collapsible Titles Should Update"

    BINDINGS = [("d", "toggle_dark", "Update"),
    ("c", "collapse_or_expand(True)", "Collapse"),
    ("e", "collapse_or_expand(False)", "Expand")]
    
    def action_collapse_or_expand(self, collapse: bool) -> None:
        for child in self.walk_children(Collapsible):
            child.collapsed = collapse

        """Create child widgets for the app."""
        yield Header(show_clock=True)
        yield VerticalScroll(
            Collapsible(
                Horizontal(
                 Placeholder(variant="text"),
                 Placeholder(variant="size"),
                 Placeholder(variant="text"),
            ),
            title="Top_Stuff",id="top"
            )
        )
        yield Footer()

    def action_toggle_dark(self) -> None:
        """An action to toggle dark mode."""
        top_collapsible_widget = self.query_one("#top")
        top_collapsible_widget.title = "UPDATED_Top_Stuff"
        self.app.title = "UPDATED_App_Title"
        self.dark = not self.dark


if __name__ == "__main__":
    app = AnApp()
    app.run()
~~~

TomJGooding:
Closes #3829

- [x] Updated documentation


TomJGooding:
Hm, I'm getting this error trying to run `make docs-serve-offline`  - anyone  come across this before?

```
RecursionError: maximum recursion depth exceeded in comparison
```

davep:
Same here, @rodrigogiraoserrao too.

TomJGooding:
Blimey was not expecting a reply that fast, thanks Dave. Have you found any workaround?

davep:
Unfortunately not; I had to make some doc changes in the blind today because of this.

TomJGooding:
Perhaps it has been a while since I updated the docs, but I think this was working _fairly_ recently. Do you happen to remember approximately when you started getting this error?

rodrigogiraoserrao:
> Have you found any workaround?

Don't try to build the docs.

davep:
> Do you happen to remember approximately when you started getting this error?

Yesterday. It's the first time I've needed to try and personally build the docs in weeks.

willmcgugan:
Docs should be fixed in https://github.com/Textualize/textual/pull/3841

src/textual/widgets/_collapsible.py
```diff
@@ -214,3 +217,8 @@ def compose_add_child(self, widget: Widget) -> None:
             widget: A Widget to add.
         """
         self._contents_list.append(widget)
+
+    def _watch_title(self, title: str) -> None:
+        if not self.is_mounted:
+            return
+        self._title.label = title
```
rodrigogiraoserrao, line 224: I'd argue (I may be wrong) that we should always update `label` in `_title`, it's the collapsible title that should determine whether it wants to render or not.
I can also see that with the current implementation of `CollapsibleTitle`, this isn't exactly easy...

Which triggered some thinking... maybe `CollapsibleTitle` should be a `Static` with two attribute strings that represent the collapsed and non-collapsed labels, and when `label` or `collapsed` are updated the watchers just update those strings and the one that is in use.
This means that the rendering of `CollapsibleTitle` can be cached.

rodrigogiraoserrao (approved):

src/textual/widgets/_collapsible.py
TomJGooding: Sorry @rodrigogiraoserrao, I'm afraid I don't fully understand what you mean. How would `renderable` be initialised if `CollapsibleTitle` was a subclass of `Static`?

rodrigogiraoserrao: Completely untested, but I mean smth like

```
class CollapsibleTitle(Static):
    title = reactive(...)
    collapsed = reactive(...)

    def watch_title(...):
        self.collapsed_title = f"> {title}"
        self.expanded_title = f"v {title}"

    def watch_collapsed(...):
        if collapsed: self.update(self.collapsed_title)
        else: self.update(self.expanded_title)

    ...
```

rodrigogiraoserrao: Maybe something for a different PR? Not sure.

TomJGooding: I've changed `CollapsibleTitle` to `Static`, hopefully I've understood correctly. I'm afraid I still don't fully understand how this works despite tests passing.

rodrigogiraoserrao: Maybe you did it already, but reading the source code for `Static` might help.
It's surprisingly short and it'll probably help.

TomJGooding: Sorry I'm probably just missing something obvious. I understand how `Static` works, but what I don't quite get is how the `renderable` is being initialised here. Aren't watch methods only called if the value changes?

src/textual/widgets/_collapsible.py
```diff
@@ -214,3 +224,8 @@ def compose_add_child(self, widget: Widget) -> None:
             widget: A Widget to add.
         """
         self._contents_list.append(widget)
+
+    def _watch_title(self, title: str) -> None:
+        if not self.is_mounted:
```
willmcgugan, line 229: Do we need this? If somebody extends this class and modifies `title` in the constructor, won't this prevent the title from updating?

AFAIK we only need this if we are using the DOM, but updating the state is not a problem.

src/textual/widgets/_collapsible.py
willmcgugan: @rodrigogiraoserrao I think I need clarification as well.

rodrigogiraoserrao: The renderable is initialised when `collapsed = reactive(True)` is set, which triggers its watch method which calls `update`.
I thought this was “obvious”! 🤔

willmcgugan: No, I don't understand your point at the beginning of this thread. I don't know what you are arguing for. Can you elaborate?

rodrigogiraoserrao: It's harder to understand what I was saying if you didn't review the code in its first iteration.
If you think this implementation looks good (with `CollapsibleTitle` inheriting from `Static`), then that means you probably agree with my point.

The `if not self.is_mounted` eventually led me to think `CollapsibleTitle` could inherit from `Static` so that it easily caches its rendering, which is always one of two different strings.

I was also advocating for an attempt at removing that `if` if possible.

src/textual/widgets/_collapsible.py
TomJGooding: Here's the trackback if this is removed:

```python
$ textual run docs/examples/widgets/collapsible.py 
╭──────────────────────────────────────────────────────────── Traceback (most recent call last) ─────────────────────────────────────────────────────────────╮
│ /home/tom/Projects/textual/docs/examples/widgets/collapsible.py:33 in compose                                                                              │
│                                                                                                                                                            │
│   30 │   def compose(self) -> ComposeResult:                                                                                                               │
│   31 │   │   """Compose app with collapsible containers."""                                                                                                │
│   32 │   │   yield Footer()                                                                                                                                │
│ ❱ 33 │   │   with Collapsible(collapsed=False, title="Leto"):                                                                                              │
│   34 │   │   │   yield Label(LETO)                                                                                                                         │
│   35 │   │   yield Collapsible(Markdown(JESSICA), collapsed=False, title="Jessica")                                                                        │
│   36 │   │   with Collapsible(collapsed=True, title="Paul"):                                                                                               │
│                                                                                                                                                            │
│ ╭─────────────────────────────── locals ────────────────────────────────╮                                                                                  │
│ │ self = CollapsibleApp(title='CollapsibleApp', classes={'-dark-mode'}) │                                                                                  │
│ ╰───────────────────────────────────────────────────────────────────────╯                                                                                  │
│                                                                                                                                                            │
│ /home/tom/Projects/textual/src/textual/widgets/_collapsible.py:182 in __init__                                                                             │
│                                                                                                                                                            │
│   179 │   │   │   disabled: Whether the collapsible is disabled or not.                        ╭───────────── locals ─────────────╮                        │
│   180 │   │   """                                                                              │         children = ()            │                        │
│   181 │   │   super().__init__(name=name, id=id, classes=classes, disabled=disabled)           │          classes = None          │                        │
│ ❱ 182 │   │   self.title = title                                                               │        collapsed = False         │                        │
│   183 │   │   self._title = CollapsibleTitle(                                                  │ collapsed_symbol = '▶'           │                        │
│   184 │   │   │   label=title,                                                                 │         disabled = False         │                        │
│   185 │   │   │   collapsed_symbol=collapsed_symbol,                                           │  expanded_symbol = '▼'           │                        │
│                                                                                                │               id = None          │                        │
│                                                                                                │             name = None          │                        │
│                                                                                                │             self = Collapsible() │                        │
│                                                                                                │            title = 'Leto'        │                        │
│                                                                                                ╰──────────────────────────────────╯                        │
│                                                                                                                                                            │
│ /home/tom/Projects/textual/src/textual/widgets/_collapsible.py:229 in _watch_title                                                                         │
│                                                                                                                                                            │
│   226 │   │   self._contents_list.append(widget)                                               ╭─────── locals ────────╮                                   │
│   227 │                                                                                        │  self = Collapsible() │                                   │
│   228 │   def _watch_title(self, title: str) -> None:                                          │ title = 'Toggle'      │                                   │
│ ❱ 229 │   │   self._title.label = title                                                        ╰───────────────────────╯                                   │
│   230                                                                                                                                                      │
╰────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
AttributeError: 'Collapsible' object has no attribute '_title'
```

TomJGooding: I've now removed this so hopefully we're almost there.

Sorry I thought this was going to be a simple PR but I think my contributing has been more of a hindrance than help here. Thanks for your patience.

rodrigogiraoserrao (approved):
Still looks good.

src/textual/widgets/_collapsible.py
rodrigogiraoserrao: It's not your fault. Your PR was in excellent shape when I first reviewed, I was the one that pushed you to do something else :P

willmcgugan: > Sorry I thought this was going to be a simple PR but I think my contributing has been more of a hindrance than help here. Thanks for your patience.

Not at all. You contributions are always welcome!

PR #3837, opened 2023-12-08
Optimizations

willmcgugan:
- Optimize Horizontal and Vertical layout
- Optimize stylesheet processing
- Optimize invoking methods

### DIFF
--- a/src/textual/dom.py
+++ b/src/textual/dom.py
@@ -1176,19 +1176,20 @@ class DOMNode(MessagePump):
         """
         return self._classes.issuperset(class_names)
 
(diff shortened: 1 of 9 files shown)


src/textual/dom.py
```diff
@@ -1209,10 +1209,11 @@ def _update_styles(self) -> None:
 
         Should be called whenever CSS classes / pseudo classes change.
         """
-        try:
-            self.app.update_styles(self)
-        except NoActiveAppError:
-            pass
+        if self._is_mounted:
```
willmcgugan, line 1212: Avoid updating styles until this widget is mounted.

PR #3844, opened 2023-12-10
Faster css

willmcgugan:
Speeds up CSS by discarding rules that would never apply as early as possible.

This is big win. `textual colors` startup time reduced from 1026ms to 630ms. Not too shabby.

### DIFF
--- a/src/textual/css/stylesheet.py
+++ b/src/textual/css/stylesheet.py
@@ -179,6 +179,7 @@ class Stylesheet:
                 for name in rule.selector_names:
                     rules_map[name].append(rule)
             self._rules_map = dict(rules_map)
+
         return self._rules_map
 
     @property
@@ -335,6 +336,7 @@ class Stylesheet:
             return
         self.source[read_from] = CssSource(css, is_default_css, tie_breaker, scope)
         self._require_parse = True
+        self._rules_map = None
 
     def parse(self) -> None:
         """Parse the source in the stylesheet.
@@ -416,7 +418,6 @@ class Stylesheet:
         self,
         node: DOMNode,
         *,
-        limit_rules: set[RuleSet] | None = None,
         animate: bool = False,
     ) -> None:
         """Apply the stylesheet to a DOM node.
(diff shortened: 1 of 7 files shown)


src/textual/css/stylesheet.py
```diff
@@ -634,31 +636,14 @@ def update_nodes(self, nodes: Iterable[DOMNode], animate: bool = False) -> None:
...
+                    apply(node.vertical_scrollbar)
                 if node.show_horizontal_scrollbar:
-                    rules = get_rules(node.horizontal_scrollbar)
-                    if rules:
-                        apply(node.horizontal_scrollbar, limit_rules=rules)
+                    apply(node.horizontal_scrollbar)
                 if node.show_horizontal_scrollbar and node.show_vertical_scrollbar:
-                    rules = get_rules(node.scrollbar_corner)
-                    if rules:
-                        apply(node.scrollbar_corner, limit_rules=rules)
```
rodrigogiraoserrao, line 664: Is this part of the optimisation process or is it just a simplification of the code because the check `if rules` was somehow superfluous?

src/textual/strip.py
```diff
@@ -199,7 +199,7 @@ def join(cls, strips: Iterable[Strip | None]) -> Strip:
         return strip
 
     def __bool__(self) -> bool:
-        return bool(self._segments)
```
rodrigogiraoserrao, line 202: This is faster!?
If so, _maybe_ it's worth adding a comment in front saying this is a micro optimisation.

What if you call `self._segments.__bool__()` yourself?

src/textual/strip.py
```diff
@@ -199,7 +199,7 @@ def join(cls, strips: Iterable[Strip | None]) -> Strip:
...
+        return not not self._segments
```
rodrigogiraoserrao, line 202:
```suggestion
        return not not self._segments  # faster than bool(...)
```

rodrigogiraoserrao (approved):

src/textual/css/stylesheet.py
willmcgugan: The `get_rules` pattern is not handled by `apply`. So its redundant here.

src/textual/strip.py
willmcgugan: It's faster because `bool()` requires a global lookup and a function call. `not not` is two op-codes, no explicit function call.

It's around 75% faster. But still unlikely to make a tangible difference.

PR #3845, opened 2023-12-10
drop importlib-metadata

branchvincent:
This is now included in all supported python versions (starting with 3.8): https://docs.python.org/3/library/importlib.metadata.html

- [ ] Updated documentation


rodrigogiraoserrao:
Hey there, Branch,

Thanks for your PR. The `importlib-metadata` page on PyPI (https://pypi.org/project/importlib-metadata/) shows a compatibility table and the version >=4.11.3 that we request is in the middle between the version that matches that of 3.8 and 3.10.
Did you check that we can really drop the third-party version and use the standard-library version in Python 3.8 and 3.9?

src/textual/__init__.py
```diff
@@ -26,19 +27,7 @@
...
-        try:
-            from importlib.metadata import version
-        except ImportError:
-            import pkg_resources
-
-            return pkg_resources.get_distribution("textual").version
-        else:
-            return version("textual")
-    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
+__version__ = version("textual")
```
rodrigogiraoserrao, line 30: Even if we use `importlib.metadata`'s `version`, we probably still want it to be lazily computed to make sure Textual's startup time isn't impacted by the retrieval of the version when it's not needed.
Does that make sense?

willmcgugan: If I recall that was the reason for doing it lazily. It was surprisingly slow.

Docs say that importlib.metadata was adding in 3.8. We still support 3.7, so I don't think we could accept this PR.

TomJGooding: > We still support 3.7

I thought this was dropped following #3766?

willmcgugan: You're right. Sorry, I'm a version behind.

willmcgugan: In that case, no objections if we restore the lazy behaviour!

branchvincent: restored the lazy import!

branchvincent:
> Did you check that we can really drop the third-party version

Yup, you're only using the simple `importlib.metadata.version` API which hasn't changed since it was introduced in [3.8](https://github.com/python/cpython/blob/v3.8.0/Lib/importlib/metadata.py#L465)

rodrigogiraoserrao (approved):
Great, thanks for this!

I hope this doesn't rub you the wrong way but I wanted to ask you why you decided to force-push your second commit as force-pushing is typically a “destructive” action.
(In this case, it broke the links to Will's and Tom's comments, see screenshot below.)

I was even going to write an explanation of why it's typically a bad idea but your GH stats and activity imply that you know your way around, so instead I'll ask if you did it on purpose for any particular reason!

Screenshot 2023-12-14 at 12 01 33

branchvincent:
Oh just out of habit from the preferences of work/other projects (i.e. they value a linear history on `main` and thus don't allow merge commits, requiring each PR to either be squash merged or preferably have tidy atomic commits that can be rebased onto `main`). Basically, the idea is that the only commit history worth preserving is that on `main` (not the intermediate states of your commits during review cycles) and so you rewrite/fixup commits in a PR in order to keep the mainline clean

But agreed that the GitHub UI doesn't always play nice with that workflow and I'll keep it in mind for this project :)

rodrigogiraoserrao:
Thanks for explaining, @branchvincent! I guess that makes sense.

PR #3848, opened 2023-12-11
fixed regular expressions for improving performance

eukub:

- [x] Updated documentation

### DIFF
--- a/src/textual/css/tokenize.py
+++ b/src/textual/css/tokenize.py
@@ -24,12 +24,12 @@ SCALAR = rf"{DECIMAL}(?:fr|%|w|h|vw|vh)"
 DURATION = r"\d+\.?\d*(?:ms|s)"
 NUMBER = r"\-?\d+\.?\d*"
 COLOR = rf"{HEX_COLOR}|{RGB_COLOR}|{HSL_COLOR}"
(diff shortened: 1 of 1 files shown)


rodrigogiraoserrao:
Hey there, thanks for your PR!

If your change improved performance, can you share the benchmark(s) you ran / by how much did performance improve?

Also, [if I'm reading the docs correctly](https://docs.python.org/3/library/re.html#regular-expression-syntax), `\w` is only equivalent to `[a-zA-Z0-9_]` if we use the flag `re.ASCII` and it doesn't look like we are, right?
Or am I missing something?

rodrigogiraoserrao:
I was looking around and your PR might be a good idea, not from the point of view of performance, but to allow Unicode characters in class names, CSS IDs, variables, etc.
We won't push this forward right now because we want to consider the possible ramifications of making this change carefully.

We'll get back to you!

eukub:
> I was looking around and your PR might be a good idea, not from the point of view of performance, but to allow Unicode characters in class names, CSS IDs, variables, etc. We won't push this forward right now because we want to consider the possible ramifications of making this change carefully.
> 
> We'll get back to you!

Great! :)

PR #3856, opened 2023-12-12
Fix link previews on Twitter

simonw:
Because Elon broke the large summary card https://twitter.com/willmcgugan/status/1734586762863710659 looks like this:

IMG_6832

Switching the card type to `summary` improves things a bit:

IMG_6833


willmcgugan:
Thanks!

Issue #3858, opened 2023-12-12
Style changes made `on_mount` no longer applying correctly in 0.45


davep:
Given this code:

```python
from textual import __version__
from textual.app import App, ComposeResult
from textual.widgets import Static

class BrokenClassesApp(App[None]):

    CSS = """
    Screen {
        align: center middle;
    }

    Static {
        width: 50%;
        height: 50%;
        border: solid;
    }

    Screen.go-red Static {
        background: red;
    }
    """

    def compose(self) -> ComposeResult:
        yield Static(f"This should have a red background\n\nTextual {__version__}")

    def on_mount(self) -> None:
        self.screen.set_class(True, "go-red")

if __name__ == "__main__":
    BrokenClassesApp().run()
```

in Textual 0.44.x the result was as I'd expect:

![Screenshot 2023-12-12 at 17 56 58](https://github.com/Textualize/textual/assets/28237/d54a76ea-a5d0-4024-84a8-5e9ed1d350c8)

but as of 0.45 it seems the class no longer applies:

![Screenshot 2023-12-12 at 17 57 09](https://github.com/Textualize/textual/assets/28237/46331264-5d8e-46f6-a498-3d14f8c102cd)

Flagging @willmcgugan for triage as this seems like an unintended consequence of some of the CSS performance tweaks.

PR #3860, opened 2023-12-12
fix update styles

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/3858


willmcgugan:
#hotfix Good catch @davep

davep:
Looking good here; thanks for the quick fix!

PR #3861, opened 2023-12-13
Add workarounds for foot terminal

danofsteel32:
See https://codeberg.org/dnkl/foot/issues/628 for why foot terminal handles Shift+Enter and Control+Enter differently than other terminals.

Without these 2 fixes pressing Shift+Enter/Control+Enter in a TextArea causes the whole App to hang for ~10 seconds before inserting the raw escape sequence.

With these 2 fixes Shift+Enter/Control+Enter insert a carriage return.

### DIFF
--- a/src/textual/_ansi_sequences.py
+++ b/src/textual/_ansi_sequences.py
@@ -370,6 +370,10 @@ ANSI_SEQUENCES_KEYS: Mapping[str, Tuple[Keys, ...] | str] = {
     "\x1bOx": "8",
     "\x1bOy": "9",
     "\x1bOM": (Keys.Enter,),
+    # Foot terminal specific
+    # https://codeberg.org/dnkl/foot/issues/628
+    "\x1b[27;2;13~": (Keys.Enter,),  # Shift+Enter
+    "\x1b[27;5;13~": (Keys.Enter,),  # Ctrl+Enter
 }
 
 # https://gist.github.com/christianparpart/d8a62cc1ab659194337d73e399004036


rodrigogiraoserrao:
Hey Daniel, thanks for the PR and the extra context.
Right now, Dave is doing some work on escape sequences in #3800.

I don't understand _a lot_ about control sequences so I need your help to understand this: if these 2 control sequences do _not_ get added specifically for foot and instead they're handled by the general system in #3800 for unknown sequences, what would happen if you hit Ctrl + Enter / Shift + Enter inside a TextArea?

danofsteel32:
I dug into this a bit more and I was able to find a workaround in foot so this doesn't actually need to be merged.
I think an FAQ like 'Why do I see `[27;5;13~` or `[27;2;13~` when I press control/shift + enter?' is a better way to handle it.
I'd be happy to write the FAQ if you point me to where it needs to go

rodrigogiraoserrao:
> I dug into this a bit more and I was able to find a workaround in foot so this doesn't actually need to be merged.

Excellent!

> I think an FAQ like 'Why do I see `[27;5;13~` or `[27;2;13~` when I press control/shift + enter?' is a better way to handle it. 

I think this is a bit niche for a F(requently)AQ, no?
Of course I may be misunderstanding something, but hitting Shift + Enter / Ctrl + Enter inside a TextArea for me just creates a next line!
(iTerm2 on MacOS).

One thing that would undoubtedly be useful is if you write a sentence or two about what you did on the foot side to circumvent this issue so it's documented here.

twrightsman:
I use foot as well and would argue against mapping these combinations to Enter by default. I would like to map shift+enter to other things in my application.

darrenburns:
> I dug into this a bit more and I was able to find a workaround in foot so this doesn't actually need to be merged.

Closing because of this ^

Issue #615, opened 2022-07-18
Is it possible to run Textual apps inside of a Jupyter Notebook?


SlappedWithSilence:
This seems like it should work, but I'm running into AsyncIO issues when I try.

```
RuntimeError                              Traceback (most recent call last)
Input In [12], in <cell line: 9>()
     11 logger.add("debug.log", level="ERROR")
     12 if load_config(CONFIG_PATH):
---> 13     UI.run(log="textual.log")

File ~\.conda\envs\basic\lib\site-packages\textual\app.py:206, in App.run(cls, console, screen, driver, **kwargs)
    203     app = cls(screen=screen, driver_class=driver, **kwargs)
    204     await app.process_messages()
--> 206 asyncio.run(run_app())

File ~\.conda\envs\basic\lib\asyncio\runners.py:33, in run(main, debug)
      9 """Execute the coroutine and return the result.
     10 
     11 This function runs the passed coroutine, taking care of
   (...)
     30     asyncio.run(main())
     31 """
     32 if events._get_running_loop() is not None:
---> 33     raise RuntimeError(
     34         "asyncio.run() cannot be called from a running event loop")
     36 if not coroutines.iscoroutine(main):
     37     raise ValueError("a coroutine was expected, got {!r}".format(main))

RuntimeError: asyncio.run() cannot be called from a running event loop
```

Glyphack:
I guess there's a workaround for this. In the newer versions of the Jupyter notebook, you can [use top level](https://ipython.readthedocs.io/en/stable/interactive/autoawait.html) `await` keyword. This is done by having a running event loop when you start the jupyter.

The conflict with Textualize here is because of [this line]:(https://github.com/Textualize/textual/blob/main/src/textual/app.py)
```
async def run_app() -> None:
  app = cls(screen=screen, driver_class=driver, **kwargs)
  await app.process_messages()

asyncio.run(run_app())
```

So either Texualize can reuse the active event loop here, or you can disable this in the Jupyter:
```
%autoawait False
```

@willmcgugan Do you have any preference on this? I would be happy to help with a patch or write documentation to clarify this.

SlappedWithSilence:
This is basically what I thought was happening. I am not a Python expert, but I am trying to improve.

I tried to patch the `run` method to allow it to re-use the existing event loop:

```
@classmethod
    async def run_notebook(
        cls,
        console: Console = None,
        screen: bool = True,
        driver: Type[Driver] = None,
        **kwargs,
    ):
[...]

PR #3868, opened 2023-12-14
Improve event loop handling around run_app

bsdz:
- [ ] Updated documentation

This resolves issue #615 as it also occurs in Panel which embeds a memray app that utilizes textual.

### DIFF
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -123,9 +123,6 @@ WINDOWS = PLATFORM == "Windows"
 if constants.DEBUG:
     warnings.simplefilter("always", ResourceWarning)
 
-# `asyncio.get_event_loop()` is deprecated since Python 3.10:
-_ASYNCIO_GET_EVENT_LOOP_IS_DEPRECATED = sys.version_info >= (3, 10, 0)
-
 LayoutDefinition = "dict[str, Any]"
 
 DEFAULT_COLORS = {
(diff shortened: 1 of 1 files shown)


willmcgugan:
This looks like it would be implementing [`run_async`](https://textual.textualize.io/api/app/#textual.app.App.run_async), which is the recommended way of integrating Textual with an existing loop.

bsdz:
> This looks like it would be implementing [`run_async`](https://textual.textualize.io/api/app/#textual.app.App.run_async), which is the recommended way of integrating Textual with an existing loop.

I see. Is the issue then, how memray integrates with textual?

willmcgugan:
> I see. Is the issue then, how memray integrates with textual?

Potentially, but I'm not clear on what the issue is. Can you provide me with more information on what you are doing and the error you are getting?

bsdz:
> > I see. Is the issue then, how memray integrates with textual?
> 
> Potentially, but I'm not clear on what the issue is. Can you provide me with more information on what you are doing and the error you are getting?

I was trying to use [Holviz Panel's builtin profiler](https://panel.holoviz.org/how_to/profiling/profile.html). One profiler option in Panel is to use [memray](https://github.com/bloomberg/memray). Panel is able to embed various web based components within a unified framework. The profilers are embeded as web components under the admin panel within Panel. It seems that this worked for memray version 1.10.0 (discovered after some investigation). Since [memray 1.11.0](https://github.com/bloomberg/memray/releases/tag/v1.11.0), textual must have replaced the web components of memray and when one tries to use this version of memray in panel, it fails with the above error. Hence my PR. All that said, it seems even with the above change in code, memray 1.11.0 still doesn't work in Panel. I've now switched to memray 1.10.0 and this seems to help (although not perfectly).

willmcgugan:
You may want to raise this issue with the memray devs. I'm sure there is a simple workaround.

bsdz:
> You may want to raise this issue with the memray devs. I'm sure there is a simple workaround.

That's fine, although I'd recommend changing your code too (as suggested in my PR). The python docs recommend using [get_running_loop](https://docs.python.org/release/3.12.1/library/asyncio-eventloop.html?highlight=get_running_loop#asyncio.get_event_loop) and my proposed approach is similar to that [suggested here](https://stackoverflow.com/a/61331974/370696).

PR #3876, opened 2023-12-14
delay underline in tabs

willmcgugan:
This adds a small delay to the underlines in tabs, to masked the pause caused when tabbed_content mounts a tab pane with many widgets.

You can see this in `textual colors`. The underline would jump half way then animate smoothly. With this change the animation is smooth from the start.

### DIFF
--- a/src/textual/widgets/_tabs.py
+++ b/src/textual/widgets/_tabs.py
@@ -593,12 +593,10 @@ class Tabs(Widget, can_focus=True):
 
                 def animate_underline() -> None:
                     """Animate the underline."""
-                    underline.animate(
-                        "highlight_start", start, duration=0.3, delay=0.05
-                    )
-                    underline.animate("highlight_end", end, duration=0.3, delay=0.05)
+                    underline.animate("highlight_start", start, duration=0.3)
+                    underline.animate("highlight_end", end, duration=0.3)
 
-                self.call_after_refresh(animate_underline)
+                self.set_timer(0.01, lambda: self.call_after_refresh(animate_underline))
             else:
                 underline.highlight_start = start
                 underline.highlight_end = end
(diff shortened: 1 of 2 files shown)


src/textual/widgets/_tabs.py
```diff
@@ -590,8 +590,13 @@ def _highlight_active(self, animate: bool = True) -> None:
...
             if animate:
-                underline.animate("highlight_start", start, duration=0.3)
-                underline.animate("highlight_end", end, duration=0.3)
+
+                def animate_underline() -> None:
+                    """Animate the underline."""
+                    underline.animate("highlight_start", start, duration=0.3)
+                    underline.animate("highlight_end", end, duration=0.3)
+
+                self.set_timer(0.01, lambda: self.call_after_refresh(animate_underline))
```
rodrigogiraoserrao, line 599: Why don't you just set `delay=0.01` on each call to `animate`?

willmcgugan: Tried that. Not sure why, but this way reduced the glitch.

rodrigogiraoserrao: _Could_ be an issue with the way animations are delayed...

rodrigogiraoserrao (approved):

rodrigogiraoserrao:
I forgot about it... But this will mess up snapshot tests in CI: https://github.com/Textualize/textual/actions/runs/7221754163/job/19677399262?pr=3872

Do we just power through?

willmcgugan:
Could have sworn the tests passed. Not so familiar with the GitHub mobile app. Can you fix with a pause or something.

rodrigogiraoserrao:
> Could have sworn the tests passed. Not so familiar with the GitHub mobile app. Can you fix with a pause or something.

Tests passed, yes. It's a new flaky test.

PR #3877, opened 2023-12-14
optimization

willmcgugan:
Avoid doing work when styles don't animate. This impacts startup time, and shaves ~30ms off `textual colors` startup time.


rodrigogiraoserrao (approved):
Looks good to me.

If I understood correctly, “the only thing” you did was take lines that were only needed for the `if` branch and move them inside the `if`, right?

willmcgugan:
Yup.

PR #3880, opened 2023-12-15
rules caching

willmcgugan:
Adds caching of rule attributes to Stylesheet.apply. I had considered this and thought it would be impossible since the cache key would be impractical to calculate. Turns out there is a compromise. You can cache siblings when they are mounted. So if you have a bunch of buttons (for example) with the same parent we only need to apply the CSS rules once.

This reduces `textual colors` by 100ms!

### DIFF
--- a/src/textual/css/stylesheet.py
+++ b/src/textual/css/stylesheet.py
@@ -439,15 +439,15 @@ class Stylesheet:
         rule_attributes: defaultdict[str, list[tuple[Specificity6, object]]]
         rule_attributes = defaultdict(list)
 
+        cache_key: tuple | None
         if cache is not None:
-            cache_key: tuple | None = (
+            cache_key = (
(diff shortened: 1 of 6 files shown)


src/textual/css/stylesheet.py
```diff
@@ -437,6 +439,23 @@ def apply(
...
 
+        if cache is not None:
+            cache_key: tuple | None = (
+                node._parent,
+                node._id,
+                node.classes,
+                node.pseudo_classes,
+                node._css_type_name,
+            )
+            assert cache_key is not None
```
rodrigogiraoserrao, line 450: If you “declare” `cache_key: tuple | None` before the `if`, can't you get rid of this `assert`?

rodrigogiraoserrao (approved):

PR #3885, opened 2023-12-16
Button optimization

willmcgugan:
Avoid calling `_update_styles` when the widget has no parent. This will be the case if a reactive is modified in the constructor. The styles are being updated anyway, so we can avoid the extra work.


willmcgugan:
#executivedecision

PR #3896, opened 2023-12-18
Move child acts as noop if trying to move before/after itself

### LINKED ISSUE #1743: Make moving a child before or after itself, a no-op

rodrigogiraoserrao:
This fixes #1743 for good.

### DIFF
--- a/src/textual/widget.py
+++ b/src/textual/widget.py
@@ -892,9 +892,30 @@ class Widget(DOMNode):
         await_mount = self.mount(*widgets, before=before, after=after)
         return await_mount
 
+    @overload
     def move_child(
         self,
         child: int | Widget,
+        *,
+        before: int | Widget,
+        after: None = None,
+    ) -> None:
+        ...
+
+    @overload
+    def move_child(
+        self,
+        child: int | Widget,
+        *,
+        after: int | Widget,
+        before: None = None,
+    ) -> None:
+        ...
+
+    def move_child(
+        self,
+        child: int | Widget,
+        *,
         before: int | Widget | None = None,
         after: int | Widget | None = None,
     ) -> None:
--- a/src/textual/widget.py
+++ b/src/textual/widget.py
@@ -941,10 +941,6 @@ class Widget(DOMNode):
         elif before is not None and after is not None:
             raise WidgetError("Only one of `before` or `after` can be handled.")
 
-        # We short-circuit the no-op, otherwise it will error later down the road.
-        if child is before or child is after:
-            return
-
         def _to_widget(child: int | Widget, called: str) -> Widget:
             """Ensure a given child reference is a Widget."""
             if isinstance(child, int):
@@ -968,6 +964,9 @@ class Widget(DOMNode):
         target = _to_widget(
             cast("int | Widget", before if after is None else after), "move towards"
         )
+
+        if child is target:
+            return  # Nothing to be done.
 
         # At this point we should know what we're moving, and it should be a
         # child; where we're moving it to, which should be within the child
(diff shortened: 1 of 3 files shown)


jepler (commented):
:+1:

src/textual/widget.py
```diff
@@ -948,6 +965,9 @@ def _to_widget(child: int | Widget, called: str) -> Widget:
             cast("int | Widget", before if after is None else after), "move towards"
```
rodrigogiraoserrao, line 965: The overloads above are smart enough to detect issues when calling the function with none of `after`/`before` or with both but they're not smart enough to infer that one of the two is not `None` inside the function body so I couldn't get rid of this `cast`.

tests/test_widget_child_moving.py
```diff
@@ -36,28 +36,66 @@ async def test_move_child_not_our_child() -> None:
...
+    Regression test for https://github.com/Textualize/textual/issues/1743
+    """
+
+    widget = Widget()
+    keys = {
+        "widget": widget,
+        "index": 0,
+    }
+    child = keys[type_of_child]
+    after = keys[after]
```
darrenburns, line 94: Why the extra lookup here? Does something like the code below not work?

```python
[
  (Widget(), Widget()),
  (Widget(), 0),
  # ...
]
```

And just directly pass in `child` and `after`?

rodrigogiraoserrao: Your proposal will generate many different `Widget` instances and I need to check what happens when the child and before/after `is` the same instance, whereas if `child, after = Widget(), Widget()`, then we get `child is not after`.

darrenburns: I see. The reason that I'm asking is that I find it hard to follow, and the use of conditionals and lookups (which are basically just conditionals) is usually considered a code smell in a parametrised test - it suggests the test is trying to do cover too many things, and there's almost always a better way. Usually, that means breaking the test up into multiple, more readable tests, even if it means duplication.

In this case, I think you're trying to shoehorn in a "a widget can be specified as a integer" test into this test, rather than covering it off in an isolated manner (a separate unit test). If you do this in a separate test, then you can let this test focus on one unit of functionality: move_child nooping.

One other thing is - what is this intended to test? It just calls `move_child` then does nothing. The title of the PR explains that it should no-op, so we could check (within reason) that that's the actual behaviour.

Maybe something like this, to ensure the DOM isn't modified and that it is in fact a null operation?

```python
screen = pilot.app.screen
await pilot.app.mount(widget)
dom_before = screen.walk_children()

screen.move_child(widget, after=widget)

dom_after = screen.walk_children()
assert dom_before == dom_after
```

rodrigogiraoserrao: Point taken regarding lookups/conditionals and parametrised tests.

Regarding the testing, I'm testing that the function call doesn't raise an error (before the PR, it did).
I can be extra safe and check that the DOM did stay the same, but given that the screen has a single `Widget` children, doesn't make a lot of sense to me.

rodrigogiraoserrao: Thanks for your feedback. I've addressed it in e03de27.

darrenburns (approved):
LGTM

PR #3900, opened 2023-12-19
docs(queries): add one period

jrycw:
We seem to be missing a period here.


rodrigogiraoserrao (approved):
Well spotted; thanks.

PR #3905, opened 2023-12-19
Fix highlighted lines in docs example

### LINKED ISSUE #3902: Broken markup in style reference for visibility
See the `visibility_container.tcss` tab

https://textual.textualize.io/styles/visibility/#__tabbed_2_3

rodrigogiraoserrao:
Fix #3902

### DIFF
--- a/docs/css_types/_template.md
+++ b/docs/css_types/_template.md
@@ -51,7 +51,7 @@ If the type has many different syntaxes, cover all of them.
 Add comments when needed/if helpful.
 -->
 
-```sass
+```css
 .some-class {
     rule: type-value-1;
     rule: type-value-2;
--- a/docs/css_types/border.md
+++ b/docs/css_types/border.md
@@ -37,7 +37,7 @@ textual borders
 
 ### CSS
 
-```sass
+```css
 #container {
     border: heavy red;
 }
--- a/docs/css_types/color.md
+++ b/docs/css_types/color.md
@@ -106,7 +106,7 @@ For example, `hsla(128, 100%, 50%, 0.5)` is the color `hsl(128, 100%, 50%)` with
 
 ### CSS
 
-```sass
+```css
 Header {
     background: red;           /* Color name */
 }
--- a/docs/css_types/horizontal.md
+++ b/docs/css_types/horizontal.md
@@ -16,7 +16,7 @@ The [`<horizontal>`](./horizontal.md) type can take any of the following values:
 
 ### CSS
 
-```sass
+```css
 .container {
     align-horizontal: right;
 }
--- a/docs/css_types/integer.md
+++ b/docs/css_types/integer.md
@@ -14,7 +14,7 @@ An [`<integer>`](./integer.md) is any valid integer number like `-10` or `42`.
 
 ### CSS
 
-```sass
+```css
 .classname {
     offset: 10 -20
 }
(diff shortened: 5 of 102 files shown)


willmcgugan:
Can you also fix the annotations.

rodrigogiraoserrao:
I can't, no.

I've tried changing the type of annotation to the simple one (without `!`).
I've tried changing the spacing between annotations.
I've tried removing different annotations to see if one specific one was breaking things.
I've tried reintroducing the annotations one by one.

In principle, the annotations are OK but docs are building wrong.

Seems to be some sort of bug with the docs tooling as I've identified other places where they're funny:

 - https://textual.textualize.io/styles/border_subtitle_align/#__tabbed_2_3
 - https://textual.textualize.io/styles/border_title_align/#__tabbed_2_3

I'm going to identify the plugin/tool that provides this functionality and figure out if there have been bug reports about this.

rodrigogiraoserrao:
Feel free to merge this fix to the highlighted lines and then a follow-up PR will take care of the annotations in this and the other examples.

rodrigogiraoserrao:
Opened #3906 to keep track of the non-rendered annotations.

willmcgugan:
I think I'd we can't do it properly in tcss we should just replace them with comments.

rodrigogiraoserrao:
> I think I'd we can't do it properly in tcss we should just replace them with comments.

It _can_ be done in tcss: https://textual.textualize.io/styles/height/#__tabbed_2_3

rodrigogiraoserrao:
We just need to investigate what's happening.

willmcgugan:
Are you seeing the same thing as me? It's removing the `/*` but leaving in the `*/`. Which looks the the syntax is all wrong. If we can't fix that, then we can just use comments.

rodrigogiraoserrao:
> Are you seeing the same thing as me? It's removing the `/*` but leaving in the `*/`. Which looks the the syntax is all wrong. If we can't fix that, then we can just use comments.

Ah, you only noticed that now?
The trailing `*/` has been there since I can remember and I _think_ mkdocs (or the tool that provides this feature) has a bug report open to fix it.

I thought you were talking about the `/* (n) */` that aren't being parsed / used to create annotations.

rodrigogiraoserrao:
In that case, do you want to get rid of all of the `/* (n) */` in CSS files and use normal comments instead?

willmcgugan:
Yeah, its a shame. But it could be confusing, so lets replace just the tcss annotations with simple comments.

TomJGooding:
Sorry to stick my nose in (as usual!) but I _think_ changing to single-line comments and ensuring they aren't on the same line as the selector might fix this - see https://github.com/Textualize/textual/issues/3906#issuecomment-1863117627

rodrigogiraoserrao:
I confirm Tom's findings, both.

If we change all listings from `sass` to `css`, annotations will work as expected and there will be no leftover `*/`.

If we _really_ want to keep listings as `sass`, then we can use `//` comments but that would require modifying Textual to allow `//` comments in TCSS.

rodrigogiraoserrao:
I suppose we want to go with the option of changing code listings from `sass` to `css`..?

TomJGooding:
> ...but that would require modifying Textual to allow `//` comments in TCSS.

Oops, sorry yeah I should have checked this was actually valid in TCSS! :facepalm:

rodrigogiraoserrao:
I've went ahead and replaced all sass with css listings.
The only place I didn't do it was in blog posts.

davep (approved):
I've not checked every single thing that was modified, but a random sample looks good to me.

PR #3912, opened 2023-12-20
Fix keyboard navigation in option list (& selection list), radio set, list view

### LINKED ISSUE #3881: Keyboard navigation should skip disabled widgets in collections of widgets
- `SelectionList` and `OptionList` allow highlighting disabled items (with keyboard & mouse) but don't allow selection.
 - `ListView` allows highlighting with keyboard but not with mouse (and doesn't allow selection).
 - As of #3872, `RadioSet` doesn't allow neither navigation nor selection.

I feel like we should disable navigation everywhere and instead <kbd>Up</kbd> / <kbd>Down</kbd> should only go through enabled children.

See https://github.com/Textualize/textual/issues/3839#issuecomment-1857885036 for more context.

rodrigogiraoserrao:
Working through this, I have some questions about the API for keyboard navigation.

 - Should keyboard navigation wrap around?

Current behaviour is to wrap in option list/selection list and radio set but not in list view, I didn't change this.

 - When assigning to the reactive that controls the highlighted item, what type of validation do we do?

Previously, the index would just be clamped. Do we clamp to the first/last enabled item?
What if setting to an index in the range but assigning to an item that's disabled?
Do we adjust the value to the next available item?

Currently, I do clamping to the first/last enabled item but if trying to set to a value in range that is disabled, I default back to the previous value.

Should we raise an error if the dev assigns to the reactive and uses an invalid index?
Should we just let the assignment work, even if targeting a disabled item?

Fixes #3881.


src/textual/_widget_navigation.py
davep, line 1: If I'm reading this correctly, the contents of this module would seem to have utility to folk building their own widgets too. Does it make sense to place them in an "internal" module? I can't help but feel that these should be exposed and documented for all.

davep, line 1: I feel that the contents of this module should have their own unit tests.

src/textual/_widget_navigation.py
```diff
@@ -0,0 +1,139 @@
...
+
+def distance(index: int, start: int, direction: Direction, wrap_at: int) -> int:
+    """Computes the distance going from `start` to `index` in the given direction
+
+    Starting at `start`, this is the number of steps you need to take in the given
+    `direction` to reach `index`, assuming there is wrapping at 0 and `wrap_at`.
+    This is also the smallest non-negative integer solution `d` to
+    `(start + d * direction) % wrap_at == index`.
+
+    The diagram below schematises the computation of `d1 = distance(2, 8, 1, 10)` and
```
davep, line 44: Nit-pick, but for consumption by the average reader of the documentation, "schematises" seems a little... unusual. I don't think I've ever seen the word before, and had to look it up to be sure exactly what it meant and that my inference was correct.

Perhaps something like "illustrates" would be more readable?

src/textual/widgets/_option_list.py
davep, line 1: First quick test of how this affects `OptionList` and it looks like some more work might need to be done when it comes to handling page up and page down. Using this code:

```python
from textual.widgets.option_list import Option

class DisabledNavigationTestApp(App[None]):

    CSS = """
    Screen {
        layout: horizontal;
    }

    OptionList {
        height: 1fr;
    }
    """

        yield OptionList(*[
            Option(f"This is option #{n}", disabled=bool(n % 2)) for n in range(1_000)
        ])

if __name__ == "__main__":
    DisabledNavigationTestApp().run()
```

depending on how tall I make the terminal, <kbd>PgDn</kbd> either works or does nothing when I'm on the first option in the list. If you size the terminal so an enabled item is the last visible option it will do nothing, if you size it so a disabled option is the last visible option, it will page down.

davep, line 1: If highlighting of a disabled option is the way we're going now (and I'm all for it; albeit I do low-level wonder if it should be optional in some way), it probably makes sense to remove any code and styling relating to:

- `option-list--option-highlighted-disabled`
- `option-list--option-hover-highlighted-disabled`

From what I can tell they can never happen once this PR is in place.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 
 - `Widget.move_child` would break if `before`/`after` is set to the index of the widget in `child` https://github.com/Textualize/textual/issues/1743
+- Keyboard navigation in `RadioSet`, `ListView`, `OptionList`, and `SelectionList`, no longer allows highlighting disabled items https://github.com/Textualize/textual/issues/3881
```
davep, line 13: I feel this isn't a `Fixed` thing but more of a breaking change thing.

davep (changes requested):
Couple of questions and the like, and what looks to be one bug.

CHANGELOG.md
rodrigogiraoserrao: Point taken.

src/textual/widgets/_option_list.py
rodrigogiraoserrao: To _me_, highlighting a disabled option makes as much sense as having “hover” effects or being able to focus a disabled widget.
Because we can't do the latter, I thought it made sense to prevent the former.

From this reasoning, it follows that those component classes can go.
Thanks for spotting that.

rodrigogiraoserrao: From my own argument it also follows that perhaps `option-list--option-hover-disabled` should go?
Otherwise it looks like disabled options can still be clicked/selected.

davep: Yup, that should go too.

src/textual/_widget_navigation.py
rodrigogiraoserrao: Agreed.
The direct PT 🇵🇹 synonym is a fairly “simple” verb: esquematizar.

src/textual/_widget_navigation.py
rodrigogiraoserrao: Point taken.

rodrigogiraoserrao: Done.

src/textual/widgets/_option_list.py
rodrigogiraoserrao: Thanks for spotting this.
Fixed in 352d3ad.

src/textual/widget_navigation.py
```diff
@@ -0,0 +1,192 @@
+"""
```
willmcgugan, line 1: Does the Textual app developer need to use this code?

I would guess not. I think this should be private, and we don't need to expose it in the api docs.

src/textual/widget_navigation.py
```diff
@@ -0,0 +1,192 @@
...
+    disabled: bool
+
+
+Direction: TypeAlias = Literal[-1, 1]
+"""Valid values to determine navigation direction.
+
+In a vertical setting, 1 points down and -1 points up.
+In a horizontal setting, 1 points right and -1 points left.
+"""
+_D = TypeVar("_D", bound=Disableable)
```
willmcgugan, line 33: I know typing generics tend to use single letter variables for typevars, but I think we should endevor to make them as descriptive as possible.

src/textual/widget_navigation.py
```diff
@@ -0,0 +1,192 @@
...
+_W = TypeVar("_W", bound="Widget")
+
+
+def distance(index: int, start: int, direction: Direction, wrap_at: int) -> int:
```
willmcgugan, line 37: Should really start with a very. At least `get_distance`, but can we come up with a more descriptive name?

willmcgugan (changes requested):
What happens if everything is disabled and we do page-up, page-down ?

src/textual/widget_navigation.py
rodrigogiraoserrao: [In a previous comment](https://github.com/Textualize/textual/pull/3912#discussion_r1439489659), Dave pointed out that if devs want to build their own widgets with sub-widgets that can be navigated with the keyboard, this might come in handy.
I hadn't thought of that and I thought this would be useful and general enough to be helpful to those.

If you really don't agree, then I'm not too fussed about it and I can make it private. 🤷

rodrigogiraoserrao:
> What happens if everything is disabled and we do page-up, page-down ?

Nothing happens because there's no option to go.

src/textual/widget_navigation.py
willmcgugan: I think we should make it private for now. Better to start private then make it public than vice versa.

rodrigogiraoserrao: Sure.

src/textual/_widget_navigation.py
rodrigogiraoserrao: Overruled in https://github.com/Textualize/textual/pull/3912#discussion_r1444451796

src/textual/widget_navigation.py
rodrigogiraoserrao: bfd413b.

src/textual/_widget_navigation.py
```diff
@@ -0,0 +1,198 @@
...
+from itertools import count
+from typing import TYPE_CHECKING, Literal, Protocol, Sequence
+
+from typing_extensions import TypeAlias, TypeVar
+
+if TYPE_CHECKING:
+    from .widget import Widget
+
+
+class Disableable(Protocol):
```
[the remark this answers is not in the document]
rodrigogiraoserrao: Thanks.

mkdocs-nav.yml
```diff
@@ -208,6 +208,7 @@ nav:
       - "api/validation.md"
       - "api/walk.md"
       - "api/widget.md"
+      - "api/widget_navigation.md"
```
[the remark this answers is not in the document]
rodrigogiraoserrao: Of course. Done.

rodrigogiraoserrao:
@willmcgugan I'd still like to have the second bullet point from the original post answered:

When assigning to the reactive that controls the highlighted item, what type of validation do we do?

Previously, the index would just be clamped. Do we clamp to the first/last _enabled_ item?
What if setting to an index in the range but assigning to an item that's disabled?
Do we adjust the value to the next available item?

It may be surprising to have something like

```py
option_list.highlighted = 3
print(option_list.highlighted)  # 7
```

Currently, I do clamping to the first/last enabled item but if trying to set to a value in range that is disabled, I default back to the previous value.
I also don't like this a lot.

Should we raise an error if the dev assigns to the reactive and uses an invalid index?
Should we just let the assignment work, even if targeting a disabled item?

rodrigogiraoserrao:
> I would err on the side of doing what the user is asking. i.e. highlight the item even if it is disabled.

Sounds reasonable.
I let the assignment go through but there's no Highlighted message being posted because the item isn't really being highlighted.

rodrigogiraoserrao:
@willmcgugan feel free to take a look when you're ready.

willmcgugan (approved):

PR #3917, opened 2023-12-21
Touch-up widget guide byte example.

### LINKED ISSUE #3913: Proposed modifications for the example of "Widgets"
This issue is intended to propose modifications to the example provided in [guide/widgets](https://textual.textualize.io/guide/widgets/#__tabbed_18_1), and the original code can be accessed [here](https://github.com/Textualize/textual/blob/ff70df87aad4d0c3ccad45709813299e04faeb03/docs/examples/guide/compound/byte03.py).

*  In reality, the `on_bit_switch_bit_changed` of `ByteEditor` receives `BitSwitch.BitChanged` but does not use it. Consequently, there is no need to pass the `bit` and `value` to `BitSwitch.BitChanged` while initializing.

* The `value`(`textual.reactive.reactive`) in `BitSwitch` appears to indicate the switch status (on/off). Therefore, it seems more natural to use `value = reactive(False)` instead of `value = reactive(0)`.

To summarize, I would suggest modifying `BitSwitch` with the following code:

```python
class BitSwitch(Widget):
    """A Switch with a numeric label above it."""

    DEFAULT_CSS = """
    BitSwitch {
        layout: vertical;
        width: auto;
        height: auto;
    }
    BitSwitch > Label {
        text-align: center;
        width: 100%;
    }
    """

    class BitChanged(Message):
        """Sent when the 'bit' changes."""

    value = reactive(False)

    def __init__(self, bit: int) -> None:
        self.bit = bit
        super().__init__()

        yield Label(str(self.bit))
        yield Switch()

    def watch_value(self, value: bool) -> None:
        """When the value changes we want to set the switch accordingly."""
        self.query_one(Switch).value = value

    def on_switch_changed(self, event: Switch.Changed) -> None:
        """When the switch changes, notify the parent via a message."""
        event.stop()
        self.value = event.value
        self.post_message(self.BitChanged())
```

rodrigogiraoserrao:
Fixes #3913.

Set `BitSwitch.value` to a Boolean because everywhere else in the code it's used as if it were a Boolean.

Change computation when value is changed to make use of the parameters of `BitChanged` event.
(Given the context of the guide, we want to show how to coordinate widgets; in a `XChanged` event it makes sense to add parameters to signal what was changed and it's very common to do that; this example does it but then doesn't use the parameters, so we'd need to remove them... or refactor the event handler to be less "wasteful" and actually use that information)

### DIFF
--- a/docs/examples/guide/compound/byte02.py
+++ b/docs/examples/guide/compound/byte02.py
@@ -31,7 +31,7 @@ class BitSwitch(Widget):
             self.bit = bit
             self.value = value
 
-    value = reactive(0)  # (1)!
+    value = reactive(False)  # (1)!
 
     def __init__(self, bit: int) -> None:
         self.bit = bit
@@ -90,9 +90,12 @@ class ByteEditor(Widget):
 
     def on_bit_switch_bit_changed(self, event: BitSwitch.BitChanged) -> None:
         """When a switch changes, update the value."""
-        value = 0
-        for switch in self.query(BitSwitch):
-            value |= switch.value << switch.bit
+        current_value = int(self.query_one(Input).value or "0")
+        value_change = 1 << event.bit
+        if event.value:  # Bit changed from 0 to 1.
+            value = current_value + value_change
+        else:  # Bit changed from 1 to 0.
+            value = current_value - value_change
         self.query_one(Input).value = str(value)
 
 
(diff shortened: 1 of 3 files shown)


willmcgugan:
Not using the values in the event is not really a problem which needs fixing. It's not broken per se. It's a good idea to add information to messages, but there is nothing that compels you to use all of them.

The lack of bitwise arithmetic is a bug in waiting. i.e. if a future update causes two BitChanged messages with a 1 bit, it will break.

I think this code is fine as a teaching aid as it is.

rodrigogiraoserrao:
Points taken.

PR #3929, opened 2023-12-26
fix ListView not scrolling correctly when updating its index before its children have been refreshed

arvidfm:
Currently `ListView` fails to scroll to a new index if you add new children and then immediately update its index to one of the new children, even if you `await` the `ListView.extend` method, because the sizing of the children isn't updated until the next refresh. This defers the scroll until after the next refresh.

- [ ] Updated documentation


willmcgugan:
Thanks! This has gone a bit stale. We will make the changes locally.

PR #3862, opened 2023-12-13
Fix decorator typing issue.

rodrigogiraoserrao:
@willmcgugan like I mentioned previously, this seems to be the issue and Pylance is happy about it but mypy complains in _other_ places about the changes I made.

In particular, running mypy (both 1.7.1 and `1.9.0+dev.0567da99784c376e723907daf4f16df2f03368c1 0567da9`) in our codebase introduces these type errors:

<details>
<summary> new mypy errors </summary>

```
src/textual/widgets/_log.py:120: error: Argument 1 has incompatible type "Callable[[Log, int, List[str]], None]"; expected "Union[Callable[[VarArg(Never), KwArg(Never)], Never], Callable[[VarArg(Never), KwArg(Never)], Coroutine[Any, Any, Never]]]"  [arg-type]
src/textual/widgets/_log.py:223: error: Invalid self argument "Self" to attribute function "_update_size" with type "Callable[[VarArg(Never), KwArg(Never)], Worker[Never]]"  [misc]
src/textual/widgets/_log.py:223: error: Argument 1 to "_update_size" of "Log" has incompatible type "int"; expected Never  [arg-type]
src/textual/widgets/_log.py:223: error: Argument 2 to "_update_size" of "Log" has incompatible type "List[str]"; expected Never  [arg-type]
src/textual/widgets/_directory_tree.py:182: error: Invalid self argument "DirectoryTree" to attribute function "_loader" with type "Callable[[VarArg(Never), KwArg(Never)], Worker[Never]]"  [misc]
src/textual/widgets/_directory_tree.py:396: error: Argument 1 has incompatible type "Callable[[DirectoryTree, TreeNode[DirEntry]], List[Path]]"; expected "Union[Callable[[VarArg(Never), KwArg(Never)], Never], Callable[[VarArg(Never), KwArg(Never)], Coroutine[Any, Any, Never]]]"  [arg-type]
src/textual/widgets/_directory_tree.py:414: error: Argument 1 has incompatible type "Callable[[DirectoryTree], Coroutine[Any, Any, None]]"; expected "Union[Callable[[VarArg(Never), KwArg(Never)], Never], Callable[[VarArg(Never), KwArg(Never)], Coroutine[Any, Any, Never]]]"  [arg-type]
src/textual/widgets/_directory_tree.py:426: error: Invalid self argument "DirectoryTree" to attribute function "_load_directory" with type "Callable[[VarArg(Never), KwArg(Never)], Worker[Never]]"  [misc]
[...]

PR #3937, opened 2023-12-30
add missing generic types to Worker apis

arvidfm:
makes pyright stop yelling at me

before:
```
> poetry run pyright
error: Type of "get_current_worker" is partially unknown
    Type of "get_current_worker" is "() -> Worker[Unknown]" (reportUnknownMemberType)
error: Type of "worker" is partially unknown
    Type of "worker" is "Worker[Unknown]" (reportUnknownVariableType)
error: Type of "get_current_worker" is partially unknown
    Type of "get_current_worker" is "() -> Worker[Unknown]" (reportUnknownMemberType)
error: Type of "worker" is partially unknown
    Type of "worker" is "Worker[Unknown]" (reportUnknownVariableType)
error: Type of "get_current_worker" is partially unknown
    Type of "get_current_worker" is "() -> Worker[Unknown]" (reportUnknownMemberType)
error: Type of "worker" is partially unknown
    Type of "worker" is "Worker[Unknown]" (reportUnknownVariableType)
error: Type of "cancel_node" is partially unknown
    Type of "cancel_node" is "(node: DOMNode) -> list[Worker[Unknown]]" (reportUnknownMemberType)
error: Type of "get_current_worker" is partially unknown
    Type of "get_current_worker" is "() -> Worker[Unknown]" (reportUnknownMemberType)
error: Type of "worker" is partially unknown
    Type of "worker" is "Worker[Unknown]" (reportUnknownVariableType)
error: Type of "cancel_node" is partially unknown
    Type of "cancel_node" is "(node: DOMNode) -> list[Worker[Unknown]]" (reportUnknownMemberType)
error: Type of "cancel_node" is partially unknown
    Type of "cancel_node" is "(node: DOMNode) -> list[Worker[Unknown]]" (reportUnknownMemberType)
11 errors, 0 warnings, 0 informations 
```

after:
```
> poetry run pyright
0 errors, 0 warnings, 0 informations
```

### DIFF
--- a/src/textual/_worker_manager.py
+++ b/src/textual/_worker_manager.py
@@ -37,7 +37,7 @@ class WorkerManager:
         """
         self._app = app
         """A reference to the app."""
-        self._workers: set[Worker] = set()
+        self._workers: set[Worker[Any]] = set()
         """The workers being managed."""
 
     def __rich_repr__(self) -> rich.repr.Result:
@@ -64,7 +64,7 @@ class WorkerManager:
         return worker in self._workers
 
     def add_worker(
-        self, worker: Worker, start: bool = True, exclusive: bool = True
+        self, worker: Worker[Any], start: bool = True, exclusive: bool = True
     ) -> None:
         """Add a new worker.
 
@@ -91,7 +91,7 @@ class WorkerManager:
         start: bool = True,
         exclusive: bool = False,
         thread: bool = False,
-    ) -> Worker:
+    ) -> Worker[Any]:
         """Create a worker from a function, coroutine, or awaitable.
 
         Args:
@@ -119,7 +119,7 @@ class WorkerManager:
         self.add_worker(worker, start=start, exclusive=exclusive)
         return worker
 
-    def _remove_worker(self, worker: Worker) -> None:
+    def _remove_worker(self, worker: Worker[Any]) -> None:
         """Remove a worker from the manager.
 
         Args:
@@ -137,7 +137,7 @@ class WorkerManager:
         for worker in self._workers:
             worker.cancel()
 
-    def cancel_group(self, node: DOMNode, group: str) -> list[Worker]:
+    def cancel_group(self, node: DOMNode, group: str) -> list[Worker[Any]]:
         """Cancel a single group.
 
         Args:
@@ -156,7 +156,7 @@ class WorkerManager:
             worker.cancel()
         return workers
 
-    def cancel_node(self, node: DOMNode) -> list[Worker]:
+    def cancel_node(self, node: DOMNode) -> list[Worker[Any]]:
         """Cancel all workers associated with a given node
 
         Args:
@@ -170,7 +170,9 @@ class WorkerManager:
             worker.cancel()
         return workers
 
-    async def wait_for_complete(self, workers: Iterable[Worker] | None = None) -> None:
+    async def wait_for_complete(
+        self, workers: Iterable[Worker[Any]] | None = None
+    ) -> None:
         """Wait for workers to complete.
 
         Args:
--- a/src/textual/_worker_manager.py
+++ b/src/textual/_worker_manager.py
@@ -14,7 +14,7 @@ from typing import TYPE_CHECKING, Any, Iterable, Iterator
 
 import rich.repr
 
-from .worker import Worker, WorkerState, WorkType
+from .worker import ResultType, Worker, WorkerState, WorkType
 
 if TYPE_CHECKING:
     from .app import App
@@ -81,7 +81,7 @@ class WorkerManager:
 
     def _new_worker(
         self,
-        work: WorkType,
+        work: WorkType[ResultType],
         node: DOMNode,
         *,
         name: str | None = "",
@@ -91,7 +91,7 @@ class WorkerManager:
         start: bool = True,
         exclusive: bool = False,
         thread: bool = False,
-    ) -> Worker[Any]:
+    ) -> Worker[ResultType]:
         """Create a worker from a function, coroutine, or awaitable.
 
         Args:
@@ -107,7 +107,7 @@ class WorkerManager:
         Returns:
             A Worker instance.
         """
-        worker: Worker[Any] = Worker(
+        worker: Worker[ResultType] = Worker(
             node,
             work,
             name=name or getattr(work, "__name__", "") or "",
--- a/src/textual/widgets/_directory_tree.py
+++ b/src/textual/widgets/_directory_tree.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 from asyncio import Queue
 from dataclasses import dataclass
 from pathlib import Path
-from typing import TYPE_CHECKING, Callable, ClassVar, Iterable, Iterator
+from typing import TYPE_CHECKING, Any, Callable, ClassVar, Iterable, Iterator
 
 from ..await_complete import AwaitComplete
 
@@ -375,7 +375,7 @@ class DirectoryTree(Tree[DirEntry]):
             )
         node.expand()
 
-    def _directory_content(self, location: Path, worker: Worker) -> Iterator[Path]:
+    def _directory_content(self, location: Path, worker: Worker[Any]) -> Iterator[Path]:
         """Load the content of a given directory.
 
         Args:
--- a/src/textual/worker.py
+++ b/src/textual/worker.py
@@ -138,7 +138,7 @@ class Worker(Generic[ResultType]):
     def __init__(
         self,
         node: DOMNode,
-        work: WorkType | None = None,
+        work: WorkType[ResultType] | None = None,
         *,
         name: str = "",
         group: str = "default",
--- a/src/textual/worker.py
+++ b/src/textual/worker.py
@@ -11,6 +11,7 @@ from contextvars import ContextVar
 from time import monotonic
 from typing import (
     TYPE_CHECKING,
+    Any,
     Awaitable,
     Callable,
     Coroutine,
@@ -30,7 +31,7 @@ if TYPE_CHECKING:
     from .dom import DOMNode
 
 
-active_worker: ContextVar[Worker] = ContextVar("active_worker")
+active_worker: ContextVar[Worker[Any]] = ContextVar("active_worker")
 """Currently active worker context var."""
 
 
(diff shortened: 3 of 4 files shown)


src/textual/widgets/_directory_tree.py
```diff
@@ -375,7 +375,7 @@ def _populate_node(self, node: TreeNode[DirEntry], content: Iterable[Path]) -> N
             )
         node.expand()
 
-    def _directory_content(self, location: Path, worker: Worker) -> Iterator[Path]:
+    def _directory_content(self, location: Path, worker: Worker[Any]) -> Iterator[Path]:
```
rodrigogiraoserrao, line 378: If I'm reading the code correctly, this `worker` instance comes from `_load_directory`, so it would be a `Worker[list[Path]]`, which is a bit more specific than `Worker[Any]`.

src/textual/_worker_manager.py
```diff
@@ -91,7 +91,7 @@ def _new_worker(
         start: bool = True,
         exclusive: bool = False,
         thread: bool = False,
-    ) -> Worker:
+    ) -> Worker[Any]:
```
rodrigogiraoserrao, line 94: Looking at this, at line 84, and line 110, shouldn't all of these be connected?
`WorkType` is a generic with `ResultType` as a variable, which looks like it should match the generic argument here and in line 110.

rodrigogiraoserrao (changes requested):
Thanks for this.

A couple of the typing changes can be improved and there are also a couple of situations where I _think_ we can do better.

For reference, this is related to #3862.

rodrigogiraoserrao:
@davep sorry to bother, but given that this is related to [my PR 3862](https://github.com/Textualize/textual/pull/3862) and that the errors I see with mypy/Pylance differ from the ones you see with pyright, could you check that this PR does reduce the number of pyright errors, please?

src/textual/worker.py
```diff
@@ -119,7 +120,7 @@ class Worker(Generic[ResultType]):
     class StateChanged(Message, bubble=False, namespace="worker"):
         """The worker state changed."""
 
-        def __init__(self, worker: Worker, state: WorkerState) -> None:
+        def __init__(self, worker: Worker[Any], state: WorkerState) -> None:
```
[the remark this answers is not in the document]
arvidfm: I don't believe so, no, not unless we make `StateChanged` generic as well. `ResultType` isn't bound here, it belongs to the outer class. To demonstrate:

```py
T = TypeVar("T")
U = TypeVar("U")


class Outer(Generic[T]):
    def  __init__(self, value: T):
        self.value = value

    class Inner(Generic[U]):
        def __init__(self, value1: T, value2: U):
            self.value1 = value1
            self.value2 = value2


if __name__ == '__main__':
    outer = Outer(True)
    reveal_type(outer.value)

    inner = Outer.Inner(5, "test")
    reveal_type(inner.value1)
    reveal_type(inner.value2)
```

Pyright will output:
```
information: Type of "outer.value" is "bool"
  error: Argument of type "Literal[5]" cannot be assigned to parameter "value1" of type "T@Outer" in function "__init__"
    Type "Literal[5]" cannot be assigned to type "T@Outer" (reportGeneralTypeIssues)
information: Type of "inner.value1" is "T@Outer"
information: Type of "inner.value2" is "str"
```

Mypy doesn't output an outright error, but we can see that it fails to infer `inner.value1`
```
note: Revealed type is "builtins.bool"
note: Revealed type is "T`-1"
note: Revealed type is "builtins.str"
```

Constructing `inner` as `Outer[int].Inner(...)` doesn't make a difference for the record.

src/textual/widgets/_directory_tree.py
arvidfm: `_directory_content` is called with the result of `get_current_worker()` which returns (and can only return) a `Worker[Any]`, so this would cause an error in `_load_directory`

src/textual/_worker_manager.py
arvidfm: Ah yes, I missed that `WorkType` was a generic type itself, I'll add the missing type parameter to it and update the signature

davep:
@rodrigogiraoserrao I had a look at this a couple of days back but couldn't see any of the errors mentioned, so made a mental note to return to this after I've had another once-over of my Emacs/eglot/pyright setup to see if there's some extra level of checking I'm missing out on (generally I see all the issues I'd expect with assorted Python code, and a failing mypy run generally tends to coincide with similar complaints from pyright).

All of which is to say: I didn't see the errors in the first place.

arvidfm:
Sorry, to clarify, the errors are from my own project - Pyright is complaining about the fact that I have code that calls third-party functions that are missing type parameters. Also, I'm not sure `reportUnknown*Type` are on by default, but it should be trivial to reproduce by writing any external code that calls one of these functions and then type checking with Pyright in strict mode.

src/textual/worker.py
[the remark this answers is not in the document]
rodrigogiraoserrao: Thanks. My typing game isn't top notch yet :P

arvidfm:
I went ahead and fixed up the `WorkType` type parameters mentioned above.

I also decided to improve the typing of `textual.work`, as a lot of type information was getting lost with the current type hints. In particular, the following errors are now caught at typecheck time:
* Decorating a module-level function with `@work`
* Decorating a method in a non-`DOMNode` class with `@work`
* Decorating a non-asynchronous method with `@work(thread=False)`
* Passing in the wrong number of arguments to workers
* Passing in the wrong argument types to workers
* Assigning the result of a worker to a variable of the wrong type

However, there is a caveat that I couldn't manage to find a solution for. The following is inherently ambiguous:
```py
class MyClass(Widget):
    @work(thread=True)
    async def my_worker() -> int:
        return 0
```
Since both asynchronous and synchronous functions are allowed when `thread=True`, decorating a coroutine like this can result in the following type inferences for `my_worker`, both technically valid:
* `Callable[[], Coroutine[None, None, ReturnType]]` where `ReturnType` is inferred as `int`
* `Callable[[], ReturnType]` where `ReturnType` is inferred as `Coroutine[None, None, int]`

In order to handle this ambiguity I ended up adding a new `is_async` argument to `work()` which (thanks to a new `@overload`) instructs the type checker to use the first inference above. My assumption was that `thread=True` is mainly used for synchronous functions, which is why I made that the default.

Still, I think the added type safety is worth it, as demonstrated in the following gist, showing the issues that Mypy and Pyright failed to report previously, and how the inferred types of the worker factories were not very informative: https://gist.github.com/arvidfm/a57eb3f451ab76c26b881c3dd515f3bc. Besides, if the user can't be bothered about type checking they don't need to specify the `is_async` argument anyway, it'll work either way. (Also, it seems Pyright is actually able to infer it correctly even without the argument? Not sure if they've got some special heuristics for that.)

src/textual/widgets/_directory_tree.py
rodrigogiraoserrao: Of course I might be misunderstanding something, but what I'm suggesting is that we narrow the type ourselves, which should work.

E.g., [the code below typechecks](https://mypy-play.net/?mypy=latest&python=3.12&gist=3f1ed4ae05b311ea040a72651cea8fb5):

```py
from typing import Any

def any_factory() -> Any:
    ...

integer: int = any_factory()
```

`get_current_worker` is being called inside a worker, so we know that the current worker is the function we're in, so we know its type, right?
That was my reasoning.

arvidfm: Sorry for not getting back to you earlier, it's been a hectic week! But yes, you're right. I have a bad habit of thinking of `Any` as the base type, in which case this would have not worked due to invariance, but it's actually an opt-out of the typing system. Personally I always avoid implicit casts from `Any`, since it means the cast goes completely unchecked and it makes it easy to introduce errors while still being under a false sense of security.

Incidentally, this reminded me of something that I'd had in the back of my mind for a while as something I've been wanting to try in practice, namely this note on `Any` vs `object` from [this page on typing best practices](https://typing.readthedocs.io/en/latest/source/best_practices.html#ergonomic-practices):

> Generally, use Any when a type cannot be expressed appropriately with the current type system or using the correct type is unergonomic.
>
> If a function accepts every possible object as an argument, for example because it’s only passed to str(), use object instead of Any as type annotation. Similarly, if the return value of a callback is ignored, annotate it with object

I actually did have a go this week to see if I could eliminate `Any` in favour of `object` in my own codebase for better typing safety, and while I wasn't able to remove all uses due to limitations in Python's typing system (notably the [lack of wildcards](https://github.com/python/typing/issues/912)), I did have decent success and I'm definitely more confident in the type checking now.

If that's something you would be interested in I'd be happy to update this PR to use `Worker[object]` instead of `Worker[Any]` where applicable, which would provide better type safety for users of the code and make clear where explicit casts need to be made. It would mean users having to add an explicit cast if they want to get the result of `get_current_worker()` as a `Worker[int]` or whatever, but I don't think that's an issue - what would you even use the result type for from the worker itself before the result has been returned? `is_cancelled` can be checked even if you don't have the specific type.

Anyway! To get back to the point of your comment, there are three options I see here, in descending order of personal preference:

* Keep the argument as `Worker[Any]` (or better, `Worker[object]`)
    * We're only using the worker to access `is_cancelled`, and I don't think there's any real reason to add more specialisation than we actually need for the function - less casting means more type safety
* Have `get_current_worker` return `Worker[object]` but let `_directory_content` take a `Worker[Path]`, with an explicit cast to silence the type checker
    * Would at least make it clear that there is casting going on, and it makes it easy to do a search for `cast` later if one wanted to eliminate casting for the sake of stronger typing guarantees in the future
* Stick to `Worker[Any]` for `get_current_worker` and rely on implicit casting to `Worker[Path]`
    * Not a fan personally! Hard to spot and easy for issues to slip in. Only real way of tracking this down in the future is to run Mypy with `--disallow-any-expr` which is impossibly noisy

Let me know your thoughts!

rodrigogiraoserrao: Thanks for your comments!
I wasn't aware of the best practice of using `object` (instead of `Any`) in the contexts listed but it does make sense.
I _think_ we'd better leave it as `Worker[Any]` to be in line with the remainder of the code base.

I'll muster the courage to review this PR and then I'll rope someone else in to also take a look!

willmcgugan:
I'm afraid we can't accept this. Too many changes makes it hard to understand what the intent was. Suggest breaking it up in to smaller updates, ideally with a discussion beforehand so we can better understand the problem(s).

PR #3940, opened 2024-01-01
calculate message namespace from __qualname__ when not specified

### LINKED ISSUE #3939: Assigning a message type to a class variable breaks message handlers
If you assign a message type to a class variable belonging to a class that derives from MessagePump, non-decorated message handlers stop working because the message type's `handler_name` incorrectly gets reassigned to use the name of the class containing the class variable as the namespace.

I ran into this when writing a general class for input prompts/modals, where I was using class variables to specify what event to treat as the "submit" event.

Simple example, where `on_button_pressed` won't be run unless `EVENT_TYPE` is removed:
```py


class MyApp(App):
    # prevents on_button_pressed from running
    EVENT_TYPE = Button.Pressed

        yield Button("Click me")

    def on_button_pressed(self) -> None:
        self.query_one(Button).label = "Clicked!"


if __name__ == '__main__':
    MyApp().run()
    # prints "on_my_app_pressed"!
    print(Button.Pressed.handler_name)
```

arvidfm:
Fixes #3939

Use `__qualname__` for the message namespace since it already contains the outer class, making the `handler_name` processing in `_MessagePumpMeta` unnecessary

### DIFF
--- a/src/textual/message.py
+++ b/src/textual/message.py
@@ -73,8 +73,13 @@ class Message:
             cls.no_dispatch = no_dispatch
         if namespace is not None:
             cls.namespace = namespace
-        name = camel_to_snake(cls.__name__)
-        cls.handler_name = f"on_{namespace}_{name}" if namespace else f"on_{name}"
+            name = f"{namespace}_{camel_to_snake(cls.__name__)}"
+        else:
+            # a class defined inside of a function will have a qualified name like func.<locals>.Class,
+            # so make sure we only use the actual class name(s)
+            qualname = cls.__qualname__.split("<locals>.")[-1]
+            name = camel_to_snake(qualname).replace(".", "_")
+        cls.handler_name = f"on_{name}"
 
     @property
     def control(self) -> Widget | None:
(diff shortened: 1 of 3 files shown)


src/textual/message.py
```diff
@@ -73,8 +73,16 @@ def __init_subclass__(
             cls.no_dispatch = no_dispatch
         if namespace is not None:
             cls.namespace = namespace
-        name = camel_to_snake(cls.__name__)
-        cls.handler_name = f"on_{namespace}_{name}" if namespace else f"on_{name}"
+            name = f"{namespace}_{camel_to_snake(cls.__name__)}"
+        else:
+            # a class defined inside of a function will have a qualified name like func.<locals>.Class,
+            # so make sure we only use the actual class name(s)
+            qualname = cls.__qualname__.rsplit("<locals>.", 1)[-1]
```
willmcgugan, line 80: This seem brittle. Is it guaranteed that CPython and all other Python's will have a qualname formatted like this?

arvidfm: It's explicitly defined in the PEP that introduced `__qualname__`: https://peps.python.org/pep-3155/#proposal

darrenburns:
Sorry for the delay on this - this is quite the gotcha!

I've confirmed it fixes the issue and tested it in a complex app. I also updated the CHANGELOG. If the tests pass I think we're good to merge.

darrenburns (approved):
LGTM

PR #3944, opened 2024-01-02
Retain `Tree.root`'s expanded state when performing a `Tree.clear`

### LINKED ISSUE #3557: Tree.clear doesn't respect root expanded state
This is quite pedantic, but caused me trouble which will make more sense in the context of an upcoming discussion r.e. 'refreshing' a Tree widget and preserving which nodes are expanded vs collapsed.

---

TLDR:

When you create a `Tree` it creates a root node with:
https://github.com/Textualize/textual/blob/3b170b074802fab825a698ca66be2196871700cf/src/textual/widgets/_tree.py#L605

which has `expand` as `False` by default:
https://github.com/Textualize/textual/blob/3b170b074802fab825a698ca66be2196871700cf/src/textual/widgets/_tree.py#L641-L646

But when you `Tree.clear` it sets `expand` to `True`:

https://github.com/Textualize/textual/blob/3b170b074802fab825a698ca66be2196871700cf/src/textual/widgets/_tree.py#L710-L716

---

This is a problem for me because I'd _theorized_ that I could capture the `is_expanded` state of all nodes, then `clear()` and recreate it, then `expand()` the relevant nodes back to restore the previous 'expansion' state...

... but this doesn't _quite_ work because the `clear()` will `expand()` the root node regardless of whether it was expanded before.

Edge case, but I _presume_ was accidental?

davep:
Fixes #3557.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - `Widget.move_child` would break if `before`/`after` is set to the index of the widget in `child` https://github.com/Textualize/textual/issues/1743
 - Fixed auto width text not processing markup https://github.com/Textualize/textual/issues/3918
+- Fixed `Tree.clear` not retaining the root's expanded state https://github.com/Textualize/textual/issues/3557
 
 ### Changed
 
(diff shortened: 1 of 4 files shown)


rodrigogiraoserrao (approved):
LGTM 🐍 🚀

PR #3946, opened 2024-01-02
nested

willmcgugan:
- Implements nested CSS
- Documents nested CSS in the guide
- Makes CSS errors a little friendlier by associating a message rather than a list of tokens types

Screenshot 2024-01-03 at 14 10 56

### DIFF
--- a/docs/guide/CSS.md
+++ b/docs/guide/CSS.md
@@ -523,7 +523,7 @@ The following produces identical results to the previous example, but adds nesti
     ```{.textual path="docs/examples/guide/css/nesting02.py"}
     ```
 
-In the first example we had a rule set that began with the selector `#questions .button`, which would match any widget with a class called "button" that are inside a container with id `questions`.
+In the first example we had a rule set that began with the selector `#questions .button`, which would match any widget with a class called "button" that is inside a container with id `questions`.
 
 In the second example, the button rule selector is simply `.button`, but it is *within* the rule set with selector `#questions`.
 The nesting means that the button rule set will inherit the selector from the outer rule set, so it is equivalent to `#questions .button`.
--- a/docs/guide/CSS.md
+++ b/docs/guide/CSS.md
@@ -502,7 +502,8 @@ The CSS is quite straightforward; there is one rule for the container, one for a
 However it is easy to imagine this stylesheet growing more rules as we add features.
 
 Nesting allows us to group rule sets which have common selectors.
-In the example above, the rules all start with `#questions`, which is a good indication that we can nest the rules.
+In the example above, the rules all start with `#questions`.
+When we see a common prefix on the selectors, this is a good indication that we can use nesting.
 
 The following produces identical results to the previous example, but adds nesting of the rules.
 
--- a/docs/guide/CSS.md
+++ b/docs/guide/CSS.md
@@ -477,13 +477,13 @@ If you use `initial` within default css, it will treat the rule as completely un
 CSS rule sets may be *nested*, i.e. they can contain other rule sets.
 When a rule set occurs within an existing rule set, it inherits the selector from the enclosing rule set.
 
-Let's put this in to practical terms.
+Let's put this into practical terms.
 The following example will display two boxes containing the text "Yes" and "No" respectively.
 These could eventually form the basis for buttons, but for this demonstration we are only interested in the CSS.
 
 === "nesting01.tcss (no nesting)"
 
-    ```sass
+    ```css
     --8<-- "docs/examples/guide/css/nesting01.tcss"
     ```
 
@@ -498,7 +498,7 @@ These could eventually form the basis for buttons, but for this demonstration we
     ```{.textual path="docs/examples/guide/css/nesting01.py"}
     ```
 
-The CSS is quite straight forward; there is a rule for the container, all buttons, and a rule each for the yes and no buttons.
+The CSS is quite straightforward; there is one rule for the container, one for all buttons, and one rule for each of the buttons.
 However it is easy to imagine this stylesheet growing more rules as we add features.
 
 Nesting allows us to group rule sets which have common selectors.
@@ -534,7 +534,7 @@ The two remaining rules are nested within the button rule, which means they will
 
 You may have noticed that the rules for the button styles contain a syntax we haven't seen before.
 The rule for the Yes button is `&.affirmative`.
-The ampersand (`&`) is known as the *nesting selector* and it tells Textual that the selector should be combined with the selector from the outer rule-set.
+The ampersand (`&`) is known as the *nesting selector* and it tells Textual that the selector should be combined with the selector from the outer rule set.
 
 So `&.affirmative` in the example above, produces the equivalent of `#questions .button.affirmative` which selects a widget with both the `button` and `affirmative` classes.
 Without `&` it would be equivalent to `#questions .button .affirmative` (note the additional space) which would only match a widget with class `affirmative` inside a container with class `button`.
@@ -544,7 +544,7 @@ For reference, lets see those two CSS files side-by-side:
 
 === "nesting01.tcss"
 
-    ```sass
+    ```css
     --8<-- "docs/examples/guide/css/nesting01.tcss"
     ```
 
@@ -554,6 +554,6 @@ For reference, lets see those two CSS files side-by-side:
     --8<-- "docs/examples/guide/css/nesting02.tcss"
     ```
 
-### Why used nesting?
+### Why use nesting?
 
-There is no requirement to use nested CSS, but it can help to group related rule-sets together (which makes it easier to edit). Nested CSS can also help you avoid some repetition in your selectors, i.e. in the nested CSS we only need to type `#questions` once, rather than four times in the non-nested CSS.
+There is no requirement to use nested CSS, but it can help to group related rule sets together (which makes it easier to edit). Nested CSS can also help you avoid some repetition in your selectors, i.e. in the nested CSS we only need to type `#questions` once, rather than four times in the non-nested CSS.
--- a/docs/guide/CSS.md
+++ b/docs/guide/CSS.md
@@ -468,3 +468,92 @@ For instance, if we have a widget with a (CSS) class called `dialog`, we could r
 
 Note that `initial` will set the value back to the value defined in any [default css](./widgets.md#default-css).
 If you use `initial` within default css, it will treat the rule as completely unstyled.
+
+
+## Nesting CSS
+
+!!! tip "Added in version 0.47.0"
+
+CSS rule sets may be *nested*, i.e. they can contain other rule sets.
+When a rule set occurs within an existing rule set, it inherits the selector from the enclosing rule set.
+
+Let's put this in to practical terms.
+The following example will display two boxes containing the text "Yes" and "No" respectively.
+These could eventually form the basis for buttons, but for this demonstration we are only interested in the CSS. 
+
+=== "nesting01.tcss"
+
+    ```sass
+    --8<-- "docs/examples/guide/css/nesting01.tcss"
+    ```
+
+=== "nesting01.py"
+
+    ```python
+    --8<-- "docs/examples/guide/css/nesting01.py"
+    ```
+
+=== "Output"
+
+    ```{.textual path="docs/examples/guide/css/nesting01.py"}
+    ```
+
+The CSS is quite straight forward; there is a rule for the container, all buttons, and a rule each for the yes and no buttons.
+However it is easy to imagine this stylesheet growing more rules as we add features.
+
+Nesting allows us to group rule sets which have common selectors.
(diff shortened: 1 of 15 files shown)


docs/guide/CSS.md
```diff
@@ -468,3 +468,92 @@ For instance, if we have a widget with a (CSS) class called `dialog`, we could r
...
+    ```python
+    --8<-- "docs/examples/guide/css/nesting01.py"
+    ```
+
+=== "Output"
+
+    ```{.textual path="docs/examples/guide/css/nesting01.py"}
+    ```
+
+The CSS is quite straight forward; there is a rule for the container, all buttons, and a rule each for the yes and no buttons.
```
davep, line 501:
```suggestion
The CSS is quite straightforward; there is a rule for the container, all buttons, and a rule each for the yes and no buttons.
```

docs/guide/CSS.md
```diff
@@ -468,3 +468,92 @@ For instance, if we have a widget with a (CSS) class called `dialog`, we could r
...
+In the second example, the button rule selector is simply `.button`, but it is *within* the rule set with selector `#questions`.
+The nesting means that the button rule set will inherit the selector from the outer rule set, so it is equivalent to `#questions .button`.
+
+### Nesting selector
+
+The two remaining rules are nested within the button rule, which means they will inherit their selectors from the button rule set *and* the outer `#questions` rule set.
+
+You may have noticed that the rules for the button styles contain a syntax we haven't seen before.
+The rule for the Yes button is `&.affirmative`.
+The ampersand (`&`) is known as the *nesting selector* and it tells Textual that the selector should be combined with the selector from the outer rule-set.
```
davep, line 537:
```suggestion
The ampersand (`&`) is known as the *nesting selector* and it tells Textual that the selector should be combined with the selector from the outer rule set.
```

`rule set` seems to mostly be used everywhere else.

docs/guide/CSS.md
```diff
@@ -468,3 +468,92 @@ For instance, if we have a widget with a (CSS) class called `dialog`, we could r
...
+
+=== "nesting02.tcss"
+
+    ```sass
+    --8<-- "docs/examples/guide/css/nesting02.tcss"
+    ```
+
+### Why used nesting?
+
+There is no requirement to use nested CSS, but it can help to group related rule-sets together (which makes it easier to edit). Nested CSS can also help you avoid some repetition in your selectors, i.e. in the nested CSS we only need to type `#questions` once, rather than four times in the non-nested CSS.
```
davep, line 559:
```suggestion
There is no requirement to use nested CSS, but it can help to group related rule sets together (which makes it easier to edit). Nested CSS can also help you avoid some repetition in your selectors, i.e. in the nested CSS we only need to type `#questions` once, rather than four times in the non-nested CSS.
```

davep:
Before this change this would (correctly) throw an error:

```

class NestTestApp(App[None]):

    CSS = """
    &Label {
        color: red;
    }
    """

        yield Label("Hello, World!")

if __name__ == "__main__":
    NestTestApp().run()
```

now it silently does nothing. It's a nonsense rule; should it perhaps still throw an error?

davep:
Given something like this:

```python

class NestTestApp(App[None]):

    CSS = """
    Label {
        &#foo {
            color: red;
        }
    """

        yield Label("Hello, World!", id="foo", classes="bar")

if __name__ == "__main__":
    NestTestApp().run()
```

(where there's a missing `}` at the end of the CSS) this results in a `StylesheetError: failed to parse css; generator raised StopIteration` error; whereas normally a missing closing brace would tend to result in a `Unexpected end of file` error. For example this still does:

```python

class NestTestApp(App[None]):

    CSS = """
    Label {
        color: red;
    """

        yield Label("Hello, World!", id="foo", classes="bar")

if __name__ == "__main__":
    NestTestApp().run()
```

docs/guide/CSS.md
rodrigogiraoserrao: I'd go even further.
As a non-native English speaker, from the sentence alone I need to read it 3 or 4 times to be sure of how many rules in total there are.
Something slightly more verbose may be exponentially clearer:

```suggestion
The CSS is quite straightforward; there is one rule for the container, one for all buttons, and one rule for each of the buttons.
```

docs/guide/CSS.md
```diff
@@ -468,3 +468,92 @@ For instance, if we have a widget with a (CSS) class called `dialog`, we could r
...
+
+
+## Nesting CSS
+
+!!! tip "Added in version 0.47.0"
+
+CSS rule sets may be *nested*, i.e. they can contain other rule sets.
+When a rule set occurs within an existing rule set, it inherits the selector from the enclosing rule set.
+
+Let's put this in to practical terms.
```
rodrigogiraoserrao, line 480:
```suggestion
Let's put this into practical terms.
```

docs/guide/CSS.md
```diff
@@ -468,3 +468,92 @@ For instance, if we have a widget with a (CSS) class called `dialog`, we could r
...
+=== "Output"
+
+    ```{.textual path="docs/examples/guide/css/nesting01.py"}
+    ```
+
+The CSS is quite straight forward; there is a rule for the container, all buttons, and a rule each for the yes and no buttons.
+However it is easy to imagine this stylesheet growing more rules as we add features.
+
+Nesting allows us to group rule sets which have common selectors.
+In the example above, the rules all start with `#questions`, which is a good indication that we can nest the rules.
```
rodrigogiraoserrao, line 505: This sets the expectation for nesting of the `#questions`, but not the `#questions .button`.
Would it be worth adding one sentence to say something like

```suggestion
In the example above, the rules all start with `#questions`, which is a good indication that we can nest the rules.
We can also see that the last three rules all start with `#questions .button`, which means we can nest the rules even further.
```

docs/guide/CSS.md
```diff
@@ -468,3 +468,92 @@ For instance, if we have a widget with a (CSS) class called `dialog`, we could r
...
+    --8<-- "docs/examples/guide/css/nesting01.tcss"
+    ```
+
+=== "nesting02.tcss"
+
+    ```sass
+    --8<-- "docs/examples/guide/css/nesting02.tcss"
+    ```
+
+### Why used nesting?
```
rodrigogiraoserrao, line 557:
```suggestion
### Why use nesting?
```

docs/guide/CSS.md
rodrigogiraoserrao, line 1: One thing that isn't clear _to me_ from reading the updated guide is how we'd nest something like

```css
A {
    ...
}

A > B {
    ...
}

A > B > C {
    ...
}
```

Will the nested selectors be `> B` and `> C` or `& > B` and `& > C`?

docs/guide/CSS.md
```diff
@@ -468,3 +468,92 @@ For instance, if we have a widget with a (CSS) class called `dialog`, we could r
...
+CSS rule sets may be *nested*, i.e. they can contain other rule sets.
+When a rule set occurs within an existing rule set, it inherits the selector from the enclosing rule set.
+
+Let's put this in to practical terms.
+The following example will display two boxes containing the text "Yes" and "No" respectively.
+These could eventually form the basis for buttons, but for this demonstration we are only interested in the CSS.
+
+=== "nesting01.tcss (no nesting)"
+
+    ```sass
```
rodrigogiraoserrao, line 486:
```suggestion
    ```css
```

After #3902 we changed to `css` in Textual CSS snippets.

docs/guide/CSS.md
```diff
@@ -468,3 +468,92 @@ For instance, if we have a widget with a (CSS) class called `dialog`, we could r
...
+However it is easy to imagine this stylesheet growing more rules as we add features.
+
+Nesting allows us to group rule sets which have common selectors.
+In the example above, the rules all start with `#questions`, which is a good indication that we can nest the rules.
+
+The following produces identical results to the previous example, but adds nesting of the rules.
+
+=== "nesting02.tcss (with nesting)"
+
+    ```sass
```
rodrigogiraoserrao, line 511:
```suggestion
    ```css hl_lines="7 15 20"
```

After #3902 we changed to `css` in Textual CSS snippets.

docs/guide/CSS.md
```diff
@@ -468,3 +468,92 @@ For instance, if we have a widget with a (CSS) class called `dialog`, we could r
...
+
+So `&.affirmative` in the example above, produces the equivalent of `#questions .button.affirmative` which selects a widget with both the `button` and `affirmative` classes.
+Without `&` it would be equivalent to `#questions .button .affirmative` (note the additional space) which would only match a widget with class `affirmative` inside a container with class `button`.
+
+
+For reference, lets see those two CSS files side-by-side:
+
+=== "nesting01.tcss"
+
+    ```sass
```
rodrigogiraoserrao, line 547:
```suggestion
    ```css
```

After #3902 we changed to `css` in Textual CSS snippets.

src/textual/css/parse.py
```diff
@@ -29,14 +30,31 @@
...
+    specificity1: Specificity3, specificity2: Specificity3
+) -> Specificity3:
+    """Add specificity together.
+
+    Args:
+        specificity1: Specificity triple.
+        specificity2: Specificity triple.
+
+    Returns:
+        Combined specificity
```
rodrigogiraoserrao, line 47:
```suggestion
        Combined specificity.
```

src/textual/css/parse.py
```diff
@@ -29,14 +30,31 @@
...
     "selector_universal": (SelectorType.UNIVERSAL, (0, 0, 0)),
     "selector_start_universal": (SelectorType.UNIVERSAL, (0, 0, 0)),
+    "nested": (SelectorType.NESTED, (0, 0, 0)),
 }
 
 
+def _add_specificity(
+    specificity1: Specificity3, specificity2: Specificity3
+) -> Specificity3:
+    """Add specificity together.
```
rodrigogiraoserrao, line 40:
```suggestion
    """Add specificities together.
```

src/textual/css/parse.py
```diff
@@ -143,14 +164,82 @@ def parse_rule_set(
...
+
+                Args:
+                    selectors1: List of selectors.
+                    selectors2: Second list of selectors.
+
+                Returns:
+                    Combined selectors.
+                """
+                if selectors2 and selectors2[0].type == SelectorType.NESTED:
+                    final_selector = selectors1[-1]
```
rodrigogiraoserrao, line 206: Don't we need to protect ourselves against `selectors1` being empty in a degenerate case with malformed CSS?

tests/css/test_nested_css.py
rodrigogiraoserrao, line 1: We're missing testing nesting with `>`.

rodrigogiraoserrao (changes requested):

docs/guide/CSS.md
willmcgugan: Why highlight those lines? Most of the lines have changed.

docs/guide/CSS.md
willmcgugan: `& > B` and `& > V`

docs/guide/CSS.md
willmcgugan: Changed the wording to suggest it is a general principle, and not just for this one thing

src/textual/css/parse.py
willmcgugan: I don't believe that can occur. The parser will ensure that there is at least one selector.

docs/guide/CSS.md
rodrigogiraoserrao: Those show the nested selectors, which are the most significant changes (most lines changed in indentation, not content).
This was just a suggestion!

docs/guide/CSS.md
rodrigogiraoserrao: Worth mentioning in the guide or not really?

docs/guide/CSS.md
willmcgugan: I'd like to stick to the convention of highlighting changed lines. Since they all changed though, it seems redundant here.

src/textual/css/parse.py
```diff
@@ -143,14 +164,82 @@ def parse_rule_set(
...
+            )
+
+            def combine_selectors(
+                selectors1: list[Selector], selectors2: list[Selector]
+            ) -> list[Selector]:
+                """Combine lists of selectors together, processing any nesting.
+
+                Args:
+                    selectors1: List of selectors.
+                    selectors2: Second list of selectors.
```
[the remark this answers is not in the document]
willmcgugan: Possibly. But from the POV of this method they don't necessarily have to be inner or outer. Only the order is significant, hence 1 and 2.

docs/guide/CSS.md
willmcgugan: Not sure if it is required, There's nothing special about the > operator.

davep (approved):
Throwing in my approval now that the couple of issues I noticed have been hit.

Issue #3961, opened 2024-01-05
Does nested CSS have lower specificity?


davep:
Consider this code:

```python
from textual.app import App, ComposeResult
from textual.widgets import Static

class BaseTester(Static, can_focus=True):
    DEFAULT_CSS = """
    BaseTester:focus {
        background: yellow;
        border: thick magenta;
    }
    """

class NonNestedCSS(BaseTester):
    DEFAULT_CSS = """
    NonNestedCSS {
        width: 1fr;
        height: 1fr;
        background: red 10%;
        border: blank;
    }

    NonNestedCSS:focus {
        background: red 20%;
        border: round red;
    }
    """

class NestedCSS(BaseTester):
    DEFAULT_CSS = """
    NestedCSS {
        width: 1fr;
        height: 1fr;
        background: green 10%;
        border: blank;

        &:focus {
            background: green 20%;
            border: round green;
        }
    }
    """

class NestedPseudoClassesApp(App[None]):

    CSS = """
    Screen {
        layout: horizontal;
    }
    """

    def compose(self) -> ComposeResult:
        yield NonNestedCSS("This isn't using nested CSS")
        yield NestedCSS("This is using nested CSS")

if __name__ == "__main__":
    NestedPseudoClassesApp().run()
```

There is a base class with a `:focus` pseudo class that sets a garish border/background combination. There are then two subclassed widgets. One (`NonNestedCSS`) overrides the garish `:focus` with something nicer using a non-nested CSS approach. There is another (`NestedCSS`) which overrides the garish `:focus` using nested CSS. When the above code is run with 0.47.0 you'll see that `NonNestedCSS` overrides the `:focus` rules just fine; on the other hand the `:focus` rules for `NestedCSS` seem to be ignored.

It feels like there should be no difference here. Initially flagging this up as something to clarify, but perhaps also a bug.

Tagging @willmcgugan for triage as it relates to a new feature.

PR #3963, opened 2024-01-05
fix for specificity

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/3961

### DIFF
--- a/src/textual/css/model.py
+++ b/src/textual/css/model.py
@@ -20,19 +20,27 @@ if TYPE_CHECKING:
 class SelectorType(Enum):
     """Type of selector."""
 
-    UNIVERSAL = 1  # i.e. *
-    TYPE = 2  # Type, e.g Widget Label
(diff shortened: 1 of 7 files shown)


src/textual/css/model.py
```diff
@@ -12,21 +12,27 @@
...
 
 class SelectorType(Enum):
-    UNIVERSAL = 1
-    TYPE = 2
-    CLASS = 3
-    ID = 4
-    NESTED = 5
+    """Type of selector."""
+
+    UNIVERSAL = 1  # i.e. *
```
davep, line 23: Small suggestion rather than anything else: perhaps these comments should be docstrings for each of the values; then anyone working *on* Textual will get the hint-text benefits from their LSP?

davep (approved):
Confirmed as solving what I ran into this morning. 👍

darrenburns (approved):

PR #3970, opened 2024-01-06
optimization

willmcgugan:
Store pseudo classes in a set rather than a list. This allows for some optimization. Not a big win, but its a step towards a larger optimization.


CHANGELOG.md
```diff
@@ -5,6 +5,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## Unreleased
+
+### Changed
+
+- Breaking change: DomNode.has_pseudo_class now accepts a single name only https://github.com/Textualize/textual/pull/3970
```
rodrigogiraoserrao, line 12:
```suggestion
- Breaking change: `DOMNode.has_pseudo_class` now accepts a single name only https://github.com/Textualize/textual/pull/3970
```

CHANGELOG.md
```diff
@@ -5,6 +5,16 @@ All notable changes to this project will be documented in this file.
...
+
+### Added
+
+- Added DomNode.has_psuedo_classes https://github.com/Textualize/textual/pull/3970
```
rodrigogiraoserrao, line 16:
```suggestion
- Added `DOMNode.has_pseudo_classes` https://github.com/Textualize/textual/pull/3970
```

src/textual/css/model.py
```diff
@@ -43,6 +46,67 @@ class CombinatorType(Enum):
...
 
+def _check_universal(name: str, node: DOMNode) -> bool:
+    """Check node matches universal selector.
+
+    Args:
+        name: Selector name.
+        node: A DOM node.
+
+    Returns:
+        `True` if the selector matches.
```
rodrigogiraoserrao, line 57:
```suggestion
        `True`.
```

src/textual/css/stylesheet.py
```diff
@@ -183,6 +189,11 @@ def rules_map(self) -> dict[str, list[RuleSet]]:
 
     @property
     def css(self) -> str:
+        """The equivalent TCSS for this stylesheet.
+
+        Note that this may not produce the same content as the file(s) used to generate the stylesheet.
+
```
rodrigogiraoserrao, line 195:
```suggestion
```

src/textual/dom.py
```diff
@@ -1276,17 +1275,27 @@ def toggle_class(self, *class_names: str) -> Self:
...
         return self
 
-    def has_pseudo_class(self, *class_names: str) -> bool:
-        """Check for pseudo classes (such as hover, focus etc)
+    def has_pseudo_class(self, class_name: str) -> bool:
+        """Check the node has the given pseudo class.
 
         Args:
-            *class_names: The pseudo classes to check for.
+            class_name: The pseudo classes to check for.
```
rodrigogiraoserrao, line 1282:
```suggestion
            class_name: The pseudo class to check for.
```

src/textual/dom.py
```diff
@@ -1276,17 +1275,27 @@ def toggle_class(self, *class_names: str) -> Self:
...
 
         Returns:
             `True` if the DOM node has those pseudo classes, `False` if not.
```
rodrigogiraoserrao, line 1285:
```suggestion
            `True` if the DOM node has the pseudo class, `False` if not.
```

src/textual/dom.py
```diff
@@ -1276,17 +1275,27 @@ def toggle_class(self, *class_names: str) -> Self:
...
+    def has_pseudo_classes(self, class_names: set[str]) -> bool:
+        """Check the node has all the given pseudo classes.
+
+        Args:
+            class_names: Set of class names to check for.
+
+        Returns:
+            `True` if all pseudo class names are present.
+        """
+        return class_names.issubset(self.get_pseudo_classes())
```
rodrigogiraoserrao, line 1298: Same as `class_names <= self.get_pseudo_classes()`.
I think it's more elegant, but I'm not going to push too hard for it.

src/textual/dom.py
```diff
@@ -1276,17 +1275,27 @@ def toggle_class(self, *class_names: str) -> Self:
         self._update_styles()
         return self
 
-    def has_pseudo_class(self, *class_names: str) -> bool:
-        """Check for pseudo classes (such as hover, focus etc)
+    def has_pseudo_class(self, class_name: str) -> bool:
```
rodrigogiraoserrao, line 1278: In light of the optimisation intent, isn't it faster to check specifically for the given pseudo class instead of iterating over the generator of all pseudo classes?

rodrigogiraoserrao (approved):
I left some comments, docs tweaks, and a suggestion.
None of them are blocking, so I'll approve right away.

src/textual/dom.py
willmcgugan: The result of `get_pseudo_classes` is not a set, so the operator won't work here.

src/textual/dom.py
willmcgugan: At the moment, the iterator is the only way we have of exposing the pseudo classes. Its only actually tests which use `has_pseudo_class` currently.

src/textual/dom.py
rodrigogiraoserrao: 🤦  I left this comment before realising it was a generator.

src/textual/dom.py
rodrigogiraoserrao: I don't the understand the second sentence _at all_... 🙃

willmcgugan: Search the code for "has_pseudo_class". It is only used by tests.

rodrigogiraoserrao: Understood.

Issue #3728, opened 2023-11-22
HowTo package Textual apps


willmcgugan:
We could use a HowTo which covers how to package a Textual app.

It shouldn't be much more than a description of how to package any Python app. The only difference (I think) is that the dev has to be conscious to include the `*.tcss` files.

We should make a repo with a very basic app (suggest the calculator example), following these guidelines: https://packaging.python.org/en/latest/tutorials/packaging-projects/

We should also mention, and link, to other packaging tools.

As with all the docs, please keep the writing tight. Omit needless words.

rodrigogiraoserrao:
Why not poetry?

willmcgugan:
Poetry is a bit polarizing. Want to stick to whatever the standard is these days.

TomJGooding:
> Want to stick to whatever the standard is these days.

Thankfully there should be one obvious way to do it, maybe ask a Dutchman?

ndegroot:
We are not in Python Packaging heaven yet. 
It's too early (alpha stage) but I (Dutchman) am using [Pixi](https://github.com/prefix-dev/pixi). It can install Python as well, can create a virtual env, but you still need 'something' to get the files at the destination. Pixi can handle that too and creates a package for an anaconda repo. I didn't use that,  Git in combination with a repo works better for my situation. Just a suggestion.

PR #3977, opened 2024-01-08
WiP: HOWTO package a Textual application

davep:
*Ignore the `(REMOVE BEFORE FLIGHT!)` text towards the end of the document, this is just for editor configuration while I edit the document; it'll be removed once ready for merge*

Provides a HOWTO document that highlights the main areas of concern when thinking about packaging a Textual application for deployment.

Because there are quite a lot of application management tools that provide packaging and publishing facilities we've decided to concentrate on relating to two of the more popular tools that people are likely to encounter, while allowing for the option to expand the document if we see fit. The main aim here however is to try and provide a fairly generic list of things to keep in mind and make sure you address.

Overall this comes down to:

- Remember to declare Textual as a dependency for your application.
- Make sure you've written an entry point for your application.
- Think of the command name the end user will run and declare that.
- Ensure that any TCSS files that your application needs are packaged too.

Two example repositories have been created to illustrate the the above:

- [`textual-calculator-hatch`](https://github.com/Textualize/textual-calculator-hatch)
- [`textual-calculator-poetry`](https://github.com/Textualize/textual-calculator-poetry)

Implements #3728.

### DIFF
--- a/docs/how-to/package-an-application.md
+++ b/docs/how-to/package-an-application.md
@@ -1,0 +1,78 @@
+# Package an application
+
+A common question on the [Textual Discord server](https://discord.gg/Enf6Z3qhVr) and [in the Textual discussion area](https://github.com/Textualize/textual/discussions) is how to package a Textual application for distribution. At its heart, this is a question of *"how do I package a Python application for distribution?"*; this a reasonably big subject that is covered by a number of tutorials on the web. A good place to start would be [in the Python documentation itself](https://packaging.python.org/en/latest/overview/).
+
+In this HOWTO we'll concentrate on the Textual-specific issues you need to keep in mind.
+
+## What packaging tool should I use?
+
+The choice of packaging tool will often come down to personal taste, and can be a polarising question. In this document we'll cover two currently-popular tools that help with this: [Hatch](https://hatch.pypa.io/latest/) and [Poetry](https://python-poetry.org/). There are a number of other choices but the issues to consider will be similar in all cases.
+
+We recommend first getting familiar with your choice of tool and working with its documentation to get your project set up for packing and distribution.
+
+## So what Textual-specifics do I need to worry about?
+
+While these are concerns for any packaged Python application (and so should be covered by the documentation of your chosen tool), there are two areas to consider with your Textual application that will result in a successful package, publish and install experience:
+
+### Application entry point and the runnable command
+
+We're going to turn the Textual [calculator example](https://github.com/Textualize/textual/blob/main/examples/calculator.py) into an application that can be packaged, deployed to [PyPi](https://pypi.org/), and installed by users using `pip` or `pipx`.
+
+The end result we want is that, after the user has installed the application, they can type `calculator` in the shell and the application will appear.
+
+#### Making the entry point
+
+Normally, somewhere in a Textual application, you'll have some code that looks like this:
+
+```python
+if __name__ == "__main__":
+    CalculatorApp().run()
+```
+
+We're going to make a small change to this to help us when it comes to telling the packaging tool how to run the application. The change is to turn the above into this:
+
+```python
+def run() -> None:
+    """Run the calculator application."""
+    CalculatorApp().run()
+
+if __name__ == "__main__":
+    run()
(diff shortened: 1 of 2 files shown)


rodrigogiraoserrao (approved):
Looks good to me BUT there are still some REMOVE BEFORE FLIGHT lines at the end of this file... 🤔

docs/how-to/package-an-application.md
```diff
@@ -0,0 +1,138 @@
...
+    ...
+    textual = "*"
+    ```
+
+!!! note
+
+    You may wish to pin the version of Textual; perhaps to a minimum version, or even to the exact version you are using at that time.
+    Opinions on best practice here do vary.
+
+### Application entry point and the runnable command
```
willmcgugan, line 49: I don't think we've defined "entry point".

How about this heading is "Making a command" or similar. And we define entry point in the sub-heading.

docs/how-to/package-an-application.md
```diff
@@ -0,0 +1,138 @@
...
+
+We're going to turn the Textual [calculator example](https://github.com/Textualize/textual/blob/main/examples/calculator.py) into an application that can be packaged, deployed to [PyPi](https://pypi.org/), and installed by users using [`pip`](https://pip.pypa.io/en/stable/) or [`pipx`](https://pipx.pypa.io/stable/).
+
+The end result we want is that, after the user has installed the package, they can type `calculator` in the shell and the application will appear.
+
+#### Making the entry point
```
willmcgugan, line 55: Lets define entry point here. Something like "An entry point is the function in your project that launches your app."

docs/how-to/package-an-application.md
```diff
@@ -0,0 +1,138 @@
...
+
+Normally, somewhere in a Textual application, you'll have some code that looks like this:
+
+```python
+if __name__ == "__main__":
+    CalculatorApp().run()
+```
+
+We're going to make a small change to this to help us when it comes to telling the packaging tool how to run the application.
```
willmcgugan, line 64: Bit wordy, and doesn't really say *what* the change is for.

Suggest "Let's modify this so we have a function that will become our entry point."

docs/how-to/package-an-application.md
```diff
@@ -0,0 +1,138 @@
...
+    run()
+```
+
+In other words:
+we've created a function called `run` that runs the calculator application;
+then we've changed the `__main__` test to run that function (we keep that because it lets us run our application with `python -m`; this can be useful during development).
+
+#### Declaring the runnable command
+
+Having done the above, we're all set to tell the packaging tool what to call the final command that the user will run, and what it should do when they run it.
```
willmcgugan, line 82: "done the above" -> "created the entry point".

docs/how-to/package-an-application.md
```diff
@@ -0,0 +1,138 @@
...
+For Hatch and Poetry it is a case of adding:
+
+```
+<application-name> = "<entry-point>"
```
willmcgugan, line 86: Generic placeholders may confuse noobs, as they might not understand the <> mean "substitute your info".

docs/how-to/package-an-application.md
```diff
@@ -0,0 +1,138 @@
...
+```
+
+to the correct section of `pyproject.toml`.
```
willmcgugan, line 89: But waht is the correct section?

docs/how-to/package-an-application.md
```diff
@@ -0,0 +1,138 @@
...
+In the case of our calculator example, the command we want is `calculator` and the entry point is the `run` function in the module `calculator.py` which is in the package `textual_calculator`;
+we specify this with `textual_calculator.calculator:run`.
```
willmcgugan, line 91: I think we should explain this syntax. i.e. that it has the import path, a colon, and the function name.

docs/how-to/package-an-application.md
davep: The next sentence in that paragraph, with the example that is under the paragraph, shows the correct section for each of the tools we're talking about.

PR #3979, opened 2024-01-08
Better tab relabelling

### LINKED ISSUE #3901: Ability to rename tabs (TabPane).
Currently this is not supported AFAIK. If one wants to change the title of a tab, they need to remove and add it again.
Fetching the `Tab` within `TabPane` and updating it doesn't fully work, for instance the underline does not adapt.

davep:
Implements a better method of changing the label of a `Tab`; fixes #3901.

This is a fairly small change, which turns `Tab.label` into a safe way of setting a new label for a `Tab`. So the question in the original issue would end up being answered with this:

```python
self.query_one(TabbedContent).get_tab("one").label = "Renamed!"
```

It also quietly supports going via `Tab.update` too but doesn't advertise it as *the* approach to take.

The PR also, of course, addresses the issue of the underline not redrawing correctly.

https://github.com/Textualize/textual/assets/28237/3c3a146e-81ef-44a3-87d9-8772982a2815


rodrigogiraoserrao (approved):
LGTM 🚀🐍

Issue #2867, opened 2023-06-30
Subscripted generics cannot be used in query selectors


tuchandra:
(I wasn't sure if a bug or feature request was more appropriate, sorry!)

## Description
The `Select` widget is generic over the type of the option values. 
- we can use `Select[SomeType]` in e.g., a widget's `compose` method
- we can't use `self.query_one(Select[SomeType])`, though; this crashes at runtime

This happens because the `isinstance` check that Textual does is incompatible with subscripted generics (it's the same reason that `isinstance(x, list[int])` doesn't work, which is detailed in [PEP 585](https://peps.python.org/pep-0585/#making-isinstance-obj-list-str-perform-a-runtime-type-check)) .

This might be a bug, because:
- the [API reference for `query_one`](https://textual.textualize.io/api/dom_node/#textual.dom.DOMNode.query_one) doesn't indicate that this won't work
- the code type checks, but fails at runtime, which indicates that the type annotation is too broad

This is probably a feature request, though, because:
- generics have a ton of sharp edges
- it's very clear why this is happening
- it's not clear how Textual can/should handle this

## Details
Here's a video, example, and diagnostics.

https://github.com/Textualize/textual/assets/11623486/540b6d56-1078-478d-8261-0b366744f4e5

<details><summary>Demo code and traceback</summary>

```python
from rich.markdown import Markdown
from textual import on
from textual.app import App, ComposeResult
from textual.containers import Horizontal
from textual.widgets import Select

class MyApp(App):
    def compose(self) -> ComposeResult:
        with Horizontal():
            yield Select[Markdown](
                (
                    ("h1", Markdown("# This is header 1")),
                    ("h2", Markdown("## This is header 2")),
                    ("h3", Markdown("### This is header 3")),
                ),
                prompt="Choose a header ...",
                id="id1",
            )
            yield Select[Markdown](
                (
                    ("bold", Markdown("**This is bold**")),
                    ("italic", Markdown("*This is italics*")),
                ),
                prompt="Choose a format ...",
                id="id2",
            )

    @on(Select.Changed, "#id1")
    def handle_id1_change(self, event: Select.Changed):
        # This will not work
        err = self.query_one("#id2", Select[Markdown])
        reveal_type(err)  # Select[Markdown] - correct, but crashes at runtime



╭──────────────────────────── Traceback (most recent call last) ─────────────────────────────╮
│ /Users/tushar.chandra/work/apis/examples/demo_subscript.py:32 in handle_id1_change         │
[...]

PR #3980, opened 2024-01-08
Handle subscripted generics in query selector

davetapley:
Same as #3805 but change my branch off `main`

### DIFF
--- a/src/textual/css/query.py
+++ b/src/textual/css/query.py
@@ -218,7 +218,13 @@ class DOMQuery(Generic[QueryType]):
         if self.nodes:
             first = self.nodes[0]
             if expect_type is not None:
-                if not isinstance(first, expect_type):
+                try:
+                    type_match = isinstance(first, expect_type)
+                except TypeError:
+                    # to handle subscripted generics like DataTable[float]
+                    type_match = isinstance(first, expect_type.__origin__)
+
+                if not type_match:
                     raise WrongType(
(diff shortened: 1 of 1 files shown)


rodrigogiraoserrao:
Hey Dave, thanks for the PR.
Would you mind adding a couple of regression tests? 🙌

davetapley:
@rodrigogiraoserrao I _can_ but it won't be pretty because:

The fix requires Python 3.9 (for https://peps.python.org/pep-0585/), but the tests also run on 3.8.

Note: #2867 is a hard crash anyway, so fix is _no worse_ than existing behavior on 3.8.

For tests can either:
1. Defer until 3.8 support is dropped (easiest)
2. Only run test on >= 3.9
3. Expect exception on 3.8, work on >= 3.9

Any preference?

davetapley:
@rodrigogiraoserrao I was, but I didn't know about `skipif`, that makes it look quite easy 😁 
I'll get some tests in ASAP.

willmcgugan:
Would I be correct in saying this wouldn't check the generic type, just the class?

i.e. the following would return all DataTable, not just float Datatables?

```python
self.query("*", "DataTable[float]")
```

willmcgugan:
I don't think this would correctly restrict the query. i.e. `self.query(DataTable[float])` would return all datatables, not just tables of floats. If we can't guarantee that the runtime matches the types, then it's not really viable.

Closing for now, but let me know if I've misunderstood or there is another way to achieve this.

PR #3988, opened 2024-01-09
Minor improvements to option list

### LINKED ISSUE #3903: get_option in SelectionList not working
The function `get_option` in `SelectionList` does not work as intended. The following raises `OptionDoesNotExist`

```
from textual.app import App
from textual.widgets import SelectionList
from textual.widgets.selection_list import Selection
class QuickApp(App):

    def compose(self):
        yield SelectionList(
            Selection("a", 1, id="a")
        )

    def on_mount(self):
        list = self.query_one(SelectionList)
        list.get_option("a")
```

I looked through the code and it looks like `_option_ids` is never set in `OptionList` but gets evaluated in `get_option_index`.

rodrigogiraoserrao:
When instantiating an option list, immediately map IDs to options so that methods like `get_option` can be used as soon as the option list is instantiated.

Fixes #3903.


willmcgugan (approved):

willmcgugan:
Merge when tests pass...

PR #4001, opened 2024-01-10
Make TextArea, MarkdownFence respect theme changes.

ggozad:
Make `TextArea`, `MarkdownFence` update their theme when the app theme (`app.dark`) changes.
Happy to add tests if necessary, would help if you pointed out something similar.

### DIFF
--- a/src/textual/widgets/_markdown.py
+++ b/src/textual/widgets/_markdown.py
@@ -503,7 +503,9 @@ class MarkdownFence(MarkdownBlock):
         self.code = code
         self.lexer = lexer
         self.theme = (
-            self._markdown.code_dark_theme if self.app.dark else self._markdown.code_light_theme
+            self._markdown.code_dark_theme
+            if self.app.dark
+            else self._markdown.code_light_theme
         )
 
     def _block(self) -> Syntax:
@@ -523,7 +525,9 @@ class MarkdownFence(MarkdownBlock):
     def _retheme(self) -> None:
         """Rerender when the theme changes."""
         self.theme = (
-            self._markdown.code_dark_theme if self.app.dark else self._markdown.code_light_theme
+            self._markdown.code_dark_theme
+            if self.app.dark
+            else self._markdown.code_light_theme
         )
         code_block = self.query_one(Static)
         code_block.update(self._block())
--- a/src/textual/widgets/_markdown.py
+++ b/src/textual/widgets/_markdown.py
@@ -18,7 +18,7 @@ from ..await_complete import AwaitComplete
 from ..containers import Horizontal, Vertical, VerticalScroll
 from ..events import Mount
 from ..message import Message
-from ..reactive import Reactive, reactive, var
+from ..reactive import reactive, var
 from ..widget import Widget
 from ..widgets import Static, Tree
 
--- a/src/textual/widgets/_markdown.py
+++ b/src/textual/widgets/_markdown.py
@@ -584,8 +584,11 @@ class Markdown(Widget):
 
     BULLETS = ["\u25CF ", "▪ ", "‣ ", "• ", "⭑ "]
 
-    light_theme: Reactive[str] = reactive("material-light")
     dark_theme: reactive[str] = reactive("material")
+    """The theme to use for code blocks when in [dark mode][textual.app.App.dark]."""
+
+    light_theme: reactive[str] = reactive("material-light")
+    """The theme to use for code blocks when in [light mode][textual.app.App.dark]."""
 
     def __init__(
         self,
--- a/src/textual/widgets/_markdown.py
+++ b/src/textual/widgets/_markdown.py
@@ -18,7 +18,7 @@ from ..await_complete import AwaitComplete
 from ..containers import Horizontal, Vertical, VerticalScroll
 from ..events import Mount
 from ..message import Message
-from ..reactive import reactive, var
+from ..reactive import Reactive, reactive, var
 from ..widget import Widget
 from ..widgets import Static, Tree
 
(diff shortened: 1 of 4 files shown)


src/textual/widgets/_text_area.py
davep, line 1: It's best to leave `TextArea` out of this PR. @darrenburns is doing a lot of work on `TextArea` right now; and next up after soft-wrapping is some work on enhanced integration with Textual theming.

ggozad: I amended leaving TextArea out.

src/textual/widgets/_markdown.py
```diff
@@ -500,19 +501,35 @@ class MarkdownFence(MarkdownBlock):
     def __init__(self, markdown: Markdown, code: str, lexer: str) -> None:
         self.code = code
         self.lexer = lexer
+        self.theme = "solarized-dark" if self.app.dark else "solarized-light"
```
davep, line 504: I feel that, ideally, there would be some method of allowing the developer using `Markdown` to specify their preference for the themes to use for light and dark mode; which might mean that there needs to be a pair of properties on the `Markdown` widget itself that supplies these.

src/textual/widgets/_markdown.py
```diff
@@ -500,19 +501,35 @@ class MarkdownFence(MarkdownBlock):
...
         super().__init__(markdown)
 
-    def compose(self) -> ComposeResult:
-        from rich.syntax import Syntax
+    def _retheme(self) -> None:
+        """Swap between a dark and light theme when the mode changes."""
+        self.theme = "solarized-dark" if self.app.dark else "solarized-light"
```
davep, line 509: If the names of the preferences for the light/dark themes move up to `Markdown`, this might be moot, but I feel that hard-coding the theme choices in two different places is something we should avoid.

src/textual/widgets/_markdown.py
```diff
@@ -500,19 +501,35 @@ class MarkdownFence(MarkdownBlock):
...
+        code_block = self.query_one(Static)
+        code_block.renderable = Syntax(
```
davep, line 511: I feel that doubling up on the code that crates the `Syntax` object it something we should avoid.

src/textual/widgets/_markdown.py
```diff
@@ -500,19 +501,35 @@ class MarkdownFence(MarkdownBlock):
...
+    def compose(self) -> ComposeResult:
         yield Static(
             Syntax(
                 self.code,
                 lexer=self.lexer,
                 word_wrap=False,
                 indent_guides=True,
                 padding=(1, 2),
-                theme="material",
+                theme=self.theme,
```
davep, line 532: I note that the theme that was used was `"material"`. but this PR changes the theme used for dark mode; was there a reason for this?

davep (changes requested):
Left a couple of comments and suggestions. I also feel that this might deserve a pair of snapshot tests too.

ggozad:
> Left a couple of comments and suggestions. I also feel that this might deserve a pair of snapshot tests too.

Thank you! I think I agree with all your comments. I am away for a few days, but will get to it asap :)

src/textual/widgets/_markdown.py
ggozad: I think it's a nice idea to specify both light/dark themes. But would you do them as properties or rather simply pass them in `__init__`? I think that I would either:
* Have one `theme` property that the developer can set on the widget that does listen to light/dark changes.
* Specify on `__init__` a light/dark theme (with defaults) and listen to light/dark changes.

src/textual/widgets/_markdown.py
ggozad: No real reason. I just thought that there is some "consistency" in choosing a theme that has both light and dark versions. But I would be happy to align with some other choice if there is one. What would you choose?

src/textual/widgets/_markdown.py
davep: I think, if I were approaching this myself, I'd have them be reactive properties on `Markdown` (as the public API for setting them). My thinking here being:

1. If someone were to change the light or dark themes for code on a "live" `Markdown`, we can watch for those changes and, if necessary (we're in dark mode, and the dark mode theme gets changed), we can query through all of the instances of `MarkdownFence` within us and get them to refresh (a pretty niche situation, I'd say, but not very costly to handle).
2. `MarkdownFence` takes a reference to its parent `Markdown` anyway, so when it needs to reference the appropriate theme, it can just reference the correct property on the markdown object.
3. `MarkdownFence` can watch for the chance in light/dark mode and switch to the most appropriate theme when necessary.

Hopefully that makes sense?

src/textual/widgets/_markdown.py
davep: For now at least I think I'd stick with "material" as the default for dark mode, so then we're not changing the look of people's applications when they're in dark mode.

ggozad: Would you have a preference for the light theme default?

ggozad:
Think all your comments are now addressed as we discussed.
Will look into tests as soon as I get some time!

ggozad:
Here are some snapshot tests, hope I figured out how to do them properly :)

src/textual/widgets/_markdown.py
ggozad: I ended up using "material-light" for consistency.

davep: Sounds like a good choice to me!

davep:
I hope to get to doing some final testing on this in the next day or so (sorry, been super busy with some other stuff); do you wan to resolve the CHANGELOG conflict, or leave it for the moment until we know we're close to merge?

ggozad:
> I hope to get to doing some final testing on this in the next day or so (sorry, been super busy with some other stuff); do you wan to resolve the CHANGELOG conflict, or leave it for the moment until we know we're close to merge?

There is zero rush, take your time :)
I rebased from main, happy to do again later if something comes up.

davep:
@ggozad Just catching up again; aiming to merge this in today. Just a heads-up that I'm making a few small tweaks ahead of that.

ggozad:
> @ggozad Just catching up again; aiming to merge this in today. Just a heads-up that I'm making a few small tweaks ahead of that.

Your tweaks are totally fine, things I should have caught beforehand :)
Please go ahead!

davep (approved):
Looking good to me and I think ready to merge.

@willmcgugan do you want a quick glance over before merge?

willmcgugan:
LGTM

PR #4012, opened 2024-01-12
Signal

willmcgugan:
This is mainly for a new Signal class, and also some fixes / new method

- Adds a Signal class, which will be part of a new "pub sub" system.
- Adds Query.blur and Query.focus
- Fixes rule appearing after nested CSS
- Adds a message_queue_size property
- Adds TabbedContent.active_pane property

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Added `DOMNode.has_pseudo_classes` https://github.com/Textualize/textual/pull/3970
 - Added `Widget.allow_focus` and `Widget.allow_focus_children` https://github.com/Textualize/textual/pull/3989
+- Added Query.blur and Query.focus https://github.com/Textualize/textual/pull/4012
+- Added MessagePump.message_queue_size https://github.com/Textualize/textual/pull/4012
+- Added TabbedContent.active_pane https://github.com/Textualize/textual/pull/4012
 
 ### Fixed
 
@@ -25,6 +28,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Live reloading of TCSS wouldn't apply CSS changes to screens under the top screen of the stack https://github.com/Textualize/textual/issues/3931
 - Fix issue with `Strip.crop` when crop window start aligned with strip end https://github.com/Textualize/textual/pull/3998
 - Fixed Strip.crop_extend https://github.com/Textualize/textual/pull/4011
+- Fixed declaration after nested rule set causing a parse error https://github.com/Textualize/textual/pull/4012
 
 
 ## [0.47.1] - 2023-01-05
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,9 +18,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Added `DOMNode.has_pseudo_classes` https://github.com/Textualize/textual/pull/3970
 - Added `Widget.allow_focus` and `Widget.allow_focus_children` https://github.com/Textualize/textual/pull/3989
-- Added Query.blur and Query.focus https://github.com/Textualize/textual/pull/4012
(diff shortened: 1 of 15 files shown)


src/textual/css/tokenizer.py
```diff
@@ -248,11 +248,14 @@ def get_token(self, expect: Expect) -> Token:
         line = self.lines[line_no]
         match = expect.match(line, col_no)
         if match is None:
+            error_line = line[col_no:].rstrip()
```
willmcgugan, line 251: Slight tweak to formatting. Checks if there is a semi-colon at the end of the line, before suggesting it.

willmcgugan:
@davep You may be able to use the Signal mechanism herein for the suspend resume notification.

darrenburns (approved):
Looks good to me. Not 100% on the name - mostly "publish" sounds a little strange to me (publish a signal?)

CHANGELOG.md
```diff
@@ -17,6 +17,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Added `DOMNode.has_pseudo_classes` https://github.com/Textualize/textual/pull/3970
 - Added `Widget.allow_focus` and `Widget.allow_focus_children` https://github.com/Textualize/textual/pull/3989
+- Added Query.blur and Query.focus https://github.com/Textualize/textual/pull/4012
```
davep, line 21:
```suggestion
- Added `Query.blur` and `Query.focus` https://github.com/Textualize/textual/pull/4012
```

CHANGELOG.md
```diff
@@ -17,6 +17,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
...
+- Added MessagePump.message_queue_size https://github.com/Textualize/textual/pull/4012
```
davep, line 22:
```suggestion
- Added `MessagePump.message_queue_size` https://github.com/Textualize/textual/pull/4012
```

CHANGELOG.md
```diff
@@ -17,6 +17,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
...
+- Added TabbedContent.active_pane https://github.com/Textualize/textual/pull/4012
```
davep, line 23:
```suggestion
- Added `TabbedContent.active_pane` https://github.com/Textualize/textual/pull/4012
```

src/textual/message.py
```diff
@@ -77,7 +77,7 @@ def __init_subclass__(
         cls.handler_name = f"on_{namespace}_{name}" if namespace else f"on_{name}"
 
     @property
-    def control(self) -> Widget | None:
+    def control(self) -> DOMNode | None:
```
davep, line 80: Possibly want to tweak the docstring to say node rather than widget now; or from the "end user" point of view does the intent of the property remain the same? (we want them to think about widgets)

src/textual/signal.py
```diff
@@ -0,0 +1,88 @@
...
+        """
+        if not node.is_running:
+            raise SignalError(
+                f"Node must be running to subscribe to a signal (has {node} been mounted)?"
+            )
+        callbacks = self._subscriptions.setdefault(node, [])
+        if callback not in callbacks:
+            callbacks.append(callback)
+
+    def unsubscribe(self, node: DOMNode) -> None:
```
davep, line 67: As I read the interface, it seems you can subscribe multiple callbacks from a single node, but on the other hand you can only unsubscribe a node (presumably unsubscribing *every* callback). Is that a correct reading and is that intentional?

willmcgugan:
> Looks good to me. Not 100% on the name - mostly "publish" sounds a little strange to me (publish a signal?)

Any suggestions for an alernative?

davep (approved):
One or two suggested tweaks to the ChangeLog, a question about the interface, but on the whole I subscribe to this PR.

src/textual/message.py
willmcgugan: Yeah, might leave it as widget in the docs. It's more meaningful for the average user.

PR #4023, opened 2024-01-15
Fix box model

rodrigogiraoserrao:
This fixes several “off by one” errors when computing `width`, `height`,  and their `max-`/`min-` variations for some combinations of unit, widget padding, and widget margin.

ASSUMPTION of this PR: the margin is _**not**_ taken into account when computing the dimensions of a given widget.

Example application where all red frames should have the same width running with the tip of `main`:

Screenshot 2024-01-15 at 15 10 46

Now running with the fix implemented:

Screenshot 2024-01-15 at 15 10 57

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - `SelectionList` option IDs are usable as soon as the widget is instantiated https://github.com/Textualize/textual/issues/3903
 - Fix issue with `Strip.crop` when crop window start aligned with strip end https://github.com/Textualize/textual/pull/3998
 - Fixed Strip.crop_extend https://github.com/Textualize/textual/pull/4011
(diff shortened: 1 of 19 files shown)


willmcgugan:
This breaks the calculator example and the dictionary example, and I wouldn't be surprised if it breaks some examples in the docs.

Given we've changed the semantics of percentage units, it is going to break a lot of projects.

If this is the way forward, we will have to document it as a breaking change. But I would have a strong preference for not breaking anything.

Why should `width: 1fr`, be different from `width: 100%` ?

rodrigogiraoserrao:
I agree it's reasonable to mark the fix as possibly breaking people's styles.

I may have misunderstood something, but I thought we agreed that the calculations were _wrong_ and that this PR fixes them; I didn't really change the semantics of anything – at least, not knowingly.
What semantics do you think changed?

As for `1fr` vs `100%`, I'm not sure I follow.
Why would they be the same if they mean different things?
`fr` is a unit that is relative to siblings whereas `%` is a unit that is relative to ancestors, so they're clearly different.

TomJGooding:
Presumably #4037 supersedes this PR?

rodrigogiraoserrao:
Ty @TomJGooding

Issue #4026, opened 2024-01-15
[IGNORE] Why doesn't the DataTable scroll programmatically?


TomJGooding:

davep:
🙈

TomJGooding:
Sorry, just testing a theory. It looks like the link is broken because of the backticks.

darrenburns:
👀

PR #4027, opened 2024-01-15
docs(faq): fix broken datatable scroll link

TomJGooding:
The FAQ link for 'Why doesn't the DataTable scroll programmatically?' is broken because of the backticks, as tested here https://github.com/Textualize/textual/issues/4026#issuecomment-1892754619

I've never used FAQtory so hopefully this fix is right - should this package also be added a Textual dev dependency?

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -321,8 +321,8 @@ There is currently a light and dark version of the design system, but more are p
 
 ---
 
-<a name="why-doesn't-the-`datatable`-scroll-programmatically"></a>
-## Why doesn't the `DataTable` scroll programmatically?
+<a name="why-doesn't-the-datatable-scroll-programmatically"></a>
+## Why doesn't the DataTable scroll programmatically?
 
 If scrolling in your `DataTable` is _apparently_ broken, it may be because your `DataTable` is using the default value of `height: auto`.
 This means that the table will be sized to fit its rows without scrolling, which may cause the *container* (typically the screen) to scroll.
--- a/questions/datatable-doesnt-scroll.question.md
+++ b/questions/datatable-doesnt-scroll.question.md
@@ -1,5 +1,5 @@
 ---
-title: "Why doesn't the `DataTable` scroll programmatically?"
+title: "Why doesn't the DataTable scroll programmatically?"
 alt_titles:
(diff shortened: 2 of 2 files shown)


willmcgugan:
Faqtory generates FAQ.md, which means this change will be overwritten if we add a new question.

I think the fix may be in Faqtory itself to remove or escape those quotes.

TomJGooding:
> Faqtory generates FAQ.md, which means this change will be overwritten if we add a new question.

Oh, I assumed I just needed to update the `*.question.md` file then `make faq`, did I miss a step?

> I think the fix may be in Faqtory itself to remove or escape those quotes.

This might be tricky to pin down what characters need removing/escaping, as I believe Markdown converters may generate these heading IDs differently?

TomJGooding:
I've been doing some tests with headings containing punctuation, and at least for GitHub and mkdocs-material the only characters that seem to be retained are hyphens and underscores.

There is however a difference between how GitHub and mkdocs-material handle hyphens.

# Hyphen-Word DoubleHyphen--Word -BeginHyphen --DoubleBeginHyphen`

This heading on GitHub will generate a heading ID with all the hyphens retained:

> hyphen-word-doublehyphen--word--beginhyphen---doublebeginhyphen

Whereas mkdocs-material will convert two or more hyphens in a row (including the replaced spaces) to one:

> hyphen-word-doublehyphen-word-beginhyphen-doublebeginhyphen

I'm happy to create a PR on FAQtory to remove other punctuation characters, but I just wanted to show that Markdown converters may generate these heading IDs differently.

TomJGooding:
> should [FAQtory] also be added a Textual dev dependency?

Today I found that multiple tests were failing on my fork of Textual, and I couldn't understand why until I remembered that I'd installed FAQtory for this PR.

Textual currently requires rich >=13.3.3, whereas FAQtory is pinned to ^12.6.0.

I'm afraid I'm not familiar enough yet with Poetry to fix myself, but should the Rich version be bumped in FAQtory and this package added to the Textual dev dependencies?

willmcgugan:
You probably don't want to install it in your venv. It is better installed as an independant tool. I might update the faqtory repo to recommend `pipx`.

PR #4030, opened 2024-01-16
External watch fix init

### LINKED ISSUE #3878: A `watch` on a reactive that is declared `init=False` fires watch method on init
[Stemming from a question on Discord](https://discord.com/channels/1026214085173461072/1033754296224841768/1185161330462838784); consider this code:

```python
from textual.widgets import Label, Log

class SomeWidget(Label):

    test_1: var[int] = var(0)
    test_2: var[int] = var(0, init=False)

    def watch_test_1(self, was: int, into: int) -> None:
        self.screen.query_one(Log).write_line(f"test_1 {was} -> {into}")

    def watch_test_2(self, was: int, into: int) -> None:
        self.screen.query_one(Log).write_line(f"test_2 {was} -> {into}")

class InitOverrideApp(App[None]):

        yield SomeWidget()
        yield Log()

if __name__ == "__main__":
    InitOverrideApp().run()
```

when run, it correctly logs that `test_1` changed, and correctly doesn't log that `test_2` changed. Now, if we add a `watch` on `test_2`:

```python
from textual.widgets import Label, Log

class SomeWidget(Label):

    test_1: var[int] = var(0)
    test_2: var[int] = var(0, init=False)

    def watch_test_1(self, was: int, into: int) -> None:
        self.screen.query_one(Log).write_line(f"test_1 {was} -> {into}")

    def watch_test_2(self, was: int, into: int) -> None:
        self.screen.query_one(Log).write_line(f"test_2 {was} -> {into}")

class InitOverrideApp(App[None]):

        yield SomeWidget()
        yield Log()

    def on_mount(self) -> None:
        def gndn() -> None:
            return
        self.watch(self.query_one(SomeWidget), "test_2", gndn)

if __name__ == "__main__":
    InitOverrideApp().run()
```

this appears to override the `init=False` declared for `test_2`, and causes `SomeWidget.watch_test_2` to fire. This can be worked around by making the `watch` setup call like this:

```python
        self.watch(self.query_one(SomeWidget), "test_2", gndn, init=False)
```

but it seems like an unintended consequence that the watch method (`SomeWidget.watch_test_2`) gets called without that.

rodrigogiraoserrao:
Fixes #3878 

When a new watcher is created, we invoke that single watcher instead of checking all of the reactive watchers.
To enable this, we extract two auxiliary functions to the module scope from the inner scope they were defined in.


src/textual/reactive.py
```diff
@@ -363,11 +374,13 @@ def _watch(
     """
     if not hasattr(obj, "__watchers"):
         setattr(obj, "__watchers", {})
-    watchers: dict[str, list[tuple[Reactable, Callable]]] = getattr(obj, "__watchers")
+    watchers: dict[str, list[tuple[Reactable, WatchCallbackType]]] = getattr(
+        obj, "__watchers"
+    )
     watcher_list = watchers.setdefault(attribute_name, [])
-    if callback in watcher_list:
+    if any(callback == callback_from_list for _, callback_from_list in watcher_list):
```
rodrigogiraoserrao, line 381: This was checking if it could find an apple in a list of tuples of oranges and apples.
Instead, we must unpack the tuples in the list and compare apples to apples.

darrenburns (changes requested):
A couple of requests, mostly around trying to make sure the tests are more robust.

tests/test_reactive.py
```diff
@@ -575,3 +575,63 @@ def callback(self):
...
+            self.watch(self.holder, "attr", self.callback)
+            self.watch(self.holder, "attr", self.callback)
+
+        def callback(self) -> None:
+            return
+
+    app = MyApp()
+    async with app.run_test():
+        pass
+    assert len(app.holder.__watchers["attr"]) == 1
```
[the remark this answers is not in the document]
rodrigogiraoserrao: Good idea.

tests/test_reactive.py
```diff
@@ -575,3 +575,63 @@ def callback(self):
...
+
+    class InitOverrideApp(App[None]):
+        def compose(self) -> ComposeResult:
+            yield SomeWidget()
+
+        def on_mount(self) -> None:
+            def nop() -> None:
+                return
+
+            self.watch(self.query_one(SomeWidget), "test_2", nop)
```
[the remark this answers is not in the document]
rodrigogiraoserrao: I can make `nop` log something as well.
When `nop` is set as a callback, it _should_ fire (but not the other watchers associated with that same reactive).
I will also make sure that both callbacks are called when the reactive is changed.

src/textual/reactive.py
```diff
@@ -36,6 +43,43 @@ class TooManyComputesError(Exception):
...
+    _rich_traceback_omit = True
+    param_count = count_parameters(watch_function)
+    if param_count == 2:
+        watch_result = cast(WatchCallbackBothValuesType, watch_function)(
+            old_value, value
+        )
+    elif param_count == 1:
+        watch_result = cast(WatchCallbackNewValueType, watch_function)(value)
+    else:
+        watch_result = cast(WatchCallbackNoArgsType, watch_function)()
```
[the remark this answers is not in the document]
darrenburns: Not sure I'm following this. The `param_count` can't be known until runtime as it gets computed dynamically, but the type checker runs before the program starts. So how can the conditionals + casts influence the type checkers behaviour if they run after the type checker?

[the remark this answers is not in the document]
rodrigogiraoserrao: That's exactly how type narrowing works in general.
```py
x: int | None = ...
if x is not None:
    y = x + 3
```

The `if` only runs at runtime and yet the type checker understands that `if x is not None` is narrowing the type of `x`, effectively concluding that `x` must be an integer inside that `if`.

However, the type checker isn't smart enough to do type narrowing with the `if` with the `param_count` check.
So, it complains, even though we know the code runs precisely because at runtime we only run the branch that works.
Precisely because _these_ `if` do not influence the type checker, I add the casts to silence it.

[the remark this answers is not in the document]
darrenburns: As you noted, the example you gave of `if x is not None` is not the same, since it's can be inferred statically.

I was mostly surprised about how it seems like a whole bunch of type definitions that are only used for casting, and their sole purpose seems to be to silence the type checker.

Not a blocker though, was just a curiosity :)

tests/test_reactive.py
```diff
@@ -598,3 +598,70 @@ def compose(self) -> ComposeResult:
...
+                logs.append("test_2_extra")
+
+            self.watch(self.query_one(SomeWidget), "test_2", watch_test_2_extra)
+
+    app = InitOverrideApp()
+    async with app.run_test():
+        assert logs == ["test_1", "test_2_extra"]
+        app.query_one(SomeWidget).test_2 = 73
+        assert logs.count("test_2_extra") == 2
+        assert "test_2" in logs
```
darrenburns, line 667: Let's make sure that the watcher fires exactly once instead of at least once.
```suggestion
        assert logs.count("test_2") == 1
```

tests/test_reactive.py
```diff
@@ -598,3 +598,70 @@ def compose(self) -> ComposeResult:
...
+            self.watch(self.holder, "attr", self.callback)
+            self.watch(self.holder, "attr", self.callback)
+
+        def callback(self) -> None:
+            nonlocal counter
+            counter += 1
+
+    app = MyApp()
+    async with app.run_test():
+        before = counter
```
darrenburns, line 626: Any reason why you do the before and after stuff here instead of `assert counter == 1` before the assignment and then `assert counter == 2` after the assignment? That would seem simpler and a more accurate test.

As it stands, this test wouldn't catch duplicate watchers firing on `init`.

src/textual/reactive.py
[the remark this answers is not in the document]
rodrigogiraoserrao: Ah, yeah. I refactored the original `WatchCallbackType` because it was huge and it makes sense to split it into 3, depending on the number of arguments.
Then, I used that refactor to silence the type checker 🤷

tests/test_reactive.py
rodrigogiraoserrao: I don't remember ever seeing that we must call watchers in any specific order. Do we..?

tests/test_reactive.py
rodrigogiraoserrao: I was testing whether the assignment triggered a single watcher but I agree your suggestion yields a more accurate test.

rodrigogiraoserrao: See 1e682c2 for the change you suggested + a fix to a bug that your suggestion uncovered.

tests/test_reactive.py
```diff
@@ -598,3 +598,68 @@ def compose(self) -> ComposeResult:
...
+    class Holder(Widget):
+        attr = var(None)
+
+    class MyApp(App[None]):
+        def __init__(self) -> None:
+            super().__init__()
+            self.holder = Holder()
+
+        def on_mount(self) -> None:
+            self.watch(self.holder, "attr", self.callback)
```
darrenburns, line 617: The test name/docstring says it's checking that we skip duplicate watchers, but the test no longer involves a duplicate watcher. It seems to just be checking the basic usage of `watch`, which is probably covered by another test.

rodrigogiraoserrao: As for the first two paragraphs, I think I deleted the second callback when I was debugging the bug I mentioned in a previous commit and then forgot to add it again, thanks for spotting that...

As for the third paragraph... Do you mean something like the test `test_external_watch_init_does_not_propagate_to_externals` added in 8b349e3?

tests/test_reactive.py
darrenburns: I don't know that it's documented, but we should ensure it's consistent I suppose 🤷

tests/test_reactive.py
darrenburns: Yup, looks good!

darrenburns (approved):
Looks good to me, thanks for making those changes

PR #4032, opened 2024-01-16
Validate ids

### LINKED ISSUE #3954: It is possible to give a widget an ID that can't be queried back
It is possible to give a widget an ID that can't then be queried back; for example:

```python

class BadIDApp(App[None]):

        yield Label("Hello, World!", id="foo&")

    def on_mount(self) -> None:
        self.notify(f"{self.query_one('#foo&')}")

if __name__ == "__main__":
    BadIDApp().run()
```

Perhaps we should validate widget IDs when they are assigned?

rodrigogiraoserrao:
Fix #3954.

There was already a check taking place but the regex used was too lenient.

### DIFF
--- a/src/textual/dom.py
+++ b/src/textual/dom.py
@@ -38,7 +38,7 @@ from .color import BLACK, WHITE, Color
 from .css._error_tools import friendly_list
 from .css.constants import VALID_DISPLAY, VALID_VISIBILITY
 from .css.errors import DeclarationError, StyleValueError
-from .css.parse import parse_declarations
+from .css.parse import parse_declarations, validate_identifier
 from .css.styles import RenderStyles, Styles
 from .css.tokenize import IDENTIFIER
 from .message_pump import MessagePump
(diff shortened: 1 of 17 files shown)


src/textual/dom.py
```diff
@@ -62,7 +62,7 @@
 
 from typing_extensions import Literal
 
-_re_identifier = re.compile(IDENTIFIER)
+_re_identifier = re.compile(f"^{IDENTIFIER}$")
```
willmcgugan, line 65: Do we need to do this, or can we use `re.fullmatch` ?

rodrigogiraoserrao: Tests pass with `re.fullmatch` instead of the formatted string identifier.
Do you want me to update the PR with `fullmatch`?

willmcgugan: Simpler perhaps. Go with full match.

rodrigogiraoserrao: See 98e28f0.

Issue #3721, opened 2023-11-21
`Input` wider than widgets sized the same


davep:
I've not dived into *why* this is, but I just noticed it while working on something else and wanted to make a note for further checking. If you have a collection of widgets, all with the same width, `Input` stands out as not being the same width as the others:

```python
from textual.app import App, ComposeResult
from textual.widgets import Input, TextArea, Static, Button

class InputVsTextArea(App[None]):

    CSS = """
    Screen > *, Screen > *:focus {
        width: 50%;
        height: 1fr;
        border: solid red;
    }
    """

    def compose(self) -> ComposeResult:
        yield Input()
        yield TextArea()
        yield Static()
        yield Button()

if __name__ == "__main__":
    InputVsTextArea().run()
```

![Screenshot 2023-11-21 at 21 19 28](https://github.com/Textualize/textual/assets/28237/dbd29d31-3560-48d0-ad72-517acc077ca3)

Possible down to something in its `DEFAULT_CSS`? Whatever the cause, it should probably really line up with the other widgets when at the same width.

rodrigogiraoserrao:
I wondered whether it was something about the box sizing, but it's not.

Also, this happens with `h` and `w` units but not with absolute widths, `vh`, and `vw`.
`fr` untested.

TomJGooding:
You've probably figured this out already, but this seems to be due to the `padding: 0 2` in the Input default CSS.

If you add the same padding to the other widgets in your example, they display the same width. This possibly suggests a bug with how the width is calculated not taking into account padding?

rodrigogiraoserrao:
I meant to say that the issue could be a rule `box-sizing` that had the wrong value and it turns out it's not that (or, at least, not just that).

It may very well just be that the style `width` doesn't take into account `padding`, which may be on purpose.

TomJGooding:
I *think* I might have narrowed it down to this line:

https://github.com/Textualize/textual/blob/a64a0d21d787016df9dde84287cb36ecda50b787/src/textual/widget.py#L991

I confess I'm struggling a bit to follow the code, but here's what seems to be happening if we take my example app below. 

1. `content_container` = 96 (100 - 4)
2. `content_width` = 48 (96 * 0.5)
3. widget width = 52 (48 + 4)

```python
from textual.app import App, ComposeResult
from textual.containers import Container
from textual.widgets import Static


class PercentWidthTestApp(App):
    CSS = """
    Container {
        width: 100;
    }

    Static {
[...]

PR #4037, opened 2024-01-17
fix for percentage dimensions

willmcgugan:
Calculating percentage dimensions didn't take in to account padding / border.

Fixes https://github.com/Textualize/textual/issues/3721

Supersedes https://github.com/Textualize/textual/pull/4023


src/textual/widget.py
```diff
@@ -1143,25 +1137,21 @@ def _get_box_model(
             styles_height = styles.height
             # Explicit height set
             content_height = styles_height.resolve(
-                sizing_container - styles.margin.totals, viewport, height_fraction
+                container - styles.margin.totals, viewport, height_fraction
             )
-            if is_border_box and styles_height.excludes_border:
```
willmcgugan, line 1148: Not sure what I was thinking re `excludes_border`. But it doesn't appear to be required.

rodrigogiraoserrao:
There's something wrong with this PR.

If you look at `dimensions_width.py` test in the superseded PR, you'll see that your PR doesn't pass that test.
Barring any misunderstanding on my part of how `max-`/`min-` units are supposed to work, the apps from `dimensions_width.py`, `dimensions_max_width.py`, and `dimensions_min_width.py`, all should look the same but they don't (I suggest looking at them on a terminal of width 80 because that's the length of the ruler).

willmcgugan:
@rodrigogiraoserrao Could you please isolate a single case that isn't working. i.e. an MRE with as few widgets as possible which displays incorrect behaviour.

rodrigogiraoserrao:
Sure, here is one: `%` isn't respected when `margin` is set.
In this case, the red frame should take up all of the horizontal width but it's off by 4

```py


    CSS = """
    Label.target {
        width: 100%;
        margin: 2;
        border: solid red;
    }
    """

        # yield Label("1234567890" * 80)
        yield Label("", classes="target")


if __name__ == "__main__":
```

Another manifestation of the same problem can be seen in the app below. The bottom label is correctly set to 50% width (via `max-width`) but the top one isn't:

```py
from textual.widgets import Label, Static


    CSS = """
    Label {
        border: solid red;
        margin: 0 4;
    }
    #first { width: 50%; }
    #second {
        width: 100%;
        max-width: 50%;
    }
    """

        yield Static("1234567890" * 8)
        yield Label("", id="first")
        yield Label("", id="second")


if __name__ == "__main__":
```

rodrigogiraoserrao:
This may be why you feel my PR changed the semantics of `%`, but we talked and we agreed that the margin is _not_ a part of the size of the widget.

When we talked, you mentioned there was some special case in the code to do something cheeky with `100%`, widgets with a single children, and margin, but we also realised that you didn't need that special case (which impacted other situations where there are % units and margin, like the one with the `50%` above).

willmcgugan:
It's behaving as I'd expect.

The percentage applies to the available width, which doesn't include margin.

If it didn't behave like this, and you set width 100% with a margin, it would be pushed off screen. It would be simpler (for us) if the percentage was an exact percentage of the container, but this would make it very hard for the dev to achieve sensible layouts.

The behaviour should be identical if you have a width of `1fr` with a single item.

rodrigogiraoserrao:
I was under the impression that the percent unit `%` was relative to the _total size of the container_ and not _“available” size of the container_, and that using margins that are too big with sizes that are too big does result in things getting pushed off.

I know we don't blindly copy browser CSS but it appears that's what it does as well:

Screenshot 2024-01-17 at 17 40 05

When we talked, we understood that the effect you're describing could be achieved by setting padding in the container and child dimensions to `fr` units, for example, precisely because `100%` + `margin` pushes things off screen.

TomJGooding:
Hope you don't mind me sticking my nose in as usual, but this behaviour with margin seems counter-intuitive to me?

**Percent size with padding :white_check_mark:**

<details>
<summary>Code</summary>

```python3


class ExampleApp(App):
    CSS = """
    Screen {
        align: center middle;
    }

    Container {
        height: 10;
        width: 10;
        background: blue;
    }

    Static {
        height: 50%;
        width: 50%;
        background: red;
        padding: 2 1;
    }
    """

        with Container():
            yield Static("5x5")


if __name__ == "__main__":
    app = ExampleApp()
    app.run()
```

</details>

image


**Percent size with padding and margin :x:**

<details>
<summary>Code</summary>

```python3


class ExampleApp(App):
    CSS = """
    Screen {
        align: center middle;
    }

    Container {
        height: 10;
        width: 10;
        background: blue;
    }

    Static {
        height: 50%;
        width: 50%;
        background: red;
        padding: 2 1;
        margin-top: 5;
        margin-left: 5;
    }
    """

        with Container():
            yield Static("5x5")


if __name__ == "__main__":
    app = ExampleApp()
    app.run()
```

</details>

image

willmcgugan:
It's a pragmatic design choice for Textual. If I have a widget with width and height 100% and apply a margin, it currently produces something useful. i.e. the widget is centered in the screen. If the margin was additive, then the widget would overflow the screen boundary. That is very unlikely to be useful for the developer.

If you do want a widget to be an exact percentage, but offset from the top left then you can use `offset` and achieve the same results.

The interaction with margin is orthogonal to the original issue. The original issue was that percentage dimensions excluded the gutter, which this PR should fix. Nothing re margins has changed. We can always change that if needed, but AFAIK nobody as found this to be a problem, and I think we should err on the side of not breaking things.

src/textual/layouts/horizontal.py
```diff
@@ -26,7 +26,9 @@ def arrange(
         add_placement = placements.append
 
         child_styles = [child.styles for child in children]
-        box_margins: list[Spacing] = [styles.margin for styles in child_styles]
+        box_margins: list[Spacing] = [
+            styles.margin for styles in child_styles if styles.overlay != "screen"
+        ]
```
rodrigogiraoserrao, line 31: I see you changed this because you probably thought of an edge case that wasn't being covered; wouldn't this warrant a test?

src/textual/layouts/horizontal.py
```diff
@@ -36,7 +38,7 @@ def arrange(
                     ]
                 )
                 + (box_margins[0].left + box_margins[-1].right),
-                max(
+                min(
```
rodrigogiraoserrao, line 41: Same comment re: test

src/textual/layouts/vertical.py
```diff
@@ -29,7 +29,7 @@ def arrange(
         ]
         if box_margins:
             resolve_margin = Size(
-                max(
+                min(
```
rodrigogiraoserrao, line 32: Same comment re: test

src/textual/widget.py
rodrigogiraoserrao: I agree it doesn't appear to be required at all.

rodrigogiraoserrao (commented):
If `margin` + `%` is to work like so, shouldn't `margin` be taken into account for `max`/`min` units, then?

E.g., regardless of the way we decide to resolve `%` units with margin, I'd expect the two placeholders below to have the same width:
Screenshot 2024-01-18 at 14 22 09

<details>
<summary>App code:</summary>

```py
from textual.widgets import Label, Placeholder


    CSS = """
    Placeholder {
        margin: 0 4;
        height: 3;
    }
    #first { 
        width: 50%;
    }
    #second {
        width: 999;
        max-width: 50%;
    }
    """

        yield Label("1234567890" * 8)
        yield Placeholder("", id="first")
        yield Placeholder("", id="second")


if __name__ == "__main__":
```

</details>

TomJGooding:
> It's a pragmatic design choice for Textual. If I have a widget with width and height 100% and apply a margin, it currently produces something useful. i.e. the widget is centered in the screen.

I promise this will be my last post on this PR, but surely this is where you should add padding to the container? I'm afraid I still don't understand the reason for changing the size of a widget based on its margin.

src/textual/layouts/horizontal.py
willmcgugan: It probably should. Going to leave it for now, because overlay isn't yet documented.

src/textual/layouts/horizontal.py
willmcgugan: It is covered by a snapshot, but it probably does deserve a more specific test. Going to leave it for now though.

rodrigogiraoserrao:
@willmcgugan you didn't say anything regarding [this comment](https://github.com/Textualize/textual/pull/4037#pullrequestreview-1829781345) so I'm not sure if this slipped past you or if it's the intended behaviour.

Issue #4039, opened 2024-01-17
Nested TCSS interacts poorly with pseudo-classes in selectors.


rodrigogiraoserrao:
Working on #3969 and #3999 we realised that pseudo-classes in nested selectors aren't correctly handled.

(In short, to the tokenizer, a selector + pseudo-class looks the same as a rule + value in a nested context.)

Examples of CSS that breaks the tokenizer:

```py
from textual.app import App, ComposeResult
from textual.widgets import Label


class NestedCSSTokenErrorApp(App[None]):
    CSS = """
    Screen {
        Label:hover {
                border: solid red;
        }
    }
    """

    def compose(self) -> ComposeResult:
        yield Label("This is class foo", classes="foo")
        yield Label("This is class bar", classes="bar")


if __name__ == "__main__":
    NestedCSSTokenErrorApp().run()
```

```py
from textual.app import App, ComposeResult
from textual.widgets import Label


class NestedCSSTokenErrorApp(App[None]):
    CSS = """
    Screen {
        Label {
            &.foo, &.bar:hover {
                border: solid red;
            }
        }

        border: solid green;
    }
    """

    def compose(self) -> ComposeResult:
        yield Label("This is class foo", classes="foo")
        yield Label("This is class bar", classes="bar")


if __name__ == "__main__":
    NestedCSSTokenErrorApp().run()
```

rodrigogiraoserrao:
See [Will's comment](https://github.com/Textualize/textual/issues/3999#issuecomment-1896259284) showing how the use of `&` “makes” some cases work.

PR #4040, opened 2024-01-17
Allow lists of nested selectors and allow styles after nested CSS blocks

### LINKED ISSUE #3969: Nesting selectors in a selector list causes a stylesheet error
If you have a stylesheet like this:

```python

    CSS = """
    Label {
        &.foo, &.bar {
            border: solid red;
        }
    }
    """

if __name__ == "__main__":
```

the result is:

```
Error in stylesheet:
 /Users/davep/develop/python/textual-sandbox/nested_css_token_error.py, NestedCSSTokenErrorApp.CSS:3:16
╭─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│   1 │                                                                                                                                               │
│   2 │   Label {                                                                                                                                     │
│ ❱ 3 │   │   &.foo, &.bar {                                                                                                                          │
│   4 │   │   │   border: solid red;                                                                                                                  │
│   5 │   │   }                                                                                                                                       │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
 • Expected selector or { (found '&.bar {').                                                                                                           
```

This feels like a bug given that it's a slightly different way of writing this:

```python

    CSS = """
    Label {
        &.foo {
            border: solid red;
        }
        &.bar {
            border: solid red;
        }
    }
    """

if __name__ == "__main__":
```

Tagging @willmcgugan for triage.

### LINKED ISSUE #3999: A style after a nested section in CSS causes a CSS parsing error
This code:

```python

class NestedParseErrorApp(App[None]):

    CSS = """
    Screen {
        Label {
            border: solid red;
        }
        border: solid green;
    }
    """

        yield Label("First", id="first")
        yield Label("Second", id="second")

if __name__ == "__main__":
    NestedParseErrorApp().run()
```

results in this error:

```
Error in stylesheet:
 /Users/davep/develop/python/textual-sandbox/nested_error.py, NestedParseErrorApp.CSS:6:15
│   4 │   │   │   border: solid red;                                                                                                                                                                                         │
│   5 │   │   }                                                                                                                                                                                                              │
│ ❱ 6 │   │   border: solid green;                                                                                                                                                                                           │
│   7 │   }                                                                                                                                                                                                                  │
│   8                                                                                                                                                                                                                        │
 • Expected selector or { (found ': solid green                                                                                                                                                                               
 • ').                                                                                                                                                                                                                        
```

whereas this works fine:

```python

class NestedParseErrorApp(App[None]):

    CSS = """
    Screen {
        border: solid green;
        Label {
            border: solid red;
        }
    }
    """

        yield Label("First", id="first")
        yield Label("Second", id="second")

if __name__ == "__main__":
    NestedParseErrorApp().run()
```

The difference being where the `border` for the `Screen` comes in relation to the nested `Label`.

rodrigogiraoserrao:
Fixes #3969 and fixes #3999.

Changes the tokenizer to accept lists of selectors in a nested context and also to accept styles after nested scopes.

While working on the above, we opened #4039 because the tokenizer isn't yet clever enough to distinguish a selector with a pseudo-class from a rule with a value in a nested context, if the selector + pseudo-class doesn't use the `&` operator. (See [my comment](https://github.com/Textualize/textual/issues/3999#issuecomment-1896164715) and the interactions under it.)


willmcgugan:
@rodrigogiraoserrao Please add some detail on what you fixed (and how). The issues you linked to don't make it clear what work is outstanding. The title is also rather vague -- what was improved?

rodrigogiraoserrao:
@willmcgugan changed the title, expanded the description, and clarified the [outstanding issue 4039](https://github.com/Textualize/textual/issues/4039).

darrenburns (changes requested):
I've added a couple more comments - sorry!

rodrigogiraoserrao:
@darrenburns ignored! But did you review at all? It doesn't show neither an accepted nor a change requested.

willmcgugan:
@rodrigogiraoserrao  Your tests are essentially snapshot tests. They are as likely to catch *any* change to the tokenizer, as they are to catch errors.

Suggest we have a test which takes CSS then checks that the expected styles are set on the widget. If that breaks in the future we know that it is a genuine issue, and not just that we have made an inconsequential changed to the tokenizer.

rodrigogiraoserrao:
> @rodrigogiraoserrao Your tests are essentially snapshot tests. They are as likely to catch _any_ change to the tokenizer, as they are to catch errors.

Isn't that true of all tokenizer tests?

> Suggest we have a test which takes CSS then checks that the expected styles are set on the widget.

Then, doesn't that mean the test can fail for many different reasons that are completely orthogonal to the thing I'm trying to test, which is that the tokenizer knows what to do with these tokens?

willmcgugan:
> Isn't that true of all tokenizer tests?

Yes, which is why we don't need more.

> Then, doesn't that mean the test can fail for many different reasons that are completely orthogonal to the thing I'm trying to test, which is that the tokenizer knows what to do with these tokens?

That's not what we should be testing. We should be testing the CSS that didn't work before, and now does. The output from the tokenizer does not prove that the CSS is processes as expected.

rodrigogiraoserrao:
@willmcgugan got it. See 3c18e47.
This is ready to be reviewed.

tests/css/test_nested_css.py
```diff
@@ -44,6 +44,54 @@ async def test_nest_app():
...
+    app = ListOfNestedSelectorsApp()
+    red = Color.parse("red")
+    async with app.run_test():
+        assert app.query_one(".foo").styles.background == red
+        assert app.query_one(".bar").styles.background == red
+        assert app.query_one(".heh").styles.background != red
+
+
+class DeclarationAfterNestedApp(App[None]):
+    # css = "Screen{Label{background:red;}background:green;}"
```
darrenburns, line 73:
```suggestion
```

darrenburns (approved):
Looks good!

Issue #4045, opened 2024-01-18
TextArea should handle missing tree-sitter-languages binaries


tconbeer:
I'm upstreaming https://github.com/tconbeer/textual-textarea/issues/198 (ultimately from a Harlequin user). 

When installed with the `syntax` extra, the TextArea assumes that all of the binaries for the "built-in" languages for tree-sitter will be present. However, this may not be the case; for example (in the lijnked issue), Conda (micromamba) seems to struggle with installing the language binaries, and they may be missing on some platforms.

In that case, initializing a TextArea with a built-in language name will raise an `OSError`, which is not handled by the widget. I think it would be better to handle this error and re-raise as a `SyntaxAwareDocumentError`, so the TextArea degrades instead of crashes.

I'll be opening a PR with a simple fix for this shortly.

Full trace from the linked issue:

```
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ /home/moritz/micromamba/envs/harlequin-issue/lib/python3.11/site-packages/textual/widget.py:3386 │
│ in _on_compose                                                                                   │
│                                                                                                  │
│   3383 │                                                                                         │
│   3384 │   async def _on_compose(self) -> None:                                                  │
│   3385 │   │   try:                                                                              │
│ ❱ 3386 │   │   │   widgets = [*self._nodes, *compose(self)]                                      │
│   3387 │   │   except TypeError as error:                                                        │
│   3388 │   │   │   raise TypeError(                                                              │
│   3389 │   │   │   │   f"{self!r} compose() method returned an invalid result; {error}"          │
│                                                                                                  │
│ /home/moritz/micromamba/envs/harlequin-issue/lib/python3.11/site-packages/textual_textarea/texta │
│ rea.py:950 in compose                                                                            │
│                                                                                                  │
│    947 │                                                                                         │
│    948 │   def compose(self) -> ComposeResult:                                                   │
│    949 │   │   with TextContainer():                                                             │
│ ❱  950 │   │   │   yield TextInput(language=self._language)                                      │
│    951 │   │   │   yield CompletionList()                                                        │
│    952 │   │   with FooterContainer():                                                           │
│    953 │   │   │   yield Label("", id="validation_label")                                        │
│                                                                                                  │
│ /home/moritz/micromamba/envs/harlequin-issue/lib/python3.11/site-packages/textual_textarea/texta │
│ rea.py:191 in __init__                                                                           │
│                                                                                                  │
[...]

PR #3933, opened 2023-12-28
Text Area wrapping

darrenburns:
Adds soft wrapping to the `TextArea` widget.

PR #4046, opened 2024-01-18
fix: handle missing tree-sitter-language binaries

tconbeer:

- [x] Updated documentation

Closes https://github.com/Textualize/textual/issues/4045

I don't think new documentation is necessary. Happy to provide if you disagree.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - `SelectionList` option IDs are usable as soon as the widget is instantiated https://github.com/Textualize/textual/issues/3903
 - Fix issue with `Strip.crop` when crop window start aligned with strip end https://github.com/Textualize/textual/pull/3998
 - Fixed Strip.crop_extend https://github.com/Textualize/textual/pull/4011
+- Fixed a crash if the `TextArea` language was set but tree-sitter lanuage binaries were not installed https://github.com/Textualize/textual/issues/4045
 
 
 ## [0.47.1] - 2023-01-05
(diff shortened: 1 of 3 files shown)


davep:
I've not looked to see if it is affected or not, but note that #3933 is close to being merged and has a *lot* of changes to all things `TextArea`; so keep in mind that this might get held back for a wee while until that's merged, or perhaps run this by @darrenburns as a small addition there (there is also a followup PR to come that will make some other QoL changes to `TextArea`, so it might make sense that it rolls into that).

tconbeer:
Thanks, Dave. I'm fine if this waits on the other PR, although a quick scan through the changes didn't reveal any obvious conflicts

darrenburns:
Strange - this doesn't happen for me if I uninstall `tree-sitter-languages`. The functionality is supposed to degrade with a logged warning in the console:

image

The intention is that the constructor of `SyntaxAwareDocument` shouldn't even be called if `tree-sitter-languages` is unavailable (I would expect `TREE_SITTER` to be `False` on the line shown below, which results in a plain `Document` being created).

https://github.com/Textualize/textual/blob/23735cdc341d07d27cece6e1c94ac27a9b155609/src/textual/widgets/_text_area.py#L625

https://github.com/Textualize/textual/blob/bbde62fc5751a03d5fe6e96ee0da469308f16295/src/textual/_tree_sitter.py#L10

So it seems the binaries can be installed and then fail in some environments?

I don't understand the root cause, but I think handling this error is still a good call regardless.

darrenburns:
Ah, I think I understand now. So the `tree-sitter-languages` package on Conda is broken such that the package is available, but doesn't actually contain the required binaries... I think? So, installation of `tree-sitter-languages` succeeds, but it's essentially an empty package 😄

darrenburns (approved):

willmcgugan:
Thanks @tconbeer !

PR #4048, opened 2024-01-19
docs(questions): recommend installing faqtory via pipx

TomJGooding:
Update internal docs to recommend installing faqtory with pipx, as per the updated faqtory readme.


willmcgugan:
Thanks

PR #4062, opened 2024-01-23
Add support for env variable TEXTUAL_ANIMATIONS

### LINKED ISSUE #3992: Env var to modify animation
Add an environment variable to set animation behaviour.

`TEXTUAL_ANIMATION` should be one of the following values:

`NONE` for no animation at all.
`BASIC` for just scrolling animation.
`FULL` for all scrolling.

Any other values should be equivalent to `FULL`

This should also be replicated in `App` with a property, so it could be set programatically.

rodrigogiraoserrao:
Adds support for the environment variable `TEXTUAL_ANIMATIONS`, which dictates which animations take place in Textual apps.
(Kind of like a log level but for animations.)

Fixes #3992.


src/textual/constants.py
```diff
@@ -4,14 +4,31 @@
...
+import enum
 import os
 
 from typing_extensions import Final
 
 get_environ = os.environ.get
 
 
-def get_environ_bool(name: str) -> bool:
+class AnimationsEnum(enum.Enum):
```
willmcgugan, line 15: We generally haven't used Enums in public interfaces, because they require an additional import step. And they are a little verbose. This also doesn't feel like the right place for this. I wouldn't think of importing `AnimationsEnum` from textual.contants.

Suggest we use a typing Literal, which is more typical for this codebase, and doesn't require an additional import. A literal gives us most of the benefits of eums.

src/textual/_animator.py
```diff
@@ -93,9 +95,19 @@ class SimpleAnimation(Animation):
...
+    """Minimum level required for the animation to take place (inclusive)."""
+
+    def __post_init__(self) -> None:
+        """Cache baseline that determines whether an animation runs."""
+        # We cache this value so we don't have to `.get` the context var repeatedly.
+        self.show_animations = active_app.get().show_animations
 
     def __call__(self, time: float) -> bool:
-        if self.duration == 0:
+        if (
```
willmcgugan, line 107: This is where we need the `app.show_animations`. Could we pass it to `__call__` ?

Alternatively, do we even need to do this logic here? The animator calls this method. Could we not perform this logic from the animator? The animator has a reference to the app.

src/textual/app.py
```diff
@@ -575,6 +576,12 @@ def __init__(
         self.set_class(self.dark, "-dark-mode")
         self.set_class(not self.dark, "-light-mode")
 
+        self.show_animations: AnimationsEnum = constants._get_textual_animations()
```
willmcgugan, line 579: Calling a method here, suggests it is not a constant. The `constants` module should really only expose values.

The name `show_animations` strongly suggests a bool, which is likely to confuse.

src/textual/css/scalar_animation.py
```diff
@@ -48,11 +52,21 @@ def __init__(
             assert duration is not None, "Duration expected to be non-None"
             self.duration = duration
 
+        self.show_animations = active_app.get().show_animations
```
willmcgugan, line 55: The animator has a reference to app, so I think we can avoid this by moving some of the logic to the animator.

src/textual/widgets/_tabs.py
```diff
@@ -590,11 +591,17 @@ def watch_active(self, previously_active: str, active: str) -> None:
             underline.highlight_end = 0
             self.post_message(self.Cleared(self))
 
-    def _highlight_active(self, animate: bool = True) -> None:
+    def _highlight_active(
+        self,
+        animate: bool = True,
+        *,
+        animate_on_level: AnimationsEnum = AnimationsEnum.BASIC,
```
willmcgugan, line 598: Seems odd to have this in the signature. I don't see anywhere that sets it, and given it is private, the user should not need to set it.

src/textual/_animator.py
```diff
@@ -93,9 +95,19 @@ class SimpleAnimation(Animation):
     final_value: object
     easing: EasingFunction
     on_complete: CallbackType | None = None
+    animate_on_level: AnimationsEnum = AnimationsEnum.BASIC
```
willmcgugan, line 98: Not keen on this name. It doesn't read well to me. Given from the context it is clearly related to animation, can it not simply be "level" ?

willmcgugan (changes requested):
When you animate something, you only really need to tell the animation system if it is a basic animation that doesn't slow down the UI, or a more gratuitous animation that is just because it looks cool.

This makes having a bool with 3 values seem superfluous.

I would think that when calling animate, you would specify "basic" or "full".

Something like.

```python
self.animate("x", 50, level="basic")
```

Where the default would be "full".

Note that I don't think we should use Enums in the public interface (see comments in the code).

src/textual/_animator.py
```diff
@@ -93,9 +95,19 @@ class SimpleAnimation(Animation):
...
+    """Minimum level required for the animation to take place (inclusive)."""
+
+    def __post_init__(self) -> None:
+        """Cache baseline that determines whether an animation runs."""
+        # We cache this value so we don't have to `.get` the context var repeatedly.
+        self.show_animations = active_app.get().show_animations
```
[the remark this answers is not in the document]
rodrigogiraoserrao: I can add a fallback to fetch the env var in case there is no active app.
As for copying it from the app, we are caching it to prevent accessing the context var every single time we compute the next “frame” of the animation.

src/textual/_animator.py
rodrigogiraoserrao: We _could_ perform the logic there but I think it's cleaner here.
If we do it on the animator, then we also need to set the attribute to its value in the animator and I think it'll look clunky there.

src/textual/app.py
rodrigogiraoserrao: Changed to `animations_level`.

src/textual/css/scalar_animation.py
rodrigogiraoserrao: Removed it from here.

src/textual/widgets/_tabs.py
rodrigogiraoserrao: Fair point.

src/textual/_animator.py
rodrigogiraoserrao: Sure.

src/textual/constants.py
rodrigogiraoserrao: Sure thing.

src/textual/_animator.py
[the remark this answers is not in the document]
rodrigogiraoserrao: My reply is no longer relevant after a first pass through the comments.

src/textual/_animator.py
willmcgugan: You may need to elaborate on "cleaner" and "clunky" which are quite subjective terms.

It does feel like it is the animators job to decide if to animate something. Perhaps an Animation object should grow a `finalize` method which sets the attribute to its final stage.

src/textual/css/scalar_animation.py
```diff
@@ -48,11 +51,18 @@ def __init__(
...
 
-    def __call__(self, time: float) -> bool:
+    def __call__(
+        self, time: float, app_animation_level: AnimationLevel = "full"
+    ) -> bool:
         factor = min(1.0, (time - self.start_time) / self.duration)
         eased_factor = self.easing(factor)
 
-        if eased_factor >= 1:
+        if (
```
willmcgugan, line 60: I guess this sets the value to its final position, but what is to prevent the animator calling it multiple times? Anything?

src/textual/widgets/_button.py
```diff
@@ -239,22 +240,29 @@ async def _on_click(self, event: events.Click) -> None:
         event.stop()
         self.press()
 
-    def press(self) -> Self:
+    def press(self, *, level: AnimationLevel = "basic") -> Self:
```
willmcgugan, line 243: After a discussion in the office, we don't think this qualifies as an animation at all

src/textual/widgets/_loading_indicator.py
```diff
@@ -54,6 +54,9 @@ def _on_mount(self, _: Mount) -> None:
         self.auto_refresh = 1 / 16
 
     def render(self) -> RenderableType:
+        if self.app.animation_level is "none":
```
willmcgugan, line 57: Good thinking.

willmcgugan (changes requested):
A few suggestions

src/textual/css/scalar_animation.py
rodrigogiraoserrao: Yeah. The `return True` tells the animator the animation was completed, which means that the animation will have its "on complete callback" called (if there's any) and then the animator gets rid of the animation:

https://github.com/Textualize/textual/blob/16656e62e47943e65d06bdb1737decdaf4ad59f3/src/textual/_animator.py#L528-L532

src/textual/css/scalar_animation.py
```diff
@@ -4,6 +4,7 @@
 
 from .._animator import Animation, EasingFunction
 from .._types import CallbackType
+from ..constants import AnimationLevel
```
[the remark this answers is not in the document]
rodrigogiraoserrao: Good call. I'll move it to `_types`, though, because that's where we have those types.
`types.py` is just to expose things to the documentation.

src/textual/widgets/_button.py
rodrigogiraoserrao: 👌

src/textual/widgets/_loading_indicator.py
rodrigogiraoserrao: Thanks 🥲

rodrigogiraoserrao:
> A few suggestions

@willmcgugan addressed 👌

willmcgugan (approved):

PR #4064, opened 2024-01-23
Application suspension

davep:
# Introduction

This PR adds two related features to Textual applications:

- An `App.suspend` context manager that allows the dev to temporarily stop application mode to run some other code; a classic example would be shelling out to an external editor.
- An action (`App.action_suspend_process`) that, where appropriate (currently GNU/Linux and macOS, perhaps other Unix and Unix-like environments), suspends the application as the foreground task.

## App.suspend

Here's a simple example of using the `App.suspend` context manager:

```python
from os import system
from pathlib import Path
from tempfile import NamedTemporaryFile

from textual import on
from textual.widgets import Button, Static

class EditInVimApp(App[None]):

    text: var[str] = var("Hello, World!", init=False)

        yield Button("Edit in vim")
        yield Static(self.text)

    def edit_in_vim(self) -> str:
        with NamedTemporaryFile(delete=False) as file_to_edit:
            file_to_edit.write(self.text.encode())
        try:
            with self.suspend():
                system(f"vim {file_to_edit.name}")
                return Path(file_to_edit.name).read_text()
        finally:
            Path(file_to_edit.name).unlink()

    @on(Button.Pressed)
    def edit_text(self) -> None:
        self.text = self.edit_in_vim()

    def watch_text(self) -> None:
        self.query_one(Static).update(self.text)

if __name__ == "__main__":
    EditInVimApp().run()
```

https://github.com/Textualize/textual/assets/28237/975bbb65-c65f-40f7-b571-781e15da452c

## Suspend as foreground task

This PR also adds support for suspending the application as the foreground task. This is done by providing `App.action_suspend_process` as an action that can be bound to a key combination, the conventional choice being <kbd>Ctrl</kbd>+<kbd>Z</kbd>. This feature is only available on Unix and Unix-like operating systems (currently tested on GNU/Linux and macOS). On Windows it is a no-op. No matter the host operating system it is also a no-op when running under Textual Web.

https://github.com/Textualize/textual/assets/28237/e7573697-763f-428a-947e-bd0fe770ab8d

## Suspend and resume signals

Because the developer may with to perform some actions as a suspend happens, or when a resume happens, this PR also makes use of the new `Signal` facility. The developer can subscribe to `App.app_suspend_signal` and `App.app_resume_signal` to be notified when a suspend or a resume happens.

Using this, an example app with code like this in it (to log when a suspend and a resume happens):

```python
    def suspending(self) -> None:
        self.query_one(Log).write_line("Suspending!")

    def resuming(self) -> None:
        self.query_one(Log).write_line("Resuming!")

    def on_mount(self) -> None:
        self.app_suspend_signal.subscribe(self, self.suspending)
        self.app_resume_signal.subscribe(self, self.resuming)
```

will log the suspends and the resumes:

https://github.com/Textualize/textual/assets/28237/4b05dfd8-8e01-4e9d-b22f-0e24784abacc

## App.suspend and SuspendNotSupported

If `App.suspend` is called in an environment that isn't supported a `SuspendNotSupported` exception will be raised. Using code like this:

```python
    @on(Button.Pressed, "#countdown")
    def countdown(self) -> None:
        try:
            with self.app.suspend():
                for n in reversed(range(10)):
                    print(n)
                    sleep(1)
        except SuspendNotSupported:
            self.notify("I'm sorry Dave, I'm afraid I can't do that", severity="error")
```

in the test application above, and then run under Textual Web, will show the above notification:

https://github.com/Textualize/textual/assets/28237/0f83d139-2da0-4453-876f-36f531b3bc46


src/textual/app.py
```diff
@@ -573,6 +594,24 @@ def __init__(
...
+        self.app_suspend_signal = Signal(self, "app-suspend")
+        """The signal that is published when the app is suspended.
+
+        When [`App.suspend`][textual.app.App.suspend] is called this signal
+        will be [published][textual.signal.Signal.publish];
+        [subscribe][textual.signal.Signal.subscribe] to this signal to
+        perform work before the suspension takes place.
+        """
+        self.app_resume_signal = Signal(self, "app-resume")
+        """The signal that is published when the app is resumes after a suspend.
```
darrenburns, line 606:
```suggestion
        """The signal that is published when the app is resumed after a suspend.
```

src/textual/app.py
```diff
@@ -3287,3 +3326,85 @@ def action_command_palette(self) -> None:
         """Show the Textual command palette."""
         if self.use_command_palette and not CommandPalette.is_open(self):
             self.push_screen(CommandPalette(), callback=self.call_next)
+
+    def _suspend_signal(self) -> None:
+        """Signal that the application is being suspended."""
+        self.app_suspend_signal.publish()
+
+    @on(Driver.SignalResume)
```
darrenburns, line 3334: I found the name `SignalResume` confusing here, as I thought this event was directly related to the signalling system. Would `Driver.Resume` suffice?

davep: The thing is, it's not a message to indicate that the driver has resumed or that it's been asked to resume; it's a message that is asking the app to publish that (Textual) signal.

darrenburns: Shouldn't messages be communicating "a thing happened", while the "what to do with them" side of things is left up to the consumer? When we prescribe an effect in an event/message name it feels like we're enforcing that all consumers must do a certain thing in response, narrowing our options for the future should we ever want to do additional things in response to something.

And if we do want to make it such that only a single behaviour should happen in response to this, perhaps it could be a method instead of an event (events to me imply the potential for multiple consumers).

davep: I'll have to dig a bit deeper then, as it wasn't clear to me how to safely call from the driver into the app. What would you suggest as the safe method of performing that?

darrenburns: I don't think I'm expressing myself well. I'm trying to understand "why are we soft-enforcing that the only thing we can do in response to an event which says 'the driver has resumed' is publish a signal"? I.e. why is the event named specifically instead of generically to leave open other options for doing things in response to it? Are you intentionally trying to discourage that?

To be clear - this is a genuine question and not a change request. I'm trying to understand this more deeply.

davep: Sure, the question makes sense; I'm just trying to work out what the better way is of having the driver call back into the app, to get it to perform a very specific act, without tightly-coupling the driver to the app, and also in such a way as to make sure the act is performed once everything is up and running again.

The issue is that I need `LinuxDriver.start_application_mode` to call [`App._resume_signal`](https://github.com/davep/textual/blob/3a91b3711d5d169e8c06c65fa718a7a6812f4100/src/textual/app.py#L3335-L3337) (so that any code subscribed to [`App.app_resume_signal`](https://github.com/davep/textual/blob/3a91b3711d5d169e8c06c65fa718a7a6812f4100/src/textual/app.py#L605-L613) gets the broadcast at the right moment) under a particular circumstance, but I also don't want to build knowledge about the internals of `App` into `LinuxDriver`. This particular message/event serves no other purpose than to tell `App` that it should broadcast this particular signal.

So what I'm asking is: if not this way, what way would be cleaner?

darrenburns: I don't think there's anything wrong with the current the approach at all. I don't find it "unclean". I worry that you think I have much more context than I do, and therefore my genuine clueless questions may be coming across as "you should've done it like this - why didn't you?!".

I'm coming at this with very little context and trying to understand it by asking questions. My context is standup, where I heard you talk about how the SIGABC needs to happen before the SIGBANANA but after the SIGXYZ or your file system will fill up with directories called "zzzcrazybananas", or something like that.

The thing about "perhaps it should be a method" was intended to be a question than a suggestion - but I missed the question mark and unfortunately I can't type with a rising intonation 🤦‍♂️

The only thing I was (poorly) asking/probing about was why the event is named Driver.SignalResume and not Driver.Resume - as it seems to limit how the event _could_ be used in the future and seemed a bit different (much more prescriptive) to how we usually name them.

So that led me to wonder if this event will ever be used for other things - and lead us to want to use the more generic name in the future. For example, would we want to handle the driver resume event and e.g. reload Textual (official) configuration or CSS in the future? Or if we want to add other side effects on "driver resume" in the future inside Textual, should that be done via subscribing to the signal?

davep: I don't anticipate there being any other uses for the message; it should very much be considered an internal. I suspect you're right, anything that may wish to react to the resume should probably subscribe to the Signal.

davep: Also I wish there was a SIGBANANA.

willmcgugan (changes requested):
Great. Just some suggestions, mostly doc related.

src/textual/app.py
```diff
@@ -3287,3 +3326,85 @@ def action_command_palette(self) -> None:
...
+            Unix-like operating systems and Microsoft Windows. Suspending is
+            not supported in Textual Web.
+        """
+        if self._driver is None:
+            return
+        if self._driver.can_suspend:
+            # Publish a suspend signal *before* we stop application mode and
+            # close the driver.
+            self._suspend_signal()
+            self._driver.stop_application_mode()
```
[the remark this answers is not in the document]
davep: Good call; done.

docs/guide/app.md
```diff
@@ -240,6 +240,66 @@ if __name__ == "__main__"
...
+    - Windows
+
+    It is currently not available when an application is being served via Textual Web.
+
+### Suspending in code
+
+To suspend your application use the [App.suspend](/api/app/#textual.app.App.suspend) context manager.
+For example, here is an application that has a button that will open an external editor, suspending the application before running it:
+
+```python hl_lines="14-15"
```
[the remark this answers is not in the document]
davep: Done.

docs/guide/app.md
```diff
@@ -240,6 +240,66 @@ if __name__ == "__main__"
...
+1. All code in the body of the `with` statement will be run while the app is suspended.
+
+### Suspending from foreground
+
+On Unix and Unix-like systems (GNU/Linux, macOS, etc) Textual has support for the user pressing a key combination to suspend the application as the foreground process.
+Ordinarily this key combination is <kbd>Ctrl</kbd>+<kbd>Z</kbd>;
+in a Textual application this is disabled by default, but an action is provided ([`action_suspend_process`](/api/app/#textual.app.App.action_suspend_process)) that you can bind in the usual way.
+For example:
+
+```python
```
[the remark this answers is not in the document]
davep: Done.

davep:
@willmcgugan Changes made and doc-tweaks merged.

willmcgugan (approved):
Nice

PR #4066, opened 2024-01-24
fix(app): change render return type to RenderResult

### LINKED ISSUE #4063: Tidy the return type of `App.render`
Currently the return type of `App.render` is `RenderableType`; in our docs and in most places we talk about `App.render` returning a `RenderResult`. `RenderResult` is an alias for `RenderableType` and is a top-level value in `textual.app`.

This can cause a lint/type warning in dev's apps if they let their IDE complete the signature when writing their own `render` method, as it will often complete to something like this:

```python
        return super().render()
```

which in turn will auto-import `RenderableType` from `textual.app`. This can then result in a warning like this:

Screenshot 2024-01-23 at 14 16 46

We should change `App.render` so that it's typed as returning `RenderResult` rather than `RenderableType`, then such completion will use the type that *is* exported from `textual.app`.

Also note that [the code in the compose/render HOWTO seems to have fallen foul of this](https://textual.textualize.io/how-to/render-and-compose/) and has that type warning.

TomJGooding:
Fixes #4063 

- [ ] Updated documentation

### DIFF
--- a/docs/blog/posts/spinners-and-pbs-in-textual.md
+++ b/docs/blog/posts/spinners-and-pbs-in-textual.md
@@ -264,10 +264,10 @@ When writing the code for this app, I realised both widgets had a lot of shared 
 That led to the code shown below (more or less) where I implemented the updating functionality in `IntervalUpdater` and then let the `IndeterminateProgressBar` and `SpinnerWidget` instantiate the correct Rich renderable.
 
 ```py hl_lines="8-15 22 30"
+from rich.console import RenderableType
 from rich.progress import Progress, BarColumn
 from rich.spinner import Spinner
 
-from textual.app import RenderableType
 from textual.widgets import Button, Static
 
 
--- a/docs/examples/how-to/render_compose.py
+++ b/docs/examples/how-to/render_compose.py
@@ -2,7 +2,7 @@ from time import time
 
 from rich.console import RenderableType
 
-from textual.app import App, ComposeResult, RenderableType
+from textual.app import App, ComposeResult
 from textual.containers import Container
 from textual.renderables.gradient import LinearGradient
 from textual.widgets import Static
--- a/docs/examples/how-to/render_compose.py
+++ b/docs/examples/how-to/render_compose.py
@@ -1,5 +1,7 @@
 from time import time
 
+from rich.console import RenderableType
+
 from textual.app import App, ComposeResult, RenderableType
 from textual.containers import Container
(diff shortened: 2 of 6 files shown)


davep:
Thanks! I hadn't dived into the detail myself as I only recorded the issue when noticing it in passing while writing something else, but I suspect there's a few other widgets and places to consider:

```sh
$ rg -l "def render.*RenderableType"
src/textual/scrollbar.py
examples/splash.py
src/textual/screen.py
src/textual/command.py
src/textual/widgets/_toast.py
src/textual/widget.py
src/textual/widgets/_input.py
src/textual/scroll_view.py
src/textual/app.py
src/textual/demo.py
src/textual/widgets/_footer.py
src/textual/widgets/_static.py
src/textual/widgets/_digits.py
src/textual/widgets/_loading_indicator.py
src/textual/widgets/_button.py
src/textual/widgets/_placeholder.py
src/textual/widgets/_switch.py
tests/utilities/render.py
tests/snapshot_tests/snapshot_apps/loading.py
docs/blog/posts/spinners-and-pbs-in-textual.md
docs/examples/how-to/render_compose.py
```

Not suggesting *all* of them need a change, but I would imagine the widgets at least do.

Also, generally, we've tended to not "fix" a blog post when they're personal devblogs; might be worth a double-check with @rodrigogiraoserrao just to be sure he's happy with the retcon.

TomJGooding:
Ah, I think I've misunderstood the issue. Do you meanthe return type of `render` should be `RenderResult`  _everywhere_, not just the App?

davep:
I believe so, yes; I'd just quickly written up the issue as a "found this, should probably go look for this all over the place" thing.

rodrigogiraoserrao:
(In this specific instance it doesn't make sense to update my blog post because I'm showing a “snapshot” of another file at a given point in time.)

TomJGooding:
Thanks @davep for clarifying, sorry my fault for misunderstanding. I was probably a bit over-eager to contribute, I've closed this PR but happy to look at this again if it would help rather than hinder...

Just to note in case someone else picks this up, there's also a `RenderResult` in `rich.console` which might be slightly annoying to avoid conflicts!

davep:
@TomJGooding No fault here and no apology needed; you appeared to be doing what this needed; I was just highlighting that it looked like there were a handful more instances of `render` that had the unhelpful type signature that might also need tidying up (and, of course, flagging up the thing about retconning devlog blog posts).

I'd say carry on if you're wanting to!

PR #4079, opened 2024-01-30
Fix CSS watcher crashing when file becomes unavailable...

### LINKED ISSUE #3996: Editing CSS with e.g. vim can crash an app
Editing the CSS of a running app (with the `--dev` switch) using e.g. `vim` causes frequent crashes when saving the changed CSS.

The code that causes it is: https://github.com/Textualize/textual/blob/4da008d5cb5d4c8485c57969b6c27b1bedeae0fc/src/textual/file_monitor.py#L32-L34

Where the `os.stat` call can raise a `FileNotFoundError` (I assume that `vim` stores a temporary version of the edited file, removes the original and renames the temporary to the original file name, meaning that the CSS file doesn't exist for a short time window).

I've added a temporary workaround which loops/sleeps a couple of times, but that's obviously not a real solution...I'm not 100% sure what the right solution would be...

darrenburns:
... and ensure it still works when the file comes back.

Fixes #3996.

### DIFF
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -1477,7 +1477,7 @@ class App(Generic[ReturnType], DOMNode):
                     #  we'll end up with partial CSS, which is probably confusing more than anything. We opt to do
                     #  nothing here, knowing that we'll retry again very soon, on the next file monitor invocation.
                     #  Related issue: https://github.com/Textualize/textual/issues/3996
-                    self.log.debug(str(error))
+                    self.log.warning(str(error))
                     return
                 stylesheet.parse()
                 elapsed = (perf_counter() - time) * 1000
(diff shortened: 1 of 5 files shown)


rodrigogiraoserrao (approved):
🐍 🚀

src/textual/app.py
```diff
@@ -1469,7 +1470,15 @@ async def _on_css_change(self) -> None:
...
                 stylesheet = self.stylesheet.copy()
-                stylesheet.read_all(css_paths)
+                try:
+                    stylesheet.read_all(css_paths)
+                except StylesheetError as error:
+                    # If one of the CSS paths is no longer available (or perhaps temporarily unavailable),
+                    #  we'll end up with partial CSS, which is probably confusing more than anything. We opt to do
+                    #  nothing here, knowing that we'll retry again very soon, on the next file monitor invocation.
+                    #  Related issue: https://github.com/Textualize/textual/issues/3996
+                    self.log.debug(str(error))
```
willmcgugan, line 1480: Could we add a friendlier message here? I think warning may be appropriate.

```python
self.log.warning(f"Unable to load new CSS; {error}")
```

willmcgugan (approved):
Just a wee request.

PR #4080, opened 2024-01-30
docs(faq): add issues with IDE terminals question

### LINKED ISSUE #4016: Add FAQ/docs about issues with IDE terminals
The issue of Textual apps not working as expected in embedded IDE terminals (or Jupyter) seems to come up quite often, most recently #3995 and #4014. Perhaps this should at least be added as a FAQ, or the docs should be more explicit that Textual is designed to work in a proper terminal emulator.

TomJGooding:
Resolves #4016

This FAQ might not get many "hits" as unfortunately people won't mention they're running the app in their IDE terminal, but this would at least provide a resource to link to. I'd welcome any feedback as this was really just a quick draft.

I completely understand the "runs anywhere" message in the docs, but I still wonder if perhaps it is worth adding some caveats.

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -321,6 +321,19 @@ There is currently a light and dark version of the design system, but more are p
 
 ---
 
+<a name="why-doesn't-textual-work-as-expected-in-the-vscode/pycharm/other-ide-terminal"></a>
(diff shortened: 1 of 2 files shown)


TomJGooding:
Closed as stale.

darrenburns:
Yeah, it's not super frequent any more.

I'm not sure why it trailed off.

PR #4081, opened 2024-01-31
Cancelling scrolling animations on new scroll_to calls

darrenburns:
Ensures that previous scrolling animations are cancelled when we call scroll_to.


darrenburns:
@willmcgugan I've updated the approach based on our chat earlier.

willmcgugan (approved):
LGTM

willmcgugan (approved):

PR #4093, opened 2024-02-01
docs(text area): fix code editor link

TomJGooding:
image


TomJGooding:
Sorry I also just spotted this, which perhaps now doesn't make sense now that TextArea is plain by default?

image

darrenburns (approved):

PR #4105, opened 2024-02-02
Only perform the SIGTOU check in the Linux driver when hooked up to a tty

### LINKED ISSUE #4104: Regression in Textual 0.48 when executing app in a subprocess
In the memray project we have several tests that execute textual in a subprocess and drives them by passing input. For example:

https://github.com/bloomberg/memray/blob/cc5463a4c4fca83b53a48331501e1753a78e39ca/tests/integration/test_main.py#L853-L874

Since yesterday's release of Textual 0.48 all of these test hang and our test suite is frozen. It looks like the app ignores the input that we are passing via `input="q"`. I can confirm that all these test work with textual<0.48.

Unfortunately this is blocking memray's CI right now

davep:
Fixes #4104.

Tagging @pablogsal for review (and thanks for the report/test of the issue).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+### Fixed
+
+- Fixed a hang in the Linux driver when connected to a pipe https://github.com/Textualize/textual/issues/4104
+
 ### Changed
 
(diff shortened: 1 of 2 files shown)


pablogsal (commented):
LGTM

davep:
@willmcgugan if we do manage a point release this weekend; might be worth rolling this as well as #4103 into it.

willmcgugan:
Thanks @davep @pablogsal Point release coming soon.

PR #4120, opened 2024-02-05
Calculate `ProgressBar` ETA from most recent updates rather than for whole period

### LINKED ISSUE #4096: `ProgressBar` ETA goes adrift if a progress bar is "reused"
Given this code:

```python
from asyncio import sleep

from textual import on, work
from textual.widgets import Button, ProgressBar

class ReuseProgressApp(App[None]):

        yield Button("Count 10 seconds")
        yield ProgressBar()

    @on(Button.Pressed)
    def start_count(self) -> None:
        progress = self.query_one(ProgressBar)
        progress.total = 100
        progress.progress = 0
        self.count()

    @work
    async def count(self) -> None:
        for _ in range(100):
            self.query_one(ProgressBar).advance()
            await sleep(0.1)

if __name__ == "__main__":
    ReuseProgressApp().run()
```

the first time it's run, the estimated time left is about right:

https://github.com/Textualize/textual/assets/28237/03027456-eee4-42c2-9c9a-2421d4c45de9

However, once finished, if the button is pressed again, the estimated time starts much higher, initially showing around 20 minutes.

https://github.com/Textualize/textual/assets/28237/fa5b1d01-627f-4773-9ab8-d4381ba43c28

Used again, it starts higher still, initially showing over an hour:

https://github.com/Textualize/textual/assets/28237/8bdc9f37-62d9-4e01-b34c-56972f62ada7

This is very heavily related to #4054 and will be fixed as part of the work on that; but I'm recording it as an issue in its own right in case anyone runs into this in the mean time. I suspect this will be a good catalyst for reworking the "ETA" calculation from scratch.

I also suspect that `ProgressBar` should probably acquire some form of "reset" method so it can be put back in an initial state.

davep:
There are two main changes in this PR:

- The method used to calculate the ETA now looks at the most recent updates to the percentage and works off how long they took (underlying this is a class that helps estimate a time to completion based on a timed window, capped at a maximum number of samples to constrain memory usage, and with a final fallback position should the window become near-exhausted).
- Added a `reset_eta` parameter to `ProgressBar.update` so that the dev can signal that the `ProgressBar` is being reused in some way so that the ETA calculation will always start fresh (fixes #4096)

The PR also includes a fix to what seemed to be a bug with how `ETAStatus` stopped refreshing itself; in that before now it looks like it never did (there looks to have been some confusion over using its own timer, but trying to stop the updates via `auto_refresh`).


tests/test_etc.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+def test_window_got_very_narrow() -> None:
+    """If the window got very narrow a best-estimate should still happen."""
+    time_to_completion = TimeToCompletion(100, time_window_size=5)
+    for n in range(5):
+        time_to_completion.record(n, n)
+    time_to_completion.record(50, 50)
+    assert len(time_to_completion) == 1
+    assert time_to_completion.estimated_time_to_complete == 50
```
darrenburns, line 64: Am I understanding correctly that this will use the sample recorded as `(50, 50)` _**and**_ the sample `(4, 4)` since that was the "last dropped sample"?

darrenburns (commented):
Asked a few questions.

Also, not enough walruses.

tests/test_etc.py
```diff
@@ -0,0 +1,64 @@
...
+        assert len(time_to_completion) == min(n, 10)
+        time_to_completion.record(n, n)
+
+
+def test_no_go_backwards() -> None:
+    """It should not be possible to go backwards in time."""
+    time_to_completion = TimeToCompletion(10)
+    time_to_completion.record(2)
+    with pytest.raises(ValueError):
+        time_to_completion.record(1)
```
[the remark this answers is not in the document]
davep: Good point, the docstring for the test is badly-worded; although now you mention it `TimeToCompletion` probably *should* check if we're going backwards in time too.

tests/test_etc.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+def test_no_go_past_end() -> None:
+    """It should not be possible to go past the destination value."""
+    with pytest.raises(ValueError):
+        TimeToCompletion(1).record(2)
```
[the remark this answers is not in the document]
davep: I could be convinced by that, although I wonder if given the lower-level nature of this code, it should be up to the caller to clamp its values?

tests/test_etc.py
```diff
@@ -0,0 +1,64 @@
...
+    """Test the time to completion calculation."""
+    time_to_completion = TimeToCompletion(100)
+    for n in range(0, 10, 2):
+        time_to_completion.record(n, n)
+    assert time_to_completion.estimated_time_to_complete == 92
+
+
+def test_estimate_no_samples() -> None:
+    """Time to completion should be 0 of no samples exist."""
+    assert TimeToCompletion(100).estimated_time_to_complete == 0
```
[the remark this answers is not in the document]
davep: I'm not sure I follow what "renders" means here; could you expand?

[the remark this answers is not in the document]
darrenburns: Renders, as in, how it appears to the end user, in their terminal window.

It won't say "ETA 0s"?

src/textual/_time_to_completion.py
```diff
@@ -0,0 +1,301 @@
...
+from __future__ import annotations
+
+from dataclasses import dataclass
+from time import monotonic
+
+from rich.repr import Result
+from typing_extensions import Self
+
+
+@dataclass
```
willmcgugan, line 12: Can this be a NamedTuple? As a tuple it would be a little more memory efficient and hashable by default.

src/textual/_time_to_completion.py
```diff
@@ -0,0 +1,301 @@
...
+
+    def _prune(self) -> Self:
+        """Prune the samples.
+
+        Returns:
+            Self.
+        """
+        if samples := self._samples:
+            # Trim off any "too old" samples.
+            oldest_time = samples[-1].moment - self._time_window_size
```
willmcgugan, line 58: Should the pruning really occur from the last sample, or should it occur from the current time?

src/textual/_time_to_completion.py
```diff
@@ -0,0 +1,301 @@
...
+        """Add a sample to the samples.
+
+        Args:
+            sample: The sample to add.
+
+        Returns:
+            Self.
+        """
+        self._samples.append(sample)
+        return self._prune()
```
willmcgugan, line 80: Do we need to call prune on every sample?

Could we defer pruning to when we need the samples? i.e. if there are 1000 samples between refreshes, could we prune just once?

tests/test_etc.py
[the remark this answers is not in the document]
davep: Ahh, gotcha; that's up to the `ProgressBar` or other body of code using this.

src/textual/_time_to_completion.py
```diff
@@ -0,0 +1,301 @@
...
+    @property
+    def _speed(self) -> float:
+        """The speed based on the recorded samples."""
+        try:
+            return self._elapsed / self._distance_covered_in_window
+        except ZeroDivisionError:
+            return self._elapsed
+
+    @property
+    def _speed_now(self) -> float:
```
willmcgugan, line 260: I don't follow the need for a "speed" and a "speed now".

willmcgugan (changes requested):
I think there may be too much emphasis on keeping samples.

We only need to discard old samples when we calculate the speed.

src/textual/_time_to_completion.py
davep: Come to think of it, can't think of a reason not to make it so.

src/textual/_time_to_completion.py
davep: It makes sense to me that it works off the last-known sample rather than off "now", as what we're making here is a heuristic for figuring out a likely time to completion, and going off "last best" makes more sense to me than going off nothing.

src/textual/_time_to_completion.py
davep: This could go either way; there could be 1,000 "refreshes" between samples too. It feels to me like, on balance, code that is showing an estimated time is more likely to get new estimates than it is to record new samples.

davep:
Seems we're going to go with a very totally different PR, so closing this unreviewed.

PR #4123, opened 2024-02-05
Directory tree reload now preserves state

### LINKED ISSUE #4056: Reload directory tree and keep state
We need a way of reloading the directory tree on demand, so that it reflects any changes to the filesystem.

- The reload should be atomic. i.e. it shouldn't permit modifications to the tree from another task. Otherwise it could get out-of-sync. Suggest we need a asyncio.Lock here.
- The highlighted node should stay highlighted, assuming the node still exists. If the node has been removed, then the highlight should go up a directory.
- Open directories should be reloaded recursively.
- The loading directory information (including subdirectories) should happen in a thread, so as not block the main loop. Note that thins should probably be done in a single thread.

Suggested implementation:

- In a thread (or threaded worker), recursively load the directories which have been expanded (assuming they exist).
- Note the highlighted node, and the open nodes.
- Update the DirectoryTree nodes to reflect the new data.
- Restore the highlighted node.

Note than since we want this to be atomic, this should be done in a coroutine, or at least return an awaitable.

Supersedes https://github.com/Textualize/textual/issues/3574

### LINKED ISSUE #4122: `DirectoryTree.clear_node` doesn't actually clear the node
The current implementation of `DirectoryTree.clear_node` assigns to `node` but then does nothing with it, which means it doesn't really clear the node passed in:

https://github.com/Textualize/textual/blob/ca2c11bdb86d7f48acadee534a1764777b444d34/src/textual/widgets/_directory_tree.py#L188-L208

rodrigogiraoserrao:
Fixes #4056.

Adds an `asyncio.Lock` to `Tree`.
Adds a private coroutine `DirectoryTree._reload` used by `DirectoryTree.reload`/`.reload_node` that reloads a subtree and preserves expanded nodes + highlighted node.

Fixes #4122.
While working on the above I noticed this bug and I fixed it en passant.

### DIFF
--- a/src/textual/widgets/_directory_tree.py
+++ b/src/textual/widgets/_directory_tree.py
@@ -192,17 +192,7 @@ class DirectoryTree(Tree[DirEntry]):
             The `Tree` instance.
         """
         self._clear_line_cache()
-        node_label = node._label
-        node_data = node.data
-        node_parent = node.parent
-        node = TreeNode(
-            self,
-            node_parent,
-            self._new_id(),
-            node_label,
-            node_data,
-            expanded=True,
-        )
+        node.remove_children()
         self._updates += 1
         self.refresh()
         return self
--- a/src/textual/widgets/_directory_tree.py
+++ b/src/textual/widgets/_directory_tree.py
@@ -5,19 +5,18 @@ from dataclasses import dataclass
 from pathlib import Path
 from typing import TYPE_CHECKING, Callable, ClassVar, Iterable, Iterator
 
-from ..await_complete import AwaitComplete
-
-if TYPE_CHECKING:
-    from typing_extensions import Self
-
 from rich.style import Style
 from rich.text import Text, TextType
 
 from .. import work
+from ..await_complete import AwaitComplete
 from ..message import Message
 from ..reactive import var
 from ..worker import Worker, WorkerCancelled, WorkerFailed, get_current_worker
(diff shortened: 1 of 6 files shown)


src/textual/widgets/_directory_tree.py
```diff
@@ -225,6 +216,88 @@ def reset_node(
...
+            # Reopen nodes that were expanded and still exist.
+            to_reopen = [node]
+            while to_reopen:
+                reopening = to_reopen.pop()
+                if not reopening.data:
+                    continue
+                if (
+                    reopening.allow_expand
+                    and (reopening.data.path in currently_open or reopening == node)
+                    and reopening.data.path.exists()
```
willmcgugan, line 262: `path.exists()` is a blocking operation. i.e. it may have to go to the disk to check if the path is on the filesystem.

Perhaps this could be factored in to the `_load_directory` functionality? By throwing an exception if it doesn't exist?

willmcgugan (changes requested):
Great work. Just one thing to consider.

src/textual/widgets/_directory_tree.py
rodrigogiraoserrao: Got it. See bffa84f.

willmcgugan (approved):

JoeZiminski:
@rodrigogiraoserrao @willmcgugan That's awesome, thanks for this!

willmcgugan:
All credit to Rodrigo.

PR #4125, opened 2024-02-05
fix(text area)!: stop escape shifting focus if default tab behaviour

### LINKED ISSUE #4110: `TextArea` still uses `Escape` to move focus on
While "out of the box" `TextArea` now uses <kbd>Tab</kbd> to shift focus, it also still uses <kbd>Escape</kbd>, ideally it would not when in this mode. Pressing <kbd>Escape</kbd> to exit a modal dialog would be common, so having <kbd>Escape</kbd> do something different given one particular widget type would be surprising for the user.

For example, this dialog uses <kbd>Escape</kbd> as the quick exit key, but instead if moves focus if you happen to press it while in the `TextArea`:

Screenshot 2024-02-02 at 21 18 15

TomJGooding:
Fixes #4110.

- [x] Updated documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,12 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
-## [0.48.2] - 2024-02-02
-
-### Fixed
-
-- Fixed a hang in the Linux driver when connected to a pipe https://github.com/Textualize/textual/issues/4104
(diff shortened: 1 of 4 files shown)


darrenburns (approved):
One small suggestion, but looks good. Thanks!

darrenburns:
Also, for the CHANGELOG I think it'll go into the "Fixed" section, and noted as a breaking change.

TomJGooding:
Done - I struggling with wording the changelog entry so hope that is okay!

(Also fixed where it looks like the changelog was out-of-sync after latest update)

darrenburns:
Thanks!

PR #4126, opened 2024-02-06
Fix a TextArea crash

darrenburns:
The TextArea widget could crash if a replacement operation occurred while a backward selection was made, and the inserted content contained fewer lines than the content being replaced.

I've fixed this and simplified some operations which don't care about the "direction" of the selection (i.e. did you click and drag from top to bottom or bottom to top with your mouse to select text).

Also fixes an issue where the selection wasn't being correctly updated after a paste occurred.

### DIFF
--- a/src/textual/widgets/_text_area.py
+++ b/src/textual/widgets/_text_area.py
@@ -2048,10 +2048,12 @@ class Edit:
 
     @property
     def top(self) -> Location:
+        """The Location impacted by this edit that is nearest the start of the document."""
         return min([self.from_location, self.to_location])
 
     @property
     def bottom(self) -> Location:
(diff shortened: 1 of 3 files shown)


rodrigogiraoserrao (approved):
LGTM 🐍 🚀

src/textual/widgets/_text_area.py
```diff
@@ -2048,6 +2046,14 @@ def after(self, text_area: TextArea) -> None:
             text_area.selection = self._updated_selection
         text_area.record_cursor_width()
 
+    @property
+    def top(self) -> Location:
```
willmcgugan, line 2050: Docstrings por favor

willmcgugan (approved):
Just a request for docstrings.

PR #4139, opened 2024-02-08
Adds `Widget.lock` attribute.

### LINKED ISSUE #4134: Add a `lock` to every widget.
We should add an `asyncio.Lock` to every widget, and a helpful docstring along with it.

It is possible for two tasks to call a single method, which can result in a race condition. This is easily fixable by adding `async with self.lock:`

Note that this is not an issue if methods on a widget are called from its own message handlers.

I'd be surprised if this task takes more than an hour!

rodrigogiraoserrao:
Adds an `asyncio.Lock` to every widget.

Fixes #4134.


willmcgugan (changes requested):
Thanks. Don't merge just yet.

willmcgugan (approved):

PR #4149, opened 2024-02-11
Add default syntax mapping to CSS theme in TextArea

darrenburns:
Copies the syntax highlighting from the dark VSCode theme in to the default CSS theme.

So now, if you set the `language` parameter on the plain `TextArea` (not using `code_editor`), you'll get some highlighting:

image

Previously, since there was no syntax highlighting style mapping, tree-sitter was parsing the file but there was no visual indication of that.

This was created with `TextArea(text=TEXT, language="markdown")` (and line numbers enabled).

### DIFF
--- a/src/textual/widgets/_text_area.py
+++ b/src/textual/widgets/_text_area.py
@@ -983,21 +983,6 @@ TextArea:light .text-area--cursor {
         selection_top_row, selection_top_column = selection_top
         selection_bottom_row, selection_bottom_column = selection_bottom
 
-        highlights = self._highlights
-        if highlights and theme:
-            line_bytes = _utf8_encode(line_string)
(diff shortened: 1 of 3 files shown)


src/textual/widgets/_text_area.py
```diff
@@ -1032,6 +1017,21 @@ def render_line(self, widget_y: int) -> Strip:
                         else:
                             line.stylize(selection_style, end=line_character_count)
 
+        highlights = self._highlights
```
darrenburns, line 1020: Just reordering the style sandwich ingredients here, no actual change to the code that is inside the diff.

rodrigogiraoserrao (approved):
LGTM 🐍🚀

Issue #4145, opened 2024-02-11
Clarify `is_end_of_document_line` in docs


TomJGooding:
The documentation for the `DocumentNavigator` says that a location "refers to a location, in document-space (in the raw document)".

I would expect the "end of line" in the example below to be at index 9. But the `is_end_of_document_line` is actually based on the _length_ of the row, meaning it is off-by-one.

Perhaps what is meant by "end of line" just needs clarifying the docs, or this might be an unintended bug?

```python
from textual.app import App, ComposeResult
from textual.widgets import TextArea

TEXT = "0123456789"


class ExampleApp(App):
    def compose(self) -> ComposeResult:
        yield TextArea.code_editor(TEXT)

    def on_mount(self) -> None:
        text_area = self.query_one(TextArea)
        expected_eol = (0, 9)
        assert text_area.navigator.is_end_of_document_line(expected_eol)


if __name__ == "__main__":
    app = ExampleApp()
    app.run()
```

TomJGooding:
Sorry I forgot to mention: I think the docstring might need correcting anyway here:

https://github.com/Textualize/textual/blob/e1749ac36a88ae7a015a5c0fc16923df9b6ee3a9/src/textual/document/_document_navigator.py#L107-L108

darrenburns:
It's not off by one - I should maybe clarify in the docs though. The end of the line is the maximum position in a line the cursor can rest on. Since you need to be able to move the cursor **_beyond_** the end of the text (otherwise inserting new text would be impossible), the "end of document line" is `len(document_line)`. It's not "index of the last character".

TomJGooding:
Thanks Darren for the explanation, I think clarifying this in the docs is a good idea.

TomJGooding:
Closed as completed by #4151 (thanks Darren!)

PR #4151, opened 2024-02-12
Text area read only

darrenburns:
Adds a read_only attribute to TextArea.

When True, you cannot edit the content.

The bulk of the work here was more clearly defining the paths that edit operations take when they come in via the API vs via a users keyboard.

### DIFF
--- a/src/textual/_text_area_theme.py
+++ b/src/textual/_text_area_theme.py
@@ -100,17 +100,16 @@ class TextAreaTheme:
                 self.gutter_style = self.base_style.copy()
 
         background_color = Color.from_rich_color(self.base_style.bgcolor)
-        if self.cursor_style is None:
-            # If the theme doesn't contain a cursor style, fallback to component styles.
-            cursor_style = get_style("text-area--cursor")
-            if cursor_style:
-                self.cursor_style = cursor_style
-            else:
-                # There's no component style either, fallback to a default.
-                self.cursor_style = Style(
-                    color=background_color.rich_color,
-                    bgcolor=background_color.inverse.rich_color,
-                )
+        # If the theme doesn't contain a cursor style, fallback to component styles.
+        cursor_style = get_style("text-area--cursor")
+        if cursor_style:
+            self.cursor_style = cursor_style
+        else:
+            # There's no component style either, fallback to a default.
+            self.cursor_style = Style(
+                color=background_color.rich_color,
+                bgcolor=background_color.inverse.rich_color,
+            )
 
         # Apply fallbacks for the styles of the active line and active line gutter.
         if self.cursor_line_style is None:
--- a/src/textual/_text_area_theme.py
+++ b/src/textual/_text_area_theme.py
@@ -100,16 +100,17 @@ class TextAreaTheme:
                 self.gutter_style = self.base_style.copy()
 
         background_color = Color.from_rich_color(self.base_style.bgcolor)
-        # If the theme doesn't contain a cursor style, fallback to component styles.
-        cursor_style = get_style("text-area--cursor")
-        if cursor_style:
-            self.cursor_style = cursor_style
-        else:
-            # There's no component style either, fallback to a default.
-            self.cursor_style = Style(
-                color=background_color.rich_color,
-                bgcolor=background_color.inverse.rich_color,
-            )
+        if self.cursor_style is None:
+            # If the theme doesn't contain a cursor style, fallback to component styles.
+            cursor_style = get_style("text-area--cursor")
+            if cursor_style:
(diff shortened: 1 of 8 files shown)


src/textual/_text_area_theme.py
```diff
@@ -100,17 +100,16 @@ def apply_css(self, text_area: TextArea) -> None:
...
-                    color=background_color.rich_color,
-                    bgcolor=background_color.inverse.rich_color,
-                )
+        # If the theme doesn't contain a cursor style, fallback to component styles.
+        cursor_style = get_style("text-area--cursor")
+        if cursor_style:
+            self.cursor_style = cursor_style
+        else:
+            # There's no component style either, fallback to a default.
+            self.cursor_style = Style(
```
willmcgugan, line 109: Style.from_colors is an alternative constructor that's a little faster.

darrenburns: The type signature for that doesn't align - it only supports `Color | None`, not `Color | str | None` like `__init__` does.

willmcgugan: Yes, but you are using Rich `Color` instances here aren't you? Should be a drop-in replacement.

Granted this doesn't look like time critical code.

darrenburns: Ah I see, yes in this particular instance I could swap it. However, it only runs once.

src/textual/document/_document_navigator.py
```diff
@@ -101,6 +101,10 @@ def is_start_of_wrapped_line(self, location: Location) -> bool:
     def is_end_of_document_line(self, location: Location) -> bool:
         """True if the location is at the end of a line in the document.
 
+        Note that the "end" of a line is equal to its length (one greater
+        than the final index), since there is a space at the end of the line
+        for the cursor to rest.
+
```
TomJGooding, line 107: Thanks for taking #4145 on boardhere! Sorry I realise this PR still a draft, but just a reminder that this docstring might also need correcting here:

```
        Returns:
            True if and only if the document is on the last line of the document.
```

darrenburns: Good spot!

tests/text_area/test_edit_via_api.py
rodrigogiraoserrao, line 1: Shouldn't these tests be modified to verify that editing _still_ works if `read_only=True`?

tests/text_area/test_selection_bindings.py
```diff
@@ -13,34 +13,48 @@
...
+    It means every test that uses the app fixture will run twice - once
+    for read_only=True and once for read_only=False.
+    """
+    return request.param
+
 
+@pytest.fixture
+async def app(read_only_mode: bool):
+    text_area_app = TextAreaApp(read_only_mode)
+    yield text_area_app
```
rodrigogiraoserrao, line 36: From the way I'm reading these two fixtures, it looks like you could just parametrise the second one and get rid of the first! 🚀

willmcgugan (approved):

tests/snapshot_tests/test_snapshots.py
```diff
@@ -879,6 +879,20 @@ def setup_selection(pilot):
     )
 
 
+def test_text_area_read_only_cursor_rendering(snap_compare):
+    def setup_selection(pilot):
+        text_area = pilot.app.query_one(TextArea)
+        text_area.theme = "css"
+        text_area.text = "Hello, world!"
+        text_area.read_only = True
```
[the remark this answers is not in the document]
darrenburns: I think we're ok here as I don't think there's anything async. If we see any flakiness we can add it.

tests/text_area/test_edit_via_api.py
darrenburns: You're right - yes. I'll add a new test instead of modifying though. Since the bindings tests have a bunch of different operations it made sense to check read_only with all of them (even found a bug there where ctrl+x to delete a line moved the cursor despite it not deleting the line on read-only mode).

In this case however, there's a pretty small surface area to check, so I'll do it on its own rather than doubling the test count.

tests/text_area/test_selection_bindings.py
darrenburns: Good spot!

tests/text_area/test_edit_via_api.py
rodrigogiraoserrao: Alright, you're the boss!

darrenburns: https://github.com/Textualize/textual/pull/4151/commits/59909a4fe7d4109cf8782f20ed182131c8fb4ae3

Issue #4111, opened 2024-02-02
Show all commands in command palette


chbndrhnns:
When I'd like to explore an app, I would want to see all available commands available in the command palette just by pulling it up, just like VS Code does it.

willmcgugan:
I think this is very good idea.

darrenburns:
Just to make sure I'm understanding correctly - does this mean when the search box is empty, we show everything?

davep:
Just to add context (not really to answer that question), this stems from [this conversation on Discord](https://discord.com/channels/1026214085173461072/1033754296224841768/1203090041803636808), as far as I can tell.

In terms of my own ideas on it: I'm not a huge fan of "show everything by default", but I can see the value in "cursor down when empty shows everything".

willmcgugan:
We might want to have some mechanism to cherry pick the commands we show, just so it doesn't get too large. But I do think they should show when the input is empty.

davep:
Picking this up... defining characteristics of the enhancement request so far:

- All applicable hits start to trickle into the display the moment the command palette opens.
- Some method of deciding what is an applicable hit needs to be decided and added.

My initial feeling is that the bulk of the work here is around deciding and implementing the latter; my gut reaction is that one or more of the following could be approaches (just thinking out loud before diving back into the code):

- It could be left totally up to the providers; if given a empty string to search for, they could decide what to yield up.
- It could be left totally up to the providers; but instead there's a method that's parallel to `Provider.search` that is more like `Provider.default` or something, and this yields up all of the commands the provider wants to provide by default.
- It could be left totally up to the providers; but it's a on/off switch at the provider level that says if it wants to take part in providing default hits.

One question stands out right away:

- What ordering to use when there's no match score to go off.

Seems that alphabetical order would make sense.

darrenburns:
> What ordering to use when there's no match score to go off.

Could we use the order that would be used when the search value is `""`?

davep:
That would mean that every `Hit` has a score of either 0 or 1 (not sure which would make more sense I guess), so the order is probably going to be the speed with which they come in. Not sure that quite makes sense (or rather, it does, but might not be useful)?

willmcgugan:
Do we really want to search? I was thinking each provider would offer up some text that tells the user what it accepts.

Like ("Help", "Show the help screen.") or ("open FILENAME", "Open a file.")

PR #4154, opened 2024-02-13
Command palette command discovery

davep:
PR to add a discovery API to the command palette, which allows command providers to offer up commands that should appear in the command palette's dropdown as soon as it's opened.

Implements #4111.

The request in #4111 was to show everything, but this could be problematic depending on the nature of the provider; so this approach adds a `discover` method, which can be omitted, and the discoverable commands will be taken from that; this gives providers the option to either offer up all their commands, or just a subset that make sense.

This PR also implements `textual._system_commands.SystemCommands.discover`.


docs/guide/command_palette.md
```diff
@@ -99,7 +99,23 @@ In the example above, the callback is a lambda which calls the `open_file` metho
...
+!!! note
+
+    Because `discover` hits are shown the moment the command palette is opened, these should ideally be quick to generate;
+    commands that might take time to generate are best left for `search` -- use `discover` to help the user easily find the most important commands.
+
+`discover` is similar to `search` but with the difference that no search value is passed to it, and it should *yield* instances of [`DiscoveryHit`][textual.command.DiscoveryHit].
+There is no matching and no match score is needed.
+The [`DiscoveryHit`][textual.command.DiscoveryHit] contains information about how the hit should be displayed, and an optional help string;
+discovery hits are sorted in ascending alphabetical order.
+It also contains a callback, which will be run if the user selects that command.
```
rodrigogiraoserrao, line 116: This paragraph doesn't flow very well but this is a subjective thing.

Your paragraph reads a bit like a list that's phrased weird, so maybe you could refactor into something similar to the suggestion below, or maybe just rewrite the prose to flow a little bit better.

```suggestion
`discover` is similar to `search` but with these differences:
 - `discover` accepts no parameters (instead of the search value);
 - `discover` yields instances of [`DiscoveryHit`][textual.command.DiscoveryHit]` (instead of instances of xxx);
 - discovery hits are sorted in ascending alphabetical order because there is no matching and no match score is generated.
Instances of [`DiscoveryHit`][textual.command.DiscoveryHit] contain information about how the hit should be displayed, an optional help string, and a callback which will be run if the user selects that command.
```

(The suggestion above is not to be taken at face value. It's just to illustrate roughly what I had in mind.)

src/textual/command.py
```diff
@@ -105,7 +111,57 @@ def __post_init__(self) -> None:
...
+    display: RenderableType
+    """A string or Rich renderable representation of the hit."""
+
+    command: IgnoreReturnCallbackType
+    """The function to call when the command is chosen."""
+
+    text: str | None = None
+    """The command text associated with the hit, as plain text.
+
+    If `display` is not simple text, this attribute should be provided by
```
rodrigogiraoserrao, line 127: What does “simple text” mean? Are you talking about the `type` of the value `display`, as in `str` versus other renderables?
Or are you talking about the contents of the text, e.g., simple words versus complicated words; short sentences vs long sentences; active voice vs passive voice?
I'm guessing you're talking about the type `str`, but that's not very clear.

src/textual/command.py
```diff
@@ -105,7 +111,57 @@ def __post_init__(self) -> None:
...
+                self.text = self.display
+            elif isinstance(self.display, Text):
+                self.text = self.display.plain
+            else:
+                raise ValueError(
+                    "A value for 'text' is required if 'display' is not a str or Text"
+                )
+
+
+Hits: TypeAlias = AsyncIterator["DiscoveryHit | Hit"]
```
rodrigogiraoserrao, line 164: Shouldn't the union be outside?
Or can it happen that we're yielding both regular hits and discovery hits?

```suggestion
Hits: TypeAlias = AsyncIterator["DiscoveryHit"] | AsyncIterator["Hit"]
```

src/textual/command.py
```diff
@@ -215,6 +274,22 @@ async def search(self, query: str) -> Hits:
...
+        """A default collection of hits for the provider.
+
+        Yields:
+            Instances of [`DiscoveryHit`][textual.command.DiscoveryHit].
+
+        Note:
+            This is different from
+            [`search`][textual.command.Provider.search] in that it should
+            yield [`DiscoveryHit`s][textual.command.DiscoveryHit] that
+            should be shown by default; before user input.
```
rodrigogiraoserrao, line 287: Maybe I just need to reread what the semicolon is used for, but all of the suggestions below read better to me than the current docstring.

```suggestion
            should be shown by default (before user input).
```
```suggestion
            should be shown before user input.
```
```suggestion
            should be shown by default.
```
```suggestion
            should be shown when there is no user input.
```

src/textual/command.py
```diff
@@ -598,23 +674,36 @@ def _stop_no_matches_countdown(self) -> None:
     _NO_MATCHES_COUNTDOWN: Final[float] = 0.5
     """How many seconds to wait before showing 'No matches found'."""
 
-    def _start_no_matches_countdown(self) -> None:
+    def _start_no_matches_countdown(self, search_value: str) -> None:
```
rodrigogiraoserrao, line 677: Isn't it cleaner to just return without scheduling a no-op callback when there is no search value?

src/textual/command.py
```diff
@@ -640,7 +729,7 @@ async def _watch__show_busy(self) -> None:
         self.query_one(CommandList).set_class(self._show_busy, "--populating")
 
     @staticmethod
-    async def _consume(hits: Hits, commands: Queue[Hit]) -> None:
+    async def _consume(hits: Hits, commands: Queue[DiscoveryHit | Hit]) -> None:
```
rodrigogiraoserrao, line 732: Same comment as before about union being inside the generic vs outside.

src/textual/command.py
```diff
@@ -650,7 +739,9 @@ async def _consume(hits: Hits, commands: Queue[Hit]) -> None:
         async for hit in hits:
             await commands.put(hit)
 
-    async def _search_for(self, search_value: str) -> AsyncGenerator[Hit, bool]:
+    async def _search_for(
+        self, search_value: str
+    ) -> AsyncGenerator[DiscoveryHit | Hit, bool]:
```
rodrigogiraoserrao, line 744: Same comment as before about union being inside the generic vs outside.

src/textual/command.py
```diff
@@ -661,7 +752,7 @@ async def _search_for(self, search_value: str) -> AsyncGenerator[Hit, bool]:
         """
 
         # Set up a queue to stream in the command hits from all the providers.
-        commands: Queue[Hit] = Queue()
+        commands: Queue[DiscoveryHit | Hit] = Queue()
```
rodrigogiraoserrao, line 755: Same comment as before about union being inside the generic vs outside.

tests/command_palette/test_discover.py
```diff
@@ -0,0 +1,34 @@
...
+
+class CommandPaletteApp(App[None]):
+    COMMANDS = {SimpleSource}
+
+    def on_mount(self) -> None:
+        self.action_command_palette()
+
+
+async def test_discovery_visible() -> None:
+    """A provider with discovery should cause the command palette to be opened right away."""
```
rodrigogiraoserrao, line 29: I genuinely don't understand what you mean.
Do you mean to say that if there's discovery, results should be shown right away..?

rodrigogiraoserrao (approved):
I left a couple of (subjective) docstring suggestions and a couple of other minor suggestions. None of them warranted I blocked the PR 🤷 

Good job 🐍 🚀

docs/guide/command_palette.md
davep: Yeah, I feel that's probably cleaner; I'll merge that and tidy it up.

src/textual/command.py
davep: This is just a copy of the pre-existing documentation for a `Hit` that Will added back in the day.

src/textual/command.py
davep: Actually that's probably more accurate (although there was a moment while I was working on this where the version I have above was more accurate; but it's settled in a place where this makes more sense now).

src/textual/command.py
davep: I'm struggling to understand this question.

tests/command_palette/test_discover.py
davep: Yes, that's what that is testing for.

src/textual/command.py
davep: Tinkering with that for a moment... doesn't actually work and has a lot of knock-on implications for the typing of the search machinery in the command palette. I think to make this work nicely there'd need to be a fully-parallel discovery pipeline, whereas at the moment I'm making full use of the pre-existing search code. Either that or to make the above work there'd need to be a lot of overloading done, perhaps.

Technically what you say here *is* correct, we should never be building up a mixture of the two.

rodrigogiraoserrao: Ok, understood. If being marginally “more correct” increases the burden on the programmer disproportionally, I won't push for this change.

tests/command_palette/test_discover.py
rodrigogiraoserrao: Ok. I didn't understand the docstring, then.

PR #4159, opened 2024-02-14
Fix active tab not coming into view plus `TabbedContent.TabActivated` not always been posted

### LINKED ISSUE #4150: TabbedContent active attribute doesn't post message TabActivated + doesn't arrange tabs
Hello,

I noticed that when using `active` reactive attribute, there is no message posted to `TabActivated`. I also noticed that if the tab bar is larger than the terminal, it doesn't correctly adjust the tabs to show what's around them as if you clicked on it.

Please see the MRE below to understand the issue:
```python
from textual import on
from textual.app import App
from textual.widgets import Button, TabbedContent, TabPane


class TabApp(App):
    def compose(self):
        with TabbedContent(id="tabbed_content"):
            for tab_id in range(50):
                yield TabPane(f"Tab_{tab_id}", id=f"tab_{tab_id}")

        yield Button("First Tab", id="first_tab")
        yield Button("Last Tab", id="last_tab")

    @on(TabbedContent.TabActivated, "#tabbed_content")
    def tab_changed(self, event: TabbedContent.TabActivated):
        self.notify(f"Tab changed {event.tab.id}")

    @on(Button.Pressed, "#first_tab")
    def on_first_tab(self):
        self.query_one("#tabbed_content", TabbedContent).active = "tab_0"

    @on(Button.Pressed, "#last_tab")
    def on_last_tab(self):
        self.query_one("#tabbed_content", TabbedContent).active = "tab_49"


app = TabApp()
app.run()
```

Thanks!

davep:
This PR fixes the two issues raised in #4150 

- When `Tabs.active` (and by extension `TabbedContent.active`) is assigned the ID of a tab that is outwith of the current view; the tab isn't scrolled into view.
- When `TabbedContent.active` is assigned `TabbedContent.TabActivated` isn't posted.

The PR also contains a couple of minor tweaks to the existing `TabbedContent` code to make type checking happier (just things in the area where I was working).


darrenburns (changes requested):
Could you add a small regression test to ensure that when `.active` is set, the correct message (and only the correct message) is posted?

rodrigogiraoserrao (approved):
Looks good to me.

davep:
> Could you add a small regression test to ensure that when `.active` is set, the correct message (and only the correct message) is posted?

Confusingly it seems there already is one (kinda), and I'm not sure how it ever passed:

https://github.com/Textualize/textual/blob/b13a215372d0b4aef605e1162faa0581615a09ad/tests/test_tabbed_content.py#L138-L163

darrenburns (approved):

PR #4162, opened 2024-02-15
Fix Changelog header version.

rodrigogiraoserrao:
Fixes the version header to match the actual package version.


willmcgugan (approved):
Can you check the release page as well? I may have copied that.

davep (approved):

rodrigogiraoserrao:
@willmcgugan good catch. Fixed there as well.

PR #4172, opened 2024-02-16
fix(text area): update code editor constructor

TomJGooding:
Update the `TextArea.code_editor` convenience constructor to include the new `read_only` and `max_checkpoints` attributes.

- [ ] Updated documentation


darrenburns:
When I added this constructor I thought to myself "I'm definitely gonna forget to update this..." 😂 

I wonder if we could add a sort of "meta" test which would catch variance between the constructors using the `inspect` module and looking at the signatures to detect if their parameters differ.

TomJGooding:
Something like this?

```python
import inspect


text_area_params = inspect.getfullargspec(TextArea.__init__).kwonlyargs
code_editor_params = inspect.getfullargspec(TextArea.code_editor).kwonlyargs

print(text_area_params)
print(code_editor_params)

assert text_area_params == code_editor_params
```

TomJGooding:
I've added a test in 4cb6244 - hopefully I've understood you correctly!

PR #4177, opened 2024-02-18
signals env var

willmcgugan:
Some changes to allow piping data in to a Textual app.

Basically when piping data in to a Textual app you don't want to the app to capture ctrl+C. This adds an env var to prevent that.


willmcgugan:
#executivedecision But let me know if you have any comments.

PR #4178, opened 2024-02-19
Remove double-up scrolling of active tab

### LINKED ISSUE #4169: Tab scrolling not working
Tabs do not scroll as they used to.

This is Textual 0.51.0 (latests at time of writing)

https://github.com/Textualize/textual/assets/554369/5254051a-d8b4-416a-80d9-396e336bf3b1

This is Textual 0.50.1

https://github.com/Textualize/textual/assets/554369/e3a0fd66-206b-47f1-9eb6-4a8d6dc4def9

I'm assuming something was broken with @davep 's recent fix to `active` reactive.

davep:
Follows on from #4159; in that PR we do the scrolling when seeing the active tab (because that could be done from code and it might not be in view and it needs to be dragged into view); this resulted in the unintended consequence of the animation being kicked off twice, presumably causing the previous instance to be forced to finish instantly, thus making it look like it didn't animate at all.

Fixes #4169

(or at least, as I'm testing it, it fixes #4169, but some doubt has be raised about this so this is first offered as a double-check)

In this video, 0.51 tab click is on the left, this PR's effect is on the right.

https://github.com/Textualize/textual/assets/28237/be75de8f-b2d9-4375-b5e2-e4526360ef26

### DIFF
--- a/src/textual/widgets/_tabs.py
+++ b/src/textual/widgets/_tabs.py
@@ -699,7 +699,6 @@ class Tabs(Widget, can_focus=True):
         tab_count = len(tabs)
         new_tab_index = (tabs.index(active_tab) + direction) % tab_count
         self.active = tabs[new_tab_index].id or ""
-        self._scroll_active_tab()
 
     def _on_tab_disabled(self, event: Tab.Disabled) -> None:
         """Re-post the disabled message."""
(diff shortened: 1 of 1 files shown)


willmcgugan (approved):
Thanks. Merge when tests pass please.

PR #4183, opened 2024-02-19
Adds a `Widget.batch` async context manager

### LINKED ISSUE #4133: Add `remove` attribute to `mount` and `mount_all`
It's common to remove some widgets and add some other widgets. To make this easier, I think we should add a `remove` attribute to the mount methods which accepts a selector and removes those widgets prior to mounting.

The remove + mount should be atomic, to avoid flicker (might need to be wrapped in `batch_update`).

API would look something like this:

```python
# Replace all MyWidgets with a single MyWidget
await self.mount(MyWidget("Foo"), remove="MyWidget")

# Remove all children and replace with a label
await self.mount(Label("Goodbye"), remove="*")
```

rodrigogiraoserrao:
The `Widget.batch` context manager locks the widget and batches app updates too.
This fixes #4133.

This PR depends on #4139 and shouldn't be merged while #4139 doesn't get the OK to be merged.

### DIFF
--- a/src/textual/widget.py
+++ b/src/textual/widget.py
@@ -6,11 +6,13 @@ from __future__ import annotations
 
 from asyncio import Lock, create_task, wait
 from collections import Counter
+from contextlib import asynccontextmanager
 from fractions import Fraction
 from itertools import islice
 from types import TracebackType
 from typing import (
     TYPE_CHECKING,
+    AsyncGenerator,
     Awaitable,
     ClassVar,
     Collection,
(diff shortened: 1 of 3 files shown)


src/textual/widget.py
```diff
@@ -3283,15 +3294,42 @@ def remove(self) -> AwaitRemove:
         await_remove = self.app._remove_nodes([self], self.parent)
         return await_remove
 
-    def remove_children(self) -> AwaitRemove:
-        """Remove all children of this Widget from the DOM.
+    def remove_children(
```
willmcgugan, line 3297: This method looks like it applies to children and their descendants? Basically equivalent to `self.query(selector).remove()`, so there doesn't seem much point in this existing.

What we could probably use is a method to remove only direct children, which would better serve the use case of removing and mounting in a batch.

willmcgugan (changes requested):

src/textual/widget.py
rodrigogiraoserrao: Understood. See 808f536.

willmcgugan (approved):

PR #4185, opened 2024-02-19
Markdown reloads when component classes change.

### LINKED ISSUE #3464: Some markdown component classes don't refresh on CSS update
The component classes `s`, `strong`, `em`, and `code_inline` aren't refreshed when CSS is reloaded and those are changed.
However, the component classes `markdown-table--header` and `markdown-table--lines` do refresh.

The video below shows this. Code is included after the video.

https://github.com/Textualize/textual/assets/5621605/d0f3d624-c510-4955-bfc8-5d53df590e1d

<details>
<summary>App code</summary>

```py
from textual.app import App

MD = """
# This is a header

| col1 | col2 |
| :- | :- |
| value 1 | value 2 |

Here's some code: `from itertools import product`.
**Bold text**
_Emphasized text_
~~strikethrough~~
"""


class MyApp(App):
    CSS_PATH = "markdown.tcss"

    def compose(self):
        yield Markdown(MD)


if __name__ == "__main__":
    MyApp().run()
```

```sass
.code_inline,
.em,
.strong,
.s,
.markdown-table--header,
.markdown-table--lines,
{
    color: blue;
}
```

</details>

rodrigogiraoserrao:
Fixes #3464.

The reason why these component classes weren't being reloaded was that markdown blocks inherit from `Static` and they had their content set when markdown was first parsed, and then they would lose access to the source markdown.

This PR moves some of the logic that builds markdown blocks into a `MarkdownBlock` method that parses the markdown tokens into a `rich.text.Text` but **_also_** keeps track of the original source token.
This way, when CSS is reloaded, we can just ask the `MarkdownBlock` to rebuild itself.

By creating this method directly in `MarkdownBlock`, we only save the markup that is necessary and only reload the nodes that may have inline styles.


src/textual/widgets/_markdown.py
```diff
@@ -118,6 +119,95 @@ async def action_link(self, href: str) -> None:
...
+        change.
+
+        See https://github.com/Textualize/textual/issues/3464 for more information.
+
+        Args:
+            token: The token from which this block is built.
+        """
+
+        self._token = token
+        style_stack: list[Style] = [Style()]
```
rodrigogiraoserrao, line 147: From `style_stack: ...` to the end of this method, I copied and pasted the code.
The only change happens when I'm getting the component classes.
Instead of getting them from `self`, I get them from `self._markdown`.

willmcgugan (approved):

Issue #3045, opened 2023-08-01
Tooltips should be removed on widget unmount and scrolling


darrenburns:
When a tooltip is visible on screen, and the widget it's associated with gets unmounted. The tooltip remains on screen, which looks kind of weird.

Also, if a tooltip is visible, and you scroll such that the widget is no longer on screen, the tooltip remains. When you finally move the cursor, it disappears.

When the parent is unmounted, or when we scroll, I believe the tooltips should be hidden immediately.

davep:
> Also, if a tooltip is visible, and you scroll such that the widget is no longer on screen, the tooltip disappears. When you finally move the cursor, it disappears.

Just to check: there's a typo here @darrenburns? This is trying to say that if you scroll the tooltip *remains*, or something else?

darrenburns:
It remains, yeah

davep:
Another situation: if a widget is `visible`, has a tooltip shown, and then it is made not `visible`.

davep:
And another situation: if a widget is `display: block`, has a tooltip shown, and then becomes `display: none`.

davep:
#4191 is working well and seems to cover all the bases. We've decided that we should open it up a little and make this based off a signal, which will be published when a layout change happens, and so they the tooltip system can subscribe to it (the idea being someone's third party tooltip system could do the same).

Also, in addition to this: tooltips don't go away if a screen is pushed on top of a screen that is currently displaying a tooltip, so we'll look at taking that into account too.

PR #4191, opened 2024-02-20
Tooltip clearing on layout changes

davep:
Fix for #3045. While the initial issue was about how the tooltip remains if the widget that it pertains to has been removed, or scrolled away, really the issue comes down to if *any* layout change has happened such that the widget isn't the one "under the mouse" any more (so including things like `visible` or `display` changing, or other widgets changing in size such that the target widget gets moved, etc).

The idea here is that if a `Screen._refresh_layout` takes place, a check is made to ensure that the widget under the cursor hasn't changed from the one relating to the tooltip. If it has, the tooltip is removed. To implement this, and to allow a "hook" that other people can hook into if they need to do similar things, we've created a signal that is published when a screen's layout is refreshed; the check is then done by subscribing to that signal.

This PR also ensures that a tooltip gets removed when a screen is pushed over it.

Tested lots with [with this body of code](https://github.com/davep/textual-sandbox/blob/main/tooltip_hell.py), which covers the main cases (`remove`, scroll, `visible` toggle and `display` toggle).

---
Also adds pilot testing for tooltips.

---
Also includes a small fix to `Pilot`'s mouse event faking code to ensure that `App.mouse_position` is properly updated.


src/textual/screen.py
```diff
@@ -783,7 +783,9 @@ def _refresh_layout(
         if self.is_current:
             self._compositor_refresh()
 
-        if not self.app._dom_ready:
+        if self.app._dom_ready:
```
darrenburns, line 786: Does this branch run when you scroll?

darrenburns, line 786: Does this method run when you scroll?

src/textual/screen.py
```diff
@@ -851,6 +853,26 @@ def _update_tooltip(self, widget: Widget) -> None:
             if tooltip.display and self._tooltip_widget is widget:
                 self._handle_tooltip_timer(widget)
 
+    def _maybe_clear_tooltip(self) -> None:
+        """Check if the widget under the tooltip still pertains to the tooltip.
+
+        If they differ, the tooltip will be removed.
```
darrenburns, line 859:
```suggestion
        """Check if the widget under the mouse cursor still pertains to the tooltip.

        If they differ, the tooltip will be removed.
```

darrenburns: (Not 100% sure if this suggestion is correct)

davep: Kinda both really; but yes technically it's the mouse cursor (although the intent is that it's "under" the tooltip).

src/textual/screen.py
davep: In testing it very much appears to (see the test code linked to above, one of the test areas is a scrolling region).

davep:
While this approach is working out well, and we're agreed that it's the way go to, we're [going to expand on this a little](https://github.com/Textualize/textual/issues/3045#issuecomment-1954457816).

darrenburns (changes requested):
Could you add a pilot test which checks that the tooltips are now cleared from the DOM when we expect them to be?

davep:
Curiously it looks like there are zero pilot tests for tooltips to begin with; I wonder if it's even possible (or if they simply never got added)? I'll give it a crack.

davep:
Tests added.

darrenburns (approved):

PR #4192, opened 2024-02-20
Map ANSI colours

darrenburns:

- [ ] Updated documentation


src/textual/_ansi_theme.py
```diff
@@ -0,0 +1,53 @@
...
+        (152, 224, 36),
+        (224, 213, 97),
+        (157, 101, 255),
+        (244, 0, 95),
+        (88, 209, 235),
+        (246, 246, 239),
+    ],
+)
+
+MONOKAI_LIGHT = TerminalTheme(
```
willmcgugan, line 28: Is this MONOKAI with the foreground and background swapped? Monokai doesn't really have a light variant AFAIK. I think we should pick a dedicated light theme.

Let's pick a theme from https://iterm2colorschemes.com/

Then use this tool to export it:

https://github.com/jeremyschulman/iterm2-colors2rgb

willmcgugan (changes requested):
Looks good. Just a request.

src/textual/_ansi_theme.py
```diff
@@ -0,0 +1,53 @@
+from rich.terminal_theme import TerminalTheme
+
+MONOKAI_DARK = TerminalTheme(
```
willmcgugan, line 3: Is this theme called "Monokai dark" or just "monokai" ?

darrenburns:
Resolved the two pending requests.

src/textual/filter.py
```diff
@@ -188,9 +190,18 @@ def __init__(self, terminal_theme: TerminalTheme):
...
-        self.terminal_theme = terminal_theme
+        self._terminal_theme = terminal_theme
+        self._truecolor_cache: LRUCache[Style, Style] = LRUCache(maxsize=1024)
+
+    @property
+    def theme(self) -> TerminalTheme:
+        return self._terminal_theme
+
+    @theme.setter
+    def theme(self, theme: TerminalTheme) -> None:
```
willmcgugan, line 201: Rather than mutate this object, can we reconstruct the `ANSIToTruecolor` when the theme changes?

That way we can keep the `lru_cache`, which is faster. Filters run on ever segment, so the extra performance is worthwhile.

willmcgugan (changes requested):
Had a thought about caching and themes.

src/textual/filter.py
darrenburns: I've made this change but for some reason CI won't kick off. I think GitHub is having trouble - will try again tomorrow.

src/textual/app.py
```diff
@@ -880,8 +887,46 @@ def watch_dark(self, dark: bool) -> None:
...
+
+    def _refresh_truecolor_filter(self, theme: TerminalTheme) -> None:
+        """Update the ANSI to Truecolor filter, if available, with a new theme mapping.
+
+        Args:
+            theme: The new terminal theme to use for mapping ANSI to truecolor.
+        """
+        filters = self._filters
+        target_index = -1
+        for index, filter in enumerate(filters):
```
willmcgugan, line 920: Can we simplify this a bit? Looks like you could just return in the `isinstance` branch.

willmcgugan (approved):
Just one request

src/textual/app.py
darrenburns: Woops, of course

PR #4195, opened 2024-02-20
fix(tree-sitter): remove slow html highlight patterns

### LINKED ISSUE #4152: TextArea crashes Python on html language
The `TextArea` widget crashes the Python process when trying to render the attached HTML file with the `html` language: [snapshot_report.html.zip](https://github.com/Textualize/textual/files/14256168/snapshot_report.html.zip). 

There typically isn't any stacktrace for me, the entire terminal window closes with a `Python quit unexpectedly` message. I was able to observe the following error though: 
image

I used the `pytest-textual-snapshot` package to create the original HTML file. I've tested this on ARM Mac. This includes with the `syntax` extras installed. When you remove the language param the file is rendered correctly.

```py
"""
TextArea Example
"""

import os
import pathlib
from typing import Any

from textual.app import ComposeResult, App


class TextAreaApp(App):

    def __init__(self, path: os.PathLike, *args: Any, **kwargs: Any) -> None:
        self.path = pathlib.Path(path)
        self.text_area = TextArea(
            language="html",
            theme="monokai",
            soft_wrap=False,
            show_line_numbers=True,
        )

        yield self.text_area

    def on_mount(self) -> None:
        self.text_area.text = self.path.read_text(encoding="utf-8")


html_path = pathlib.Path(".") / "snapshot_report.html"
app = TextAreaApp(path=html_path)

if __name__ == "__main__":
    app.run()
```

TomJGooding:
Fixes #4152 by removing all `((element (start_tag (tag_name) @_tag)` patterns from the `html.scm` highlights query file.

These patterns will cause a segfault on relatively large documents and even just one seems a massively expensive operation from some quick testing.

All tests pass after removing these and I couldn't see they were actually used anywhere in syntax highlighting, but please correct me if I'm wrong!

- [ ] Updated documentation


willmcgugan:
Much appreciated, @TomJGooding !

juftin:
Very nice work on this, thank you!

TomJGooding:
No problem, glad I can contribute occasionally rather than just being a nuisance!

PR #4205, opened 2024-02-24
Decorator madness

willmcgugan:
Decorators for both watch and compute. Works like this:

```python
class MyWidget(Widget):

    count = reactive(0)

    @count.watch
    def _watch(self, new_count:int):
        ...
```

The old naming convention will continue to work.

No docs update in this PR. This is part of a body of work that includes a new Action mechanism. Will update to docs together.

### DIFF
--- a/src/textual/reactive.py
+++ b/src/textual/reactive.py
@@ -39,6 +39,8 @@ if TYPE_CHECKING:
 
 ReactiveType = TypeVar("ReactiveType")
 ReactableType = TypeVar("ReactableType", bound="DOMNode")
+WatchMethodType = TypeVar("WatchMethodType")
+ComputeMethodType = TypeVar("ComputeMethodType")
 
 
 class ReactiveError(Exception):
@@ -47,6 +49,54 @@ class ReactiveError(Exception):
 
 class TooManyComputesError(ReactiveError):
     """Raised when an attribute has public and private compute methods."""
+
+
+class WatchDecorator(Generic[WatchMethodType]):
+    """Watch decorator."""
+
+    def __init__(
+        self, watches: list[tuple[WatchMethodType, bool]] | None = None
+    ) -> None:
+        self._watches = watches
+
+    @overload
+    def __call__(self, *, init: bool = True) -> WatchDecorator[WatchMethodType]: ...
+
+    @overload
+    def __call__(self, method: WatchMethodType) -> WatchMethodType: ...
+
+    def __call__(
+        self, method: WatchMethodType | None = None, *, init: bool = True
+    ) -> WatchMethodType | WatchDecorator[WatchMethodType]:
+        if method is None:
+            return self
+        assert hasattr(method, "__name__")
+        if not method.__name__.startswith("watch_"):
+            if self._watches is not None:
+                self._watches.append((method, init))
+        return method
+
+
+class ComputeDecorator(Generic[WatchMethodType]):
+    """Watch decorator."""
+
+    def __init__(self, reactive: Reactive | None = None) -> None:
+        self._reactive = reactive
+
+    @overload
+    def __call__(self, *, init: bool = True) -> ComputeDecorator[WatchMethodType]: ...
+
+    @overload
+    def __call__(self, method: WatchMethodType) -> WatchMethodType: ...
+
+    def __call__(
+        self, method: WatchMethodType | None = None, *, init: bool = True
+    ) -> WatchMethodType | ComputeDecorator[WatchMethodType]:
+        if method is None:
+            return self
+        if not method.__name__.startswith("compute_"):
+            self._reactive._compute_method = method
+        return method
 
 
 async def await_watcher(obj: Reactable, awaitable: Awaitable[object]) -> None:
@@ -118,6 +168,8 @@ class Reactive(Generic[ReactiveType]):
         self._always_update = always_update
         self._run_compute = compute
         self._owner: Type[MessageTarget] | None = None
+        self._watches: list[tuple[Callable, bool]] = []
+        self._compute_method: Callable | None = None
 
     def __rich_repr__(self) -> rich.repr.Result:
         yield self._default
@@ -170,6 +222,8 @@ class Reactive(Generic[ReactiveType]):
         _rich_traceback_omit = True
         for name, reactive in obj._reactives.items():
             reactive._initialize_reactive(obj, name)
+            for watch_method, init in reactive._watches:
+                obj.watch(obj, name, watch_method.__get__(obj), init=init)
 
     @classmethod
     def _reset_object(cls, obj: object) -> None:
@@ -233,8 +287,14 @@ class Reactive(Generic[ReactiveType]):
         if not hasattr(obj, internal_name):
             self._initialize_reactive(obj, self.name)
 
-        if hasattr(obj, self.compute_name):
-            value: ReactiveType
+        value: ReactiveType
+        if self._compute_method is not None:
+            old_value = getattr(obj, internal_name)
+            value = self._compute_method.__get__(obj)()
+            setattr(obj, internal_name, value)
+            self._check_watchers(obj, self.name, old_value)
+            return value
+        elif hasattr(obj, self.compute_name):
             old_value = getattr(obj, internal_name)
             value = getattr(obj, self.compute_name)()
             setattr(obj, internal_name, value)
@@ -326,21 +386,35 @@ class Reactive(Generic[ReactiveType]):
             obj: Reactable object.
         """
         _rich_traceback_guard = True
-        for compute in obj._reactives.keys():
-            try:
-                compute_method = getattr(obj, f"compute_{compute}")
-            except AttributeError:
+        for name, reactive in obj._reactives.items():
+            print(name, reactive)
+            if reactive._compute_method is not None:
+                compute_method = reactive._compute_method.__get__(obj)
+            else:
                 try:
-                    compute_method = getattr(obj, f"_compute_{compute}")
+                    compute_method = getattr(obj, f"compute_{name}")
                 except AttributeError:
-                    continue
+                    try:
+                        compute_method = getattr(obj, f"_compute_{name}")
+                    except AttributeError:
+                        continue
             current_value = getattr(
-                obj, f"_reactive_{compute}", getattr(obj, f"_default_{compute}", None)
+                obj, f"_reactive_{name}", getattr(obj, f"_default_{name}", None)
             )
             value = compute_method()
-            setattr(obj, f"_reactive_{compute}", value)
+            setattr(obj, f"_reactive_{name}", value)
             if value != current_value:
-                cls._check_watchers(obj, compute, current_value)
+                cls._check_watchers(obj, name, current_value)
+
+    @property
+    def watch(self) -> WatchDecorator:
+        """A decorator to make a method a watch method."""
+        return WatchDecorator(self._watches)
+
+    @property
+    def compute(self) -> ComputeDecorator:
+        """A decorator to make a method a compute method."""
+        return ComputeDecorator(self)
 
 
 class reactive(Reactive[ReactiveType]):
--- a/tests/test_reactive.py
+++ b/tests/test_reactive.py
@@ -705,3 +705,41 @@ async def test_external_watch_init_does_not_propagate_to_externals() -> None:
         assert logs == ["first", "second"]
         app.query_one(SomeWidget).test_var = 73
         assert logs == ["first", "second", "first", "second"]
+
+
+async def test_watch_decorator():
+    """Test watchers defined via a decorator."""
+
+    class WatchApp(App):
+        count = reactive(0, init=False)
+
+        watcher_call_count = 0
+
+        @count.watch
+        def _(self, value: int) -> None:
+            self.watcher_call_count = value
+
+    app = WatchApp()
+    async with app.run_test():
+        app.count += 1
+        assert app.watcher_call_count == 1
+        app.count += 1
+        assert app.watcher_call_count == 2
+        app.count -= 1
+        assert app.watcher_call_count == 1
+        app.count -= 1
+        assert app.watcher_call_count == 0
+
+
+async def test_reactive_compute_decorator_first_time_set():
+    class ReactiveComputeFirstTimeSet(App):
+        number = reactive(1)
+        double_number = reactive(None)
+
+        @double_number.compute
+        def _double_number(self):
+            return self.number * 2
+
+    app = ReactiveComputeFirstTimeSet()
+    async with app.run_test():
+        assert app.double_number == 2
--- a/examples/calculator.py
+++ b/examples/calculator.py
@@ -47,8 +47,7 @@ class CalculatorApp(App):
         """Compute switch to show AC or C button"""
         return self.value in ("", "0") and self.numbers == "0"
 
-    @show_ac.watch
-    def _show_ac(self, show_ac: bool) -> None:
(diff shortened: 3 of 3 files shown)


src/textual/reactive.py
```diff
@@ -50,6 +53,108 @@ class TooManyComputesError(ReactiveError):
...
+    ) -> WatchMethodType | WatchDecorator[WatchMethodType]:
+        _rich_traceback_omit = True
+        if method is None:
+            return self
+        assert self._reactive is not None
+        assert hasattr(method, "__name__")
+        if not method.__name__.startswith("watch_"):
+            if self._reactive._watch_method is not None:
+                raise RuntimeError(
+                    "Only a single method may be decorator with watch (per-reactive)."
```
rodrigogiraoserrao, line 83:
```suggestion
                    "Only a single method may be decorated with watch (per-reactive)."
```

src/textual/reactive.py
```diff
@@ -50,6 +53,108 @@ class TooManyComputesError(ReactiveError):
     """Raised when an attribute has public and private compute methods."""
 
 
+class WatchDecorator(Generic[WatchMethodType]):
+    """Watch decorator.
+
+    Decorate a method to make it a watcher.
+
```
rodrigogiraoserrao, line 60: Good try!

```suggestion
```

src/textual/reactive.py
```diff
@@ -50,6 +53,108 @@ class TooManyComputesError(ReactiveError):
...
+            self._reactive._watch_method = method
+            self._reactive._watch_method_init = init
+        return method
+
+
+class ComputeDecorator(Generic[ComputeMethodType]):
+    """Watch decorator.
+
+    Decorate a widget method to make it a compute method.
+
```
rodrigogiraoserrao, line 94:
```suggestion
```

src/textual/reactive.py
```diff
@@ -50,6 +53,108 @@ class TooManyComputesError(ReactiveError):
...
+                raise RuntimeError(
+                    "Only a single method may be decorator with watch (per-reactive)."
+                )
+            self._reactive._watch_method = method
+            self._reactive._watch_method_init = init
+        return method
+
+
+class ComputeDecorator(Generic[ComputeMethodType]):
+    """Watch decorator.
```
rodrigogiraoserrao, line 91:
```suggestion
    """Compute decorator.
```

src/textual/reactive.py
```diff
@@ -50,6 +53,108 @@ class TooManyComputesError(ReactiveError):
...
+                )
+            self._reactive._compute_method = method
+        return method
+
+
+class ValidateDecorator(Generic[ValidateMethodType]):
+    """Validate decorator.
+
+    Decorate a Widget method to make it a validator for the attribute/
+
```
rodrigogiraoserrao, line 127:
```suggestion
    Decorate a Widget method to make it a validator for the attribute.
```

src/textual/reactive.py
```diff
@@ -50,6 +53,108 @@ class TooManyComputesError(ReactiveError):
     """Raised when an attribute has public and private compute methods."""
 
 
+class WatchDecorator(Generic[WatchMethodType]):
```
rodrigogiraoserrao, line 56: This (and the `Validate`/`ComputeDecorator`) kind of reminded me of the typing struggles I had with the work decorator.
Maybe you did it already but it may be worth running `make typecheck` and making sure this introduces as little typing errors as possible.

src/textual/reactive.py
```diff
@@ -50,6 +53,108 @@ class TooManyComputesError(ReactiveError):
...
+
+    @overload
+    def __call__(self, method: WatchMethodType) -> WatchMethodType: ...
+
+    def __call__(
+        self, method: WatchMethodType | None = None, *, init: bool = True
+    ) -> WatchMethodType | WatchDecorator[WatchMethodType]:
+        _rich_traceback_omit = True
+        if method is None:
+            return self
```
rodrigogiraoserrao, line 77: If we use the decorator as `@some_reactive.watch(init=False)` we need to save the value of `init=False` _somewhere_ before returning the decorator, otherwise `init` is always used as `True`.

To test this, try using `some_reactive.watch(init=False)` around a method and print the value of `init` after the assertions. The value `True` will be printed.

This means we also need a test to make sure that `init=False` is respected when specified in the watch decorator.

src/textual/reactive.py
```diff
@@ -50,6 +53,108 @@ class TooManyComputesError(ReactiveError):
...
+        assert self._reactive is not None
+        assert hasattr(method, "__name__")
+        if not method.__name__.startswith("watch_"):
```
rodrigogiraoserrao, line 80: I'm a bit confused. If the decorated method has a name starting with `_watch`, we don't do the duplicate method checking / raise the error and instead just override the previous decorated watch method?
Similar comments apply to the compute and validate decorators.

In other words, if `test` is a reactive, then the first option below raises an error but the 2nd one doesn't?

```py
class MyApp(App[None]):
    test = reactive(42)

    @test.watch
    def _(self):
        pass

    @test.watch
    def _2(self):
        pass
```

```py
class MyApp(App[None]):
    test = reactive(42)

    @test.watch
    def _(self):
        pass

    @test.watch
    def watch_2(self):
        pass
```

src/textual/reactive.py
```diff
@@ -50,6 +53,108 @@ class TooManyComputesError(ReactiveError):
...
+            if self._reactive._watch_method is not None:
+                raise RuntimeError(
```
rodrigogiraoserrao, line 82: When this is raised, we don't get the pretty-printed Textual tracebacks.
I'm assuming you're aware of that?

Screenshot 2024-02-27 at 16 20 00

Similar comments apply to the compute and validate decorators.

src/textual/reactive.py
```diff
@@ -50,6 +53,108 @@ class TooManyComputesError(ReactiveError):
...
+        self._reactive = reactive
+
+    @overload
+    def __call__(self, *, init: bool = True) -> ComputeDecorator[ComputeMethodType]: ...
+
+    @overload
+    def __call__(self, method: ComputeMethodType) -> ComputeMethodType: ...
+
+    def __call__(
+        self, method: ComputeMethodType | None = None, *, init: bool = True
```
rodrigogiraoserrao, line 107: It looks like `init` is not needed in the compute decorator, which means the overloads are superfluous.
If `init` _is_ needed, where is it used?

src/textual/reactive.py
```diff
@@ -50,6 +53,108 @@ class TooManyComputesError(ReactiveError):
...
+    def __call__(
+        self, *, init: bool = True
+    ) -> ValidateDecorator[ValidateMethodType]: ...
+
+    @overload
+    def __call__(self, method: ValidateMethodType) -> ValidateMethodType: ...
+
+    def __call__(
+        self, method: ValidateMethodType | None = None, *, init: bool = True
+    ) -> ValidateMethodType | ValidateDecorator[ValidateMethodType]:
```
rodrigogiraoserrao, line 143: It looks like `init` is not needed in the validate decorator, which means the overloads are superfluous.
If `init` _is_ needed, where is it used?

src/textual/reactive.py
```diff
@@ -278,6 +400,8 @@ def __set__(self, obj: Reactable, value: ReactiveType) -> None:
         public_validate_function = getattr(obj, f"validate_{name}", None)
         if callable(public_validate_function):
             value = public_validate_function(value)
+        if self._validate_method:
+            value = self._validate_method.__get__(self)(value)
```
rodrigogiraoserrao, line 404: This feels like it's ok to have 2+ validate methods but the validate decorator can only be used once per reactive. Am I reading this correctly and is this the intended behaviour?

src/textual/reactive.py
```diff
@@ -341,21 +465,76 @@ def _compute(cls, obj: Reactable) -> None:
...
+        Example:
+            ```python
+            class MyWidget(Widget):
+                count = reactive(0)
+                @count.watch
+                def _count_changed(self, count:int) -> None:
+                    # Called when count changes
+                    ...
+            ```
+
```
rodrigogiraoserrao, line 500:
```suggestion
```

tests/test_reactive.py
rodrigogiraoserrao, line 1: We're missing, at least, a test that makes sure that `init=False` in the watch decorator is respected.

rodrigogiraoserrao (changes requested):
I'm excited for these changes but I think there are a couple of loose ends we must tie before merging.

src/textual/reactive.py
willmcgugan: There is one, because I couldn't figure out how to type an unbound method. Any suggestions?

src/textual/reactive.py
willmcgugan: Good catch. Turns out the init wasn't needed at all. I've removed it.

src/textual/reactive.py
willmcgugan: This is to exclude match methods defined via the naming convention. Without it, a decorated watch method may be called twice.

It may be better to prohibit decorating a watch method. What do you think?

src/textual/reactive.py
willmcgugan: Yes, because it occurs at import time, before the traceback hooks. I don't want to set up the pretty tracebacks at import time, because that would force it on users who just import `textual`.

src/textual/reactive.py
willmcgugan: Copypasta. Removed.

src/textual/reactive.py
willmcgugan: Why bother? 🤷‍♂️

src/textual/reactive.py
willmcgugan: You can't have two validate methods via the naming convention, because they will clash. i.e. If you have two `validate_foo` methods, only the second one will be called.

You *can* have a validate method defined via the naming convention, and one via the decorators. 

I could prohibit that, but then it may break existing code.

It is a bit of a wart. I can see us removing the naming convention before 1.0.

src/textual/reactive.py
rodrigogiraoserrao: I think decorating a watch method will look confusing but if we end up dropping the naming convention either way, is it worth creating a special rule now and then remove it?

src/textual/reactive.py
rodrigogiraoserrao: Ok, good!

src/textual/reactive.py
rodrigogiraoserrao: Ok, thanks for explaining.

src/textual/reactive.py
rodrigogiraoserrao: Copypasta carbonara?

src/textual/reactive.py
rodrigogiraoserrao: Right.

src/textual/reactive.py
rodrigogiraoserrao: I noticed it en passant and making the suggestion is almost effortless 🤷‍♂️

src/textual/reactive.py
rodrigogiraoserrao: I think I do have a suggestion.
The decorators (Watch / Compute / Validate) shouldn't be generic over the type of the method they're decorating; rather, they should be generic over the type of the reactive involved because that's what can really change.

For example, for my compute methods, I know I'll always be decorating methods with type `Callable[[Any], ReactiveType]`.
So, what _maybe_ I can have is an alias for compute methods and then use that.

The suggestion below seems to eliminate an error you get with `make typecheck` and it also gets rid of some IDE errors I get in several places.
However, I'm not 100% sure this is better.
It _feels_ better, but I wouldn't bet my life on it.

```py
# ...

ComputeMethodAlias: TypeAlias = Callable[[Any], ReactiveType]  # <--

# ...

class ComputeDecorator(Generic[ReactiveType]):  # <--
    def __init__(self, reactive: Reactive[ReactiveType] | None = None) -> None:  # <-- better typing here
        self._reactive = reactive

    @overload
    def __call__(self) -> ComputeDecorator[ReactiveType]: ...  # <--

    @overload
    def __call__(
        self, method: ComputeMethodAlias[ReactiveType]  # <--
    ) -> ComputeMethodAlias[ReactiveType]: ...  # <--

    def __call__(
        self, method: ComputeMethodAlias[ReactiveType] | None = None  # <--
    ) -> ComputeMethodAlias[ReactiveType] | ComputeDecorator[ReactiveType]:  # <--
        # ...

# ...

@rich.repr.auto
class Reactive(Generic[ReactiveType]):
    # ...
    def __init__(...):
        # ...
        self._compute_method: ComputeMethodAlias[ReactiveType] | None = None  # <--  better typing

    # ...

    @property
    def compute(self) -> ComputeDecorator[ReactiveType]:  # <--
        # ...
```

tests/test_reactive.py
```diff
@@ -709,6 +709,268 @@ def second_callback() -> None:
...
+
+async def test_watch_decorator_call():
+    """Test watchers defined via a decorator, that is called."""
+
+    class WatchApp(App):
+        count = reactive(0, init=False)
+
+        watcher_call_count = 0
+
+        @count.watch()
```
rodrigogiraoserrao, line 748: It was _very_ difficult for me to spot the difference.
Would it be worth adding a comment `# <--` to make the diff obvious so that no one deletes this test in the future by mistake?

tests/test_reactive.py
```diff
@@ -709,6 +709,268 @@ def second_callback() -> None:
...
+        app.count -= 1
+        assert app.watcher_call_count == 3
+        assert app.count == 1
+        app.count -= 1
+        assert app.watcher_call_count == 4
+        assert app.count == 0
+
+
+async def test_watch_decorator_init_true():
+    """Test watchers defined via a decorator, that is called."""
```
rodrigogiraoserrao, line 769:
```suggestion
    """Test watchers defined via a decorator, that is called, when the reactive has init set."""
```

darrenburns:
With this change, how would we override what a watcher does in a widget subclass?

If I have a `count` reactive declared in a widget which has a watcher defined, then I subclass that and I want my own watcher, how would I achieve this? In the child class, `@count.watch` doesn't work because `count` is not defined.

For example, this fails to run because `NameError: name 'count' is not defined` 

```python
    class Parent(Widget):
        count: Reactive[int] = reactive(0, init=False)

        @count.watch
        def _count_parent(self, new_value: int) -> None:
            print("parent watcher")

    class Child(Parent):

        @count.watch
        def _count_parent(self, new_value: int) -> None:
            print("child watcher")
```

`@Parent.count.watch` doesn't work because of the "only a single method may be decorated with watch".

Even if that's solved, there's still a bit of a developer experience issue that @davep mentioned, whereby we no longer know what the parent watch method is called without going looking for it.

willmcgugan:
Going to park this for now. @darrenburns 's observation re inheritance may kill this idea entirely.

yggdr:
You can work around the `NameError` issue with the metaclass' `__prepare__` method, e.g.
```python
class M(type):
         @classmethod
         def __prepare__(cls, name, bases, **kwargs):
             for base in bases:
                 if base.__name__ == 'A':
                     return dict(a=base.a)
             return super().__prepare__(name, bases, **kwargs)

class A(metaclass=M):
    a=1

class B(A):
    print(a)
```
will print out `1`.
When properly generalising this, I would create the MRO from the given list of bases, then go through them checking for attributes of type `textual.reactive.reactive`, and add the last one.

Issue #4179, opened 2024-02-19
`grid-rows: auto` seems to be broken when `column-span`/`Pretty` is used


mzebrak:
Looks like there is a rendering issue when `column-span` (or `Pretty` widget) is used together with `grid-rows: auto`.

Consider such a  example:

```python
from __future__ import annotations

from textual.app import App, ComposeResult
from textual.binding import Binding
from textual.containers import Grid
from textual.widgets import Footer, Pretty, Static


class MyGrid(Grid):
    """A simple grid with 2 columns and 2 rows."""


class MyApp(App):
    BINDINGS = [
        Binding("a", "add", "Add a new item"),
    ]

    DEFAULT_CSS = """
    MyGrid {
        height: auto;
        grid-rows: auto;
        grid-size: 2;
        grid-gutter: 1;
        background: green;
    }

    Pretty {
        background: red;
    }

    Static {
        background: blue;
    }
"""

    def compose(self) -> ComposeResult:
        with MyGrid():
            yield Pretty(["Something that should have colspan=2"])
            yield Static("")
            yield Static("I want this")
            yield Static("... in a single row")
        yield Footer()

    def action_add(self):
        pretty = self.query_one(Pretty)
        before = pretty._renderable._object
        pretty.update([*before, before[0]])


MyApp().run()
```

Results (correctly) in:

![image](https://github.com/Textualize/textual/assets/73518558/e2521b70-c2d8-4f43-9aed-333a3c12fe5e)
![image](https://github.com/Textualize/textual/assets/73518558/dd913fb4-78d9-4a7c-906c-c4100b4dbb92)

But not quite correct when we change the terminal size a bit (adds a lot of free green space):

![image](https://github.com/Textualize/textual/assets/73518558/dd3e7b78-11b6-4b5b-9ae6-0a6849d9a661)

When using column-span:

```python
from __future__ import annotations

from textual.app import App, ComposeResult
from textual.binding import Binding
from textual.containers import Grid
from textual.widgets import Footer, Pretty, Static


class MyGrid(Grid):
    """A simple grid with 2 columns and 2 rows."""


class MyApp(App):
    BINDINGS = [
        Binding("a", "add", "Add a new item"),
    ]

    DEFAULT_CSS = """
    MyGrid {
        height: auto;
        grid-rows: auto;
        grid-size: 2;
        grid-gutter: 1;
        background: green;
    }

    Pretty {
        column-span: 2;
        background: red;
    }

    Static {
        background: blue;
[...]

PR #4211, opened 2024-02-26
Fix content size cache

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/4179

Forgive the lack of a test. The code is covered. But it could use https://github.com/Textualize/textual/issues/4212 for a fuller test.


rodrigogiraoserrao:
@willmcgugan I feel like I am missing something.
I still see the issue mentioned in #4179.

The video below shows that at widths 81 & 82 I have the big green region that I shouldn't.
For widths <=80, wrapping is fine. For widths >= 83, everything is shown in a single row.

https://github.com/Textualize/textual/assets/5621605/8d18786d-d8c4-46cb-b38c-2c99467c48ad

rodrigogiraoserrao:
(With #4217 merged, a fuller test can be written. I was going to write it myself but as per my comment above, it looks like the original issue hasn't been solved.)

willmcgugan:
Looks like another issue. Only occurs when there is a gutter on the grid.

PR #4219, opened 2024-02-26
Grid gutter fix

willmcgugan:
Possible fix for grid gutter issue.


rodrigogiraoserrao (approved):
Looks good to me.

With a terminal width of 81 or 82 characters, the app I introduced as a snapshot test would break in the same way as the app of the OP in the original issue.
(There would be a tall green region between two consecutive rows of the grid.)

With the fix, the gap between rows matches the grid gutter specified.

Issue #3869, opened 2023-12-14
Ensure all widget messages are posted even when triggered by the developer


rodrigogiraoserrao:
We should not prevent a message from being posted just because the state of the widget was changed programmatically, and even though the programmer already knows the state changed.

E.g., `TabbedContent` should post `TabActivated` when the reactive attribute `active` is assigned to.

See #2500 & #3417 for more examples and context.

TomJGooding:
There's a number of issues so I thought it might be a good idea to start a list of messages that might need looking at:

- [ ] TabbedContent.TabActivated
- [ ] Select.Changed
- [ ] TextArea.Changed

TomJGooding:
Presumably a `Changed` event shouldn't post on initialisation? How would you approach this if  `init=False`isn't an option as there's other functionality in the watch method?

davep:
Looking at every widget, working out which are N/A for this and which have properties that can cause events:

- `Button` - N/A
- `Checkbox` - N/A - (see also `ToggleButton`)
- `Collapsible`
  - [ ] `Expanded`
  - [ ] `Collapsed`
- `ContentSwitcher` - N/A
- `DataTable`
  - `CellHighlighted` - TBD
  - `CellSelected` - TBD
  - `RowHighlighted` - TBD
  - `RowSelected` - TBD
  - `ColumnHighlighted` - TBD
  - `ColumnSelected` - TBD
  - `HeaderSelected` - TBD
  - `RowLabelSelected` - TBD

*WIP*

davep:
*[WiP]*

Now whittling things down; here are widgets and their messages that *don't* post messages when something happens in code that mimics something the user might do:

## `Collapsible`

Neither `Collapsible.Expanded` not `Collapsible.Collapsed` are posted when `Collapsible.collapsed` is changed.

davep:
Starting to work through the widgets (as documented above), I seemed to run into a problem where, when looking at the "missing" messages from `DataTable`, it made no sense that they weren't getting posted when updating the `cursor_coordinate` from code. Reading the code there was no difference between that reactive being changed by code run from bindings, vs that being changed by any other code.

I've since gone on to see the same issue with `Input` too. Given that `Input` is more straightforward, I think I can illustrate what I'm seeing with an example built around that.

Consider this code:

```python
from textual import on
from textual.app import App, ComposeResult
from textual.containers import Vertical
from textual.widgets import Button, Input

class Child(Vertical):

[...]

PR #4222, opened 2024-02-27
Correct the sender of a reactive-posted message

davep:
While investigating #3869 it became apparent that there was an edge case where it was possible to have messages that should bubble, not bubble, depending on how and where the message was posted in respect to a reactive watcher.

See https://github.com/Textualize/textual/issues/3869#issuecomment-1964300873

This fixes that.


rodrigogiraoserrao (approved):
Ahh, this makes sense 💡

PR #4226, opened 2024-02-27
Animation docs fix

darrenburns:
Fixes the animation examples in the docs.
Removes some wait_for_idles that I don't think are required?
Fix some docstrings.


willmcgugan (approved):
Looks good. Slim chance those `wait_for_idle`s were fixing a flaky test.

PR #4231, opened 2024-02-27
fix(select): stop option highlighted bubbling

### LINKED ISSUE #4224: `Select` `OptionHighlighted` message leakage
The `Select` widget makes use of `OptionList`; when the user opens the `Select` to pick a value, `OptionHighlighted` messages "leak" out of the `Select`. This could result in unexpected messages for the developer who has an `OptionList` and a `Select` in their DOM and who doesn't expect to actually have two `OptionList`s that might send messages.

I think we should either stop the `OptionHighlighted` from bubbling outside of the `Select`, or we should turn them into `SelectionHighlighted` if it's felt necessary that they can be handled.

To illustrate:

```python
from textual import on
from textual.widgets import OptionList, Select, Log

class LeakySelectApp(App[None]):

        yield Log()
        yield OptionList(*(f"This is option {n}" for n in range(10)))
        yield Select[int](((f"This is selection {n}", n) for n in range(10)))

    @on(OptionList.OptionHighlighted)
    def log_some_option(self, event: OptionList.OptionHighlighted) -> None:
        self.query_one(Log).write_line(f"{event}")

if __name__ == "__main__":
    LeakySelectApp().run()
```

I think the dev would reasonably expect to *only* get calls to `log_some_option` from the `OptionList` they composed in.

TomJGooding:
Fixes #4224 by posting a `Select.SelectionHighlighted` message  when the highlighted option changes, instead of `OptionList.OptionHighlighted`  bubbling.

I've followed the suggestion in the linked issue for the name of the message, but I wonder if this could simply be `Select.Highlighted`, similar to the `ListView` widget? I think using mixed terminology of "option" and "selection" might be a little confusing.

- [ ] Updated documentation

### DIFF
--- a/src/textual/widgets/_select.py
+++ b/src/textual/widgets/_select.py
@@ -271,6 +271,26 @@ class Select(Generic[SelectType], Vertical, can_focus=True):
             """The Select that sent the message."""
             return self.select
 
+    class SelectionHighlighted(Message):
+        """Posted when the highlighted option changes.
+
+        This message can be handled using a `on_select_selection_highlighted` method.
+        """
+
+        def __init__(
+            self, select: Select[SelectType], value: SelectType | NoSelection
+        ) -> None:
+            super().__init__()
+            self.select = select
+            """The select widget."""
+            self.value = value
+            """The value of the highlighted option."""
+
+        @property
+        def control(self) -> Select[SelectType]:
+            """The Select that sent the message."""
+            return self.select
+
     def __init__(
         self,
         options: Iterable[tuple[RenderableType, SelectType]],
@@ -517,6 +537,14 @@ class Select(Generic[SelectType], Vertical, can_focus=True):
             self.expanded = False
 
         self.call_after_refresh(update_focus)  # Prevents a little flicker
+
+    @on(SelectOverlay.OptionHighlighted)
+    def _on_select_overlay_option_highlighted(
+        self, event: SelectOverlay.OptionHighlighted
+    ) -> None:
+        event.stop()
+        value = self._options[event.option_index][1]
+        self.post_message(self.SelectionHighlighted(self, value))
 
     def action_show_overlay(self) -> None:
         """Show the overlay."""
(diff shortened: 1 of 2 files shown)


TomJGooding:
I'm marking this as ready to review although I haven't updated the docs or changelog yet as I just want to check the proper message name.

davep:
In conversation in the office yesterday we'd concluded that the most sensible approach here was to simply swallow the message; there didn't seem to be much use for a parallel message given the drop down is a detail of Select that's "internal" to how it works.

TomJGooding:
Fair enough. I just thought a `Highlighted` message could be useful if you wanted to display some details while traversing the options. For example (using the old OptionHighlighted for ease):

```python
from textual import on
from textual.widgets import OptionList, Select, Static


class ExampleApp(App):
    CSS = """
    Select, #option-detail {
        height: 50%;
    }
    """

        yield Select[int]([(str(n), n) for n in range(3)])
        yield Static(id="option-detail")

    @on(OptionList.OptionHighlighted)
    def display_option_details(
        self,
        event: OptionList.OptionHighlighted,
    ) -> None:
        highlighted_selection = str(event.option.prompt)
        option_detail = self.query_one("#option-detail", Static)
        if highlighted_selection == "0":
            option_detail.update("Don't try dividing by zero!")
        if highlighted_selection == "1":
            option_detail.update("One is the loneliest number.")
        elif highlighted_selection == "2":
            option_detail.update("Two is the only even prime number.")


if __name__ == "__main__":
    app = ExampleApp()
    app.run()
```

TomJGooding:
Sorry Dave, but just to confirm should I close this PR? I wasn't quite sure if your :+1: reaction might mean you agreed a highlighted message might still be useful?

(Just as an aside, if you think I need to back off and leave these issues to maintainers, please just let me know. My small contributions here are trying to be helpful, but obviously I don't always have the full context with discussions at Textual towers.)

davep:
Ahh, sorry, no, often I'll just mark a message like that so I and the author know I've read it; especially if I've got nothing to add.

As for if `Select` should do this or not, that's not really up to me I'm afraid.

TomJGooding:
Closed as superseded by #4280.

PR #4235, opened 2024-02-28
Expose `Message._set_sender` as a utility method

davep:
Also have it return `self` to allow for useful chaining.

This will be useful in at least one situation that will be fixed once #3869 concludes (in that case a `TreeNode` sends a message on behalf of `Tree`, via `Tree`, but when the caller of that message is the active message pump and so ends up being seen as the sender -- the `TreeNode` method really needs to say that its parent `Tree` is the sender).

### DIFF
--- a/src/textual/message.py
+++ b/src/textual/message.py
@@ -98,8 +98,8 @@ class Message:
             sender: The sender.
 
         Note:
-            When creating a message the sender is automatically set;
-            normally there will be no need for this method to be called.
+            When creating a message the sender is automatically set.
+            Normally there will be no need for this method to be called.
             This method will be used when strict control is required over
             the sender of a message.
 
--- a/src/textual/message.py
+++ b/src/textual/message.py
@@ -8,6 +8,7 @@ from __future__ import annotations
 from typing import TYPE_CHECKING, ClassVar
 
 import rich.repr
+from typing_extensions import Self
 
 from . import _time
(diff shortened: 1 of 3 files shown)


src/textual/message.py
```diff
@@ -90,9 +91,17 @@ def _set_forwarded(self) -> None:
         """Mark this event as being forwarded."""
         self._forwarded = True
 
-    def _set_sender(self, sender: MessagePump) -> None:
-        """Set the sender."""
+    def set_sender(self, sender: MessagePump) -> Self:
```
willmcgugan, line 94: Could use a little more detail in the docstring. I doubt the app developer would know what the sender was.

willmcgugan (changes requested):

src/textual/message.py
```diff
@@ -90,9 +91,23 @@ def _set_forwarded(self) -> None:
...
+        """Set the sender of the message.
+
+        Args:
+            sender: The sender.
+
+        Note:
+            When creating a message the sender is automatically set;
+            normally there will be no need for this method to be called.
+            This method will be used when strict control is required over
+            the sender of a message.
```
willmcgugan, line 104:
```suggestion
            When creating a message the sender is automatically set.
            Normally there will be no need for this method to be called.
            This method will be used when strict control is required over
            the sender of a message.
```

Issue #4073, opened 2024-01-29
Rendering issue with soft hyphens (~~long/wrapped text in `OptionList`~~)


davep:
At the suggestion of @willmcgugan I'm recording this *without* an MRE (although I have tried to make one; see below). In an application I'm currently toying with, over the weekend, I ran into this curious rendering bug:

![Screenshot 2024-01-28 at 12 42 56](https://github.com/Textualize/textual/assets/28237/ff95e951-c313-4b31-9887-e529d92d7b97)

Note the effect seen to the right of question 9.

The widget is an `OptionList`, the `Option`s within are made up of Rich renderables which in turn are made up of a `Group` that contains two `Table.grid`s.

I've attempted to isolate the issue from the wider application, adding all the relevant styles (also doing it as a `ModalScreen` as this was part of such):

```python
from rich.table import Table
from rich.console import Group

from textual.app import App, ComposeResult
from textual.containers import Vertical
from textual.screen import ModalScreen
from textual.widgets import OptionList

NAMES = [
    "Cardiff",
    "Swansea",
    "Aberdare",
    "Llanfairpwllgwyngyllgogerychwyrndrobwllllantysiliogogogoch",
    "Mountain Ash",
    "Hirwaun"
]

class Names(ModalScreen[None]):

    CSS = """
    Names {
        align: center middle;

        Vertical {
            width: auto;
            max-width: 60%;
            height: auto;
            max-height: 80%;
            padding: 1 2 0 2;
            background: $surface;
            border: panel $primary;
            border-title-color: $accent;

            OptionList {
                height: 1fr;
                margin: 1 0 1 0;
                &> .option-list--option {
                    padding: 0 5 1 0;
                }
            }
        }
    }
    """

    def gridify(self, n: int, name: str) -> Group:
        question = Table.grid()
        question.add_column(width=3, justify="right")
        question.add_column(width=1)
        question.add_column(ratio=1)
        question.add_row(str(n), "", f"{name} is the name of a place in Wales." )
        answer = Table.grid()
        answer.add_column(width=4)
        answer.add_column(ratio=1)
        answer.add_row("", "[red]Here is the[/] [green]second line[/]")
        return Group(question, answer)

    def compose(self) -> ComposeResult:
        with Vertical():
            yield OptionList(*[self.gridify(n, name) for n, name in enumerate(NAMES)])

class TruncatedTextOptionApp(App[None]):

    CSS = """
    Screen#_default {
        background: red;
    }
    """

[...]

PR #4239, opened 2024-02-28
Update rich to use, at least, 13.7.1.

rodrigogiraoserrao:
Rich 13.7.1 introduces the fix for #4073.


davep (approved):

willmcgugan:
This isn't a good idea. We can leave the Rich dependancy as is, and new installs will still get the latest.

willmcgugan:
I don't think the issue this fixes is worth the potential dependancy issues it will cause.

PR #4243, opened 2024-02-29
Fix `TabbedContent` type errors

davep:
Following on from #4234, fixes a couple of type errors in `TabbedContent` relating to adding and removing panes.

### DIFF
--- a/src/textual/widgets/_tabbed_content.py
+++ b/src/textual/widgets/_tabbed_content.py
@@ -15,7 +15,7 @@ from ..await_complete import AwaitComplete
 from ..css.query import NoMatches
 from ..message import Message
 from ..reactive import reactive
-from ..widget import Widget
+from ..widget import AwaitMount, Widget
 from ._content_switcher import ContentSwitcher
 from ._tabs import Tab, Tabs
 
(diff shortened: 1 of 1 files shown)


src/textual/widgets/_tabbed_content.py
```diff
@@ -430,13 +432,19 @@ def add_pane(
         pane = self._set_id(pane, tabs.tab_count + 1)
         assert pane.id is not None
         pane.display = False
+
+        async def _add_part(awaitable: AwaitComplete | AwaitMount) -> None:
```
willmcgugan, line 436: This doesn't seem right. If you need to wrap those awaitables in a coroutine, it suggests that maybe `AwaitComplete` is type incorrectly.

Maybe `AwaitComplete` needs to accept an `Awaitable`?

src/textual/widgets/_tabbed_content.py
```diff
@@ -430,13 +432,19 @@ def add_pane(
...
-                after=after if after is None else ContentTab.add_prefix(after),
+            _add_part(
+                tabs.add_tab(
+                    ContentTab(pane._title, pane.id),
+                    before=before if before is None else ContentTab.add_prefix(before),
+                    after=after if after is None else ContentTab.add_prefix(after),
+                )
             ),
-            self.get_child_by_type(ContentSwitcher).mount(pane),
+            _add_part(self.get_child_by_type(ContentSwitcher).mount(pane)),
```
willmcgugan, line 447: This actually seems like a flaw. The `AwaitCompete` class runs its awaitables concurrently. This looks like something that should run in serial.

src/textual/widgets/_tabbed_content.py
davep: So the code was wrong anyway, you mean, and the hint tweak is moot? If so I'll close this. Was just a passing tidy-up anyway.

willmcgugan: It may be correct, but the typing error stems from `AwaitComplete`

davep: Okay. I'll close this and flag up that AwaitComplete might need a wee bit more tinkering.

PR #4247, opened 2024-03-01
Adjust RichLog.highlighter type to allow for custom highlighters.

xavierog:
Without an explicit type, tools like mypy assume RichLog.highlighter must be a ReprHighlighter instead of a Highlighter.

Checklist:

- [x] Updated documentation
- [ ] Updated CHANGELOG.md (where appropriate) -- is this relevant/appropriate here?


willmcgugan (approved):
thanks

PR #4252, opened 2024-03-04
Add support for pseudo-classes in nested TCSS.

### LINKED ISSUE #4039: Nested TCSS interacts poorly with pseudo-classes in selectors.

rodrigogiraoserrao:
Fixes #4039 and supersedes #4163.

To be able to disambiguate between selector:pseudo-class and declaration:rule-value in nested TCSS (see #4039 for the original issue and #4163 for a first attempt at solving this) we establish that selectors with widget type names always start with an upper case letter A-Z or an underscore _ whereas declarations always start with a lower case letter a-z.

When a user creates a widget subclass that doesn't conform to this, we issue a 'SyntaxWarning' to let the user know. Because we do this with the standard module 'warnings', the warning can easily be supressed if the user has a good reason to create a widget subclass with a name starting with a lower case letter (which is valid Python, just unhelpful to Textual).


src/textual/widget.py
```diff
@@ -243,8 +247,35 @@ def __get__(self, obj: Widget, objtype: type[Widget] | None = None) -> str | Non
...
+        **kwargs: Any,
+    ) -> _WidgetMetaSub:
+        """Hook into widget subclass creation to check the subclass name."""
+        if not name[0].isupper() and not name.startswith("_"):
+            warnings.warn(
+                SyntaxWarning(
+                    f"Widget subclass {name!r} should be capitalised or start with '_'."
+                ),
+                stacklevel=2,
+            )
```
rodrigogiraoserrao, line 273: Because this is from the std library, suppressing these warnings is “easy”.
Even so, do we add something like a Boolean `textual.widget.SHOW_WIDGET_NAME_WARNING` to make it even easier for users to suppress the warning?

davep: (just looking at the code in the comment above; so might be missing context)

Is it intentional that `_foo` seen as valid? (I'm mostly guessing yes, but wanted to be sure).

rodrigogiraoserrao: `_foo` is valid and that's ok in the context in which this restriction was introduced.

Ideally, the wording of the warning (and of the comments around the code) should give that away.

I'm open to making it invalid as well if we think it makes sense to be a bit more restrictive with the goal of looking more coherent (given that `foo` raises the warning but `Foo` doesn't).

davep: Nah, I think it's fine; just wanted to be 100% sure that the whole package of code was doing what motivated it.

src/textual/css/tokenize.py
```diff
@@ -30,6 +30,19 @@
 VARIABLE_REF = r"\$[a-zA-Z0-9_\-]+"
 
 IDENTIFIER = r"[a-zA-Z_\-][a-zA-Z0-9_\-]*"
+SELECTOR_TYPE_NAME = r"[A-Z_][a-zA-Z0-9_]*"
+"""Selectors representing Widget type names should start with upper case or '_'.
+
+The fact that a selector starts with an upper case letter or '_' is relevant in the
+context of nested CSS to help determine whether xxx:yyy is a declaration + value or a
+selector + pseudo-class."""
+DECLARATION_NAME = r"[a-z][a-zA-Z0-9_\-]*"
```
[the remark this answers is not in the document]
rodrigogiraoserrao: Yes. The more permissive we are the more cases we can catch to provide a better error message.
E.g., with this regex we can catch `backgr0und: red` and ask if the user meant `background`.

src/textual/widget.py
```diff
@@ -243,8 +247,35 @@ def __get__(self, obj: Widget, objtype: type[Widget] | None = None) -> str | Non
         return title.markup
 
 
+_WidgetMetaSub = TypeVar("_WidgetMetaSub", bound="_WidgetMeta")
+
+
+class _WidgetMeta(_MessagePumpMeta):
```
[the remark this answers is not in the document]
rodrigogiraoserrao: Indeed we can.

rodrigogiraoserrao:
> Could I see screenshots of the error cases?

All error messages are what you would expect. There are tests that check we didn't regress & I added tests that check in nested contexts the messages are the same.

Here's a screenshot of the case that was unhelpful in the previous PR and which is now as good as it gets:

Screenshot 2024-03-05 at 10 46 03

willmcgugan (approved):

Issue #3908, opened 2023-12-20
SelectionList values must be hashable for no “good” reason


rodrigogiraoserrao:
As it stands, if you try to create a `SelectionList` with values that are not hashable, you get an error.
That's because we use a dictionary to keep track of the options that have been selected.
Changing this to a set of indices should lift the seemingly arbitrary restriction that the values must be hashable while also making it straightforward & fast to do the selection bookkeeping and to return a sorted list of selections to the user.

TomJGooding:
I might be completely off base here, but is this now resolved with the recent changes to `OptionList`?

rodrigogiraoserrao:
You had me wondering for a minute.
It's not solved because it's the selection list that contains this dictionary: `self._selected: dict[SelectionType, None] = {}` inside `SelectionList.__init__`.

davep:
As of the time of writing #4256 adds code that further leans on that hashable nature; either this should be the moment to tackle this, or when this does get tackled that further change should be taken into account.

rodrigogiraoserrao:
@davep On the one hand I'm not aware of how you use the hashable nature of selection list values in the linked PR; on the other hand this issue isn't to say we must absolutely 100% lift this restriction!

When I opened the issue it looked simple enough to lift the restriction by changing a dictionary to use indices into a list instead of using values directly.

Maybe your PR is reasonable enough that this becomes a documented restriction and this issue can be closed. I'd be fine with that. 🤷

davep:
In this PR, given that values were considered to be valid as dictionary keys anyway, it leans into that when it needs to create a method of mapping values back to the index of an item; so it's pretty much similar to what you highlight above only it actually has an associated value too.

I sense you're right that it deserves cleaning up unless there's a penalty in doing so; but I thought it prudent to mention it here for anyone who picks this up.

rodrigogiraoserrao:
Understood. (I think a good paper trail is always a good idea.)

PR #4256, opened 2024-03-05
Message posting clean-up

davep:
Assorted tweaks and fixes as fallout from the work done in #3869. Where possible/sensible extra tests have been added too.

The most significant change here, I think, is to `SelectionList`, which needed a little bit of extra work because it became necessary to find back the index of a value. *(while this is in review I'm going to give that part one last run over as I want to be double sure it's a good approach -- I think this also possibly either provides a "good" reason for #3908 or might actually make it less good)*


src/textual/widgets/_selection_list.py
```diff
@@ -679,6 +699,16 @@ def add_options(
...
                 )
+
+        # Add the new items to the value mappings.
+        self._values = {
+            **self._values,
+            **{
+                option.value: index
+                for index, option in enumerate(cleaned_options, start=self.option_count)
+            },
+        }
```
rodrigogiraoserrao, line 710: Isn't `.update` better in terms of speed while being the same in terms of readability?

rodrigogiraoserrao (approved):
Looks good to me.
I'm not going to pretend I went through all the code of all widgets to make sure you didn't miss anything, but the changes you did make look good.

src/textual/widgets/_selection_list.py
davep: Off the top of my head I don't know the speed difference; it feels like you are right, do you have a reference?

rodrigogiraoserrao: I don't have a reference, no, it's just intuition maybe.
I was just surprised you spelled this out 🤷

willmcgugan: I'd probably go with update. It would be faster, but doubt the difference would be measurable. It's more that `update` expresses intent better.

willmcgugan (approved):

PR #4264, opened 2024-03-06
Clean render return type

### LINKED ISSUE #4063: Tidy the return type of `App.render`

davep:
This PR tweaks the type hinting for the `render` methods of various widgets, to ensure they follow the same hinting as is used [in our documentation](https://textual.textualize.io/guide/widgets/#custom-widgets); the aim being to make the code a little less surprising to the reader and also to ensure that any type hint information shown in an IDE, or auto-completed when inheriting from a Textual widget and overriding the `render` method, matches what would be expected.

Fixes #4063.

### DIFF
--- a/src/textual/widget.py
+++ b/src/textual/widget.py
@@ -32,14 +32,17 @@ from rich.console import (
     ConsoleRenderable,
     JustifyMethod,
     RenderableType,
-    RenderResult,
-    RichCast,
 )
+from rich.console import RenderResult as RichRenderResult
+from rich.console import RichCast
 from rich.measure import Measurement
 from rich.segment import Segment
 from rich.style import Style
 from rich.text import Text
 from typing_extensions import Self
+
+if TYPE_CHECKING:
+    from .app import RenderResult
 
 from . import constants, errors, events, messages
 from ._animator import DEFAULT_EASING, Animatable, BoundAnimator, EasingFunction
@@ -152,7 +155,7 @@ class _Styled:
 
     def __rich_console__(
         self, console: "Console", options: "ConsoleOptions"
-    ) -> "RenderResult":
+    ) -> "RichRenderResult":
         style = console.get_style(self.style)
         result_segments = console.render(self.renderable, options)
 
@@ -3375,7 +3378,7 @@ class Widget(DOMNode):
             with self.app.batch_update():
                 yield
 
-    def render(self) -> RenderableType:
+    def render(self) -> RenderResult:
         """Get text or Rich renderable for this widget.
 
         Implement this for custom widgets.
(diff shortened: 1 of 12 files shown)


src/textual/widget.py
```diff
@@ -32,15 +32,18 @@
     ConsoleRenderable,
     JustifyMethod,
     RenderableType,
-    RenderResult,
-    RichCast,
 )
+from rich.console import RenderResult as RichRenderResult
```
rodrigogiraoserrao, line 36: Why are these two imports separated from the other `rich.console` imports?
Why not the suggestion I show below?

```py
from rich.console import (
    Console,
    ConsoleOptions,
    ConsoleRenderable,
    JustifyMethod,
    RenderableType,
    RenderResult as RichRenderResult,
    RichCast,
)
```

src/textual/widgets/_input.py
```diff
@@ -47,7 +52,7 @@ def __init__(self, input: Input, cursor_visible: bool) -> None:
 
     def __rich_console__(
         self, console: "Console", options: "ConsoleOptions"
-    ) -> "RenderResult":
+    ) -> "RichRenderResult":
```
rodrigogiraoserrao, line 55:
```suggestion
    ) -> RichRenderResult:
```

rodrigogiraoserrao (approved):
Looking at this “simple” PR, I can't help but feel like you spent a _lot_ of time making sure you were getting all of the instances and to make sure you don't mix the app and Rich's `RenderResult` 😬

src/textual/widget.py
davep: I'll let you have that argument with black and/or whatever else it is that rewrote my code from that.

davep:
> Looking at this “simple” PR, I can't help but feel like you spent a lot of time making sure you were getting all of the instances and to make sure you don't mix the app and Rich's RenderResult 😬

Yup, that was fun. As I worked through this too I also got to thinking that, in retrospect, it might have been nicer if `RenderResult` and `ComposeResult` originated in `textual.types` too. Possibly. Maybe. Perhaps.

src/textual/widget.py
rodrigogiraoserrao: Ah ok, I understand 🤣

Issue #4263, opened 2024-03-06
Input and TextArea have cursor blinking even though terminal window is not focused


willmcgugan:
### Discussed in https://github.com/Textualize/textual/discussions/4259

<div type='discussions-op-text'>

<sup>Originally posted by **villekr** March  5, 2024</sup>
Following simple app has Input and TextArea widgets. Focus transitions between them as expected. However, when I **focus some other window** than iTerm2 running textual app then cursor still keeps on blinking on the previously focused widget. I would expect that by default focus is removed alltogether from app when other window gets focus. Or is this something app needs to be specifically configured? 

<del>I added also _on_app_focus annd _on_app_blur event handlers for debugging, but they are not invoked. Maybe these are only relevant for textual-web based on comments (# Required by textual-web to manage focus in a web page.)</del>

As this might be platform/terminal related here's the configuration I have.

macOS Sonoma: Version 14.3
iTerm2: Build 3.4.23
Python 3.12
Textual 0.52.1

```
from textual import events
from textual.app import App, ComposeResult
from textual.containers import Vertical
from textual.widgets import Input, TextArea


class AppFocus(App):

    def compose(self) -> ComposeResult:
        with Vertical():
            yield TextArea()
            yield Input()


if __name__ == "__main__":
    AppFocus().run()

```</div>

willmcgugan:
We should implement these ANSI codes. Should be straightforward in theory. Enable the escape sequences, and convert them to AppFocus and AppBlur.

Talk to @willmcgugan if more clarification is needed.

davep:
With an eye on the motivation for this: it would seem that there's benefit to the widgets in the focus chain knowing when `AppFocus` and `AppBlur` happens; I wonder if perhaps those events should be posted to the focused widget and allowed to bubble?

willmcgugan:
What is the benefit?

davep:
Just thinking about if there's utility for it when it comes to a random developer wanting to implement functionality on a widget that's related to having focus or not, and wanting to take into account if there emulator window has focus (while I recognise that it's not done with `Focus` and `Blur` with `Input` and `TextArea`, I'm thinking about any situations we've not considered yet that might be akin to the original motivation for the question that resulted in this issue).

willmcgugan:
The behaviour (in Textual-web) is that when the window loses focus, the focused widget is blurred. When the window regains focus, the previously blurred widget is focused again.

[...]

PR #4265, opened 2024-03-06
Enable `AppFocus` and `AppBlur` in terminal emulators

davep:
This enables support for receiving and handling FocusIn and FocusOut sequences, and turns then into `AppFocus` and `AppBlur` events. Support for enabling this is added to both the "Linux" and Windows drivers. From now on, when a terminal emulator window that supports it sends the FocusIn and FocusOut sequences the app will receive `AppFocus` and `AppBlur`.

An automatic effect of this now is that, when the window loses focus, focus will be removed from the widget that has focus. This was already an established feature of Textual applications running via textual-web; it just now extends it to the drivers used on GNU/Linux, macOS and Windows.

The `AppFocus` and `AppBlur` events could be useful too.

https://github.com/Textualize/textual/assets/28237/33bf3b9e-ab05-4c8c-89de-01da79cccace

It should be noted that this is technically a breaking change.

Implements #4263 (which in turn stems from #4259).

It also fixes a problem with the current implementation of `AppFocus` and `AppBlur` that is mentioned in a comment in #4263.

---

A natural question is going to be, how widely supported is this? So far the following have been tested and shown to support this:

- macOS
  - Kitty
  - iTerm2
  - Contour
  - terminal.app
  - WezTerm
  - rio
  - Tabby
  - VSCode Terminal
- GNU/Linux
  - GNOME Terminal
  - alacritty (via @TomJGooding)
  - urxvt (via @TomJGooding)
  - xterm
  - Konsole
- Windows 11
  - Terminal
  - Hyper
  - conemu (via @TomJGooding)

It has been noted by @TomJGooding that iTerm2 has a setting for turning this off[^1]; when turned off the enable/disable sequence appears to be swallowed and ignored.

It would appear that vterm under Emacs doesn't support this, but running my test app had the same effect as running it under iTerm with support turned off (that is, the enable/disable sequences seem to be swallowed and ignored).

**NOTE:** This is marked as and should be treated as a breaking change. This can potentially affect application styling in a way that would not have been an issue before. For example, before this change, this application would always look the same no matter if the terminal emulator it was within had focus or not:

```python

class OneWidgetApp(App[None]):

    CSS = """
    Screen {
        align: center middle;
    }

    Input:focus {
        border: round red;
        width: 50%;
        background: pink;
    }
    """

        yield Input(placeholder="Please give your answer")

if __name__ == "__main__":
    OneWidgetApp().run()
```

from now on the layout of the application will look very different when focus moves away form the window.

[^1]: Because of course it does! What doesn't iTerm have a setting for?!?

### DIFF
--- a/src/textual/_xterm_parser.py
+++ b/src/textual/_xterm_parser.py
@@ -20,6 +20,8 @@ _re_terminal_mode_response = re.compile(
 )
 _re_bracketed_paste_start = re.compile(r"^\x1b\[200~$")
 _re_bracketed_paste_end = re.compile(r"^\x1b\[201~$")
+_re_focusin = re.compile(r"^\x1b\[I$")
+_re_focusout = re.compile(r"^\x1b\[O$")
 
 
 class XTermParser(Parser[events.Event]):
@@ -201,6 +203,14 @@ class XTermParser(Parser[events.Event]):
                     sequence = new_sequence
 
                     self.debug_log(f"sequence={sequence!r}")
+
+                    if _re_focusin.match(sequence):
+                        on_token(events.AppFocus())
+                        break
+
+                    if _re_focusout.match(sequence):
+                        on_token(events.AppBlur())
+                        break
 
                     bracketed_paste_start_match = _re_bracketed_paste_start.match(
                         sequence
--- a/src/textual/_xterm_parser.py
+++ b/src/textual/_xterm_parser.py
@@ -20,9 +20,11 @@ _re_mouse_event = re.compile("^" + re.escape("\x1b[") + r"(<?[\d;]+[mM]|M...)\Z"
 _re_terminal_mode_response = re.compile(
     "^" + re.escape("\x1b[") + r"\?(?P<mode_id>\d+);(?P<setting_parameter>\d)\$y"
 )
-_re_bracketed_paste_start = re.compile(r"^\x1b\[200~$")
-_re_bracketed_paste_end = re.compile(r"^\x1b\[201~$")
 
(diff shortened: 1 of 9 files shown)


src/textual/_xterm_parser.py
```diff
@@ -20,6 +20,8 @@
 )
 _re_bracketed_paste_start = re.compile(r"^\x1b\[200~$")
 _re_bracketed_paste_end = re.compile(r"^\x1b\[201~$")
+_re_focusin = re.compile(r"^\x1b\[I$")
```
willmcgugan, line 23: It doesn't look like this needs to be a regex. Could it not be a simple string?

Ditto for the bracketed paste stuff. Unless I'm missing something, it looks like they could also just be strings.

davep: I was going to ask you that very question in the morning, oddly enough. Couldn't for the life of me see why the bracketed paste checking used a regex but followed the pattern with a mental note to ask why it was done that way (although looking at fe151a7f25cfd7f1134ebafbddc7eeade1c18ccb it seems I'd have been asking the wrong person anyway).

If there was no reason I'll make them simple strings; makes way more sense to me.

willmcgugan: Probably just following a pattern. But yeah, let's go with strings. And if you wouldn't mind changing the bracketed paste ones while you are there...

davep: Will do.

willmcgugan (approved):

davep:
Adding for future reference if someone comes looking: this is also supported by `tmux` if `focus-events` is set to `on`.

Issue #4054, opened 2024-01-20
Revised progress bar ETA


willmcgugan:
The ProgressBar "ETA" is quite naive. I *think* it assumes a constant speed, which is rarely the case for things you would want a progress bar for.

We need a smarter way of calculating the ETA which emphasizes more recent data points. There are a few ways of doing this. You could see the Rich progress bars for an examples, but there may also be better algorithms. Suggesting asking Google or ChatGPT.

The implementation could also use another look. It adds a Horizontal container, where the widget itself could have a horizontal layout. 

1d

davep:
While making a start on this I ran into some confusing results and ended up uncovering #4096; so I think this will result in a revamp of how `ProgressBar` does the calculations even in the most simplistic of situations.

Some form of "please reset this `ProgressBar` for reuse" method might also be a good idea.

davep:
WiP on this has addressed #4096, but also uncovered a further issue where the `ETAStatus` appeared to be updating when it wasn't necessary; this has now been identified and fixed too.

An updated method of calculating the ETA is now in place and seems to produce reasonable results; for constant-time progress the timer looks spot on; and for accelerating, decelerating and reasonably random progress the results look good too, and generally more/less optimistic/pessimistic where appropriate.

Work on this is having further knock-on effects on how the documentation screen shots are generated, and also their use in snapshot tests; this will likely need some more work on how the docs for `ProgressBar` are handled, and might require removal of that code from snapshot tests. If the latter happens we'll need to do a bit of work on producing unit tests for the ETA calculation (which might require isolating the ETA code somewhat).

A question has also been raised about the newer approach taken to slide a window over the updates to perform the calculation.

+1d

PR #4271, opened 2024-03-07
improved eta

willmcgugan:
Improved ETA for progress bar

- Reduces refreshes of the ProgressBar. Previously the widget would repaint every time `update` was called. Even if the output remained the same.
- Improved calculation of ETA based on recent samples. ETA is extrapolated based on this speed.
- Added a `Clock` object to return relative time, and allow for easier mocking. This is public, but as yet undocumented. This might become a public interface if we want devs to be able to swap out algorithms.

Closes https://github.com/Textualize/textual/issues/4054

### DIFF
--- a/src/textual/eta.py
+++ b/src/textual/eta.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import bisect
+from math import ceil
 from operator import itemgetter
 from time import monotonic
 
@@ -112,4 +113,4 @@ class ETA:
         time_since_sample = time - recent_time
         remaining = 1.0 - (recent_progress + speed * time_since_sample)
         eta = max(0, remaining / speed)
-        return round(eta)
+        return ceil(eta)
--- a/src/textual/eta.py
+++ b/src/textual/eta.py
@@ -64,12 +64,12 @@ class ETA:
             # Keep at least 10 samples
             return
         prune_time = self._samples[-1][0] - self.estimation_period
-        index = bisect.bisect_left(self._samples, prune_time)
+        index = bisect.bisect_left(self._samples, (prune_time, 0))
         del self._samples[:index]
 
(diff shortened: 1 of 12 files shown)


src/textual/eta.py
```diff
@@ -0,0 +1,126 @@
...
+
+import bisect
+from math import ceil
+from time import monotonic
+
+import rich.repr
+
+
+@rich.repr.auto(angular=True)
+class ETA:
```
davep, line 11: Just because I'll never be able to live with myself if I *don't* mention it on record.... Textual no-TLA rule!

src/textual/eta.py
```diff
@@ -0,0 +1,126 @@
...
+
+    @property
+    def first_sample(self) -> tuple[float, float]:
+        """First sample, or `None` if no samples."""
+        assert self._samples, "Assumes samples not empty"
+        return self._samples[0]
+
+    @property
+    def last_sample(self) -> tuple[float, float]:
+        """Last sample, or `None` if no samples."""
```
davep, line 40: The docstring here doesn't match the typing.

src/textual/widgets/_progress_bar.py
```diff
@@ -188,79 +162,23 @@ class ETAStatus(Label):
         content-align-horizontal: right;
     }
     """
+    eta: reactive[float | None] = reactive[Optional[float]](None)
+    """Estimated number of seconds till completion."""
```
davep, line 166: Perhaps mention that `None` is an option and why.

davep (approved):
Couple or so things worth a wee note and perhaps a tweak.

src/textual/eta.py
willmcgugan: This is an established abbreviation, where there is no doubt about how it is abbreviated.

There is precedence: "css" and not "cascading_style_sheet".

PR #4276, opened 2024-03-09
Compute fast

willmcgugan:
Noticed a quick optimization win. By recording the compute methods on class creation we can avoid work whenever a reactive is set.


willmcgugan:
#executivedecision Feel free to review retrospecitvely.

rodrigogiraoserrao (commented):
I think your optimisation could be even faster if you removed what looks like a bit of redundant work.

PR #4285, opened 2024-03-12
Ensure TabbedContent only handles tab messages intended for it

### LINKED ISSUE #4233: `TabbedContent` message mixup if `Tabs` inside a `TabPane`
I suspect there are a few issues here (or rather a few operations that might suffer from a similar problem) but I think this example perfectly illustrates the confusion that could arise. Consider the following code:

```python
from textual import on
from textual.widgets import TabbedContent, Tabs, TabPane, Button

class TabbedContentMessageStealingApp(App[None]):

    CSS = """
    TabbedContent, TabPane {
        height: 1fr;
        width: 1fr;
    }
    """

        with TabbedContent():
            with TabPane("Test"):
                yield Button("Disable", id="disable")
                yield Tabs("One", "Two", "Three", id="test-tabs")

    @on(Button.Pressed, "#disable")
    def disable_test(self) -> None:
        self.query_one("#test-tabs", Tabs).disable("tab-1")

if __name__ == "__main__":
    TabbedContentMessageStealingApp().run()
```

There is a `Tabs` within a `TabPane` within `TabbedContent`. There is a `Button` that, when pressed, will disable the first `Tab` of the `Tabs` that the developer has added. When pressed, you would expect the `Tab` with the title `One` to grey out and become unusable.

Instead what happens is *every* tab becomes disabled, as does the `Button` too:

Screenshot 2024-02-28 at 11 33 14

This will, of course, be due to the message leaking up to the `TabbedContent` and its first tab getting disabled due to it having a similar ID.

Compare what happens with this code, where I give the `TabPane` an ID that avoids the clash.

```python
from textual import on
from textual.widgets import TabbedContent, Tabs, TabPane, Button

class TabbedContentMessageStealingApp(App[None]):

    CSS = """
    TabbedContent, TabPane {
        height: 1fr;
        width: 1fr;
    }
    """

        with TabbedContent():
            with TabPane("Test", id="avoid-me"):
                yield Button("Disable", id="disable")
                yield Tabs("One", "Two", "Three", id="test-tabs")

    @on(Button.Pressed, "#disable")
    def disable_test(self) -> None:
        self.query_one("#test-tabs", Tabs).disable("tab-1")

if __name__ == "__main__":
    TabbedContentMessageStealingApp().run()
```

Screenshot 2024-02-28 at 11 35 19

It feels like something should be done to isolate the `TabbedContent` from any `Tabs` that might be in one or more of its panes.

davep:
Fixes #4233.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Markdown component classes weren't refreshed when watching for CSS https://github.com/Textualize/textual/issues/3464
 - Rename `Switch.action_toggle` to `action_toggle_switch` to fix clash with `DOMNode.action_toggle` https://github.com/Textualize/textual/issues/4262
 - Fixed `OptionList.OptionHighlighted` leaking out of `Select` https://github.com/Textualize/textual/issues/4224
+- Fixed `Tab` enable/disable messages leaking into `TabbedContent` https://github.com/Textualize/textual/issues/4233
 
 ### Changed
 
--- a/src/textual/widgets/_tabbed_content.py
+++ b/src/textual/widgets/_tabbed_content.py
@@ -611,6 +611,8 @@ class TabbedContent(Widget):
 
     def _on_tabs_tab_disabled(self, event: Tabs.TabDisabled) -> None:
         """Disable the corresponding tab pane."""
+        if event.tabs.parent is not self:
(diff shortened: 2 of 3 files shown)


rodrigogiraoserrao (approved):
You're fixing all the tabs bugs!

davep:
Oh, I'm sure I haven't got *all* of them! ;-)

PR #4291, opened 2024-03-13
Ensure scrollbars get Show/Hide messages

### LINKED ISSUE #4274: Scrollbars may trap the mouse if hidden while the user is scrolling
# Current behaviour

If a scrollable widget is programmatically hidden while the user is scrolling by dragging the scrollbar handle, all mouse events are seemingly blocked (they are actually trapped by the invisible scrollbar).

This MRE reproduces the issue by programmatically collapsing a `Collapsible` that contains a scrollable `RichLog`:

```python
#!/usr/bin/env python3
from textual.widgets import Label, Collapsible, RichLog

INSTRUCTIONS = """[u]Instructions:[/u]
1. Using your mouse, grab the scrollbar handle of the currently focused Collapsible
2. Move it up and down -- [b]do NOT release the mouse button[/b]
3. A few seconds later, the timer should collapse the Collapsible
4. Release the mouse button
5. Try clicking somewhere else, e.g. on the second collapsible
[u]Result:[/u] the mouse is "trapped" by the hidden scrollbar.
[i]Bonus:[/i] Using Tab and Enter, expand the Collapsible and move your mouse.
"""
FILLER = 'Lorem ipsum dolor sit amet, consectetur adipiscing elit\n'
COLLAPSIBLE_TITLE='Collapse me by pressing Enter while moving the scrollbar with your mouse!'

class MouseTrap(App):
		yield Label(INSTRUCTIONS)
		for _ in range(2):
			with Collapsible(title=COLLAPSIBLE_TITLE, collapsed=False):
				richlog = RichLog()
				richlog.styles.height = 15
				richlog.write(FILLER * 150)
				yield richlog

	def on_ready(self) -> None:
		self.set_timer(7, self.collapse)

	def collapse(self) -> None:
		self.query('Collapsible').first(Collapsible).collapsed = True

if __name__ == "__main__":
	app = MouseTrap()
	app.run()
```

# Expectations

The scrollbar handle should be "released" as soon as the scrollbar is hidden so that mouse events resume their normal course.

# Textual Diagnostics

## Versions

| Name    | Value  |
|---------|--------|
| Textual | 0.52.1 / commit f55610ee6b98715961926404b61cd03c5fb4dabe |
| Rich    | 13.7.0 |

## Python

## Operating System

## Terminal

| Name                 | Value         |
|----------------------|---------------|
| Terminal Application | tmux (3.4)    |
| TERM                 | tmux-256color |
| COLORTERM            | truecolor     |
| FORCE_COLOR          | *Not set*     |
| NO_COLOR             | *Not set*     |

## Rich Console options

| size           | width=283, height=58 |
| max_width      | 283                  |
| max_height     | 58                   |

davep:
The idea here being that while scrollbars are attached to widgets, they exist outwith of the DOM, and so don't seem to take part in the usual flow of show/hide messaging. This change checks, when posting a show/hide message, if the widget receiving the message has scrollbars and if it does the message is also sent to them too.

Fixes #4274.

### DIFF
--- a/src/textual/widget.py
+++ b/src/textual/widget.py
@@ -3706,7 +3706,17 @@ class Widget(DOMNode):
             self.scroll_page_right()
             event.stop()
 
+    def _on_show(self, event: events.Show) -> None:
+        if self.show_horizontal_scrollbar:
+            self.horizontal_scrollbar.post_message(event)
+        if self.show_vertical_scrollbar:
+            self.vertical_scrollbar.post_message(event)
+
     def _on_hide(self, event: events.Hide) -> None:
+        if self.show_horizontal_scrollbar:
+            self.horizontal_scrollbar.post_message(event)
+        if self.show_vertical_scrollbar:
+            self.vertical_scrollbar.post_message(event)
         if self.has_focus:
             self.blur()
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Fixed `OptionList.OptionHighlighted` leaking out of `Select` https://github.com/Textualize/textual/issues/4224
 - Fixed `Tab` enable/disable messages leaking into `TabbedContent` https://github.com/Textualize/textual/issues/4233
 - Fixed a style leak from `TabbedContent` https://github.com/Textualize/textual/issues/4232
+- Fixed active hidden scrollbars not releasing the mouse https://github.com/Textualize/textual/issues/4274
 
 ### Changed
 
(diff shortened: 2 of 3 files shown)


src/textual/widget.py
```diff
@@ -3706,7 +3706,17 @@ def _on_scroll_right(self, event: ScrollRight) -> None:
             self.scroll_page_right()
             event.stop()
 
+    def _on_show(self, event: events.Show) -> None:
+        if self.show_horizontal_scrollbar:
+            self.horizontal_scrollbar.post_message(event)
+        if self.show_vertical_scrollbar:
+            self.vertical_scrollbar.post_message(event)
```
rodrigogiraoserrao, line 3713: This should obviously be in the reverse order 🙄

rodrigogiraoserrao (approved):
LGTM 🐍 🚀

src/textual/widget.py
rodrigogiraoserrao: I needed to complain about something.

PR #4295, opened 2024-03-14
Fix `TextArea` holding on to focus when hidden while doing a mouse selection

### LINKED ISSUE #4292: Hiding a `TextArea` while mouse-based selection is happening can trap the cursor
*Initially discovered and reported in a comment in #4274 by @TomJGooding*

Somewhat following on from #4274, although very likely an easier fix (I think it just needs a `Hide` handler to release the mouse and do any other housekeeping necessary to `TextArea` selection), if you start a mouse selection within `TextArea` and then hide the `TextArea`, the mouse will be trapped and won't respond anywhere else.

To illustrate:

```python
from pathlib import Path
from textual.widgets import Collapsible, TextArea


class GreedyScrollbarApp(App[None]):
    BINDINGS = [("f1", "collapse")]

    CSS = """
    TextArea {
        height: 10;
    }
    """

        for n in range(5):
            with Collapsible(title=f"This is collapsible #{n}", collapsed=bool(n)):
                yield TextArea(Path(__file__).read_text(), language="python")

    def action_collapse(self) -> None:
        target.collapsed = not (target := self.query(Collapsible).first()).collapsed


if __name__ == "__main__":
    GreedyScrollbarApp().run()
```

Start a mouse selection in the first `TextArea`, press <kbd>F1</kbd> while still selecting with the mouse, then notice nothing else will respond to mouse clicks.

davep:
Discovered by @TomJGooding and similar to #4274 -- if a `TextArea` was in a mouse selection state and the `TextArea` was hidden, the selection didn't get ended and so the mouse remained captured by the (now hidden) `TextArea`.

Fixes #4292.

### DIFF
--- a/src/textual/widgets/_text_area.py
+++ b/src/textual/widgets/_text_area.py
@@ -1498,12 +1498,20 @@ TextArea {
             selection_start, _ = self.selection
             self.selection = Selection(selection_start, target)
 
-    async def _on_mouse_up(self, event: events.MouseUp) -> None:
+    def _end_mouse_selection(self) -> None:
         """Finalize the selection that has been made using the mouse."""
         self._selecting = False
         self.release_mouse()
         self.record_cursor_width()
         self._restart_blink()
+
+    async def _on_mouse_up(self, event: events.MouseUp) -> None:
+        """Finalize the selection that has been made using the mouse."""
+        self._end_mouse_selection()
+
+    async def _on_hide(self, event: events.Hide) -> None:
+        """Finalize the selection that has been made using the mouse when thew widget is hidden."""
+        self._end_mouse_selection()
 
     async def _on_paste(self, event: events.Paste) -> None:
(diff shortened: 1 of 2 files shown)


src/textual/widgets/_text_area.py
```diff
@@ -1498,13 +1498,21 @@ async def _on_mouse_move(self, event: events.MouseMove) -> None:
...
         self._selecting = False
         self.release_mouse()
         self.record_cursor_width()
         self._restart_blink()
 
+    async def _on_mouse_up(self, event: events.MouseUp) -> None:
+        """Finalize the selection that has been made using the mouse."""
+        self._end_mouse_selection()
+
+    async def _on_hide(self, event: events.Hide) -> None:
```
rodrigogiraoserrao, line 1512: These two should definitely be in reverse order. 🙄

rodrigogiraoserrao (approved):
LGTM 🐍 🚀

PR #4297, opened 2024-03-14
Let mouse scroll events pass through disabled widgets

### LINKED ISSUE #4242: Disabled widgets prevent mouse scrolling
# Context

My application used to disable entire Collapsible Contents, which is how I noticed this issue in the first place.

Notes:

- in this issue, "scrolling" and "scroll using the mouse" are both short for "scrolling using the mouse scroll wheel or any similar device"
- this issue is similar to https://github.com/Textualize/textual/issues/1912 but applies to enabled containers.

# Expectations

When an enabled scrollable container contains disabled widgets, disabled widgets should not prevent scrolling the container.

# Current behaviour

The MRE below reflects the issue. It alternates enabled and disabled widgets in a way that prevents effective scrolling.

textual-scroll-mre

```python
#!/usr/bin/env python3

class MyBoringWidget(Static, can_focus=True):
    DEFAULT_CSS = "MyBoringWidget { height: 90%; background: $primary; border: tall yellow; }"

class ScrollMRE(App):
    DEFAULT_CSS = "Static { text-align: center; }"
        instructions = 'Place your cursor in the middle of the app and try scrolling to the bottom using the mouse'
        yield Static(instructions)
        for i in range(10):
            yield MyBoringWidget(disabled=bool(i%2))
            yield Static(instructions)

if __name__ == "__main__":
    app = ScrollMRE()
    app.run()
```

# Textual Diagnostics

## Versions

| Name    | Value  |
|---------|--------|
| Textual | 0.52.1 |
| Rich    | 13.7.0 |

## Python

## Operating System

## Terminal

| Terminal Application | tmux (3.4)     |

## Rich Console options

| size           | width=232, height=54 |
| max_width      | 232                  |
| max_height     | 54                   |

davep:
All mouse events are stopped when they hit a disabled widget; this is mostly correct, except for mouse scroll events (scroll wheel, trackpad swipes, etc). This PR ensures that those particular events always pass through.

Fixes #4242.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Fixed a style leak from `TabbedContent` https://github.com/Textualize/textual/issues/4232
 - Fixed active hidden scrollbars not releasing the mouse https://github.com/Textualize/textual/issues/4274
 - Fixed the mouse not being released when hiding a `TextArea` while mouse selection is happening https://github.com/Textualize/textual/issues/4292
+- Fix mouse scrolling not working when mouse cursor is over a disabled child widget https://github.com/Textualize/textual/issues/4242
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


rodrigogiraoserrao (approved):
LGTM 🐍 🚀

PR #4298, opened 2024-03-14
Enable unsetting active tab pane / tab

### LINKED ISSUE #4241: `Tabs.Cleared` documentation a little ambiguous
The `Tabs.Cleared` message [is documented](https://textual.textualize.io/widgets/tabs/#textual.widgets._tabs.Tabs.Cleared) as being *"Sent when there are no active tabs. This can occur when Tabs are cleared, or if all tabs are hidden"*. This suggests that `Tabs.cleared` or hiding every single `Tab` are the only routes to causing this message to be posted.

From what I can see simply setting `Tabs.active` to an empty string will cause this message to be posted too. The documentation should probably make it clear that this will also be a cause.

A followup question, which might need to spin off into another issue, is: does it make sense that `Tabs.active` can be set (outwith of the workings of `Tabs`) to an empty string? And if the answer is *yes*, the next question would be *"why does doing this cause `Tabs` to behave as if the last-active-tab were still the active tab?"*

rodrigogiraoserrao:
Goes over `TabbedContent` and `Tabs` and allows unsetting their respective tab pane/tab.
Although we currently don't know if it makes a lot of sense (Will did mention a custom widget/renderable/thingy situation where it could make sense to have no active tab) we won't prevent the programmers to make use of this flexibility in their apps.

Went through both widgets to make sure the `Cleared` messages are sent in the correct places and not in duplicate.

Also made a small change so that if there's no active tab, then pressing right/left will respectively activate the first or last tab.

Fixed the documentation as reported in #4241.

Fixes #4241.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   - When the terminal window loses focus, the currently-focused widget will also lose focus.
   - When the terminal window regains focus, the previously-focused widget will regain focus.
 - TextArea binding for <kbd>ctrl</kbd>+<kbd>k</kbd> will now delete the line if the line is empty https://github.com/Textualize/textual/issues/4277
+- The active tab (in `Tabs`) / tab pane (in `TabbedContent`) can now be unset https://github.com/Textualize/textual/issues/4241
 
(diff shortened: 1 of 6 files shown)


davep (approved):
I have full confidence that this will be the last change to `TabbedContent` ever.

PR #4300, opened 2024-03-14
fix(datatable): change max-height back to 100%

### LINKED ISSUE #4286: `DataTable` scrolling issues with `max-height: 100vh`
This follows on from discussion #4281 posted by @morelop about unexpected scrollbar behaviour with the `DataTable`.

Here's a slightly modified example from this discussion, where I've added borders to help show the container and table sizing. If you resize your terminal, the horizontal scrollbar isn't visible but also because of the `Footer` you can't scroll to the final row into view.

image

I found this a bit surprising as I thought the DataTable had a `max-height: 100%` as per the FAQ - https://textual.textualize.io/FAQ/#why-doesnt-the-datatable-scroll-programmatically. But it seems this was changed to `100vh` in #3566, because "the previous value of 100% broke with horizontal layouts."

If you change the DataTable back to `max-height: 100%`, the example above works as expected. 

- Is the current behaviour with `max-height: 100vh` expected? Contrary to the note in the PR, this actually seems to break layouts based on this simple example.
- If this _is_ the expected behaviour, should the FAQ be updated?

```python
from textual.widgets import DataTable, Footer


class ExampleApp(App):
    CSS = """
    Horizontal {
        border: solid red;
    }

    DataTable {
        border: solid pink;
    }
    """

        with Horizontal():
            yield DataTable()
        yield Footer()

    def on_mount(self) -> None:
        for col in range(0, 30):
            table.add_column(f"col{col}")
        for row in range(0, 30):
            data = [f"{row}x{col}" for col in range(0, 30)]
            table.add_row(*data)


if __name__ == "__main__":
    app = ExampleApp()
    app.run()
```

TomJGooding:
The `max-height` of the `DataTable` was changed to 100vh rather than 100% in #3566, because at the time this caused issues with auto height containers, as described in #2975.

However this issue was later fixed in #3814. This PR changes the `max-height` of the `DataTable` back to 100%, as 100vh will actually break common layouts and seems no longer required to workaround this auto height issue.

Fixes #4286.

- [ ] Updated documentation


willmcgugan:
Thanks

PR #4305, opened 2024-03-18
watch with async

willmcgugan:
A potential resolution for https://github.com/Textualize/textual/discussions/4268#discussioncomment-8823867

I wasn't able to reproduce this, but I think it was down to setting the title reactive from a task, when the widget wasn't fully mounted.

By making the watch method async, it should ensure that the title is updated after mount.


willmcgugan:
#executivedecision

PR #4313, opened 2024-03-19
Fix a crash in `run_action` when an action is an empty tuple

### LINKED ISSUE #4248: Malformed @click links crash Textual application
# Context

I was just trying out the `@click` syntax.

# Expected behaviour

If I create e.g. a Label with incorrect markup, I should get a MarkupError immediately.

# Current behaviour

Some `@click` syntax do not raise a MarkupError and make Textual applications crash when clicked, namely:
- `[@click]`
- `[@click=]`
- `[@click=()]`

MRE:
```python
#!/usr/bin/env python3

class ClickMRE(App):
        yield Label("[@click]click me and crash[/]") # ValueError in run_action on click
        yield Label("[@click=]click me and crash[/]") # ValueError in run_action on click
        yield Label("[@click=()]click me and crash[/]") # ValueError in run_action on click
        #yield Label("[@click=(,)]click me and crash[/]") # MarkupError in render at startup
        yield Label("[@click=foobar]click me[/]") # ok, nothing happens
        yield Label("[@click=foobar()]click me[/]") # ok, nothing happens
        yield Label("[@click=toggle_dark]click me[/]") # ok, works fine
        yield Label("[@click=toggle_dark()]click me[/]") # ok, works fine

if __name__ == "__main__":
    app = ClickMRE()
    app.run()
```

# Textual Diagnostics

## Versions

| Name    | Value  |
|---------|--------|
| Textual | 0.52.1 |
| Rich    | 13.7.0 |

## Python

## Operating System

## Terminal

| Terminal Application | tmux (3.4)     |

## Rich Console options

| size           | width=212, height=34 |
| max_width      | 212                  |
| max_height     | 34                   |

davep:
Fixes #4248.


src/textual/app.py
```diff
@@ -3008,11 +3006,16 @@ async def _broker_event(
...
+        if isinstance(action, str) or (isinstance(action, tuple) and len(action) == 2):
             await self.run_action(action, default_namespace=default_namespace)  # type: ignore[arg-type]
         elif callable(action):
             await action()
         else:
+            if isinstance(action, tuple) and self.debug:
+                # It's a tuple and made it this far, which means it'll be a
+                # malformed action. This is a no-op, but let's log that
+                # anyway.
+                log.warning(f"{event_name} event has an empty a action!")
```
willmcgugan: This could be more helpful. It will say something like "click event has an empty action", which might be quite mysterious in the logs.

```python
Can't parse action from {action!r}; check formatting of links.
```

willmcgugan (changes requested):

src/textual/app.py
davep: Perhaps I've misunderstood this, and something akin to this was my initial take, but it'll always be `()`, which won't really relate to what the dev wrote in their code, and which could end up being more confusing than helpful (they may go looking for `click=foo()` when the issue it's it's just `click=`, etc).

willmcgugan: You're right, it should really be the raw string pulled from the meta.

davep: Digging deeper into this, I don't see any "raw string" in the meta; all I see is an empty tuple in each breaking situation. For example, for all of:

- `[@click]`
- `[@click=]`
- `[@click=()]`

the meta is `{"@click": ()}`.

Moreover, I can't see to come up with an action that can't be parsed that doesn't throw an error right away (way before the application has got into a state where user input is permitted). Perhaps I'm missing the obvious here but this seems to be very much about capturing (and ignoring) an empty action, and logging it.

willmcgugan: Ok, but lets attempt a better error message. Suggest "Can't parse @click action from style meta; check your console markup syntax."

willmcgugan (approved):

Issue #4141, opened 2024-02-09
A Widget with {margin: 1} style in a ScrollableContainer causes infinite resizes and scrollbars flicker


labo-China:
When presenting a ScrollableContainer with a Widget that has styles above in it, there will have a infinite resizes.
Example:
```from textual.app import App, ComposeResult
from textual.containers import ScrollableContainer
from textual.widget import Widget
from textual.widgets import Static

class SuspiciousWidget(Widget):
    DEFAULT_CSS = 'SuspiciousWidget {margin: 1}'

class A(App):
    def compose(self) -> ComposeResult:
        with ScrollableContainer():
            yield Static()
            yield SuspiciousWidget()


A().run()
```

Screen Record: 
![Peek 2024-02-09 16-54](https://github.com/Textualize/textual/assets/62234752/9a7fdc5e-615a-45ed-9dec-748e7212616c)

'textual diagnose' result:

# Textual Diagnostics

## Versions

| Name    | Value  |
|---------|--------|
| Textual | 0.50.0 |
| Rich    | 13.7.0 |

## Python

| Name           | Value                                  |
|----------------|----------------------------------------|
| Version        | 3.11.6                                 |
| Implementation | CPython                                |
| Compiler       | GCC 13.2.1 20230801                    |
| Executable     | /home/labo/prog/NCSbox/venv/bin/python |

## Operating System

| Name    | Value                                                      |
|---------|------------------------------------------------------------|
| System  | Linux                                                      |
| Release | 6.7.4-zen1-1-zen                                           |
| Version | (HASH)1 ZEN SMP PREEMPT_DYNAMIC Mon, 05 Feb 2024 22:07:37 +0000 |

## Terminal

| Name                 | Value          |
|----------------------|----------------|
| Terminal Application | *Unknown*      |
| TERM                 | xterm-256color |
| COLORTERM            | truecolor      |
| FORCE_COLOR          | *Not set*      |
| NO_COLOR             | *Not set*      |

## Rich Console options

| Name           | Value               |
|----------------|---------------------|
| size           | width=75, height=27 |
| legacy_windows | False               |
| min_width      | 1                   |
| max_width      | 75                  |
| is_terminal    | True                |
| encoding       | utf-8               |
| max_height     | 27                  |
| justify        | None                |
| overflow       | None                |
| no_wrap        | False               |
| highlight      | None                |
| markup         | None                |
| height         | None                |

labo-China:
[...]

PR #4315, opened 2024-03-20
fix for flicker

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/4141

Fix for flicker where scrollbars "ping pong" between two states. The root of the problem was where both scrollbars were "auto".

Removed the "stabilized_scrollbar" attribute. That may have worked at some point, but didn't really do anything.


rodrigogiraoserrao (changes requested):
This doesn't handle the cases where both scrollbars are `auto` and they both need to appear and the second one only appears on the second pass (the one that's guarded behind the extra `if` this PR adds).
For example, take the app below:

```py


class A(App[None]):
    CSS = """
    Screen {
        overflow: auto auto;
    }
    #one, #two {
        width: 80;
    }
    #one {
        height: 1;
        background: red;
    }

    #two {
        height: 100%;
        background: blue;
    }
    """

        yield Label("one", id="one")
        yield Label("two", id="two")


if __name__ == "__main__":
    A().run()
```

If you set your terminal to have width equal to 80 (or set `width: xxx` to the width of your terminal) then we should have a vertical scrollbar _and_ a horizontal one.
If you make the terminal wider, then you have only the vertical scrollbar which is correct.
If you shorten the terminal, then the horizontal scrollbar shows up but the vertical one disappears!
The recording below shows this.

https://github.com/Textualize/textual/assets/5621605/51d5ff36-c2b7-4fe9-9b7e-2d46c62809be

willmcgugan:
#executivedecision

@rodrigogiraoserrao If you can break this, let me know.

rodrigogiraoserrao:
It's definitely an improvement but now there's flickering... 😢 

E.g., run this app and press <kbd>A</kbd>. More often than not you can see the horizontal scrollbar flicker.
Do I open a follow-up issue?

I said it before and I'll say it again... I'm sure there's an issue with the way in which we compute whether scrollbars are needed or not. 🤷 

```py


class A(App[None]):
    CSS = """
    Screen {
        overflow: auto auto;
    }
    #one {
        height: 1;
    }
    #two {
        height: 100%;
    }
    """

        yield Widget(id="one")

    def key_a(self) -> None:
        self.mount(Widget(id="two"))


if __name__ == "__main__":
    A().run()
```

Issue #4024, opened 2024-01-15
Incorrect width calculation with buttons in a horizontal container that has `width: auto`


davep:
This might be less specific that the subject suggests, but so far it's the only combination I've been able to use to create this problem. Given the following code, and given a sufficiently-sized terminal (in respect to width), you would expect every `Horizontal` in the result to reach the edge of the screen and either clip its content (for `width: 1fr`) or have itself be clipped (`width: auto`).

```python
from textual.app import App, ComposeResult
from textual.containers import Horizontal
from textual.widgets import Button, Label

class HorizontalWidthAutoApp(App[None]):

    CSS = """
    Horizontal {
        border: solid red;
        height: auto;

        &.auto {
            width: auto;
        }

        Button, Label {
            margin-left: 1;
        }
    }
    """

    def compose(self) -> ComposeResult:
        with Horizontal(classes="auto"):
            for n in range(10):
                yield Button(f"{n} This is a very wide button {n}")
        with Horizontal():
            for n in range(10):
                yield Button(f"{n} This is a very wide button {n}")
        with Horizontal(classes="auto"):
            for n in range(10):
                yield Label(f"{n} This is a very wide label {n}")
        with Horizontal():
            for n in range(10):
                yield Label(f"{n} This is a very wide label {n}")

if __name__ == "__main__":
    HorizontalWidthAutoApp().run()
```

The result, however, is different from what would be expected:

![Screenshot 2024-01-15 at 15 29 02](https://github.com/Textualize/textual/assets/28237/be83bd86-5b53-43b6-823b-175fd6f31cb9)

In this case the `width: auto` `Horizontal` that contains `Button`s seems to come up short.

PR #4317, opened 2024-03-20
Fixed auto-sizing of buttons in auto containers

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/4024


davep (approved):
There's a comment describing what's going on that's made stale by this, which deserves an update.

PR #4323, opened 2024-03-21
Add a snapshot test for the Welcome widget

davep:
Noticed there was at least one widget that didn't have a test.


rodrigogiraoserrao (approved):
Good catch.

Issue #4321, opened 2024-03-20
Multiple whitespaces shown in Markdown renders


jamescooke:
## Current behaviour

Normally, when Markdown is rendered by Textual, multiple whitespaces are squashed into a single whitespace.

For example this markdown:

```md
Hello
World
```

Is rendered as:

![Screenshot from 2024-03-20 22-12-40](https://github.com/Textualize/textual/assets/781059/49a7e71e-7b6d-487c-bdc1-92c98f33caf7)

However, with Textual, once multiple whitespaces are inside or next to a Markdown inline text format (for example emphasis or bold) or link, then the collapse behaviour stops and multiple whitespaces are shown.

Therefore, given the markdown document:

```md
Multiple white spaces   _are_   flattened.

[No    extra     spaces      here              ](http://example.com).

_Definitely     no extra     spaces          here_.

**No   bold   ones either              TOO**.
```

When this is rendered with [markdown-it](https://markdown-it.github.io/), then it appears as:

![Screenshot from 2024-03-20 21-58-09](https://github.com/Textualize/textual/assets/781059/f4684d4f-1421-43fe-839e-54cd96ba7798)

However, when embedding that document into the [Markdown widget `markdown.py` example](https://textual.textualize.io/widgets/markdown/#__tabbed_1_2) from the docs as follows:

```py
from textual.app import App, ComposeResult
from textual.widgets import Markdown

EXAMPLE_MARKDOWN = """\
Multiple white spaces   _are_   flattened.

[No    extra     spaces      here              ](http://example.com).

_Definitely     no extra     spaces          here_.

**No   bold   ones either              TOO**.
"""


class MarkdownExampleApp(App):
    def compose(self) -> ComposeResult:
        yield Markdown(EXAMPLE_MARKDOWN)


if __name__ == "__main__":
    app = MarkdownExampleApp()
    app.run()
```

Then, when run with `python markdown.py`, this renders as:

![Screenshot from 2024-03-20 22-01-29](https://github.com/Textualize/textual/assets/781059/9c09886f-e9a4-44fb-a29c-1d82d1d8034a)

## Expected behaviour

Multiple whitespaces inside bold, emphasis and links (plus any other similar structures) are all squashed when Textual renders it. Render matches markdown-it as closely as possible.

## Additional information

Possibly related bug: #2805 which dealt with newlines breaking up links.

Textual Diagnostics:

### Versions

| Name    | Value  |
|---------|--------|
| Textual | 0.53.1 |
[...]

PR #4324, opened 2024-03-21
Squish whitespace in `Markdown`

davep:
Collapse whitespace within the text of a markdown document when displaying it with `Markdown`. With this change multiple concurrent instances of space and tab are collapsed down to a single space.

Implements #4321.

### DIFF
--- a/src/textual/widgets/_markdown.py
+++ b/src/textual/widgets/_markdown.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import re
 from pathlib import Path, PurePath
 from typing import Callable, Iterable, Optional
 
@@ -146,10 +147,14 @@ class MarkdownBlock(Static):
         self._token = token
         style_stack: list[Style] = [Style()]
         content = Text()
+        repeating_whitespace = re.compile(r"[ \t]+")  # Space or tabs get reduced.
         if token.children:
             for child in token.children:
                 if child.type == "text":
-                    content.append(child.content, style_stack[-1])
+                    content.append(
+                        re.sub(repeating_whitespace, " ", child.content),
+                        style_stack[-1],
+                    )
                 if child.type == "hardbreak":
                     content.append("\n")
                 if child.type == "softbreak":
--- a/tests/snapshot_tests/snapshot_apps/markdown_whitespace.py
+++ b/tests/snapshot_tests/snapshot_apps/markdown_whitespace.py
@@ -1,0 +1,49 @@
+from textual.app import App, ComposeResult
+from textual.widgets import Markdown
+
+MARKDOWN = (
+    """\
+X X
+
+X  X
+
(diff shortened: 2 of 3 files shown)


willmcgugan:
Will this have any impact on code blocks, where you wouldn't want whitespace collapsed?

src/textual/widgets/_markdown.py
```diff
@@ -146,10 +147,14 @@ def build_from_token(self, token: Token) -> None:
         self._token = token
         style_stack: list[Style] = [Style()]
         content = Text()
+        repeating_whitespace = re.compile(r"[ \t]+")  # Space or tabs get reduced.
```
rodrigogiraoserrao, line 150: Would it be worth it (performance-wise) to compile this in the module scope and then use it here?

rodrigogiraoserrao, line 150:
```suggestion
        repeating_whitespace = re.compile(r"[ \t]+")  # Spaces and tabs get reduced.
```

rodrigogiraoserrao: Yes, I'm _this_ annoying 😆

willmcgugan, line 150: Wouldn't `"\s+"` included spaces and tabs?

There isn't much benefit in compiling the regex here. The `re` module has its own cache. I'd suggest moving the `re.compile` to module scope, or omitting and and let the `re` module do the caching.

rodrigogiraoserrao (approved):
Left two comments for your consideration but none of them are blocking.

willmcgugan (changes requested):
Suggestion re regexes

src/textual/widgets/_markdown.py
davep: > Wouldn't "\s+" included spaces and tabs?

I didn't see any benefit in matching line breaks. Should we?

davep: I'd read it as it's making plural "space or tab", but I'm happy to change it.

rodrigogiraoserrao: Ah, maybe it's a non-native speaker thing.
I thought the missing `s` was a typo: the translation would be wrong unless both nouns were in the plural and for some reason it felt weird in English too. 
The “and” suggestion was just because it sounded better...
But they were just _suggestions_!

davep: That's the difference between proper English and Yorkshire I guess. ;-)

willmcgugan: I think linebreaks are converted to spaces anyway in markdown. There are also a few other whitespace characters covered by `s`

davep: In that case I've dropped it to just that.

willmcgugan (approved):

PR #4331, opened 2024-03-24
fix(button): render with console markup

### LINKED ISSUE #4328: Button doesn't render Rich Markup correctly
If I'm reading the docs correctly, the `Button` widget *should* accept Rich Markup for its `label`, but in 0.53.1, it doesn't seem to do this.

Here's my test app:

```python3
from textual.app import App
from textual.widgets import Button, SelectionList
import sys

class MyApp(App):
	def compose(self):
		if len(sys.argv) > 1:
			yield SelectionList(('one', 1), ('two', 2))
		yield Button(label='[red]hell[/]o')

MyApp().run()
```

And here are my findings regarding versions:

# Textual 0.39.0 and 0.40.0
## Two `Widget`s
If run with any argument, the app behaves exactly as expected, i.e. renders red "hell" and uncolored "o":

v0 40 0-double

## One `Widget`
However if the Button is the only thing to be rendered, i.e. when you run it without any arguments, then the meaning of "red" becomes somehow inverted:

v0 40 0-single

# Textual 0.41.0 – 0.51.3
## Two `Widget`s:
Here markup seems totally ignored:
v0 41 0-double

## One `Widget`:
When run without arguments, we still see the inverted text, even in the latest version:
v0 41 0-single

# Thoughts
I would like to add Rich Markup to `Button` labels to indicate hotkeys for those buttons nicely, however for some reason latest version doesn't let me do that. How can I get Rich Markup to work on `Button` labels?

What causes the weird inversion behaviour depending on how many `Widget`s the app is rendering, is that normal?

Am I doing something else wrong?

# Diagnostics
My `textual diagnose` output (on version 0.53.1):

# Textual Diagnostics

## Versions

| Name    | Value  |
|---------|--------|
| Textual | 0.53.1 |
| Rich    | 13.7.1 |

## Python

| Name           | Value                                            |
|----------------|--------------------------------------------------|
| Version        | 3.11.2                                           |
| Implementation | CPython                                          |
| Compiler       | GCC 12.2.0                                       |
| Executable     | /home/goblin/venv/bin/python3 |

## Operating System

| Release | 6.1.0-11-amd64                                      |
| Version | #1 SMP PREEMPT_DYNAMIC Debian 6.1.38-4 (2023-08-08) |

## Terminal

## Rich Console options

| size           | width=80, height=24 |
| max_width      | 80                  |
| max_height     | 24                  |

TomJGooding:
Fixes #4328.

I've added a new snapshot test - here's a screenshot showing the difference after/before this change:

image

- [ ] Updated documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,8 +12,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Fixed a crash in `TextArea` when undoing an edit to a selection the selection was made backwards https://github.com/Textualize/textual/issues/4301
 - Fixed issue with flickering scrollbars https://github.com/Textualize/textual/pull/4315
 - Fix progress bar ETA not updating when setting `total` reactive https://github.com/Textualize/textual/pull/4316
+- Fixed `Button` not rendering correctly with console markup https://github.com/Textualize/textual/issues/4328
 
-### Changed 
+### Changed
 
 - ProgressBar won't show ETA until there is at least one second of samples https://github.com/Textualize/textual/pull/4316
 
--- a/src/textual/widgets/_button.py
+++ b/src/textual/widgets/_button.py
@@ -227,7 +227,7 @@ class Button(Widget, can_focus=True):
     def render(self) -> RenderResult:
         assert isinstance(self.label, Text)
         label = self.label.copy()
-        label.stylize(self.rich_style)
+        label.stylize_before(self.rich_style)
         return HorizontalPad(
             label,
             1,
--- a/tests/snapshot_tests/snapshot_apps/button_markup.py
+++ b/tests/snapshot_tests/snapshot_apps/button_markup.py
@@ -1,0 +1,14 @@
+from textual.app import App, ComposeResult
+from textual.widgets import Button
+
+
+class ButtonsWithMarkupApp(App):
+    def compose(self) -> ComposeResult:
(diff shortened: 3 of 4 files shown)


TomJGooding:
Something I noticed is that the shade is red is different comparing against v0.40.0, but it looks like there must have been other changes to how the markup is rendered as `Static` and `Label` show the same difference...

### Textual v0.40.0

image


### Now

image

```python
from textual.widgets import Button, Label, Static


class ButtonsWithMarkupApp(App):
        yield Button("[italic red]Focused[/] Button")
        yield Button("[italic red]Blurred[/] Button")
        yield Button("[italic red]Disabled[/] Button", disabled=True)

        yield Static("[italic red]Test[/] Static")
        yield Label("[italic red]Test[/] Label")


if __name__ == "__main__":
    app = ButtonsWithMarkupApp()
    app.run()
```

willmcgugan:
> Something I noticed is that the shade is red is different comparing against v0.40.0, but it looks like there must have been other changes as `Static` and `Label` show the same difference...

We changed the default dark ansi theme from DIMMED_MONOKAI to MONOKAI. Personally I prefer the dimmed version, but I was outnumbered. 🤷

rodrigogiraoserrao (approved):
Looks good to me.

willmcgugan (approved):

PR #4335, opened 2024-03-25
Delay suggestions until an `Input` is modified

### LINKED ISSUE #3811: The `Input` suggester should only show suggestions if an active edit is happening
Somewhat following on from #3806, I feel that the suggester API for `Input` is a wee bit too eager to make a suggestion. Consider the following code:

```python
from textual.suggester import SuggestFromList

class SuggestTooSoon(App[None]):

        yield Input("TeX", suggester=SuggestFromList([
            "textual"
        ], case_sensitive=False))

if __name__ == "__main__":
    SuggestTooSoon().run()
```

imagine this is a field in a dialog where the `Input` has been populated from some external database. The value in the field is the correct value. On the other hand the current approach suggests it might not be:

Screenshot 2023-12-05 at 11 09 13

The proposal here is that we modify the suggester API so that it doesn't make a visible suggestion until the user makes an edit. The "correct" value in that `Input` is `TeX` -- no suggestion needed -- but the moment the user backspaces the `X`, or adds a `t`, then it makes sense that a suggestion is made.

It may be that the current behaviour is desirable in some situations, in which case it would be fine to make this uneager approach an option.

davep:
Fixes #3811.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Fixed issue where narrow TextArea would repeatedly wrap due to scrollbar appearing/disappearing https://github.com/Textualize/textual/pull/4334
 - Fix progress bar ETA not updating when setting `total` reactive https://github.com/Textualize/textual/pull/4316
 
-### Changed 
+### Changed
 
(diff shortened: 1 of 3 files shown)


rodrigogiraoserrao (approved):
Sounds reasonable to me 🤷

PR #4341, opened 2024-03-26
docs: fix changelog

TomJGooding:
Spotted the wrong year in a few places in the changelog.

Also I think my addition to `Document` might have just missed being included the latest release - will check there aren't any others that have become out-of-sync.


willmcgugan:
Can't believe its March and I'm still doing that!

TomJGooding:
I _think_ this should be correct with unrelased updates, but please do give it a check!

willmcgugan (approved):

PR #4366, opened 2024-03-30
Allow pygments text area

royatt:
### Rational

Some project I'm working on includes a custom scripting language and a corresponding `Pygments` lexer, I would like to use `TextArea`; but my language sadly cannot be highlighted with it because I did not implement a `tree-sitter` binding for it.

This relatively small change  allows me to easily subclass `TextArea` and make it use a `Syntax` object (and consequently my pygments lexer)

### Note

The example provided in the 2nd commit shows a simplified example of how that would be done, and is mainly there to show why the first commit is needed. I am totally willing to drop that example commit if this use-case doesn't seem too main stream.


guy-av:
+1, I've got a CLI project that was built on top of `pygments` lexers which I'm trying to modernize, and could really use this change

willmcgugan:
It looks like this is highlighting a line at a time. Which would prevent highlighting tokens that span multiple lines. Consider a multi line string for example.

royatt:
> It looks like this is highlighting a line at a time. Which would prevent highlighting tokens that span multiple lines. Consider a multi line string for example.

That is definitely a valid point, but I am not trying to find a full replacement for tree-sitter, but rather simplify how I would achieve this low hanging fruit; as it's very much non-intrusive and doesn't change the code very much and still covers all my use cases well

willmcgugan:
If it fits your needs, then that's great. But I think it is better left as something to add to your own project.

Ideally, we would have a more comprehensive solution to use Pygments lexers in the TextArea.

darrenburns:
I think this is a good change - it'll make applying arbitrary highlights far simpler in subclasses (for example, drawing a background behind some text that matches a Cmd+F search query in an editor).

I've removed the Pygments example, for the reason @willmcgugan touched on, and also added a docstring.

darrenburns (approved):

PR #4367, opened 2024-03-30
fix: patch mixing width 100% with width auto

### LINKED ISSUE #4354: Merlin example app broken
The merlin example app is broken, likely because of recent changes to the way widths are resolved.

The problem is that there is no way to have a container with two children, with the container having its size set to `auto` and one of the children saying “be as large as the container” while the container hugs the other child widget.
(I think @davep merged a related PR recently but I can't seem to find it.)

In the merlin app, the issue is in the `SwitchLabel` widget.
The switch label widget contains a switch and we want the label to be as wide as the switch so that we can centre the label but there's no way of making the label automatically as wide as the switch without fixing both widths to a fixed absolute value.

After fixing the merlin example app we can merge #4036 that adds the example apps to the tests.

### LINKED ISSUE #4360: `width: 100%` mixed with `width: auto` amongst children has changed recently
This has been mentioned and reported in a couple of places, and @willmcgugan is aware and considering the issue, but I thought it a good idea to have a central issue recorded for it (if only so there's somewhere to point anyone seeing the same problem with their apps at).

Given this code:

```python

class Width100PCentApp(App[None]):

    CSS = """
    Vertical {
        border: solid red;
        width: auto;

        Label {
            border: solid green;
        }

        #first {
            width: 100%;
        }

        #second {
            width: auto;
        }
    }
    """

        with Vertical():
            yield Label("I want to be 100% of my parent", id="first")
            yield Label("I want my parent to be wide enough to wrap me and no more", id="second")

if __name__ == "__main__":
    Width100PCentApp().run()
```

until very recently this would result in the container being wide enough to contain the widest child, and the more-narrow child growing to fit the width of the container. Visually something like this:

Screenshot 2024-03-29 at 08 43 26

As of (I think) 0.54, this has changed, and now the `width: 100%` within a `width: auto` container is causing the container to grow too wide:

Screenshot 2024-03-29 at 08 43 49

This has been the cause of https://github.com/Textualize/trogon/issues/77 and https://github.com/Textualize/frogmouth/issues/98 as well as #4354. I've not checked yet but I strongly suspect this will also have broken `ModalScreen` dialogs on a couple of my main applications too, if used with 0.54.

TomJGooding:
Fixes #4360 by reverting the change in f094588. This change was intended to fix the issue with `Button` widths in #4024, but seems to have had wider unintended consequences. Instead this PR fixes the button width issue by adding a `get_content_width` override to button, which passes all tests.

There may be a wider issue with how relative sizes are handled in Textual, but this patch at least seems to provide a temporary fix.

This will also fix #4354 and other issues mentioned in #4360.

- [ ] Updated documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 
 - Fix priority bindings not appearing in footer when key clashes with focused widget https://github.com/Textualize/textual/pull/4342
+- Patch for issues mixing `width: 100%` with `width: auto` https://github.com/Textualize/textual/issues/4360
 
 ### Changed
 
--- a/src/textual/_layout.py
+++ b/src/textual/_layout.py
@@ -160,7 +160,8 @@ class Layout(ABC):
         if not widget._nodes:
             width = 0
         else:
-            arrangement = widget._arrange(Size(container.width, 0))
+            # Use a size of 0, 0 to ignore relative sizes, since those are flexible anyway
+            arrangement = widget._arrange(Size(0, 0))
             return arrangement.total_region.right
         return width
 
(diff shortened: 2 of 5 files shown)


TomJGooding:
This is failing in CI as `Cannot install tree-sitter-languages.` Not sure if recent changes today may have caused this.

TomJGooding:
Closed in favour of #4369.

PR #4369, opened 2024-03-31
width fix

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/4360

### DIFF
--- a/src/textual/widgets/_button.py
+++ b/src/textual/widgets/_button.py
@@ -4,6 +4,7 @@ from functools import partial
 from typing import TYPE_CHECKING, cast
 
 import rich.repr
+from rich.cells import cell_len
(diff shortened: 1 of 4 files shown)


src/textual/widgets/_button.py
```diff
@@ -203,6 +205,13 @@ def __init__(
         self.active_effect_duration = 0.3
         """Amount of time in seconds the button 'press' animation lasts."""
 
+    def get_content_width(self, container: Size, viewport: Size) -> int:
+        try:
+            return max([cell_len(line) for line in self.label.plain.splitlines()]) + 2
```
TomJGooding, line 210: Good catch. Might it be worth adding a snapshot test for buttons with multi-line labels?

Issue #4287, opened 2024-03-12
Include default CSS in docs?


TomJGooding:
Obviously all the `DEFAULT_CSS` is available by checking the source code, but perhaps this should somehow be included in the docs? If the default CSS was more discoverable, this might help reduce the number of issues/questions about styling.

TomJGooding:
Apparently I didn't search far back enough in the issues, as this has already discussed in https://github.com/Textualize/textual/issues/1705#issuecomment-1410740095

Let me know if this should be closed as a duplicate.

TomJGooding:
I think this would be even more useful following the changes with themes.

You really need to understand all the variables in the default CSS to correctly change the styling.  (Honestly, even after checking the source for the default CSS I now struggle a bit.)

PR #4371, opened 2024-04-01
Added a md that explains default css

Danu8902:

- [x] Updated documentation


TomJGooding:
The Textual docs already has a [guide to CSS](https://textual.textualize.io/guide/CSS/) and a [reference for the various styles](https://textual.textualize.io/styles/). Is there anything here you think is missing?

TomJGooding:
Perhaps this PR is linked to #4287? The `DEFAULT_CSS` this issue was about is different to what you're describing as "default CSS" here. This relates to the default styling of widgets in Textual, such as the `min-width: 16` for example, rather than a general guide to CSS which already exists in the docs.

davep:
I sense that the more fundamental issue with this PR is that it seems to have nothing to do with Textual CSS (and, if I'm honest, reads like the output of a large language model).

PR #4384, opened 2024-04-04
added insert, pop and remove_items methods to ListView

MuongKimhong:
Extend ListView's methods to make it easier to work with ListView.

New methods are:
1. `insert()` : insert new items to ListView at specified index
2. `pop()` : remove last item from ListView or remove item at specified index
3. `remove_items()`: remove multiple items from ListView at specified indices

### DIFF
--- a/src/textual/widgets/_list_view.py
+++ b/src/textual/widgets/_list_view.py
@@ -225,12 +225,12 @@ class ListView(VerticalScroll, can_focus=True, can_focus_children=False):
 
         Returns:
             An awaitable that yields control to the event loop
-                until the DOM has been updated with the new child item. 
+                until the DOM has been updated with the new child item.
         """
         await_mount = self.mount(*items, before=index)
         return await_mount
 
-    def pop(self, index: Optional[int]=None) -> AwaitRemove:
+    def pop(self, index: Optional[int] = None) -> AwaitRemove:
         """Remove last ListItem from ListView or
            Remove ListItem from ListView by index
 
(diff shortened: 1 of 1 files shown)


src/textual/widgets/_list_view.py
```diff
@@ -216,6 +216,56 @@ def clear(self) -> AwaitRemove:
...
+    def remove_items(self, indices: Iterable[int]) -> AwaitRemove:
+        """Remove ListItems from ListView by indices
+
+        Args:
+            indices: index(s) of ListItems to remove from ListView
+
+        Returns:
+            An awaitable object that waits for the direct children to be removed. 
+        """
+        if len(self._nodes) == len(indices):
```
davep, line 259: If there are two items in the `ListView` and the developer passed `[0, 0]` as the `indices` this will have unintended consequences.

MuongKimhong: Thank for pointing it out! I have removed that block as it's not necessary.

src/textual/widgets/_list_view.py
```diff
@@ -216,6 +216,53 @@ def clear(self) -> AwaitRemove:
...
+
+        Args:
+            indices: index(s) of ListItems to remove from ListView
+
+        Returns:
+            An awaitable object that waits for the direct children to be removed. 
+        """
+        items_to_remove = []
+        for index in indices:
+            items_to_remove.append(self.query("ListItem")[index])
```
willmcgugan, line 261: You could cache that query. Otherwise it would be performed once per loop.

MuongKimhong: please have a review

willmcgugan:
Thanks

Issue #4391, opened 2024-04-05
Select constructor shouldn't send Changed message


willmcgugan:
### Discussed in https://github.com/Textualize/textual/discussions/4368

<div type='discussions-op-text'>

<sup>Originally posted by **villekr** March 31, 2024</sup>
After change in https://github.com/Textualize/textual/issues/3869 all widgets send 'Changed' events even during the initialization. What is the correct way to ignore these 'Changed' events during the initialization? @rodrigogiraoserrao

Let's say I have the following situation where 'secondary' Select-widget's selection depends on the selected value of 'primary' Select-widget. Initially both have some values. 

Expected:
- primary = "drink"
- secondary = "juice"

Actual:
- primary = "drink"
- secondary = "water"

```python
from textual import on
from textual.app import App, ComposeResult
from textual.widgets import Header, Select

ITEMS = {"food": ["pizza", "burger", "salad"], "drink": ["water", "soda", "juice"]}


class SelectApp(App):
    CSS_PATH = "select.tcss"
    primary = "drink"
    secondary = "juice"

    def compose(self) -> ComposeResult:
        primary_items = [(item, item) for item in ITEMS]
        secondary_items = [(item, item) for item in ITEMS[self.primary]]
        yield Header()
        yield Select(
            primary_items,
            value=self.primary,
            prompt="Select Category",
            id="primary",
            classes="settings_option",
            allow_blank=False,
        )
        yield Select(
            secondary_items,
            value=self.secondary,
            prompt="Select Item",
            id="secondary",
            classes="settings_option",
            allow_blank=False,
        )
        self.title = f"{self.primary=} {self.secondary=}"

    @on(Select.Changed)
    def select_changed(self, event: Select.Changed) -> None:
        if event.control.id == "primary":
            self.primary = event.value
            items = ITEMS[event.value]
            widget: Select = self.query_one("#secondary")
            widget.set_options((item, item) for item in items)
        if event.control.id == "secondary":
            self.secondary = event.value
        self.title = f"{self.primary=} {self.secondary=}"


if __name__ == "__main__":
    app = SelectApp()
    app.run()

```</div>

PR #4392, opened 2024-04-05
prevent messages from constructor

willmcgugan:
https://github.com/Textualize/textual/issues/4391

Allows use of self.prevent to prevent messages from being sent on_mount

Original issue may be fixed with the following change to `compose`:

```python
from textual import on
from textual.widgets import Header, Select

ITEMS = {"food": ["pizza", "burger", "salad"], "drink": ["water", "soda", "juice"]}


class SelectApp(App):
    CSS_PATH = "select.tcss"
    primary = "drink"
    secondary = "juice"

        primary_items = [(item, item) for item in ITEMS]
        secondary_items = [(item, item) for item in ITEMS[self.primary]]
        yield Header()
        with self.prevent(Select.Changed): 
            yield Select(
                    primary_items,
                    value=self.primary,
                    prompt="Select Category",
                    id="primary",
                    classes="settings_option",
                    allow_blank=False,
                )
                yield Select(
                    secondary_items,
                    value=self.secondary,
                    prompt="Select Item",
                    id="secondary",
                    classes="settings_option",
                    allow_blank=False,
                )
        self.title = f"{self.primary=} {self.secondary=}"

    @on(Select.Changed)
    def select_changed(self, event: Select.Changed) -> None:
        if event.control.id == "primary":
            self.primary = event.value
            items = ITEMS[event.value]
            widget: Select = self.query_one("#secondary")
            widget.set_options((item, item) for item in items)
        if event.control.id == "secondary":
            self.secondary = event.value
        self.title = f"{self.primary=} {self.secondary=}"


if __name__ == "__main__":
    app = SelectApp()
    app.run()

```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [0.56.0] - Unreleased
+
+### Changed
+
+- self.prevent can be used in a widget constructor to prevent messages on mount https://github.com/Textualize/textual/pull/4392
+
 ## [0.55.1] - 2024-04-2
 
 ### Fixed
--- a/src/textual/message_pump.py
+++ b/src/textual/message_pump.py
@@ -144,6 +144,7 @@ class MessagePump(metaclass=_MessagePumpMeta):
         """
         self._next_callbacks: list[events.Callback] = []
         self._thread_id: int = threading.get_ident()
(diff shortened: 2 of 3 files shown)


TomJGooding:
Sorry just to check I understand this correctly, it is the responsibility of the developer to prevent the changed event?/

willmcgugan:
> Sorry just to check I understand this correctly, it is the responsibility of the developer to prevent the changed event?

Yes, the majority of the time you do want to handle messages for any changes. The alternative would be for every widget to have some kind of `send_change_events` switch.

TomJGooding:
Thanks Will for confirming. I just wasn't sure based on the title of your issue if actually a `Changed` message shouldn't be post on initialisation at all, but I think you could argue either way!

PR #4405, opened 2024-04-09
Fix minor mistake in docs/styles/padding.md

habaneraa:
It seems there is a little mistake in the document [here](https://textual.textualize.io/styles/padding/#see-also). The "padding" should be "margin". That link is correct though.

- [ ] Updated documentation


willmcgugan:
Thanks

PR #4424, opened 2024-04-17
docs: update TabActivated on decorator selector

### LINKED ISSUE #4423: Docs change - use `pane` instead of `tab` for an event listener on tab change
We tried adding an event listener for a tab change using the docs here, but couldn't get it to work: https://textual.textualize.io/guide/events/#applying-css-selectors-to-arbitrary-attributes. Should the docs be updated to use `pane` instead of `tab`, or are we doing something wrong?  Specifically this snippet:

```python
@on(TabbedContent.TabActivated, tab="#home")
def home_tab(self) -> None:
    self.log("Switched back to home tab.")
```

I got it working, I think it's related to the breaking change described in https://github.com/Textualize/textual/blob/main/CHANGELOG.md#0460---2023-12-17

> Breaking change: tab is no longer a @on decorator selector for TabbedContent.TabActivated -- use pane instead https://github.com/Textualize/textual/pull/3815

TomJGooding:
Fixes #4423.

- [x] Updated documentation

### DIFF
--- a/docs/guide/events.md
+++ b/docs/guide/events.md
@@ -268,7 +268,7 @@ The `on` decorator also accepts selectors as keyword arguments that may be used 
 The snippet below shows how to match the message [`TabbedContent.TabActivated`][textual.widgets.TabbedContent.TabActivated] only when the tab with id `home` was activated:
 
 ```py
-@on(TabbedContent.TabActivated, tab="#home")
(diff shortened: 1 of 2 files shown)


willmcgugan:
Thanks, Tom.

PR #4427, opened 2024-04-18
fix: make Document.end zero-indexed

### LINKED ISSUE #4426: Document.end line number is off by one (1-indexed)
`Document.start` and `Document.end` were introduced in https://github.com/Textualize/textual/pull/4267

`Document.start` maps to `(0, 0)`, but `Document.end` is 1-indexed, with its line number computed from `Document.line_count`. This is confusing, and breaks with the standard from other functions that return or accept a `Location`.

For example, this is the current behavior:

```py
>>> from textual.widgets.text_area import Document
>>> doc = Document("foo\nbar")
>>> doc.end
(2, 3)
```

Expected:
```py
>>> doc.end
(1, 3)
```

# Textual Diagnostics

## Versions

| Name    | Value  |
|---------|--------|
| Textual | 0.56.4 |
| Rich    | 13.7.1 |

## Python

| Name           | Value                                                                            |
|----------------|----------------------------------------------------------------------------------|
| Version        | 3.8.10                                                                           |
| Implementation | CPython                                                                          |
| Compiler       | GCC 9.3.0                                                                        |
| Executable     | /home/tco/.cache/pypoetry/virtualenvs/textual-textarea-mNcIuqFZ-py3.8/bin/python |

## Operating System

| Name    | Value                               |
|---------|-------------------------------------|
| System  | Linux                               |
| Release | 5.15.146.1-microsoft-standard-WSL2  |
| Version | #1 SMP Thu Jan 11 04:09:03 UTC 2024 |

## Terminal

| Name                 | Value            |
|----------------------|------------------|
| Terminal Application | Windows Terminal |
| TERM                 | xterm-256color   |
| COLORTERM            | truecolor        |
| FORCE_COLOR          | *Not set*        |
| NO_COLOR             | *Not set*        |

## Rich Console options

| size           | width=120, height=30 |
| max_width      | 120                  |
| max_height     | 30                   |

tconbeer:
Fixes #4426

This changes `Document.end` to return a zero-indexed `Location`.

The only downstream dependency of this property in the Textual codebase is `TextArea.clear`, which is not impacted by this change (as tested in `tests/text_area/test_edit_via_api.py::test_clear`).

In the Changelog, I've marked this as a fix, although it's possible it's a breaking change for some app developers.

- [x] Updated documentation (this behavior is not documented directly, so no changes were made)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Fixed `Integer` validator missing failure description when not a number https://github.com/Textualize/textual/issues/4413
 - Fixed a crash in `DataTable` if you clicked a link in the border https://github.com/Textualize/textual/issues/4410
+- Fixed an off-by-one error in the line number of the `Document.end` property https://github.com/Textualize/textual/issues/4426
 
 ### Added
 
--- a/src/textual/document/_document.py
+++ b/src/textual/document/_document.py
@@ -351,7 +351,7 @@ class Document(DocumentBase):
     def end(self) -> Location:
         """Returns the location of the end of the document."""
         last_line = self._lines[-1]
-        return (self.line_count, len(last_line))
+        return (self.line_count - 1, len(last_line))
 
     def get_index_from_location(self, location: Location) -> int:
         """Given a location, returns the index from the document's text.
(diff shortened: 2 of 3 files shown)


darrenburns (approved):
Thanks Ted!

TomJGooding:
🤦 Sorry my bad!

PR #4436, opened 2024-04-21
fix(text area): end mouse selection only if currently selecting

### LINKED ISSUE #4434: All TextAreas have blinking cursors after switching back to a previously selected tab
Example:

```
from textual.widgets import TabbedContent, TabPane, TextArea

class TabbedTextAreaApp(App[None]):

        with TabbedContent():
            for tab in ("one", "two"):
                with TabPane(f"{tab}.py"):
                    for _ in range(3):
                        yield TextArea(language="python")

if __name__ == "__main__":
    TabbedTextAreaApp().run()
```

To reproduce behavior: Start program, click on tab `two`, then click on `one`. Notice that all the text areas have blinking cursors.
A workaround I've done is to set `cursor_blink` to `False` and then back to `True` on `TabActivated`.

Code_fxryGjsZcx

# Textual Diagnostics

## Versions

| Name    | Value  |
|---------|--------|
| Textual | 0.56.4 |
| Rich    | 13.7.1 |

## Python

| Name           | Value            |
|----------------|------------------|
| Version        | 3.8.10           |
| Implementation | CPython          |
| Compiler       | GCC 9.4.0        |
| Executable     | /usr/bin/python3 |

## Operating System

| Name    | Value                                        |
|---------|----------------------------------------------|
| System  | Linux                                        |
| Release | 4.4.0-19041-Microsoft                        |
| Version | #3996-Microsoft Thu Jan 18 16:36:00 PST 2024 |

## Terminal

| Terminal Application | vscode (1.88.1) |

## Rich Console options

| size           | width=203, height=31 |
| max_width      | 203                  |
| max_height     | 31                   |

TomJGooding:
The `TextArea` should only invoke the `_end_mouse_selection` method if we're actually currently selecting with the mouse.

This will fix #4434 where currently the cursor blink is restarted after the `TextArea` is hidden in the `TabbedContent`.

- [ ] Updated documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## Unreleased
+
+### Fixed
+
+- Fixed `TextArea` to end mouse selection only if currently selecting https://github.com/Textualize/textual/pull/4436
+
 ## [0.57.1] - 2024-04-20
 
 ### Fixed
--- a/src/textual/widgets/_text_area.py
+++ b/src/textual/widgets/_text_area.py
@@ -1520,11 +1520,13 @@ TextArea {
 
     async def _on_mouse_up(self, event: events.MouseUp) -> None:
         """Finalize the selection that has been made using the mouse."""
-        self._end_mouse_selection()
+        if self._selecting:
+            self._end_mouse_selection()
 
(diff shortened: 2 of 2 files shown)


darrenburns:
Thanks @TomJGooding - very minor, but would the `if self._selecting` maybe be better off inside the `_end_mouse_selection` if it's always required?

TomJGooding:
@darrenburns No problem, that's now updated. I did wonder that myself but thought the other way might better express the intent.

darrenburns (approved):

darrenburns:
@TomJGooding Thanks for making that change and for the fix!

PR #4452, opened 2024-04-27
fix(list view): fix bugs with initial index

### LINKED ISSUE #4449: `ListView` initial index is off by one
Setting the `initial_index` of a `ListView` seems to be off by one, introduced somewhere in v0.48.2:

```python
from textual.widgets import Label, ListItem, ListView


class ExampleApp(App):
        list_items = [ListItem(Label(str(i))) for i in range(10)]
        yield ListView(
            *list_items,
            initial_index=9,
        )


if __name__ == "__main__":
    app = ExampleApp()
    app.run()
```

image

TomJGooding:
Fixes #4449.

There's currently a problem with how the `ListView._index` is set in `__init__`:

https://github.com/Textualize/textual/blob/a5cab15cc10143fb49771231c35cbe5665083f9c/src/textual/widgets/_list_view.py#L120-L125

The `self._nodes` aren't actually available yet, which means this will always just return  `initial_index - 1` (the anchor).  This causes the off-by-one error in the original issue, but also means there's no actual checking for the next enabled item!

- [ ] Updated documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## Unreleased
+
+### Fixed
+
+- Fixed `ListView` bugs with the initial index https://github.com/Textualize/textual/pull/4452
+
 ## [0.58.0] - 2024-04-25
 
 ### Fixed
--- a/src/textual/widgets/_list_view.py
+++ b/src/textual/widgets/_list_view.py
@@ -119,9 +119,10 @@ class ListView(VerticalScroll, can_focus=True, can_focus_children=False):
         )
         # Set the index to the given initial index, or the first available index after.
         self._index = _widget_navigation.find_next_enabled(
-            self._nodes,
-            anchor=initial_index - 1 if initial_index is not None else None,
+            children,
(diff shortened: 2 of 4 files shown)


TomJGooding:
`mypy` complains that `Argument 1 to "find_next_enabled" has incompatible type "tuple[ListItem, ...]"; expected "Sequence[Disableable]"`, but it also complains about the `NodeList` so I'm not sure if this is an issue?

I am a bit confused why this change caused the `test_keyboard_navigation_with_disabled_items` test to fail. This apparently expected the initial highlighted to be `None`, but this shouldn't be the case if the index is:

```python
# Set the index to the given initial index, or the first available index after.
```

I've updated that test accordingly.

TomJGooding:
Sorry also to just to note, this assumes that "the first available index after" will wrap around. For example, where the `initial_index=8` in the added test, the highlighted index will be 1.

darrenburns (approved):
Thanks :) Looks good! Was this just an issue inside `compose`? I think `ListView` was from before `compose` was a thing, so it would only have been tested in an `on_mount` when it was made - could be wrong though 😄

TomJGooding:
> Thanks :) Looks good! Was this just an issue inside `compose`? I think `ListView` was from before `compose` was a thing, so it would only have been tested in an `on_mount` when it was made - could be wrong though 😄

Sorry @darrenburns I'm not sure what you mean. This handling of disabled items was added in PR #3912, which is why I was a little confused why the test seemed to contradict this.

PR #4464, opened 2024-04-29
fix(selection list): update indexes after option removed

### LINKED ISSUE #4462: SelectionList weird behavior when removing options in its own event handler until freeze
When you interact with a SelectionList and remove the option in one of the event handlers one of the following happens:
- Wrong options is removed
- Option just deselects
- Entire component just freezes (not the entire app;  most likely occurs with the first or last option in the Selection List)

Example code:
```python
from textual.app import App
from textual.widgets import SelectionList
from textual.widgets.selection_list import Selection
import typing as t
from string import ascii_letters


class MyApp(App[str]):
    def compose(self) -> t.Iterable[Widget]:
        yield SelectionList(
            *[Selection(letter, letter, True, letter) for letter in ascii_letters],
        )

    def on_selection_list_selection_toggled(
        self, event: SelectionList.SelectionToggled[str]
    ) -> None:
        id_ = event.selection.id
        if id_ is not None:
            event.selection_list.remove_option(id_)


if __name__ == "__main__":
    app = MyApp()
    app.run()
```

# Textual Diagnostics

## Versions

| Name    | Value  |
|---------|--------|
| Textual | 0.58.0 |
| Rich    | 13.7.1 |

## Python

| Name           | Value                                                |
|----------------|------------------------------------------------------|
| Version        | 3.11.2                                               |
| Implementation | CPython                                              |
| Compiler       | GCC 12.2.0                                           |
| Executable     | /home/jan/Documents/dev/trove-setup/.venv/bin/python |

## Operating System

| Release | 6.1.0-17-amd64                                      |
| Version | #1 SMP PREEMPT_DYNAMIC Debian 6.1.69-1 (2023-12-30) |

## Terminal

| Terminal Application | vscode (1.88.1) |

## Rich Console options

| size           | width=208, height=17 |
| max_width      | 208                  |
| max_height     | 17                   |

Video:

https://github.com/Textualize/textual/assets/51451323/eeff467e-6c05-4a02-bac7-b987b65f8275

TomJGooding:
Fixes #4462.

I've added a basic test that the index is updated, but I'm not sure if you need a regression test that actually replicates the issue above?

- [ ] Updated documentation

### DIFF
--- a/src/textual/widgets/_selection_list.py
+++ b/src/textual/widgets/_selection_list.py
@@ -646,6 +646,11 @@ class SelectionList(Generic[SelectionType], OptionList):
         option = self.get_option_at_index(index)
         self._deselect(option.value)
         del self._values[option.value]
+        # Decrement index of options after the one we just removed.
+        self._values = {
+            option_value: option_index - 1 if option_index > index else option_index
+            for option_value, option_index in self._values.items()
+        }
         return super()._remove_option(index)
 
     def add_options(
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## Unreleased
+
+### Fixed
+
+- Fixed `SelectionList` issues after removing an option https://github.com/Textualize/textual/pull/4464
+
 ## [0.58.0] - 2024-04-25
 
 ### Fixed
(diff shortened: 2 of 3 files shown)


jvllmr (commented):
I tested your change with the example code from my issue and it works wonderfully. Thanks a lot! 💯

src/textual/widgets/_selection_list.py
```diff
@@ -646,6 +646,11 @@ def _remove_option(self, index: int) -> None:
         option = self.get_option_at_index(index)
         self._deselect(option.value)
         del self._values[option.value]
+        # Decrement index of options after the one we just removed.
```
darrenburns, line 649: Maybe a separate issue (or not an issue at all if I'm missing something), but this makes me wonder why the `highlighted` index wouldn't be decremented if it appears below the removed options.

darrenburns: Actually from a skim, I'm guessing it might be that the highlight is applied based on the ID of the option rather than the index. It's just converted to an ID via a validator.

Issue #4461, opened 2024-04-29
Giving focus to a widget not scrolled into view doesn't always scroll it into view


davep:
Using the following code:

```python
from textual.app import App, ComposeResult
from textual.containers import VerticalScroll
from textual.widgets import Label, Input

class Inner(Label, can_focus=True):
    DEFAULT_CSS = """
    Inner {
        width: 1fr;
        border: solid grey;
        &:focus {
            border: thick green;
        }
    }
    """

class Top(VerticalScroll, can_focus=False):

    def compose(self) -> ComposeResult:
        for n in range(100):
            yield Inner(f"{n}. This is a label that can get focus")

class ScrollFocusIssueApp(App[None]):

    AUTO_FOCUS = "Input"

    def compose(self) -> ComposeResult:
        yield Top()
        yield Input()

    def on_mount(self) -> None:
        self.query_one(Top).scroll_end(animate=False)

if __name__ == "__main__":
    ScrollFocusIssueApp().run()
```

often, but not always, if I <kbd>tab</kbd> or <kbd>shift</kbd>+<kbd>tab</kbd> from the `Input` to the `Inner` that *isn't* currently visible in the `Top` widget, the focused widget isn't scrolled into view. A subsequent change of focus "further" in the same direction will then result in the subsequently-focused widget being brought into view.

So, using the above, I can often cause the problem like this:

1. Run up the application
2. Press <kbd>tab</kbd>
3. Sometimes, but not always, the focused widget doesn't scroll into view.
4. If it did scroll into view...
5. <kbd>shift</kbd>+<kbd>tab</kbd> twice
6. For me, without fail, the focused widget isn't scrolled into view

darrenburns:
I'm pretty sure this is something to do with the animation that occurs when you `focus_next`. If you switch off animation where `scroll_to_center` is called inside the focus logic in `screen.py`, things seem to work consistently.

darrenburns:
I can't reproduce the issue in Elia though in a chat with only a couple of messages - I'd have expected it to exhibit the same behaviour given this MRE - I wonder if it's related to the distance that needs to be scrolled or the number of children.

TomJGooding:
I've not dived too deep into this yet, but if you add a debug log here:

```diff
diff --git a/src/textual/screen.py b/src/textual/screen.py
index b84fbc277..6dd3d0a43 100644
--- a/src/textual/screen.py
+++ b/src/textual/screen.py
[...]

PR #4469, opened 2024-05-01
Fix scroll center

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/4461

There were two issue. The intermittent non-scrolling was a race condition. It also wasn't always scrolling to the center as intended, that was a geometry miscalculation.


willmcgugan:
@davep If you have a moment, you might want to check this against your code.

davep:
@willmcgugan Looking good with the sample code from the issue and also with the originating app that made me look into this in the first place.

PR #4471, opened 2024-05-02
Support zero-indexed line numbers in TextArea

royatt:
Possible alternatives:

- Add this as a keyword argument - IMO this use-case will not be common enough to deserve it's own typed keyword argument being there until the end of  time
- Make the ``show_line_numbers`` kw accept an ``int`` (with ``-1`` signifying ``False``) - won't add another kw but will terribly hurt readability
- Split up the source code in such a way it won't require overriding the entire ``render_line`` method  to do this in one's own project - not worth it, will probably end up being very invasive, while this approach is fairly elegant

Something I still consider adding:

typing this class variable as ``: Literal[0, 1]`` to increase clarity of what it's use case really is

### DIFF
--- a/docs/widgets/text_area.md
+++ b/docs/widgets/text_area.md
@@ -353,6 +353,8 @@ the `show_line_numbers` attribute to `True` or `False`.
 
 Setting this attribute will immediately repaint the `TextArea` to reflect the new value.
 
+You can also change the start line number (the topmost line number in the gutter) by setting the `line_number_start` reactive attribute.
+
 ### Extending `TextArea`
 
 Sometimes, you may wish to subclass `TextArea` to add some extra functionality.
@@ -506,6 +508,7 @@ A detailed view of these classes is out of scope, but do note that a lot of the 
 | `theme`                | `str`                    | `"css"`       | The theme to use.                                                |
 | `selection`            | `Selection`              | `Selection()` | The current selection.                                           |
 | `show_line_numbers`    | `bool`                   | `False`       | Show or hide line numbers.                                       |
+| `line_number_start`    | `int`                    | `1`           | The start line number in the gutter.                            |
 | `indent_width`         | `int`                    | `4`           | The number of spaces to indent and width of tabs.                |
 | `match_cursor_bracket` | `bool`                   | `True`        | Enable/disable highlighting matching brackets under cursor.      |
 | `cursor_blink`         | `bool`                   | `True`        | Enable/disable blinking of the cursor when the widget has focus. |
--- a/src/textual/widgets/_text_area.py
+++ b/src/textual/widgets/_text_area.py
@@ -301,6 +301,9 @@ TextArea {
 
     Changing this value will immediately re-render the `TextArea`."""
 
+    line_number_start: Reactive[int] = reactive(1, init=False)
+    """The line number the first line should be."""
+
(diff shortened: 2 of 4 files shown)


darrenburns:
I don't really understand the benefit of this - why would you want to have the line numbers be zero indexed?

davep:
Chiming in with a perspective: I can't personally see any benefit in zero-based line numbers, as a specific feature, but I *can* see benefits to being able to set the origin number of the line numbers. For example, I might want to create a breakout editor where I'm working on a specific few lines of code, and I want to maintain the line numbers; another reason might be I want to do some sort of narrowing effect.

So perhaps if such a feature were added in some way, it would make sense that it's a way of setting the number of the "first visible line"?

darrenburns:
@davep Yeah, I can see that being useful. Rich `Syntax` offers something similar IIRC.

royatt:
> I don't really understand the benefit of this - why would you want to have the line numbers be zero indexed?

The use case I was going for is I have a sort of a query language embedded into a ``TextArea`` that you can submit and then it shows you results (kinda like that cool SQL TUI project named ``harlequin``).

I also have this cool feature where when the language grammar is parsed; if you have a syntax error you get a toast with the error the parser created.

The problem is, the parser shows line number based errors which are zero-index based :(

I did some googling (for this problem in text editors/ programming in general) and found some more use cases, so I'm not totally alone on this it seems (albeit, yes - This might not be a very intuitive thing to add)

royatt:
> Chiming in with a perspective: I can't personally see any benefit in zero-based line numbers, as a specific feature, but I *can* see benefits to being able to set the origin number of the line numbers. For example, I might want to create a breakout editor where I'm working on a specific few lines of code, and I want to maintain the line numbers; another reason might be I want to do some sort of narrowing effect.
> 
> So perhaps if such a feature were added in some way, it would make sense that it's a way of setting the number of the "first visible line"?

Hmm, this can be fairly easily transformed into an instance attribute with no further cost if you think it's better

willmcgugan:
I think this is probably best done as a reactive on the widget itself. Possibly also exposed in the constructor.

Suggest we call it `line_number_start`

merriam:
The use case I could possibly imagine:

My program is popping up a text area to modify one little section of a large annoyingly formatted data file.   For example, "change some parameters for the machine Gandalf, but don't let anything else change".   The relevant section would be loaded into the buffer, the buffer would be edited in a TextArea, and, after some validation, the annoying file is updated.   In that case, setting the starting number of a text area would be helpful.

willmcgugan:
@royatt I think this is a good idea, but its probably best done as an argument on the constructor. Would you like to make that change?

royatt:
> @royatt I think this is a good idea, but its probably best done as an argument on the constructor. Would you like to make that change?

Hey, it would take me a couple of days to get to it but yes I'd like to do it

royatt:
> @royatt I think this is a good idea, but its probably best done as an argument on the constructor. Would you like to make that change?

Just pushed an amended commit where `line_number_start` is a reactive keyword argument

darrenburns:
I added a snapshot test and updated the docs.

Works great:

https://github.com/Textualize/textual/assets/5740731/cd0298bf-b3cf-4d15-abfa-b9332ff1cd0e

darrenburns (approved):
Looks good to me!

Issue #4489, opened 2024-05-07
Clicking `Select` doesn't properly collapse overlay


TomJGooding:
It looks like there might have been an unintended change to the behaviour of the `Select` widget.

You can reproduce this using the [basic example](https://textual.textualize.io/widgets/select/#basic-example) from the docs. Try clicking the `Select` to expand the overlay, then click it again. This should collapse the overlay, but it now quickly pops back up again.

I haven't dived into details yet but it looks like the relevant commit is ab9ea3f.

TomJGooding:
I don't understand why yet, but I found if you click exactly on the ▲ arrow everything works as expected.

willmcgugan:
I noticed that. Weird. It's something to do with the mechanism that hides the drop down with the widget is blurred. But I haven't dug up any more than that...

TomJGooding:
Comparing the logs from `textual console`, if you click the arrow the `SelectOverlay` seems to be hidden **before** the `Static` arrow is hidden. For reasons I don't understand yet, this means the `Click` event which triggers the `Toggle` message is never actually fired, yet everything works as expected. Whereas clicking anywhere apart from the arrow **also** then fires the `Click` event, causing the overlay to expand again after just being collapsed.

PR #4499, opened 2024-05-09
Fix select

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/4489

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Fixed `SelectionList` issues after removing an option https://github.com/Textualize/textual/pull/4464
 - Fixed `ListView` bugs with the initial index https://github.com/Textualize/textual/pull/4452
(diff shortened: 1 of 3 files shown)


willmcgugan:
@TomJGooding I think this is the fix...

TomJGooding:
Wouldn't this introduce a new issue where the overlay would stay expanded if you clicked to focus another widget?

willmcgugan:
You're absolutely right. Pushed an alternative fix, which doesn't break hiding the drop down on blur.

TomJGooding:
This seems to have done the trick after from a quick test!

Just for my own knowledge, is this suppression of the click really only required because the overlay concept doesn't play nicely with focusing?

willmcgugan:
> Just for my own knowledge, is this suppression of the click really only required because the 'overlay' concept doesn't play nicely with how the focus is handled?

Pretty much. The interaction with click / blur was proving to be difficult to unpick. I suspect it could be done much simpler with pure CSS, but I didn't get that far.

PR #4505, opened 2024-05-11
Allow more easily customisable scrollbar glyphs

### LINKED ISSUE #4500: Add the ability to customise scrollbar bar segments
Currently, the characters that make up a scrollbar [are hardcoded](https://github.com/Textualize/textual/blob/e21f57529d4e70bf421acb7cb5e7d041a43e5a2f/src/textual/scrollbar.py#L101-L104). It would be nice if they could be overridden. Personally, I need to support poor fonts, so need to replace missing glyphs. But it would be neat to use braille characters for a unique style.

Thoughts on making them a classvar, so they can be monkeypatched in for now? I can do that PR if it's an OK change.

mon:
- [x] Updated documentation (not needed I think?)
- [x] Updated CHANGELOG.md (not needed, I don't think it warrants a changelog?)

Closes #4500

This allows a user to easily customise scrollbar display. For example, when using the Consolas font, scrollbars display poorly:
cmd_NfNfGUSQfX

By reducing the characters to those supported, the display is more consistent:
```python
ScrollBarRender._vertical_bars = ["▄", "█", " "]
```
cmd_4MxcLegZxq

By using some lightly shaded characters, the scrollbars gain some flair:
```python
ScrollBarRender._vertical_bars = ["░", "▒", "▓", " "]
```
cmd_1HZHPAXOts

The blank glyph can also be replaced:
```python
ScrollBarRender._vertical_bars = ["░", "▓", " "]
ScrollBarRender._blank_glyph = "▒"
```
cmd_HieEhLsuqH

This is easier than replacing the entire `ScrollBarRender.render_bar` function.

### DIFF
--- a/src/textual/scrollbar.py
+++ b/src/textual/scrollbar.py
@@ -70,11 +70,11 @@ class ScrollTo(ScrollMessage, verbose=True):
 
 
 class ScrollBarRender:
-    _vertical_bars: ClassVar[list[str]] = ["▁", "▂", "▃", "▄", "▅", "▆", "▇", " "]
+    VERTICAL_BARS: ClassVar[list[str]] = ["▁", "▂", "▃", "▄", "▅", "▆", "▇", " "]
     """Glyphs used for vertical scrollbar ends, for smoother display."""
-    _horizontal_bars: ClassVar[list[str]] = ["▉", "▊", "▋", "▌", "▍", "▎", "▏", " "]
+    HORIZONTAL_BARS: ClassVar[list[str]] = ["▉", "▊", "▋", "▌", "▍", "▎", "▏", " "]
     """Glyphs used for horizontal scrollbar ends, for smoother display."""
-    _blank_glyph: ClassVar[str] = " "
+    BLANK_GLYPH: ClassVar[str] = " "
     """Glyph used for the main body of the scrollbar"""
 
     def __init__(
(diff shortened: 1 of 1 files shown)


TomJGooding:
I try to avoid Windows if I can help it, but unlike classic conhost I think the [Windows Terminal](https://apps.microsoft.com/detail/9n0dx20hk701) has font fallback. That's not any criticism of this PR and you're probably already aware, but just thought I should mention it as there's likely some other glyphs used in Textual which are missing in Consolas.

mon:
Yes, I personally use Windows Terminal with a nerd font, so textual apps look _great_ on my machine.

However, I often give TUI tools to people without python, so I compile them to an .exe with pyinstaller. Double clicking that will always open in cmd.exe, and the default font is Consolas.

src/textual/scrollbar.py
```diff
@@ -70,6 +70,13 @@ def __rich_repr__(self) -> rich.repr.Result:
 
 
 class ScrollBarRender:
+    _vertical_bars: ClassVar[list[str]] = ["▁", "▂", "▃", "▄", "▅", "▆", "▇", " "]
```
willmcgugan, line 73: If these are meant to be overridden, then they probably shouldn't be prefixed with an underscore. I might also put them in all-caps given they are constants to the class where they are defined.

It might also be worth considering changing them to a tuple and typing them as `tuple[str, str, str, str, str, str, str, str]`. A bit verbose, but it would allow the typechecker to mandate there are 8 items.

willmcgugan (changes requested):

src/textual/scrollbar.py
willmcgugan: Ah, in that case a list is fine!

willmcgugan (approved):

PR #4508, opened 2024-05-12
query children

willmcgugan:
Adds a `query_children` method. Which is virtually the same as `query`, but only finds immediate children.


darrenburns (approved):
Looks good 👍

PR #4515, opened 2024-05-15
feat!: remove confusing app workers alias

### LINKED ISSUE #3619: `DOMNode.workers` is a confusing alias
Currently, `DOMNode.workers` is a very confusing alias for `App.workers`.

When you type `self.workers` inside a widget, it _really, really, REALLY looks_ like you are talking about the workers started by/associated with the given widget.
But you're actually talking about all app workers.

See https://github.com/Textualize/textual/issues/3615#issuecomment-1786745548 for some extra context.

I suggest we rename `DOMNode.workers` to `DOMNode.app_workers` and/or make it so that `DOMNode.workers` retrieves all workers started by/associated with the given widget.

TomJGooding:
Closes #3619. 

> I suggest we rename `DOMNode.workers` to `DOMNode.app_workers` and/or make it so that `DOMNode.workers` retrieves all workers started by/associated with the given widget.

There doesn't seem much benefit of just renaming this property if only a shortcut for `self.app.workers`?

This PR simply removes this confusing alias, but I'm happy to look into workers only tied to the given widget with a bit of guidance.

- [ ] Updated documentation

### DIFF
--- a/src/textual/command.py
+++ b/src/textual/command.py
@@ -1025,7 +1025,7 @@ class CommandPalette(_SystemModalScreen[CallbackType]):
 
     def _cancel_gather_commands(self) -> None:
         """Cancel any operation that is gather commands."""
-        self.workers.cancel_group(self, self._GATHER_COMMANDS_GROUP)
+        self.app.workers.cancel_group(self, self._GATHER_COMMANDS_GROUP)
 
     @on(Input.Changed)
     def _input(self, event: Input.Changed) -> None:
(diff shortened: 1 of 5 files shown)


darrenburns:
I agree with this change. The alias is definitely confusing. However it'll need @willmcgugan to approve.

willmcgugan:
I think `Widget.workers` should refer to a pool of workers for the self widget, *if* that is useful. Otherwise, we can remove the alias altogether.

I do want to do another pass on workers in the near future.

TomJGooding:
Closed as stale.

Issue #3690, opened 2023-11-16
Confusion behaviour with `@click` links


davep:
I feel the way that an action is found for a `@click` link is unintuitive. Consider this code:

```python
from textual.app import App, ComposeResult
from textual.containers import Vertical
from textual.screen import Screen
from textual.widgets import Label

class ClickLabel(Label):

    def action_click(self) -> None:
        self.notify("ClickLabel.action_click reporting in!")

class Clicker(Vertical):

    def compose(self) -> ComposeResult:
        yield Label("1: [@click=app.click]Click on this for an app action to fire[/]")
        yield Label("2: [@click=screen.click]Click on this for a screen action to fire[/]")
        yield Label("3: [@click=click]Click on this for this container's action to fire[/]")
        yield ClickLabel("4: [@click=click]Click on this for the custom widget's action to fire[/]")

    def action_click(self) -> None:
        self.notify("Clicker.action_click reporting in!")


class MainScreen(Screen):

    def compose(self) -> ComposeResult:
        yield Clicker()

    def action_click(self) -> None:
        self.notify("Screen.action_click reporting in!")


class ClickBaitApp(App[None]):

    def on_mount(self) -> None:
        self.push_screen(MainScreen())

    def action_click(self) -> None:
        self.notify("App.action_click reporting in!")


if __name__ == "__main__":
    ClickBaitApp().run()
```

Here's what happens when you click on each of those links:

1. The App-level action is fired. This is expected.
2. The screen-level action is fired. This is expected.
3. The App-level action is fired; this is (I think) unexpected.
4. The widget-level action is fired; this is expected.

I feel the way that situation 3 is handled is unintuitive. Surely it would make more sense to have actions that handle clicks looked for going up the DOM? Or if not that, how would we at least introduce a `parent` namespace too? I feel it's obvious an intuitive that someone would want to compose a collection of widgets into a container and have `@click` links in the child widgets handled by an action on the container.

valentingregoire:
I have another simple example where this bug occurrs:

```python
class ViewerApp(App):
    def on_mount(self) -> None:
        self.push_screen(ViewerScreen())

    def action_next_song(self) -> None:
        self.screen.action_next_song()


class ViewerScreen(Screen):
    BINDINGS = [
        ("q", "app.quit", "Quit"),
        ("n", "next_song", "Next Song"),
    ]

    current_song_index: reactive[int] = reactive(0, recompose=True)

    def action_next_song(self) -> None:
[...]

PR #4516, opened 2024-05-16
dynamic bindings

willmcgugan:
Stab at dynamic bindings

Wedding present for @darrenburns

### DIFF
--- a/docs/guide/actions.md
+++ b/docs/guide/actions.md
@@ -185,10 +185,10 @@ We will also demonstrate what the footer will show if we return `None` from `che
     2. Returning `None` disables they key in the footer rather than hides it.
     3. Returning `None` disables they key in the footer rather than hides it.
 
-=== "actions07.tcss"
+=== "actions06.tcss"
 
     ```css title="actions06.tcss"
-    --8<-- "docs/examples/guide/actions/actions07.tcss"
+    --8<-- "docs/examples/guide/actions/actions06.tcss"
     ```
 
 === "Output"
--- a/docs/guide/actions.md
+++ b/docs/guide/actions.md
@@ -140,7 +140,7 @@ It should return one of the following values:
 
 - `True` to show the key and run the action as normal.
 - `False` to hide the key and prevent the action running.
-- `None` to disable the key (show dimmed), and prevent he action running.
+- `None` to disable the key (show dimmed), and prevent the action running.
 
 Let's write an app to put this in to practice:
 
--- a/docs/guide/actions.md
+++ b/docs/guide/actions.md
@@ -118,8 +118,86 @@ Textual supports the following action namespaces:
 
 - `app` invokes actions on the App.
 - `screen` invokes actions on the screen.
+- `focused` invokes actions on the currently focused widget (if there is one).
 
 In the previous example if you wanted a link to set the background on the app rather than the widget, we could set a link to `app.set_background('red')`.
+
+
+## Dynamic Actions
+
+!!! tip "Added in version 0.61.0"
+
+There may be situations where an action is temporarily unavailable due to some internal state within your app.
+For instance, consider an app with a fixed number of pages and actions to go to the next and previous page.
+It doesn't make sense to go to the previous page if we are on the first, or the next page when we are on the last page.
+
+We could easily add this logic to the actions, but the [footer][textual.widgets.Footer] would still display the keys even if they wouldn't do anything.
+The user may wonder why the app is showing keys that don't appear to work.
+
+We can solve this issue by implementing the [`check_action`][textual.dom.DOMNode.check_action] on our app, screen, or widget.
+This method is called with the name of the action and any parameters, prior to running actions or refreshing the footer.
+It should return one of the following values:
+
+- `True` to show the key and run the action as normal.
+- `False` to hide the key and prevent the action running.
+- `None` to disable the key (show dimmed), and prevent he action running.
+
+Let's write an app to put this in to practice:
+
+=== "actions06.py"
+
+    ```python title="actions06.py" hl_lines="27 32 35-43"
+    --8<-- "docs/examples/guide/actions/actions06.py"
+    ```
+
+    1. Prompts the footer to refresh, if bindings change.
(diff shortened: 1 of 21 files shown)


docs/guide/actions.md
```diff
@@ -118,10 +118,88 @@ Textual supports the following action namespaces:
...
+We could easily add this logic to the action methods, but the [footer][textual.widgets.Footer] would still display the keys even if they would have no effect.
+The user may wonder why the app is showing keys that don't appear to work.
+
+We can solve this issue by implementing the [`check_action`][textual.dom.DOMNode.check_action] on our app, screen, or widget.
+This method is called with the name of the action and any parameters, prior to running actions or refreshing the footer.
+It should return one of the following values:
+
+- `True` to show the key and run the action as normal.
+- `False` to hide the key and prevent the action running.
+- `None` to disable the key (show dimmed), and prevent he action running.
```
TomJGooding, line 143:
```suggestion
- `None` to disable the key (show dimmed), and prevent the action running.
```

docs/guide/actions.md
```diff
@@ -118,10 +118,88 @@ Textual supports the following action namespaces:
...
+
+Let's write an app to put this in to practice:
```
TomJGooding, line 145:
```suggestion
Let's write an app to put this into practice:
```

docs/guide/actions.md
```diff
@@ -118,10 +118,88 @@ Textual supports the following action namespaces:
...
+    ```
+
+    1. The `bindings=True` causes the footer to refresh when `page_no` changes.
+    2. Returning `None` disables they key in the footer rather than hides it.
+    3. Returning `None` disables they key in the footer rather than hides it.
+
+=== "actions07.tcss"
+
+    ```css title="actions06.tcss"
+    --8<-- "docs/examples/guide/actions/actions07.tcss"
```
TomJGooding, line 191: I this can be removed as the CSS is still `actions06.tcss`?

docs/guide/actions.md
```diff
@@ -118,10 +118,88 @@ Textual supports the following action namespaces:
...
+
+=== "actions07.py"
+
+    ```python title="actions06.py" hl_lines="17 36 38"
+    --8<-- "docs/examples/guide/actions/actions07.py"
+    ```
+
+    1. The `bindings=True` causes the footer to refresh when `page_no` changes.
+    2. Returning `None` disables they key in the footer rather than hides it.
+    3. Returning `None` disables they key in the footer rather than hides it.
```
TomJGooding, line 186:
```suggestion
    3. Returning `None` disables the key in the footer rather than hide it.
```

docs/guide/actions.md
```diff
@@ -118,10 +118,88 @@ Textual supports the following action namespaces:
...
+
+
+=== "actions07.py"
+
+    ```python title="actions06.py" hl_lines="17 36 38"
+    --8<-- "docs/examples/guide/actions/actions07.py"
+    ```
+
+    1. The `bindings=True` causes the footer to refresh when `page_no` changes.
+    2. Returning `None` disables they key in the footer rather than hides it.
```
TomJGooding, line 185:
```suggestion
    2. Returning `None` disables the key in the footer rather than hide it.
```

TomJGooding (commented):
Just a few suggestions where I spotted some typos in the docs - hope you don't mind!

TomJGooding:
Sorry also just a quick question which I've seen asked before relating to dynamic bindings.

What would be the most practical way with this new feature of hiding bindings when an `Input` or `TextArea` is currently focussed?

```python
from textual.widgets import Input, Static, Footer, Button


class HelpScreen(ModalScreen):
    BINDINGS = [("escape", "dismiss", "Exit Help Screen")]

    CSS = """
    Static {
        height: 50%;
        width: 50%;
        background: $surface;
        text-align: center;
        content-align: center middle;
    }
    """

        yield Static("This is an example modal screen")


class ExampleApp(App):
    CSS = """
    Screen {
        align: center middle;
    }

    Input {
        width: 16;
    }
    """

    BINDINGS = [("?", "help_screen", "Help")]

        yield Button()
        yield Input(placeholder="Input")
        yield Footer()

    def action_help_screen(self) -> None:
        self.push_screen(HelpScreen())


if __name__ == "__main__":
    app = ExampleApp()
    app.run()
```

TomJGooding:
I wonder if an exception should be added for <kbd>Ctrl</kbd>+<kbd>c</kbd> to quit the app, and maybe other default actions? If you wanted to hide all bindings when a particular widget was focussed for example, you could easily forget to add this check.

darrenburns:
😍

willmcgugan:
> Sorry also just a quick question which I've seen asked before relating to dynamic bindings.
> 
> What would be the most practical way with this new feature of hiding bindings when an `Input` or `TextArea` is currently focussed? Here's a contrived example where lots of bindings obviously won't work while the `Input` is focussed.
>

The issue there is that inputs capture most keys with the Key event, which bypasses bindings.

You could do the following:

```python
    def check_action(self, name, params):
        if name == "notify_binding" and isinstance(self.focused, Input):
            return False
        return True
```

But that does mean you would have to modify the app. Ideally there would be a mechanism for the widget itself to declare it will capture those keys.

willmcgugan:
> I wonder if an exception should be added for Ctrl+c to quit the app, and maybe other default actions? If you wanted to hide all (custom) bindings when a particular widget was focussed for example, you could easily forget to add this check.

I can't see that being much of an issue. As long as you default to return `True`. But maybe I have misunderstood. Can you give me some example code?

TomJGooding:
Perhaps just worth a note in the docs explaining that this also applies to builtin actions?

For example using the `Input` example above, let's say you had a long list of of 'printable' key bindings you wanted to hide. You might think you could do something like this (rather than listing all the actions) and then not understand why you can't quit the app:

```python
    def check_action(self, name, params):
        isinstance(self.focused, Input):
            return False
        return True
```

davep:
Glancing over this just now, I wonder if #3690 is at least partially satisfied by this PR.

Issue #4517, opened 2024-05-17
Footer in ModalScreen lists ignored App keybindings


twrightsman:
When I add a `Footer` to a `ModalScreen` it lists the screen's keybindings as well as the App's keybindings, even though the App's keybindings are disabled. I would expect only the ModalScreen's keybindings to be displayed?

![ModalScreen showing App keybindings too](https://github.com/Textualize/textual/assets/6300973/4a6da96c-a790-49d3-aa0c-35dd5aaf1799)

Here is a diff to [`modal02.py`](https://textual.textualize.io/guide/screens/#__tabbed_4_3) that shows this:

```diff
@@ -15,6 +15,8 @@
 class QuitScreen(ModalScreen):
     """Screen with a dialog to quit."""
 
+    BINDINGS = [("f", "request_foo", "Foo")]
+    
     def compose(self) -> ComposeResult:
         yield Grid(
             Label("Are you sure you want to quit?", id="question"),
@@ -22,6 +24,7 @@
             Button("Cancel", variant="primary", id="cancel"),
             id="dialog",
         )
+        yield Footer()
 
     def on_button_pressed(self, event: Button.Pressed) -> None:
         if event.button.id == "quit":
@@ -29,6 +32,9 @@
         else:
             self.app.pop_screen()
 
+    def action_request_foo(self) -> None:
+        print("I pity thee")
+
 
 class ModalApp(App):
     """An app with a modal dialog."""
```

twrightsman:
Fixed in 0.61

PR #4518, opened 2024-05-17
Return correct bindings inside modal screens

twrightsman:
- [x] Updated documentation

Fixes https://github.com/Textualize/textual/issues/4517

I'm not sure if the docstring for `App.namespace_bindings` should add detail on only returning bindings up to the next modal screen, but happy to amend if so!


willmcgugan:
I fixed this issue in a recent refactor. Now merged in main. Thanks for pointing it out!

PR #4535, opened 2024-05-21
Fix header icon not summoning command palette

### LINKED ISSUE #4534: Header icon broken since 0.61.0
As already reported in Discord: since upgrading to `textual==0.61.0`, clicking the header icon does no longer open the command palette.
The palette can still be summoned by hitting the keyboard shortcut.

MRE: 
```py
from textual import __version__
from textual.widgets import Label, Header

class CommandPaletteApp(App[None]):

        yield Header()
        yield Label(f"This is Textual v{__version__}")

if __name__ == "__main__":
    CommandPaletteApp().run()
```

Clicking the header icon with the mouse results in the following message being logged:
```
<action> 'command_palette' has no target. Could not find methods '_action_command_palette' or 'action_command_palette'
```

Darren and Dave were able to narrow it down do this line of code:
https://github.com/Textualize/textual/blob/964d988ac99fff1cad744b592844e4a5646023f5/src/textual/widgets/_header.py#L37

That should now be:
```
await self.run_action("app.command_palette")
```

panki27:
Closes #4534 

Literally a one line change suggested in Discord.
This should *probably* get its own regression test, but I got no idea how to do that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [0.62.1] - 2024-05-21
+
+- Fixed header icon not opening command palette since 0.61 https://github.com/Textualize/textual/issues/4534
+
 ## [0.62.0] - 2024-05-20
 
 ### Added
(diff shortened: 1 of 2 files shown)


TomJGooding:
TIL that clicking the header icon will open the command pallette.

If you wanted to add a regression test for this,, here's a quick example:

```python
# tests/test_header.py

async def test_clicking_header_icon_launches_command_palette():
    """Regression test for https://github.com/Textualize/textual/issues/4534"""

    class MyApp(App):
        def compose(self):
            yield Header()

    app = MyApp()
        await pilot.click(Header, offset=(1, 0))
        assert isinstance(app.screen, CommandPalette)
```

willmcgugan:
Thanks. I already fixed that in another branch. Your change was spot on BTW

Issue #4542, opened 2024-05-22
Footer issue on 0.63.0


darrenburns:
Just tried out 0.63.0 and seems like if you have multiple bindings for a single key they both show.

Previously only the first binding was displayed.

```python
Binding("o,ctrl+o", "options", "Options", key_display="^o"),
```

<img width="465" alt="image" src="https://github.com/Textualize/textual/assets/5740731/6a0fe1c0-27bb-4258-803b-1833c0338414">

PR #4543, opened 2024-05-22
fix multi keys

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/4542

### DIFF
--- a/src/textual/widgets/_footer.py
+++ b/src/textual/widgets/_footer.py
@@ -1,7 +1,11 @@
 from __future__ import annotations
 
+from collections import defaultdict
+
 import rich.repr
 from rich.text import Text
+
+from textual.binding import Binding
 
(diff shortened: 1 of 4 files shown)


src/textual/widgets/_footer.py
```diff
@@ -132,8 +135,15 @@ def compose(self) -> ComposeResult:
...
             if binding.show
         ]
-        self.styles.grid_size_columns = len(bindings)
+        action_to_bindings: defaultdict[str, list[tuple[Binding, bool]]] = defaultdict(
+            list
+        )
         for binding, enabled in bindings:
+            action_to_bindings[binding.action].append((binding, enabled))
+
+        self.styles.grid_size_columns = len(bindings)
```
TomJGooding, line 144: Should this instead be the length of `action_to_bindings'?

PR #4545, opened 2024-05-22
fix(footer): exclude multi bindings for grid size

TomJGooding:
Related to the fix for multiple bindings in #4543, I think the `Footer` grid size is wrong as it includes those multiple bindings?

- [ ] Updated documentation


willmcgugan:
Thanks. I don't think that would have broken anything, as long as the number of columns is larger than the number of widgets, but it is more correct!

willmcgugan (approved):

PR #4583, opened 2024-06-01
Fix `Markdown.goto_anchor` not scrolling the heading into view

davep:
`Markdown.goto_anchor` used to scroll the heading into view, there is code in there to do this. As was pointed out on Discord today, it no longer does. This PR changes how the request to scroll is done and now does the scrolling again.

Useful code to test before and after:

```python

class MarkdownGotoApp(App[None]):

        yield Markdown(
            "\n\n".join(
                f"# Heading {n}\n\n" +
                ("XXXXX " * 1000)
                for n in range(100)
            )
        )

    def on_mount(self) -> None:
        if self.query_one(Markdown).goto_anchor("heading-50"):
            self.notify("Markdown thinks we found and went to the anchor")
        else:
            self.notify("The anchor could not be found", severity="error")


if __name__ == "__main__":
    MarkdownGotoApp().run()
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -99,7 +99,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Dependency issue
 
-## [0.60.0] - 2024-05-14 
+## [0.60.0] - 2024-05-14
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


davep:
Aaaaaaand the tests failed because GitHub threw up while trying to set up a test environment and it seems I no longer have rights to rerun a test. ಠ_ಠ

willmcgugan (approved):

willmcgugan:
Thanks

PR #4601, opened 2024-06-04
fix(footer): add space between binding key and description

### LINKED ISSUE #4557: New `Footer` missing space between key and description
Originally discussed in #4555, it looks like the new `Footer` doesn't have any space between the key and the description, which doesn't look great if you've added a background style for the key.

```python
from textual.widgets import ClassicFooter, Footer, Label


class FooterApp(App):
    BINDINGS = [("space", "switch_footer", "Switch Footer")]

    CSS = """
    Screen {
        align: center middle;
    }

    Footer {
        background: $accent;
    }

    Footer > FooterKey {
        background: $accent;
        color: $text;
        .footer-key--key {
            background: $accent-darken-2;
            color: $text;
        }
    }
    """

    footer_type = "classic"

        yield Label(
            f"{self.footer_type.title()} Footer",
            id="footer-type-label",
        )
        yield ClassicFooter()

    def action_switch_footer(self) -> None:
        if self.footer_type == "classic":
            self.query_one(ClassicFooter).remove()
            self.mount(Footer())
            self.footer_type = "new"
        else:
            self.query_one(Footer).remove()
            self.mount(ClassicFooter())
            self.footer_type = "classic"

        self.query_one(Label).update(
            f"{self.footer_type.title()} Footer",
        )


if __name__ == "__main__":
    app = FooterApp()
    app.run()
```

TomJGooding:
Fixes #4557.

I figured this was a simple enough change to just make a PR and see this change was acceptable.

Here's some quick before and after screenshots. I purposely chose the ugliest example with the default footer styling, but for the most part an extra space isn't really too noticeable.

image

image

- [ ] Updated documentation

### DIFF
--- a/src/textual/widgets/_footer.py
+++ b/src/textual/widgets/_footer.py
@@ -94,7 +94,10 @@ class FooterKey(Widget):
             )
         else:
             label_text = Text.assemble(
-                (f" {key_display} ", key_style), (description, description_style), " "
+                (f" {key_display} ", key_style),
+                " ",
+                (description, description_style),
+                " ",
             )
         label_text.stylize_before(self.rich_style)
         return label_text
--- a/tests/snapshot_tests/snapshot_apps/footer_spacing.py
+++ b/tests/snapshot_tests/snapshot_apps/footer_spacing.py
@@ -1,0 +1,19 @@
+from textual.app import App, ComposeResult
+from textual.widgets import Footer
+
+
+class FooterApp(App):
+    CSS = """
+    Footer > FooterKey .footer-key--key{
+        background: red;
+    }
+    """
+    BINDINGS = [("q", "quit", "Quit the app")]
+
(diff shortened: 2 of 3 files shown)


willmcgugan:
I couldn't accept a solution that always added that extra space. One of the main motivations around the Footer redesign was to be more compact.

How about the key component class takes the padding from the Style? That would make it customizable.

TomJGooding:
@willmcgugan No problem, to be honest I expected you would say that!

> How about the key component class takes the padding from the Style? That would make it customizable.

Sorry perhaps I'm missing something obvious, but I don't understand how this might be implemented?

willmcgugan:
The component style could have padding, which you could look up like this:

```python
self.get_component_style("footer-key--key").padding
```

The user could then add the space with something like this:

```css
Footer>footer-key--key {
    padding-right: 1;
}
```

PR #4624, opened 2024-06-08
Account for scrollbar in DataTable page-up/down

johanfforsberg:
If the content of a DataTable is wide enough to require a horizontal scrollbar, the page-up/down actions use the wrong height to figure out how far to move the cursor. At least this is what I think because it's inconsistent, e.g. the Tree widget does not behave like this.


darrenburns:
Thank you - I included your commit in https://github.com/Textualize/textual/pull/4633 which is now merged, and will likely be part of a release today.

PR #4631, opened 2024-06-09
Keys

willmcgugan:
Implements Kitty's keyboard protocol https://sw.kovidgoyal.net/kitty/keyboard-protocol/#modifiers

This should enable a few more keys, and modifier combinations. Also sane escape key detection.

iTerm implements an earlier version of this protocol, but it seems backwards compatible. Kitty and a number of other more modern terminals implement it fully.

Of course, Terminal.app doesn't. Nothing should break, but it does mean that some keys and combinations will be different. Not sure what we can do about that, other than document it. But it would be a shame not to support it if we can.


willmcgugan:
@darrenburns @davep You may find this interesting.

davep:
Nice. I'll try and find some time to play.

PR #4650, opened 2024-06-14
recompose

willmcgugan:
This is an attempt to allow recompose to keep some widgets.

If a recomposed widget has the same id, it will have its state copied from the newly composed widget, rather than being removed.

### DIFF
--- a/src/textual/_compose.py
+++ b/src/textual/_compose.py
@@ -76,3 +76,31 @@ def compose(node: App | Widget) -> list[Widget]:
         app._compose_stacks.pop()
         app._composed.pop()
     return nodes
+
+
+def recompose(node: App | Widget) -> tuple[list[Widget], set[Widget]]:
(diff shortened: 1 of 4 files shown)


merriam:
I'm looking at this.  I see  "let's compose a new widget with stuff from previous widgets" and  my analyst hairs rise on the back of my neck.

So, first, what is the use case?  I can only think of cases where a hidden widget or ContentSwitcher should be in design time.  As the layout is not interactive in the API, I cannot see where 'I need to change what I didn't know'?  There are cases for copying trees, in order to keep the code clean, but replacing?

Second, what is state?  If I change a Label to a Static does the value stay?   The formatting?  Only the formatting not derived from DEFAULT_CSS?  Current focus?  Is this meaning of State different from all other meanings?

Third, is there a more traditional approach?  Something like `shippingAddressWidget = copy_widgets(prefix='ship-', from_prefix='bill-")`?

PR #4651, opened 2024-06-14
fix(footer): enable spacing between key and description

### LINKED ISSUE #4557: New `Footer` missing space between key and description

TomJGooding:
This is my (second) attempt to fix #4557. It isn't pretty but I'm still not sure how this should be implemented. Perhaps I'm missing something obvious, but the fact that the Footer currently uses 'raw' spaces suggests adding padding might not be so simple?

- [ ] Updated documentation


TomJGooding:
I'm actually now wondering if this whole section should be adapted for CSS styling with padding and margins:

https://github.com/Textualize/textual/blob/e9ad400559c14645e56b7a98a467295c6817686b/src/textual/widgets/_footer.py#L91-L98

Maybe *something* like this. Rather than hardcoding `f" {key_display} "` on line 97, instead this could be defined using padding. The added space after the description could be defined using margin.

What do you think? 

```python
        label_text = Text.assemble(
            # Key display (with padding)
            (" " * key_padding.left, key_style),
            (key_display, key_style),
            (" " * key_padding.right, key_style),
            # Add spacing to right of the key
            (" " * key_margin.right),
            # Description display
            (description, description_style),
            # Add spacing to right of the description
            (" " * description_margin.right),
        )
```

willmcgugan:
Seems reasonable. Although one thing to be aware of is that margins overlap. So the space between key and description should be the maximum of the two margins.

TomJGooding:
Sorry also just to note after diving deeper into this: I spotted the 'default' footer uses a raw space after the description, whereas in the 'compact' style instead uses the grid gutter...?

https://github.com/Textualize/textual/blob/4a40027ef6406a46866574af20a4b1d7c358bc78/src/textual/widgets/_footer.py#L123-L125

TomJGooding:
I've gone ahead and changed that hardcoded space to the `grid-gutter'. This caused multiple snapshot tests to fail despite no visual changes (example below) so I updated the snapshots. If you're happy to merge this PR, I wonder if worth giving projects like memray a heads up on this given the spate of other changes to the footer.

image

TomJGooding:
I've reverted the commit that updated the snapshot tests. I wanted to check if #4675 might account now for these changes but unfortunately not.

darrenburns (changes requested):
Seems like there's a rendering issue - see my other comment. Would be good to investigate this before merging.

TomJGooding:
Good catch, thanks Darren! I might be wrong but is the fix as simple as adding `layout=True` to the reactive?

darrenburns:
@TomJGooding Sadly that doesn't seem to fix it 😢

TomJGooding:
Sorry to clarify, I mean the `compact` reactive in the `FooterKey`(not the Footer). This seems to fix the example above from a quick test.

TomJGooding:
Thanks Darren, good catch again as I hadn't spotted I'd changed the hover styling.

I expect that is why the "the 'default' footer uses a raw space after the description, whereas in the 'compact' style instead uses the grid gutter" then!

darrenburns:
Yeah I'm guessing that's why. I tried playing around with the CSS for a few minutes and seem to get every combination of spacing apart from the one I'm aiming for 😂

TomJGooding:
Thanks again Darren for catching these issues, in hindsight I probably should have added more snapshot tests before making changes.

I will double-check with a fresh head tomorrow, but hopefully these are resolved with my latest commits.

darrenburns (approved):
Looks good, thanks!

PR #4661, opened 2024-06-18
Refresh input when placeholder reactive is updated

darrenburns:
More of a discussion - I'm not sure why this is required.

When I don't do this explicit refresh and the reactive is changed, the placeholder doesn't always refresh immediately.

If it's a reactive, shouldn't the widget be repainting without the need for this explicit refresh?


willmcgugan:
I'm very surprised this is necessary. Refreshing is the default behavior. Some caching on the placeholder text maybe?

PR #4671, opened 2024-06-24
Allow header clock to display dates

### LINKED ISSUE #4666: Travel back in time to 1900
This is not really a bug - more of a small feature request.

The HeaderClock is, well, a clock. It displays the time. It would be nice if it could display the date as well.  I spent more time than I will admit trying to figure out why, when I set `time_format` on `Header()` to `"%c"` the time is displayed as `'Mon Jan  1 22:36:22 1900'`.  In fact, it is because line 92 in `header.py` says:

```python
return Text(datetime.now().time().strftime(self.time_format))
```

When a `datetime` is converted by `time()`, all date info is stripped.

The `time()` seems unnecessary. I propose that the line be replaced by

```python
return Text(datetime.now().strftime(self.time_format))
```

That would allow date formatting strings to be used as well, for those of us who can't remember what day it is. You might also need to change the default `width` on `HeaderClockSpace` to `auto`.

I'll do a PR if you really want one :)

lawrenceakka:

- [x] Updated documentation

Presently, the header clock can only display the time (not unreasonably for a clock). But sometimes a date is useful as well.  Default width is changed to allow for longer strings when dates are displayed

Closes #4666

### DIFF
--- a/src/textual/widgets/_header.py
+++ b/src/textual/widgets/_header.py
@@ -52,7 +52,7 @@ class HeaderClockSpace(Widget):
     DEFAULT_CSS = """
     HeaderClockSpace {
         dock: right;
-        width: 10;
+        width: auto;
         padding: 0 1;
     }
     """
@@ -89,7 +89,7 @@ class HeaderClock(HeaderClockSpace):
         Returns:
             The rendered clock.
         """
-        return Text(datetime.now().time().strftime(self.time_format))
+        return Text(datetime.now().strftime(self.time_format))
 
 
 class HeaderTitle(Widget):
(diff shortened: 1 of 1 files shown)


src/textual/widgets/_header.py
```diff
@@ -52,7 +52,7 @@ class HeaderClockSpace(Widget):
     DEFAULT_CSS = """
     HeaderClockSpace {
         dock: right;
-        width: 10;
+        width: auto;
```
willmcgugan, line 55: Does this cause the title to change position, with a long date?

PR #4673, opened 2024-06-25
Mention more source examples, as discussed in discussion #4653, point 1.

merriam:

- [X] Updated documentation
- [n/a] Updated CHANGELOG.md (where appropriate)

This is a documentation only update.

This mentions more source examples in introductory documents, per discussion #4653, point 1.   Specifically, it mentions that there are examples in `examples/` and `docs/examples` across `docs/getting_started.md`, `docs/guide/index.md`, and `docs/widget_gallery.md`

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -104,7 +104,7 @@ cd textual/examples/
 python code_browser.py ../
 ```
 
-
+Also, the tree of directories under `docs/examples/` contains hundreds of examples for all parts of Textual. 
 
 
 ## Need help?
--- a/docs/guide/index.md
+++ b/docs/guide/index.md
@@ -7,3 +7,5 @@ Welcome to the Textual Guide! An in-depth reference on how to build apps with Te
 Most of the code in this guide is fully working&mdash;you could cut and paste it if you wanted to.
 
 Although it is probably easier to check out the [Textual repository](https://github.com/Textualize/textual) and navigate to the `docs/examples/guide` directory and run the examples from there.
+
+The tree of directories under `docs/examples/` contains of hundreds of examples for all parts of Textual.  There are also complete applications in Textual's `examples/` directory. 
--- a/docs/widget_gallery.md
+++ b/docs/widget_gallery.md
@@ -9,6 +9,11 @@ Welcome to the Textual widget gallery.
 
 We have many more widgets planned, or you can [build your own](./guide/widgets.md).
 
+Most of the code demonstrating each widget is fully working&mdash;you could cut and paste it if you wanted to.
+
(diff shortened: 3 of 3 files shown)


docs/getting_started.md
```diff
@@ -104,7 +104,7 @@ cd textual/examples/
 python code_browser.py ../
 ```
 
-
+Also, the tree of directories under `docs/examples/` contains hundreds of examples for all parts of Textual.
```
willmcgugan, line 107: These are examples for the docs. They don't make much sense out of context. i.e. I wouldn't want to give the impression that the examples are the recommended approach for apps.

docs/guide/index.md
```diff
@@ -7,3 +7,5 @@ Welcome to the Textual Guide! An in-depth reference on how to build apps with Te
 Most of the code in this guide is fully working&mdash;you could cut and paste it if you wanted to.
 
 Although it is probably easier to check out the [Textual repository](https://github.com/Textualize/textual) and navigate to the `docs/examples/guide` directory and run the examples from there.
+
+The tree of directories under `docs/examples/` contains of hundreds of examples for all parts of Textual.  There are also complete applications in Textual's `examples/` directory.
```
willmcgugan, line 11: Same comment, no need to mention docs/examples here. But examples/ is probably a good idea.

docs/widget_gallery.md
```diff
@@ -9,6 +9,11 @@ Welcome to the Textual widget gallery.
 
 We have many more widgets planned, or you can [build your own](./guide/widgets.md).
 
+Most of the code demonstrating each widget is fully working&mdash;you could cut and paste it if you wanted to.
```
willmcgugan, line 12: I don't think the repetition is necessary.

willmcgugan (changes requested):

docs/widget_gallery.md
merriam: The repetition is only found if you happen to follow a particular path through the documentation.   As people use different paths, it is very possible that the widget reference will be the only time they see the comment.  While code avoids repetition for correctness, documentation embraces it for effectiveness.

docs/getting_started.md
merriam: In practice, a large number of programmers read the code over the docs.   That is, grep the entire set of examples for the usage of some function rather than start with the docs of the widget which, somewhere in the page, may mention the function.   

Let me see if better wording would clarify this.

willmcgugan:
This may give the impression that the short examples for widgets are the best way to structure a larger app. Searching the docs is pretty much guaranteed to give better results than grepping these examples.

I'll add a note in getting started, but I think its best to leave the other edits. Thanks for your time on this.

PR #4674, opened 2024-06-25
Automatically detect terminal background color - fixes #2419

### LINKED ISSUE #2419: Automatic choice of dark/light theme based on terminal's colours
In lots of terminals, it's possible to detect the background colour of the terminal with something like this:
```
#!/bin/bash

oldstty=$(stty -g)
stty raw -echo min 0 time 0
printf "\033]11;?\033\\"
result=""
while [ -z "$result" ]; do
    read -r result
done
stty "$oldstty"

echo "$result" | sed -E 's/.*(rgb:[a-f0-9/]{14}).*/\1/'
```

Then you get output like `rgb:0000/0000/0000`, `rgb:ffff/ffff/ffff`, `rgb:158e/193a/1e75`, etc.

Textual could use this to select a dark or light theme by default. This would be nice as apps would behave and look as users expect them without any configuration.

I would prefer this to happen automatically but, if you think this is not good as some terminals don't support this, then textual/rich could at least provide code to detect this. 

I mention it in https://discord.com/channels/1026214085173461072/1033752510848053399/1075543323516866662 a while ago, but then forgot to open an issue.

dhallas:
Add support for automatically detecting the background color of the terminal using operating system command escape sequence. The response is then parsed and used to determine if the background is light or dark.

- [ X] Docstrings on all new or modified functions / classes 
- [ X] Updated documentation
- [ X] Updated CHANGELOG.md (where appropriate)


dhallas:
I would like to get some feedback on the approach I have taken to add support for this.

I have only tested this with "Kitty" on Mac OS, more testing is needed.

Should the auto light/dark detection be configurable? And maybe off by default?

dhallas:
@willmcgugan  - can you take a look at this MR to check if I am heading in the right direction?

src/textual/_xterm_parser.py
```diff
@@ -31,6 +31,10 @@
 """Sequence received when the terminal receives focus."""
 FOCUSOUT: Final[str] = "\x1b[O"
 """Sequence received when focus is lost from the terminal."""
+BG_COLOR: Final[str] = "\x1b]11;rgb:"
```
willmcgugan, line 34: We have a policy in Textual of no abbreviations, so BG should be BACKGROUND.

I think this would be better done as a regex.

Do you have a reference to the docs? Can you be certain it is sent in a two hex characters?

src/textual/app.py
```diff
@@ -3275,6 +3275,15 @@ async def _on_app_blur(self, event: events.AppBlur) -> None:
         self.app_focus = False
         self.screen.refresh_bindings()
 
+    def _is_dark_color(self, r: int, g: int, b: int) -> bool:
+        # perceived brightness formula
+        perceived_brightness = 0.2126 * r + 0.7152 * g + 0.0722 * b
+        return perceived_brightness < 128
+
+    async def _on_background_color(self, event: events.BackgroundColor) -> None:
```
willmcgugan, line 3283: I wouldn't want to do this automatically, unless the dev explicitly requests this behaviour.

Sending the message is probably fine, until we have a mechanism for the dev to request automatic light / dark mode.

willmcgugan (changes requested):
What happens on terminals that don't support this?

If you have a Mac, you might want to test terminal.app

src/textual/drivers/linux_driver.py
```diff
@@ -248,6 +248,7 @@ def on_terminal_resize(signum, stack) -> None:
         self.flush()
         self._key_thread = Thread(target=self._run_input_thread)
         send_size_event()
+        self.write("\x1b]11;?\x07")  # Detect background color
```
arcivanov, line 251: AFAIK some terminals respond with nothing when they see a sequence they don't recognize/support, i.e. if you send requests `<supported 1>;<unsupported 2>;<supported 3>;` the responses will be `<response 1>;<response 3>` only. If anything expects order guarantee for requests it may end up waiting forever. 
Is this case handled (I didn't do an in-depth review, just asking)?

dhallas:
> What happens on terminals that don't support this?
> 
> If you have a Mac, you might want to test terminal.app

You tested this with the terminal.app and to me pleasant surprise it just worked, meaning that the terminal actual supports this escape sequence and responds correctly.

src/textual/drivers/linux_driver.py
dhallas: I try to handle this case gracefully by only parsing the RGB colors if the length is as expected. But, please take a good look at the code to ensure I do not have missed anything. I would also like to look into writing some unit tests for this, including the case you mention here

src/textual/_xterm_parser.py
dhallas: I will fix the abbreviation. Here is a link to some documentation on the escape sequence:

https://www.xfree86.org/current/ctlseqs.html

Search for "Operating System Controls"

Issue #4382, opened 2024-04-03
Potential bug with bindings not appearing after 0.55.0


mzebrak:
I'm naming this issue as "potential" since I don't know what is the expected behavior, but since there was nothing about it in the release note (and this change is breaking), wanted to show it.

For me, it looks like a bug because things specified at the Screen level have a higher priority than at the App level (I mean even without explicitly setting the priority=True)

Appears after 0.55.0, specifically talking, I'm suspecting this change:
>  Fix priority bindings not appearing in footer when key clashes with focused widget https://github.com/Textualize/textual/pull/4342

Consider such an MRE:
```python
from __future__ import annotations

from textual.app import App, ComposeResult
from textual.binding import Binding
from textual.screen import Screen
from textual.widgets import Footer, Static


class ScreenWithNoBindingVisible(Screen):
    def compose(self) -> ComposeResult:
        yield Static(self.__class__.__name__)
        yield Footer()


class ScreenThatShowedBindingBefore_0_55_0(ScreenWithNoBindingVisible):
    BINDINGS = [
        Binding("f1", "notify", "Hey, I was visible before 0.55.0!"),
    ]


class BindingOverrideApp(App):
    BINDINGS = [
        Binding("f1", "notify", show=False),
        Binding("p", "pop_screen", "Pop screen"),
    ]

    def on_mount(self) -> None:
        self.push_screen(ScreenWithNoBindingVisible())
        self.push_screen(ScreenThatShowedBindingBefore_0_55_0())

    def action_notify(self):
        self.notify("F1 pressed!")


BindingOverrideApp().run()
```

Pre 0.55.0 (tested on 0.54.0):
![image](https://github.com/Textualize/textual/assets/73518558/91fa666c-96af-4ec9-8a5b-562438efad64)

On 0.55.0:
![image](https://github.com/Textualize/textual/assets/73518558/7f9b1429-5f49-46f9-84ab-b36719bac4b5)

Of course it's fixable by changing the following line like:
```python
Binding("f1", "notify", "Hey, I was visible before 0.55.0!", priority=True),
```

but this is what's this issue about - previously it worked like a charm without any explicit setting.
I'm curious what you think, is this a bug or is it the default behavior from now on?

From my side - previous behavior was expected - if I disabled something globally, then I could enable it on some screen/widget (which is a more local setting) just by overriding BINDINGS with `show=True`. Worked with no surprises.

darrenburns:
[...]

PR #4695, opened 2024-07-01
test(bindings): add regression test for #4382

TomJGooding:
This adds a regression test for #4382, where the visibility of bindings in the app should be overridden by bindings defined on a screen.

- [ ] Updated documentation


darrenburns:
Thanks Tom!

PR #4708, opened 2024-07-05
prune

willmcgugan:
Hopefully more reliable pruning of DOM nodes.

When a node is removed, it requests that it's children are removed, before sending itself an Unmount message.


darrenburns (approved):

src/textual/widgets/_list_view.py
```diff
@@ -262,7 +262,7 @@ def remove_items(self, indices: Iterable[int]) -> AwaitRemove:
         for index in indices:
             items_to_remove.append(items[index])
 
-        await_remove = self.app._remove_nodes(items_to_remove, self)
+        await_remove = items.remove()
```
jkoelker, line 265: I believe this changes the behavior of this functions to removing all `ListItem`s instead of the ones at the specific `indices`.

Issue #4705, opened 2024-07-04
Several command palette issues on 0.71.0


darrenburns:
I found 3 issues with the command palette in the latest release.

1. The help text doesn't appear for the initial "discovery" hits. When you start to search, the help text starts to appear.
2. There's something wrong with scrolling when you use the keyboard to move the cursor through the option list. In the example below, scrolling seems to break after you've started searching.
3. The initial commands listed (DiscoveryHits) are sorted alphabetically, which seems wrong (I want control over the order - it's easy enough for me to sort my commands if that's what I require). Normal hits from searches are not sorted (as expected).

```python
from typing import Any
from textual.app import App, ComposeResult
from textual.command import DiscoveryHit, Hit, Hits, Provider
from textual.widgets import Label


class MyProvider(Provider):
    @property
    def commands(self) -> list[tuple[str, Any, str]]:
        return [(f"option-{i}", lambda: 1, f"This is option {i}") for i in range(30)]

    async def discover(self) -> Hits:
        for name, runnable, help_text in self.commands:
            yield DiscoveryHit(
                name,
                runnable,
                help_text,
            )

    async def search(self, query: str) -> Hits:
        matcher = self.matcher(query)
        for name, runnable, help_text in self.commands:
            if (match := matcher.match(name)) > 0:
                yield Hit(
                    match,
                    matcher.highlight(name),
                    runnable,
                    help=help_text,
                )


class OptionListScrolling(App[None]):
    COMMANDS = {MyProvider}

    def compose(self) -> ComposeResult:
        yield Label("hey!")


if __name__ == "__main__":
    app = OptionListScrolling()
    app.run()

```

TomJGooding:
> The help text doesn't appear for the initial "discovery" hits. When you start to search, the help text starts to appear.

It looks like you just need a small tweak to your `DiscoveryHit`:

```python
    async def discover(self) -> Hits:
        for name, runnable, help_text in self.commands:
            yield DiscoveryHit(
                name,
                runnable,
                help=help_text,
            )

TomJGooding:
> There's something wrong with scrolling when you use the keyboard to move the cursor through the option list.

From a quick look at the recent changes to the `OptionList`, I'm not sure the "line position for the start of the option" is correctly tracked here?

https://github.com/Textualize/textual/blob/37f0f26f1c9380a9fe45866dd383005686df7839/src/textual/widgets/_option_list.py#L363-L364

If `OptionList

[...]

PR #4709, opened 2024-07-07
fix(option list): fix scrolling with multiline options

TomJGooding:
The recent changes to the `OptionList` widget unfortunately have broken scrolling where options span multiple lines.

For example, try scrolling the [option of tables example in the docs](https://textual.textualize.io/widgets/option_list/#options-as-rich-renderables) using the up/down keys. This is also the cause of the `CommandPalette` issues described in #4705.

The `OptionList` had quite a overhaul with the recent update, so I'm creating this PR as only a draft for now. If I have understood the changes correctly, I'm not sure the "line position for the start of the option" is correctly tracked here?

https://github.com/Textualize/textual/blob/37f0f26f1c9380a9fe45866dd383005686df7839/src/textual/widgets/_option_list.py#L363-L364

- [ ] Updated documentation

### DIFF
--- a/src/textual/widgets/_option_list.py
+++ b/src/textual/widgets/_option_list.py
@@ -20,7 +20,7 @@ from ..scroll_view import ScrollView
 from ..strip import Strip
 
 if TYPE_CHECKING:
-    from typing_extensions import Self
+    from typing_extensions import Self, TypeAlias
 
 
 class DuplicateID(Exception):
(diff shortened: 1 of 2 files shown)


willmcgugan:
Thanks, @TomJGooding This does seem to be the fix. Any objections if I merge?

TomJGooding:
I have no objections if you're happy with the fix! 

I was wondering if the `OptionList` data structures might need re-evaluating, but that's your call. This will at least provide a fix in the meantime.

willmcgugan:
They might. This whole class is a candidate for a rewrite, tbh.

PR #4720, opened 2024-07-10
Don't sort "discover" items in the command palette

darrenburns:
I think it's reasonable to give the user control over the order of items displayed in the command palette when it's first opened (I want to group related commands).

If they want to sort, they can easily call `sorted` themselves.

Closes https://github.com/Textualize/textual/issues/4705 (fixes the final bullet point - the other problems are already resolved as of `0.72.0`)


darrenburns:
Cheers @TomJGooding - I've removed that line now!

willmcgugan (approved):

Issue #3130, opened 2023-08-21
Button background stuck in hover style if disabled programatically 


TomJGooding:
### Is there an existing issue for this?

I found this issue #790 closed last November, but it isn't quite describing the same thing.

- [x] I have searched the open issues

### Description of the bug

If you are hovering a `Button` that is then disabled programmatically, the button background colour seems to be stuck in the hover style.

### To Reproduce

```python
from textual.app import App, ComposeResult
from textual.containers import Center
from textual.widgets import Button, Footer, Label


class ExampleApp(App):
    CSS = """
    Screen {
        align: center middle;
    }
    """

    BINDINGS = [("space", "toggle_button", "Toggle Button")]

    def compose(self) -> ComposeResult:
        with Center():
            yield Label("Hover the button then hit space")
        with Center():
            yield Button("Enabled", id="disable-btn")
        yield Footer()

    def action_toggle_button(self) -> None:
        self.app.bell()
        button = self.query_one("#disable-btn", Button)
        if button.disabled is False:
            button.disabled = True
            button.label = "Disabled"
        else:
            button.disabled = False
            button.label = "Enabled"


if __name__ == "__main__":
    app = ExampleApp()
    app.run()
```

TomJGooding:
I've found a workaround by setting `mouse_over` to False when disabling the button - despite the documentation saying this is read only!

Should something similar be added to the  base Widget class...?

```python
from textual.app import App, ComposeResult
from textual.containers import Center
from textual.widgets import Button, Footer, Label


class ExampleApp(App):
    CSS = """
    Screen {
        align: center middle;
    }
    """

    BINDINGS = [("space", "toggle_button", "Toggle Button")]

    def compose(self) -> ComposeResult:
        with Center():
            yield Label("Hover the button then hit space")
        with Center():
            yield Button("Enabled")
        yield Footer()

    def action_toggle_button(self) -> None:
        button = self.query_one(Button)
        if button.disabled is False:
            button.disabled = True
            button.label = "Disabled"
            # Workaround for https://github.com/Textualize/textual/issues/3130
            button.mouse_over = False
        else:
            button.disabled = False
            button.label = "Enabled"


if __name__ == "__main__":
    app = ExampleApp()
    app.run()
```

klott:
This is biting me too, thanks for the workaround!

TomJGooding:
This issue has just come up again in #4198, weirdly only a couple of days after the linked PR has apparently  been abandoned!

[...]

PR #4724, opened 2024-07-11
Fix stuck hover state

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/3130

### DIFF
--- a/src/textual/widget.py
+++ b/src/textual/widget.py
@@ -3239,6 +3239,7 @@ class Widget(DOMNode):
         from .app import ScreenStackError
 
         if disabled and self.mouse_over:
+            # Ensure widget gets a Leave if it is disabled while hovered
             self._message_queue.put_nowait(events.Leave())
         try:
             screen = self.screen
--- a/src/textual/widget.py
+++ b/src/textual/widget.py
@@ -3234,19 +3234,23 @@ class Widget(DOMNode):
         """Update from CSS if has focus state changes."""
         self._update_styles()
 
-    def watch_disabled(self) -> None:
+    def watch_disabled(self, disabled: bool) -> None:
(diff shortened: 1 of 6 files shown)


src/textual/widget.py
```diff
@@ -3234,19 +3234,23 @@ def watch_has_focus(self, value: bool) -> None:
         """Update from CSS if has focus state changes."""
         self._update_styles()
 
-    def watch_disabled(self) -> None:
+    def watch_disabled(self, disabled: bool) -> None:
         """Update the styles of the widget and its children when disabled is toggled."""
         from .app import ScreenStackError
 
+        if disabled and self.mouse_over:
+            self._message_queue.put_nowait(events.Leave())
```
darrenburns, line 3242: This block all makes sense to me other than posting the `Leave` event, because the mouse has not left the widget.

darrenburns (changes requested):
I think we should avoid posting a `Leave` event, as it feels like a workaround and the events are part of the public API. There's probably a way we could do this internally?

src/textual/widget.py
willmcgugan: It's because when a widget is disabled, it no longer receives Enter and Leave events. So a widget that is hovered and programmatically disabled will have received an Enter, but will never get a corresponding Leave. So whatever the widget does on enter is stuck even if the cursor is no longer over the widget.

Sending the Leave here ensures that the widget goes back to its unhovered state when disabled.

darrenburns: Makes sense after the explanation in the first paragraph, thanks. Probably worth a small update to the Leave docstring.

willmcgugan: Done

darrenburns (approved):

PR #4732, opened 2024-07-12
Reorder reactive comparison to short-circuit uncomparable types

kdkavanagh:
If `always_update`, dont attempt to compare new vs old.  Useful for types with cannot be compared or have otherwise complicated comparison logic that the user has indicated is unnecessary for the `reactive` instance


willmcgugan:
Thanks, fixed in https://github.com/Textualize/textual/pull/4731/

PR #4736, opened 2024-07-14
fix(listview): fix remove_items

### LINKED ISSUE #4735: `ListView.remove_items()` removes all items
Thanks @jkoelker for reporting this in https://github.com/Textualize/textual/pull/4708#discussion_r1677011195!

> I believe this changes the behavior of this functions to removing all `ListItem`s instead of the ones at the specific `indices`.

https://github.com/Textualize/textual/blob/66f2e0cf45ba4ee20b2a7d150bc8ab5f034d1f54/src/textual/widgets/_list_view.py#L262-L266

TomJGooding:
Fixes #4735 

Widget.remove_children now accepts an iterable of Widgets.

### DIFF
--- a/src/textual/widgets/_list_view.py
+++ b/src/textual/widgets/_list_view.py
@@ -258,11 +258,8 @@ class ListView(VerticalScroll, can_focus=True, can_focus_children=False):
             An awaitable object that waits for the direct children to be removed.
         """
         items = self.query("ListItem")
-        items_to_remove = []
-        for index in indices:
-            items_to_remove.append(items[index])
-
-        await_remove = self.app._prune(*items_to_remove, parent=self)
+        items_to_remove = [items[index] for index in indices]
+        await_remove = self.remove_children(items_to_remove)
         return await_remove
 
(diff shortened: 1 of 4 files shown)


src/textual/widgets/_list_view.py
```diff
@@ -262,7 +262,7 @@ def remove_items(self, indices: Iterable[int]) -> AwaitRemove:
         for index in indices:
             items_to_remove.append(items[index])
 
-        await_remove = items.remove()
+        await_remove = self.app._prune(*items_to_remove, parent=self)
```
willmcgugan, line 265: I wanted to avoid this, because widgets should be able to be written with the public API only.

Ideally we can do this through a DOMQuery, although there isn't an elegant way of excluding widgets (just selectors).

One possibility that comes to mind is that DOMQuery.remove grows an `exclude_widgets` parameter, but I am open to suggestions.

willmcgugan: Tom, hope you don't mind, I'm going to make that change to this PR

Issue #4737, opened 2024-07-14
DataTable: Switching focus is slow when table contains lots of data


ihabunek:
I have a use case where I put data from a database into the data table and i noticed that if cells contain a lot of text, switching focus to and from the data table is very slow.

In the following example, when `tab` is pressed, it takes ~2 seconds to switch focus from the Input to the DataTable. Scrolling is reasonably fast once focused.

```py
from textual.app import App
from textual.widgets import DataTable, Input

NUM_LINES = 100
NUM_COLS = 20
NUM_ROWS = 50

COLS = [f"col {n}" for n in range(NUM_COLS)]
# Each cell contains a bunch of text with line breaks
CELL = "\n".join("much data wow" for _ in range(NUM_LINES))
ROW = [CELL for _ in range(NUM_COLS)]
ROWS = [ROW for _ in range(NUM_ROWS)]


class ReproApp(App):
    CSS = "DataTable { height: 80%; }"

    def compose(self):
        yield Input("Pressing TAB to switch focus is slow!")
        dt = DataTable()
        dt.add_columns(*COLS)
        dt.add_rows(ROWS)
        yield dt


if __name__ == "__main__":
    app = ReproApp()
    app.run()
```

# Textual Diagnostics

## Versions

| Name    | Value  |
|---------|--------|
| Textual | 0.72.0 |
| Rich    | 13.7.1 |

## Python

| Name           | Value                                        |
|----------------|----------------------------------------------|
| Version        | 3.12.3                                       |
| Implementation | CPython                                      |
| Compiler       | GCC 14.0.1 20240411 (Red Hat 14.0.1-0)       |
| Executable     | /home/ihabunek/.virtualenvs/pgtui/bin/python |

## Operating System

| Name    | Value                                               |
|---------|-----------------------------------------------------|
| System  | Linux                                               |
| Release | 6.9.5-200.fc40.x86_64                               |
| Version | #1 SMP PREEMPT_DYNAMIC Sun Jun 16 15:47:09 UTC 2024 |

## Terminal

| Name                 | Value       |
|----------------------|-------------|
| Terminal Application | Kitty       |
| TERM                 | xterm-kitty |
| COLORTERM            | truecolor   |
| FORCE_COLOR          | *Not set*   |
| NO_COLOR             | *Not set*   |

## Rich Console options

| Name           | Value                |
|----------------|----------------------|
| size           | width=182, height=48 |
| legacy_windows | False                |
| min_width      | 1                    |
[...]

PR #4748, opened 2024-07-15
Datatable performance improvements

darrenburns:
Massively improves performance when cells contain multi-line content.

Fixes https://github.com/Textualize/textual/issues/4737 (also halves the startup time of that app!)

### DIFF
--- a/src/textual/widgets/_data_table.py
+++ b/src/textual/widgets/_data_table.py
@@ -154,21 +154,21 @@ class CellKey(NamedTuple):
         yield "column_key", self.column_key
 
 
-def _find_nth_newline(string: str, n: int) -> int:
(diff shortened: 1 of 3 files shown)


src/textual/widgets/_data_table.py
```diff
@@ -152,22 +154,68 @@ def __rich_repr__(self):
         yield "column_key", self.column_key
 
 
-def default_cell_formatter(obj: object) -> RenderableType:
+def _find_nth_newline(string: str, n: int) -> int:
```
willmcgugan, line 157: Sorry, have to veto the single letter variable...

willmcgugan (approved):
Great. Just a pedantic request.

PR #4772, opened 2024-07-18
feat(tree): enable add at location

### LINKED ISSUE #4049: Add index to Tree widget's add
Presently there is no way to add a new tree node except as the last node, accordingly:

Would you accept an optional `index`here?
https://github.com/Textualize/textual/blob/ce962196c73b2b87c9f812f66a6f6ded3f2aaf4b/src/textual/widgets/_tree.py#L320-L327

If set it would change this from `append` to `insert` and pass `index`:
https://github.com/Textualize/textual/blob/ce962196c73b2b87c9f812f66a6f6ded3f2aaf4b/src/textual/widgets/_tree.py#L344

TomJGooding:
Add `before` and `after` parameters to the `TreeNode.add` and `TreeNode.add_leaf` methods to enable positioning a new node.

Closes #4049.

- [ ] Updated documentation


src/textual/widgets/_tree.py
```diff
@@ -330,34 +336,101 @@ def add(
...
+
+        if before is not None:
+            if isinstance(before, int):
+                insert_index = before
+            elif isinstance(before, TreeNode):
+                try:
+                    insert_index = self.children.index(before)
+                except ValueError:
+                    raise AddNodeError(
+                        "Request to add before an unknown node that is not a child of this node"
```
darrenburns, line 367: Struggled to mentally parse the error message a bit here.

```suggestion
                        "The node specified for 'before' does not exist in this Tree."
```

src/textual/widgets/_tree.py
```diff
@@ -330,34 +336,101 @@ def add(
...
+            if isinstance(after, int):
+                insert_index = after + 1
+                if after < 0:
+                    insert_index += len(self.children)
+            elif isinstance(after, TreeNode):
+                try:
+                    insert_index = self.children.index(after) + 1
+                except ValueError:
+                    raise AddNodeError(
+                        "Request to add after an unknown node that is not a child of this node"
```
darrenburns, line 384: Same as above.

darrenburns (changes requested):
Minor request to change the wording of an error message, but otherwise looks good to me! Will be very useful - thanks!

src/textual/widgets/_tree.py
darrenburns: Actually on reading the code in more detail I think my suggestion isn't clear.

Maybe something like `f"cannot insert before {before}, as this node only has {len(children)} children"`?

darrenburns: Haha yeah, I realised that after posting. See my other comment 😄

TomJGooding: > f"cannot insert before {before}, as this node only has {len(children)} children"

Sorry but I'm not sure that's quite right either. The length of the children isn't relevant, rather that the `TreeNode` passed as the `before` isn't a child node of this node?

darrenburns: ah sorry, i thought we were in the `isinstance(before, int)` branch 🤦 

how about:

```suggestion
                        "The node specified for `before` is not a child of this node"
```

TomJGooding: Sorry also just to clarify, if the `before`/`after` argument is an index that is out of range, this also won't raise an error (the same as Python's `insert`)

darrenburns: Ah ok - that seems fine to me (I wasn't aware that's how insert worked)

willmcgugan:
Merged the CHANGELOG.

Thanks, Tom. I think we're going to squeeze this in to 0.73.0, which I'll release at EOD.

src/textual/widgets/_tree.py
TomJGooding: Thanks, that wording is much better! Updated in 04ac1ca (sorry I tried adding you as co-author but always forget that second hyphen!)

darrenburns (approved):

PR #4774, opened 2024-07-18
Gradient progress bar

willmcgugan:
Adds gradient to progress bars. Also optimizes the color.Gradient class.

https://github.com/user-attachments/assets/765d2b3a-7793-4d58-9082-4d3ea9ad80cd

### DIFF
--- a/docs/examples/widgets/progress_bar_gradient.py
+++ b/docs/examples/widgets/progress_bar_gradient.py
@@ -1,0 +1,34 @@
+from textual.app import App, ComposeResult
+from textual.color import Gradient
+from textual.containers import Center, Middle
+from textual.widgets import ProgressBar
+
+
+class ProgressApp(App[None]):
+    """Progress bar with a rainbow gradient."""
+
+    def compose(self) -> ComposeResult:
+        gradient = Gradient.from_colors(
+            "#881177",
+            "#aa3355",
+            "#cc6666",
+            "#ee9944",
+            "#eedd00",
+            "#99dd55",
+            "#44dd88",
+            "#22ccbb",
+            "#00bbcc",
+            "#0099cc",
+            "#3366bb",
+            "#663399",
+        )
+        with Center():
+            with Middle():
+                yield ProgressBar(total=100, gradient=gradient)
+
+    def on_mount(self) -> None:
+        self.query_one(ProgressBar).update(progress=70)
+
+
+if __name__ == "__main__":
+    ProgressApp().run()
--- a/src/textual/renderables/bar.py
+++ b/src/textual/renderables/bar.py
@@ -86,7 +86,7 @@ class Bar:
         if half_end:
             highlight_bar.append(Text(half_bar_right, style=highlight_style, end=""))
 
-        if self.gradient:
+        if self.gradient is not None:
             _apply_gradient(highlight_bar, self.gradient, width)
         output_bar.append(highlight_bar)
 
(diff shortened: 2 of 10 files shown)


docs/examples/widgets/progress_bar_gradient.py
```diff
@@ -0,0 +1,34 @@
...
+            "#99dd55",
+            "#44dd88",
+            "#22ccbb",
+            "#00bbcc",
+            "#0099cc",
+            "#3366bb",
+            "#663399",
+        )
+        with Center():
+            with Middle():
```
darrenburns, line 26: Bit of an aside, but I often find myself defining a `CenterMiddle`.

src/textual/widgets/_progress_bar.py
```diff
@@ -60,17 +61,23 @@ class Bar(Widget, can_focus=False):
...
         classes: str | None = None,
         disabled: bool = False,
         clock: Clock | None = None,
+        gradient: Gradient | None = None,
     ):
         """Create a bar for a [`ProgressBar`][textual.widgets.ProgressBar]."""
         self._clock = (clock or Clock()).clone()
         super().__init__(name=name, id=id, classes=classes, disabled=disabled)
+        self.gradient = gradient
+        self.set_reactive(Bar.gradient, gradient)
```
darrenburns, line 80: Duplicate setter?

src/textual/renderables/bar.py
```diff
@@ -96,45 +106,25 @@ def __rich_console__(
...
+        text: A Text object.
+        gradient: A Textual gradient.
+        width: Width to apply gradient to, or None for full width.
+    """
+    from_color = Style.from_color
+    get_rich_color = gradient.get_rich_color
+
+    max_width = width - 1
+    if not max_width:
+        return
```
darrenburns, line 122: If a renderable is resized down to width 1, should we maybe sample the "first" colour from the gradient instead of doing nothing? Feels like this would maybe flicker out.

docs/examples/widgets/progress_bar_gradient.py
willmcgugan: Not a bad idea.

src/textual/renderables/bar.py
willmcgugan: Good catch. Somebody will do that.

src/textual/renderables/bar.py
```diff
@@ -112,13 +112,14 @@ def _apply_gradient(text: Text, gradient: Gradient, width: int) -> None:
...
         text: A Text object.
         gradient: A Textual gradient.
-        width: Width to apply gradient to, or None for full width.
+        width: Width of gradient.
     """
     from_color = Style.from_color
     get_rich_color = gradient.get_rich_color
 
     max_width = width - 1
     if not max_width:
```
darrenburns, line 121: Also just to check - if width == 0, then max_width will be -1 and we'll go into the loop which I don't think is intentional?

willmcgugan: The width should never be 0. It's the not the width of the renderable, but the range that is covered by the gradient.

Here it is the width of the renderable, and if that's 0, it won't render.

I could put in an assert, to validate that.

willmcgugan: But defensive code is good code. I'll add a handler for that case.

willmcgugan: Done.

darrenburns (approved):

Issue #4718, opened 2024-07-09
Integer validator does not work as expected permitting floating point numbers


arcivanov:
The culprit lies in this piece of code:

```python
        # We know it's a number, but is that number an integer?
        is_integer = float(value).is_integer()
        if not is_integer:
            return ValidationResult.failure([Integer.NotAnInteger(self, value)])
```

The problem is that `float("1.").is_integer()` and `float("1.0").is_integer()` are all `True` (thankfully `float("1.1").is_integer()` is not).

This issue is not purely cosmetic either. `int("1.")` or `int("1.0")`  all raise exceptions:

```
>>> int("1.0")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: invalid literal for int() with base 10: '1.0'
>>> int("1.")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: invalid literal for int() with base 10: '1.'
```

Therefore a validator will validate a string as an integer that cannot be parsed as an integer by the `int` type.

merriam:
Hmm.... Funky.

This is an error, meaning the code should be wrong.   On the other hand, there is no way to cause the error to express, because `_input.py` restricts the dot.   That is, you cannot type the '.' because the `_RESTRICT_TYPES` regular expression for integers is just `r"[-+]?\d*"`.    This means "000032" is fine, but "1_000_000" is not.   In practice, there is no sequence of +/- followed by digits in Python that is not an integer.

There is also this error code as well for in the float validator,`if float_value in {math.nan, math.inf, -math.inf}...`which will not trigger for a NaN, as `float('NaN') != float('NaN')`.  It too is an error, but not expressible because of regex.
 
Let me cogitate on input validation for a minute, and, baring insight, I'll put in a PR next week.

PR #4784, opened 2024-07-22
Fix Number and Float validation errors, allow underscores, add tests.  Fix issue #4718

merriam:

- [ n/a] Docstrings on all new or modified functions / classes 
- [ tbd] Updated documentation
- [ tbd] Updated CHANGELOG.md (where appropriate)

Here are some changes to the Validation routines:

- Number validator no longer accepts 'nan'.
- Number regex no longer accepts 'e' as first character of a number.
- Integers no longer accept scientific notation.  That is, "1.23e2" happens to resolve to an integer if first viewed as a float, but cannot be cast with `int(value)`.
- Integers and floats allow underscores, per the Python standard.  That is, "1_000_000" is a valid integer and can be typed in a "integer" type field.  "12." cannot be typed or validated.
- 
The tbd on documentation:  does this warrant a note in the CHANGELOG.md?   Also, I started cleaning up the Input Widet documentation and its worth a different commit.


darrenburns (approved):
LGTM

@merriam Any change which could affect a developer using Textual or an end-user using a Textual app warrants a changelog entry. Since this affects the behaviour of the Input, we'd need a changelog entry for it.

merriam:
Sorry, I hadn't realized it was waiting for CHANGELOG.   Anything else needed?

PR #4788, opened 2024-07-22
Update snapshots

darrenburns:
Updates to use latest textual-snapshot, regenerate snapshots into separate files, switch to xdist when running locally (via the Makefile), update pytest-cov.


willmcgugan (approved):
Taking your word for it!

PR #4793, opened 2024-07-23
run commands regardless of the bindings checking order

bjesus:
Currently, when a `Binding` is marked with `priority=True`, it isn't clickable with the mouse because `_check_bindings` will only run un-prioritized bindings. This removes the check and runs actions regardless of their priority status. I'm not sure why this check was there to begin with honestly, so hopefully I'm not breaking anything else.


TomJGooding:
Thanks for reporting this and your PR. I've created issue #4794 to help track this with my attempt at an MRE, but please let me know if I've misunderstood the issue.

It looks like this change causes a number of tests to fail which suggests this may not be the correct fix?

Issue #4790, opened 2024-07-22
Screen.dismiss doesn't call the callback function if it's called with no args


darrenburns:
e.g. `screen.dismiss()`

or in a Binding `Binding(..., action="screen.dismiss", ...)`.

Confirmed with this MRE:

```python
from textual.app import App, ComposeResult
from textual.screen import Screen
from textual.widgets import Static


class MyScreen(Screen):
    BINDINGS = [("space", "screen.dismiss()", "Dismiss")]

    def compose(self) -> ComposeResult:
        yield Static("Hello")


class MyApp(App):
    def on_mount(self) -> None:
        def callback(sd):
            self.notify("Callback")

        self.push_screen(MyScreen(), callback)

    def compose(self) -> ComposeResult:
        yield Static("Default Screen")


if __name__ == "__main__":
    app = MyApp()
    app.run()
```

willmcgugan:
This does seem like it was by design.

From the dismiss docstring:

>        If `result` is provided and a callback was set when the screen was [pushed][textual.app.App.push_screen], then
>       the callback will be invoked with `result`.

I think the rationale might have been that the callback exists to receive the argument to `dismiss()`. And if you don't supply one, then you don't need the callback.

On reflection supplying a callback and then discarding it feels like a wart.

PR #4795, opened 2024-07-24
allow None in Screen callback

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/4790

Previously, calling `Screen.dismiss` with no arguments would not call the Screen callback. Now, it will call the callback with a `None` argument.

This feels like a more sensible default, as the dev can decide what to do with the screen result.


willmcgugan:
#executivedecision

Issue #4794, opened 2024-07-23
Clicking priority binding in `Footer` doesn't call the correct action


TomJGooding:
Originally reported in PR #4793 - please correct me @bjesus if I've misunderstood the issue.

The example below has a priority binding in the `App`, which should be checked prior to the binding to the same key in the `TestWidget`.

When the `TestWidget` is focused:
- Pressing the <kbd>b</kbd> key calls the correct action from the app :white_check_mark:
- Clicking the binding in the `Footer` instead calls the overridden action in the widget  :x:

After moving focus by clicking the button (or pressing <kbd>Tab</kbd>):
- Pressing the <kbd>b</kbd> key calls the correct action from the app :white_check_mark:
- Clicking the binding in the `Footer` now seems to do nothing :x:

```python
from textual.app import App, ComposeResult
from textual.binding import Binding
from textual.widget import Widget
from textual.widgets import Button, Footer


class TestWidget(Widget, can_focus=True):
    BINDINGS = [
        Binding("b", "widget_binding", "Overridden Binding"),
    ]

    DEFAULT_CSS = """
    TestWidget {
        border: tall $background;
        width: 50%;
        height: 50%;
        content-align: center middle;

        &:focus {
            border: tall $accent;
        }
    }
    """

    def action_widget_binding(self) -> None:
        self.app.notify("Overridden Binding", severity="error")


class PriorityBindingApp(App):
    BINDINGS = [
        Binding("b", "app_binding", "Priority Binding", priority=True),
    ]

    CSS = """
    Screen {
        align: center middle;
    }
    """

    def compose(self) -> ComposeResult:
        yield TestWidget()
        yield Button("Move Focus")
        yield Footer()

    def action_app_binding(self) -> None:
        self.app.notify("Priority Binding")


if __name__ == "__main__":
    app = PriorityBindingApp()
    app.run()

```

PR #4798, opened 2024-07-24
improve simulate key

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/4794

The culprit was `App.simulate_key` which circumvented some logic re priority bindings.


willmcgugan:
#executivedecision

PR #4817, opened 2024-07-29
Fix sporadic NoMatches error reported from Header on app shutdown

slawek-es:
During application shutdown, particularly when running via `App.run_test()`, it seems that the Header's watcher callback on `title` or `sub_title` is called while the Header's children, such as the `HeaderTitle`, are already unmounted. This results in a `NoMatches` error being raised.

We fix this by ignoring the `NoMatches` error, like we did with `NoScreen` in the recent "hardedning" commit 604b04db2.

- [x] Updated documentation

### DIFF
--- a/src/textual/widgets/_header.py
+++ b/src/textual/widgets/_header.py
@@ -7,6 +7,7 @@ from datetime import datetime
 from rich.text import Text
 
 from ..app import RenderResult
+from ..css.query import NoMatches
 from ..dom import NoScreen
 from ..events import Click, Mount
 from ..reactive import Reactive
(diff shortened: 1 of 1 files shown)


mzebrak:
I tried to follow this issue of `NoMatches` in the header because we have a similar workaround for this problem in our application, I feel responsible for writing it and questioning the changes proposed here.

I may be wrong, but I think this MR is more like a workaround and will hide more significant problems with --probably-- internal task handling?

Please see: https://github.com/Textualize/textual/discussions/4268#discussioncomment-8744192 which reproduced this `NoMatches` error. 

Later it resulted in https://github.com/Textualize/textual/pull/4305 I tested it and I can still reproduce NoMatches in the header including these changes.

But such a `NoMatches` try-catch will hide that error forever when there might be something else  (bad) going on. In that case, most developers will probably need to also workaround similar issues with try-catch also, which is not so promising?

PR #4819, opened 2024-07-29
Open url

darrenburns:

- [ ] Updated documentation


src/textual/app.py
```diff
@@ -3679,3 +3679,12 @@ def action_suspend_process(self) -> None:
             # NOTE: There is no call to publish the resume signal here, this
             # will be handled by the driver posting a SignalResume event
             # (see the event handler on App._resume_signal) above.
+
+    def open_url(self, url: str) -> None:
```
willmcgugan, line 3683: I think we should add a `blank` boolean. If True it would open a tab, otherwise it would open in the same tab. Default should be True.

Locally, this attribute would be ignored.

src/textual/drivers/web_driver.py
```diff
@@ -192,13 +195,12 @@ def _on_meta(self, packet_type: str, payload: bytes) -> None:
             packet_type: Packet type (currently always "M")
             payload: Meta payload (JSON encoded as bytes).
         """
-        payload_map = json.loads(payload)
+        payload_map: dict[str, Any] = json.loads(payload)
```
willmcgugan, line 198: Would `object` be better than Any here?

willmcgugan (changes requested):
Great, one request.

src/textual/app.py
darrenburns: How about `new_tab`? `blank` is a bit of a strange name to me - I get that's what JavaScript calls it but I don't really understand why.

darrenburns: Also, not sure how this API would work when not running via web UI. There's no concept of "same" tab.

willmcgugan: IIRC the target attribute specified an iframe, but _blank was a special value to mean a new tab.

You're right though, we don't need to stick to that. +1 for new_tab

When running locally we can just ignore it. As long as its documented I think that's fine.

willmcgugan (approved):

PR #4820, opened 2024-07-30
Change StringKey to simple strings.  Fix ordering issue

merriam:

- [?] Updated documentation
- [n/a] Updated CHANGELOG.md (where appropriate)

DataTable's StringKey was recently exposed, meaning users could try to compare them.  Unfortunately, this was causing an a comparison between str and None:    `StringKey("a string") < StringKey()`.   To solve the problem and simplify the class, either a StringKey value is the supplied string or the string representation of its `id()`.

Test included, if you want to see it break before applying this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - `events.Enter` and `events.Leave` events now bubble. https://github.com/Textualize/textual/pull/4818
 - Renamed `Widget.mouse_over` to `Widget.mouse_hover` https://github.com/Textualize/textual/pull/4818
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


darrenburns:
Thanks @merriam. I've pushed a couple of minor updates. Reasoning is in the commit messages - let me know if the changes seem correct to you.

merriam:
Your changes lgtm.

Thank you for your edits.

kdkavanagh:
Will there be an issue in the event that an id is reused by python (due to a prior StringKey getting gc'd)?  Python makes no guarantees that object IDs are unique across the lifetime of the process, only unique relative to the other active objects in memory at any instant

merriam:
Good catch!  Python does not make the guarantee!

That said, I cannot see how it can manifest as a bug.   The id representation is only for empty StringKeys.

Also, I have given up on trying to get the PR accepted; feel free to resubmit.

Issue #4825, opened 2024-07-31
mutate_reactive doesn't work with data binding and recompose=False


Banbury:
Note that in the following example recompose is set to False on both reactive members.
Adding an item to the list in on_mount is working as expected, and `watch_msg` is called. But changing an element in `on_button_pressed`, doesn't call `watch_msg`. Apparently only changing the size of the list, is considered a change.

```python
from textual.app import App, ComposeResult
from textual.widgets import Static, Button
from textual.reactive import reactive
from textual.events import Mount

class TestWidget(Static):
    msg: reactive[list[str]] = reactive(list, recompose=False)
    count = reactive(0)

    def compute_count(self):
        return len(self.msg)

    def watch_msg(self, old, new):
        if new and len(new) > 0:
            self.mount(Static(self.msg[0]["text"]))

class TestApp(App):
    msg: reactive[list[str]] = reactive(list, recompose=False)

    def on_mount(self, event: Mount) -> None:
        self.msg.append( { "text": "Hello" } )
        self.mutate_reactive(TestApp.msg)

    def compose(self) -> ComposeResult:
        yield TestWidget().data_bind(TestApp.msg)
        yield Button("Click me", id="button")

    def on_button_pressed(self, event: Button.Pressed) -> None:
        self.msg[0] = {"text": "World!"}
        self.mutate_reactive(TestApp.msg)

if __name__ == "__main__":
    app = TestApp()
    app.run()```

PR #4828, opened 2024-08-01
mutate via data bind

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/4825

### DIFF
--- a/src/textual/dom.py
+++ b/src/textual/dom.py
@@ -41,7 +41,7 @@ from .css.parse import parse_declarations
 from .css.styles import RenderStyles, Styles
 from .css.tokenize import IDENTIFIER
 from .message_pump import MessagePump
-from .reactive import Reactive, ReactiveError, _watch
+from .reactive import Reactive, ReactiveError, _Mutated, _watch
 from .timer import Timer
 from .walk import walk_breadth_first, walk_depth_first
 from .worker_manager import WorkerManager
@@ -334,7 +334,7 @@ class DOMNode(MessagePump):
                     """Set bound data."""
                     _rich_traceback_omit = True
                     Reactive._initialize_object(self)
-                    setattr(self, variable_name, value)
+                    setattr(self, variable_name, _Mutated(value))
 
                 return setter
 
--- a/src/textual/reactive.py
+++ b/src/textual/reactive.py
@@ -42,6 +42,13 @@ ReactiveType = TypeVar("ReactiveType")
 ReactableType = TypeVar("ReactableType", bound="DOMNode")
 
 
+class _Mutated:
+    """A wrapper to indicate a value was mutated."""
+
+    def __init__(self, value: Any) -> None:
+        self.value = value
+
+
 class ReactiveError(Exception):
     """Base class for reactive errors."""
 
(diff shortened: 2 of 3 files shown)


src/textual/dom.py
```diff
@@ -334,7 +335,10 @@ def setter(value: object) -> None:
                     """Set bound data."""
                     _rich_traceback_omit = True
                     Reactive._initialize_object(self)
-                    setattr(self, variable_name, value)
+                    # Copy the bound value
```
darrenburns, line 338: I feel like this should be documented since copying can be expensive or potentially crash the program.

I'm guessing shallow copying means it also wouldn't fix the confusion for e.g. a list of mutable objects. For example, if one of the objects in the list is mutated and I call `mutate_reactive`, the prior confusing scenario still arises.

src/textual/reactive.py
```diff
@@ -42,6 +42,13 @@
 ReactableType = TypeVar("ReactableType", bound="DOMNode")
 
 
+class _Mutated:
+    """A wrapper to indicate a value was mutated."""
+
+    def __init__(self, value: Any) -> None:
```
darrenburns, line 48: We should probably update the `always_update` docstring in reactive since we're adding a case where it's now ignored.

src/textual/dom.py
willmcgugan: I'm reconsidering the copy.

It's not clear what the dev would expect when modifying a child attribute that has been bound. I'm not sure if they would expect the changes to be reflected on the parent. But I don't know if the would expect the inverse either.

Going to remove the copy, until I have a better idea what it would fix (if anything)

src/textual/reactive.py
willmcgugan: I don't think that's the case. This wrapper is only ever used in binding to force an update, so it wouldn't change the behavior of `always_update`...

willmcgugan: Not sure if we are on the same wavelength.

When you call `mutate_reactive`, you are basically saying the value has changed. So I would expect everything to behave as if it has. All the watchers should fire, including any attached via `data_bind`.

Other than that change, everything will behave as it always did.

darrenburns: > When you call mutate_reactive, you are basically saying the value has changed. So I would expect everything to behave as if it has. All the watchers should fire, including any attached via data_bind.

I'm talking about the case where you call `mutate_reactive` but the value _hasn't_ changed and `always_update=False`.

When I set `always_update=False` I expect that if the value of the reactive before and after the change is the same, then the watcher will _not_ fire. However, when calling `mutate_reactive` it sounds like that is being ignored and the watcher will fire.

If I have a reactive defined like this:

```python
active_users = reactive(["darren", "will"], always_update=False)
```

And some part of my code "mutates" the list:

```python
active_users[0] = "darren"
self.mutate_reactive(Foo.active_users)
```

The value before and the value after happen to be equal after the mutation. Since `always_update=False`, I would expect the watchers to _not_ fire because the new value equals the old value. This is implied by the `always_update` docstring:

```
always_update: Call watchers even when the new value equals the old value.
```

darrenburns (approved):

PR #4861, opened 2024-08-09
tweak key reporting

willmcgugan:
Some tweaks to how keys are reported (with the Kitty protocol).

- If "shift" is held with another modifier it is reported as "shift", rather than a capital letter. Previously if you press shift + ctrl + e, it is reported as `"ctrl+E"`. Now it is reported as `"ctrl+shft+e"`.

I think this makes sense, as a such a combination is unlikely to be a printable key

- Removed `caps_lock` and `num_lock` modifiers. This resulted in shift + key, and caps_lock + key being considered as different keys -- which I don't think makes sense. If this is ever a requirement, I think we should add booleans to the Key event.

### DIFF
--- a/src/textual/_xterm_parser.py
+++ b/src/textual/_xterm_parser.py
@@ -275,6 +275,7 @@ class XTermParser(Parser[Message]):
             key_tokens: list[str] = []
             if modifiers:
                 modifier_bits = int(modifiers) - 1
+                # Not convinced of the utility in reporting caps_lock and num_lock
                 MODIFIERS = (
                     "shift",
                     "alt",
@@ -282,12 +283,17 @@ class XTermParser(Parser[Message]):
                     "super",
                     "hyper",
                     "meta",
-                    "caps_lock",
-                    "num_lock",
+                    "",  # caps_lock
+                    "",  # num_lock
                 )
                 for bit, modifier in zip(range(8), MODIFIERS):
-                    if modifier_bits & (1 << bit):
+                    if modifier and modifier_bits & (1 << bit):
                         key_tokens.append(modifier)
+                if key_tokens:
+                    if key.isupper():
+                        key_tokens.append("shift")
+                        key = key.lower()
+
             key_tokens.sort()
             key_tokens.append(key)
             yield events.Key(


darrenburns:
Will the break existing apps with bindings for `ctrl+E` for example? Will those apps need to update their `Binding`s to `ctrl+shift+e`?

willmcgugan:
It would. I've added a note in the changelog.

darrenburns:
Just had a thought about this - isn't it a bit strange to have this reported like this but not have "E" be reported as "shift+e"? I'm not sure which is better, but I'd definitely expect it to be consistent.

willmcgugan:
> Just had a thought about this - isn't it a bit strange to have this reported like this but not have "E" be reported as "shift+e"? I'm not sure which is better, but I'd definitely expect it to be consistent.

There isn't a simple one to one mapping of shift + letter. It depends on your keyboard and locale. e.g. shift+3 will give a `£` in GB layout or a `#` in US layout. I think that's also true of letter keys with some international layouts. There may also be multiple ways of producing a key. For instance, if capslock is on, you can produce capital letters without holding shift.

I think its safer to trust that when the keyboard produces a key, that's what the user intended, and we don't want to make assumptions about how it was produced.

willmcgugan:
Actually, I think I just talked myself out of part of this update. When the keyboard reports an alphanumeric letter, Kitty does not report that the shift modifier was pressed.

It __does__ report the shift modifier for non alphanumeric keys. i.e. `shift+space` reports that the shift modifier was held.

So I don't think Textual should be trying to deduce that shift was pressed at all. So `ctrl+shift+E` should be reported as `ctrl+E`.

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Added `tooltip` to Binding https://github.com/Textualize/textual/pull/4859
 
+### Changed
+
+- Change to how keys with modifiers are reported. When shift + another modifier is pressed, then the key name will contain "shift", and the key will be *lower case*. i.e. `ctrl+shift+E` and not `ctrl+E` https://github.com/Textualize/textual/pull/4861
```
darrenburns, line 16: Should it be this?

```suggestion
- Change to how keys with modifiers are reported. When shift + another modifier is pressed, then the key name will contain "shift", and the key will be *lower case*. i.e. `ctrl+shift+e` and not `ctrl+E` https://github.com/Textualize/textual/pull/4861 
```

darrenburns, line 16: Does this need updating to match the latest state of the PR?

darrenburns (approved):

PR #4867, opened 2024-08-11
snapshots

willmcgugan:
Displays a keybinding to the command palette on the right of the Footer, and adds an option to the Footer to disable it.l

Screenshot 2024-08-11 at 17 01 31


TomJGooding:
Sorry but this is a bit ugly and I dont' understand what problem it is trying to resolve?

willmcgugan:
Tweaked the CP link to look like the others.

darrenburns:
Just testing this out - when the terminal is small it makes the Footer look pretty confusing. It's not clear that the rest of the footer is going "under" the command palette option:

image

Although IMO enabling this by default feels a bit developer-unfriendly. It almost feels like an ad. From the perspective of someone spinning up a new Textual app, I think I would always disable this since by default the command palette doesn't have many useful builtin commands and so I wouldn't want to advertise it to end users. As an app developer, I would want to enable this *after* I add my own useful commands, rather than having to switch it off at first and then re-enable it.

willmcgugan:
Do you have any suggestions for an alternative way to advertise to the user that there is a command palette?

darrenburns:
If you're concerned that not many apps are using it, it could mean that it's too hard to work with and the API needs work rather than people are not aware of it. To be honest, it's quite a substantial a bit of boilerplate to make it useful at the moment. Maybe that's the issue? I don't know 🤷 

If you mean advertise to the end user of an app, then the developer can put the binding in the footer if they want. If you mean advertise to developers, then the same way as we advertise any other feature - talk about it in the docs/use it in the tutorial/tweet about it. I think this is a broader disagreement though (I think the palette should actually be disabled by default), so I'm happy to leave it here and let you decide the path.

willmcgugan:
Going to merge. Always happy to review once we get more feed back.

darrenburns:
I think you've missed my concern, which was only around having this visible by default.

> Developers can be educated by making it more prominent in the docs

But that's not what we're doing here - we're putting it in the footer of every Textual app by for the sole purpose of advertising just because we don't _think_ people know about it. That's _so_ unfriendly to developers IMHO.

Another thing to consider is that a huge number of Textual apps are actually just really small apps that might not warrant a command palette. Even if they know it exists they might not want it. For example, in Elia, I knew about it, but I chose not to use it because there wasn't enough functionality I could put into it.

> but users running a TUI can easy miss it even if it is in the apps help

They won't miss it if it's in the footer though, which the developer (educated via docs) could add if they _choose_ to show it to end users.

---

Since we're going with this - could we at least remove the "ring the terminal bell" command? 😂

willmcgugan:
My point re the docs was that we can educate developers, but educating users is harder. We must be failing hard somewhere because I don't know of any app out there that makes use of the command palette. 

Let's see how this is received. I suspect folk will leave it in, especially once we have a few more default commands.

Yeah, we can remove the bell command!

Issue #4870, opened 2024-08-12
v0.76.0 breaks buttons with emojis


charles-001:
I upgraded to v0.76.0 and the button layout I use is broken now. It adds a bar to the right of some of them.

<img width="516" alt="Screenshot 2024-08-12 at 4 30 58 AM" src="https://github.com/user-attachments/assets/b2e24364-58bc-428a-84fd-f9a7dac371cf">

MRE:
```python
from textual.app import App, ComposeResult
from textual.containers import Horizontal
from textual.widgets import Button


class MREApp(App):
    CSS = """
        Button {
            text-style: none;
        }
    """

    def __init__(self):
        super().__init__()

    def compose(self) -> ComposeResult:
        with Horizontal():
            yield Button("⏪  Back", id="back_button")
            yield Button("⏸️  Pause", id="pause_button")
            yield Button("⏩  Forward", id="forward_button")
            yield Button("🔍  Seek", id="seek_button")


if __name__ == "__main__":
    app = MREApp()
    app.run()
````

Thanks!

darrenburns:
Are you 100% sure this is due to upgrading to `v0.76.0`?

charles-001:
Yes. I can downgrade to `v0.75.0` and it's back to normal.

darrenburns:
Thanks! Could you run `textual diagnose` and paste the output here?

darrenburns:
Actually scrap that - I can reproduce! No need.

PR #4873, opened 2024-08-12
disable key disambiguate

willmcgugan:
This seems to break output on some terminals.

Going to disable for now, and come back to it later.

https://github.com/Textualize/textual/issues/4870


paul-ollis:
I believe that the 'key disambiguation' (the one commented out by this change) sequence is incorrect - missing an equals '=' character.

The code has: "\x1b[1;u"

It should be:   "\x1b[=1;1u"  ("\x1b[1;u" might be OK. I am not sure based on reading https://sw.kovidgoyal.net/kitty/keyboard-protocol/#progressive-enhancement. So I went with "explicit is better than implicit".)

willmcgugan:
You're absolutely right. I think my brain omitted the `=` as though it was punctuation!

Issue #4856, opened 2024-08-09
Command palette cancel


willmcgugan:
If you bring up the command palette (with `ctrl+\`) you have to press ESC twice to cancel it.

This makes sense when there is an auto-complete menu showing. But I think that a single ESC should dismiss the command palette in the following circumstances:

1. There is nothing in the input
2. There are no items in the auto-complete

MarcosLima98:
Could you please assign this issue to me? I want to work on it.

willmcgugan:
Sorry, we have a PR for this one. Will add more GFIs soon.

PR #4874, opened 2024-08-12
command palette dismiss logic

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/4856

Screwed up the diff a bit here. Most of the updated snapshots are from a previous PR.


src/textual/command.py
```diff
@@ -735,6 +737,8 @@ def _show_no_matches() -> None:
                 # discovered is a situation we don't need to highlight.
                 self._list_visible = False
 
+        self._hit_count = 0
```
darrenburns, line 740: Minor but `_start_no_matches_countdown` seems like a strange place for this (seems out of scope). I'd put it inside `gather_commands` where we check `if option_list.options_count == 0`. Take it or leave it 😄

darrenburns (approved):
One minor comment but looks good.

PR #4875, opened 2024-08-12
Deliver file

darrenburns:

- [ ] Updated documentation


pyproject.toml
```diff
@@ -47,6 +47,9 @@ rich = ">=13.3.3"
 typing-extensions = "^4.4.0"
 tree-sitter = { version = "^0.20.1", optional = true }
 tree-sitter-languages = { version = "1.10.2", optional = true }
+platformdirs = "^4.2.2"
+msgpack = "^1.0.8"
+msgpack-types = "^0.3.0"
```
darrenburns, line 52: 3 new dependencies for a relatively niche feature 😅

darrenburns:
@willmcgugan Feel free to review. Would appreciate 👀s on this.

src/textual/app.py
```diff
@@ -3693,3 +3691,139 @@ def open_url(self, url: str, *, new_tab: bool = True) -> None:
         """
         if self._driver is not None:
             self._driver.open_url(url, new_tab)
+
+    def deliver_text(
```
willmcgugan, line 3695: Suggest an explicit `name` attribute for the download.

src/textual/app.py
```diff
@@ -3693,3 +3691,139 @@ def open_url(self, url: str, *, new_tab: bool = True) -> None:
...
+
+        If running in a terminal, this will save the file to the user's
+        downloads directory.
+
+        If running via a web browser, this will initiate a download via
+        a single-use URL.
+
+        Args:
+            path_or_file: The path or file-like object to save.
+            save_location: The directory to save the file to. If path_or_file
```
willmcgugan, line 3717: Should be `save_directory`?

src/textual/drivers/web_driver.py
```diff
@@ -223,6 +241,51 @@ def on_meta(self, packet_type: str, payload: dict[str, object]) -> None:
...
+            except KeyError:
+                log.error("Protocol error: deliver_chunk_request missing key or size")
+                return
+
+            deliveries = self._deliveries
+            deliveries_lock = self._deliveries_lock
+
+            binary_io: BinaryIO | None = None
+            try:
+                with deliveries_lock:
```
willmcgugan, line 259: Think we can live without the lock since we aren't running in a worker noe.

src/textual/drivers/web_driver.py
```diff
@@ -223,6 +241,51 @@ def on_meta(self, packet_type: str, payload: dict[str, object]) -> None:
...
+                    chunk = binary_io.read(requested_size)
+                    log.debug(f"Delivering chunk {delivery_key!r} of len {len(chunk)}")
+                    self.write_packed(("deliver_chunk", delivery_key, chunk))
+                    if not chunk:
+                        log.info(f"Delivery complete for {delivery_key}")
+                        binary_io.close()
+                        with deliveries_lock:
+                            del deliveries[delivery_key]
+                except Exception:
+                    binary_io.close()
```
willmcgugan, line 278: This is repeated above. Maybe put this in a `finally:` block ?

willmcgugan (commented):
Looks good.

src/textual/drivers/web_driver.py
darrenburns: I only want to close it if there's an exception, not always.

src/textual/driver.py
```diff
@@ -196,3 +185,73 @@ def no_automatic_restart(self) -> Iterator[None]:
...
+            mode = "wb"
+        else:
+            mode = "w"
+
+        def save_file_thread():
+            with open(save_path, mode) as destination_file:
+                read = binary.read
+                write = destination_file.write
+                while True:
+                    data = read(1024)
```
willmcgugan, line 243: This can probably be much larger. 64K would be a good starting point.

Could we use `shutil.copyfileobj` here? I imagine that is well tuned.

willmcgugan (changes requested):
Looks good. Just small potatoes.

src/textual/driver.py
darrenburns: I was getting typing errors with that method, so I just copied the implementation of it. I'll bump it to 64

src/textual/driver.py
```diff
@@ -196,3 +185,73 @@ def no_automatic_restart(self) -> Iterator[None]:
...
+            mime_type: *web only* The MIME type of the file. This will be used to
+                set the `Content-Type` header in the HTTP response.
+        """
+        if isinstance(binary, BinaryIO):
+            mode = "wb"
+        else:
+            mode = "w"
+
+        def save_file_thread():
+            with open(save_path, mode) as destination_file:
```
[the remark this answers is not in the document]
darrenburns: I'm catching exceptions and passing them to the App inside a message now, as well as writing an `ERROR` log via Textual's logging system.

willmcgugan:
Just to say, we should hold off merging this until everything is in place.

willmcgugan (approved):

JaceyPenny:
FYI @darrenburns, just wanted to point out that the version `platformdirs = ^4.2.2` breaks our poetry build -- I believe since you only need the `user_downloads_path`, you could loosen the version restriction to `>=3.6.0` ([this commit](https://github.com/platformdirs/platformdirs/commit/85b0b67eb33a835b797c1d52dab4a06c5554b7ee) on the `platformdirs` repo is tagged all the way down to 3.6.0). I can handle this change in another PR if you like.

willmcgugan:
PR would be welcome.

PR #4876, opened 2024-08-12
Key panel widget

willmcgugan:
Adds a `KeyPanel` which can be summoned via the command palette.


src/textual/widgets/_key_panel.py
```diff
@@ -0,0 +1,137 @@
...
+    BindingsTable {
+        width: auto;
+        height: auto;
+    }
+    """
+
+    upper_case_keys = reactive(False)
+    """Upper case key display."""
+    ctrl_to_caret = reactive(True)
+    """Convert 'ctrl+' prefix to '^'."""
```
darrenburns, line 34: Just thinking out loud for potential future updates: I'd love if this was:

```python
ctrl_repr = reactive("^")
and_repr = reactive("")
```

Then I could choose the styling easily. e.g. if I wanted emacs style bindings:

```python
foo.ctrl_repr = 'c'
foo.and_repr = '-'
```

Which would render `ctrl+x` as `c-x`.

darrenburns: However, I think this would be defined at the app level - so we don't need to define this for both our Footer and for the Bindings table.

src/textual/command.py
```diff
@@ -1068,10 +1067,15 @@ def _select_or_command(
         if event is not None:
             event.stop()
         if self._list_visible:
+            command_list = self.query_one(CommandList)
             # ...so if nothing in the list is highlighted yet...
-            if self.query_one(CommandList).highlighted is None:
+            if command_list.highlighted is None:
                 # ...cause the first completion to be highlighted.
                 self._action_cursor_down()
+                # If there is one option, assume the user wants to select it
```
darrenburns, line 1075: Isn't this subject to the same issue as what we discussed earlier regarding having the first option always highlighted by default? If they can come in asynchronously, then when the user presses enter there may only be 1, but this doesn't mean there will always be one.

I think in VSCode and the browser URL bar, it also does things asynchronously, but it just selects the one that was at the top when the user presses enter.

src/textual/css/_style_properties.py
```diff
@@ -599,6 +599,37 @@ def __set__(self, obj: StylesBase, dock_name: str | None):
...
+    def __set__(self, obj: StylesBase, dock_name: str | None):
+        """Set the Dock property.
+
+        Args:
+            obj: The ``Styles`` object.
+            dock_name: The name of the dock to attach this widget to.
+        """
+        _rich_traceback_omit = True
+        if obj.set_rule("split", dock_name):
+            obj.refresh(layout=True)
```
darrenburns, line 630: I think this is so similar to dock that the dock rule should perhaps be updated to take a second (optional) argument which does this behaviour: `dock: right split;` or something like that.

darrenburns:
Also want to suggest that we have some way of showing which "level" of the DOM keybindings belong to. This sidebar looks a bit confusing in Posting for example, because I can't tell which widget a binding belongs to.

e.g. I have a binding called `Expand section`. If I have the `Tree` focused, it's not clear if that means "expand the current node of the tree", or "expand the tree section of the overall app". I think if they were somehow grouped with the widget `name` (if available) or something it'd be easier to read in large apps with many bindings (where the key panel is itself most beneficial).

darrenburns:
Another thing that I just noticed in the `lazygit` popup is that you can click the bindings and it runs the action, similar to our `Footer` widget.

src/textual/command.py
willmcgugan: That does introduce a 100ms window where something could change. I'll see if I can remove that.

willmcgugan: I've removed the timer for now. In a future update, I'll implement the logic to highlight the first item.

willmcgugan: https://github.com/Textualize/textual/issues/4902

src/textual/css/_style_properties.py
willmcgugan: https://github.com/Textualize/textual/issues/4903

PR #4878, opened 2024-08-13
Remove logging

darrenburns:


src/textual/screen.py
```diff
@@ -284,7 +284,6 @@ def _watch_stack_updates(self):
 
     def refresh_bindings(self) -> None:
         """Call to request a refresh of bindings."""
-        self.log.debug("Bindings updated")
```
darrenburns, line 287: This log is super noisy and low value, and as far as I know there is no way of switching it off without also disabling debug logs in my own code.

src/textual/widgets/_footer.py
```diff
@@ -232,7 +232,6 @@ def compose(self) -> ComposeResult:
                 Footer.ctrl_to_caret,
                 Footer.compact,
             )
-        self.log(bindings)
```
darrenburns, line 235: Think this was left accidentally from a recent update. It prints a LOT to the console 😄

willmcgugan (approved):

Issue #4879, opened 2024-08-13
Investigate flaky `test_command_palette_dismiss_escape_no_results` snapshot test


TomJGooding:
The `test_command_palette_dismiss_escape_no_results` snapshot test seems to be a bit flaky in CI:

https://github.com/Textualize/textual/actions/runs/10354774951/job/28660835280
https://github.com/Textualize/textual/actions/runs/10368429707/job/28702009036
https://github.com/Textualize/textual/actions/runs/10369324019/job/28704826672
https://github.com/Textualize/textual/actions/runs/10372095452/job/28713984227

TomJGooding:
Hopefully fixed in #4880 but time will tell!

PR #4880, opened 2024-08-13
tests: fix flaky command palette snapshot test

TomJGooding:
Investigating possible fix for #4879.

Maybe I've just been lucky but managed to go three-for-three!


willmcgugan:
Thanks Tom!

PR #4886, opened 2024-08-15
chore(issue template): clarify textual command requires devtools

TomJGooding:
The bug report issue template asks contributors to run the `textual diagnose` command. Unfortunately most bug reports don't provide this info and I've seen a few comments that this command doesn't work.

Clarifying that this command requires the devtools installed hopefully will help!


darrenburns (approved):
Thanks!

Issue #4887, opened 2024-08-15
`COMMAND_PALETTE_BINDING` doesn't change the binding


TomJGooding:
Sorry perhaps I am missing something or this change is still unfinished, but setting the `COMMAND_PALETTE_BINDING` added in #4867 doesn't seem to change the command palette binding?

```python
from textual.app import App, ComposeResult
from textual.widgets import Footer


class NewPaletteBindingApp(App):
    COMMAND_PALETTE_BINDING = "ctrl+backslash"

    def compose(self) -> ComposeResult:
        yield Footer()


if __name__ == "__main__":
    app = NewPaletteBindingApp()
    app.run()
```

willmcgugan:
In reterosepect, they way I did it would never have worked.

Issue #4857, opened 2024-08-09
Change command palette key


willmcgugan:
`ctrl+\` is awkward on some keyboards. Maybe it should just be `ctrl+p`?

phershbe:
I found this on a website that aggregates issues tagged as `good first issue` and would like to work on it if you are sure that you want to make the change.

Am I correct in assuming that it is basically as simple as changing the binding in `app.py` and then updating the corresponding documentation and tests?

TomJGooding:
I assumed that <kbd>Ctrl</kbd>+<kbd>\\</kbd> was chosen because this is unlikely to clash with other bindings?

If this is binding is going to change  to something like <kbd>Ctrl</kbd>+<kbd>p</kbd>, I'd suggest adding something to the FAQ as I suspect this will create some issues!

willmcgugan:
> Am I correct in assuming that it is basically as simple as changing the binding in `app.py` and then updating the corresponding documentation and tests?

It is likely as simple as that, yes!

willmcgugan:
> I assumed that Ctrl+\ was chosen because this is unlikely to clash with other bindings?
> 
> If the command palette binding is going to change to something like Ctrl+p, I'd suggest adding something to the FAQ as I suspect this will create some issues!

Pretty much, but on some keyboards that combo is tricky. ctrl+p has a higher likelyhood of clashing, but it is often used for a command palette, and muscle memory is hard to change.

A FAQ entry may be a good idea. As would a note how to change it in the docs.

TomJGooding:
It looks like the command palette binding was changed in #4867.

This change and the new `COMMAND_PALETTE_BINDING` variable probably needs adding to the CHANGELOG?

PR #4890, opened 2024-08-17
fix command palette key

willmcgugan:
Started out as a fix for the command palette key binding. Turned in to a refactor and an alternative centralised way of defining how keys are displayed in footer and key panel.

Fixes https://github.com/Textualize/textual/issues/4887
Fixes https://github.com/Textualize/textual/issues/4901

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,17 +16,20 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Added "Show keys" option to system commands to show a summary of key bindings. https://github.com/Textualize/textual/pull/4876
 - Added "split" CSS style, currently undocumented, and may change. https://github.com/Textualize/textual/pull/4876
 - Added `Region.get_spacing_between` https://github.com/Textualize/textual/pull/4876
+- Added `App.COMMAND_PALETTE_KEY` to change default command palette key binding
+- Added `App.get_key_display`
 
 ### Changed
 
 - Removed caps_lock and num_lock modifiers https://github.com/Textualize/textual/pull/4861 
 - Keys such as escape and space are now displayed in lower case in footer https://github.com/Textualize/textual/pull/4876
+- Changed default command palette binding to ctrl+p
+- Removed `ctrl_to_caret` and `upper_case_keys` from Footer. These can be implemented in `App.get_key_display`.
 
 ### Fixed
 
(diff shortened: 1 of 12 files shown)


TomJGooding (commented):
Thanks Will for taking another look at this!

I don't think this needs a 'fixed' entry in the CHANGELOG, as the command palette binding change wasn't yet released?

However as mentioned in https://github.com/Textualize/textual/issues/4857#issuecomment-2286803691, there's currently nothing in the CHANGELOG for this binding change. which might be breaking change if apps have their own bindings for <kbd>Ctrl</kbd>+<kbd>p</kbd>?

tests/snapshot_tests/snapshot_apps/command_palette_key.py
```diff
@@ -0,0 +1,15 @@
+from textual.app import App, ComposeResult
+from textual.widgets import Footer
+
+
+class NewPaletteBindingApp(App):
+    COMMAND_PALETTE_BINDING = "ctrl+backslash"
+    COMMAND_PALETTE_DISPLAY = "ctrl+\\"
```
TomJGooding, line 7: Perhaps I've misunderstood this new `COMMAND_PALETTE_DISPLAY` variable , but why does the snapshot show `^/` rather than `ctrl+\`?

willmcgugan: There is some work in https://github.com/Textualize/textual/pull/4876 that will fix that.

TomJGooding: It might be worth adding that work in this PR if possible? This might help debugging later, as really the snapshot is not as expected.

CHANGELOG.md
```diff
@@ -16,17 +16,20 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Added "Show keys" option to system commands to show a summary of key bindings. https://github.com/Textualize/textual/pull/4876
 - Added "split" CSS style, currently undocumented, and may change. https://github.com/Textualize/textual/pull/4876
 - Added `Region.get_spacing_between` https://github.com/Textualize/textual/pull/4876
+- Added `App.COMMAND_PALETTE_KEY` to change default command palette key binding
```
darrenburns, line 19: Remember to link this PR

darrenburns (approved):

PR #4894, opened 2024-08-18
Disallow `Screen` instances in `App.SCREENS`

### LINKED ISSUE #4893: Don't allow screen instances in App.SCREENS
Textual currently allows a screen instance to be specified in the `SCREENS` classvar. This works fine when launching an app, until a second instance of the app runs. Re-using a previous instance breaks the app.

See the example here: https://textual.textualize.io/guide/screens/#creating-a-screen

Note how the BSOD is given as an instance in a classvar, i.e. `SCREENS = {"bsod": BSOD()}` This would cause the instance to be shared if there is ever more than a single App.

The easiest workaround is to disallow passing instances, and require a string or callable. So the above example should be `SCREENS = {"bsod": BSOD}`

Requirements:

- Adjust typing to not allow a screen instance in `SCREENS` and `MODES`
- Reject screen instances at runtime, by throwing a `TypeError`.
- Update docs and examples accordingly.
- Tests to ensure behavior when not using a string or callable.

ZeroIntensity:
cc @willmcgugan 
Closes #4893.

It's worth noting that while implementing this, I discovered a second issue: `add_mode` appends to the `MODES` classvar, meaning it applies to all instances of the `App` -- if this is the desired behavior, then it should probably be made a classmethod. Otherwise, maybe an attribute to hold per-instance modes?


src/textual/app.py
```diff
@@ -562,7 +562,20 @@ def __init__(
         """Number of lines to scroll in the Y direction with wheel or trackpad."""
 
         self._installed_screens: dict[str, Screen | Callable[[], Screen]] = {}
+        for var_name, dct in (("SCREENS", self.SCREENS), ("MODES", self.MODES)):
+            for screen_name, v in dct.items():
+                if (
```
willmcgugan, line 567: I think we can simplify a little. The only legal types are a `str` or anything callable.

Maybe this?

```python
if not (isinstance(screen_name, str) or callable(screen_name)):
```

src/textual/app.py
```diff
@@ -562,7 +562,20 @@ def __init__(
...
+                    isinstance(v, Screen)
```
willmcgugan, line 568: I think we can simplify a little. The only legal types are a `str` or anything callable.

Maybe this?

```python
if not (isinstance(screen_name, str) or callable(screen_name)):
```

src/textual/app.py
```diff
@@ -562,7 +562,20 @@ def __init__(
         """Number of lines to scroll in the Y direction with wheel or trackpad."""
 
         self._installed_screens: dict[str, Screen | Callable[[], Screen]] = {}
+        for var_name, dct in (("SCREENS", self.SCREENS), ("MODES", self.MODES)):
```
willmcgugan, line 565: We have "no abbreviates" rule in Textual. So `var_name` and `dct` would have to be something descriptive.

The constructor may not be the best place for this. As you are checking classvars, it seems redundant to do this work every time the App is instantiated.

Suggest `__init_subclass__` is the right place for this check. That way it would run just once, at import time...

src/textual/app.py
```diff
@@ -562,7 +562,20 @@ def __init__(
...
+                    isinstance(v, Screen)
+                    or (not callable(v))
+                    and (not isinstance(v, str))
+                ):
+                    raise TypeError(
+                        f"{var_name} should contain a Screen type or callable, not an instance"
+                        f" (got instance of {type(v).__name__} for {screen_name!r})"
+                    )
         self._installed_screens.update(**self.SCREENS)
+        self.modes: dict[str, str | Callable[[], Screen]] = self.MODES.copy()
```
willmcgugan, line 577: I think `modes` should be private. It probably isn't of much use for the dev, so `self._modes` is probably better.

If we want it to be public later, we could add a property.

willmcgugan (changes requested):
Added some requests. I think the documentation may need an update. There are a few examples which use the old Screen instances in the classvar. There may also be some copy change required...

src/textual/app.py
ZeroIntensity: Ah, there was a test that accessed `MODES` to check changes made from `add_mode`, so I figured it was documented behavior. Will move it `_modes`.

src/textual/app.py
ZeroIntensity: I'm going to turn this into two checks: one that raises a `TypeError` for something wrong, and possibly a `ValueError` if it's an *instance* of `Screen` that has a nicer error message (rather than `x is the wrong type`) -- I don't want you guys to get bug reports complaining that instances no longer work!

ZeroIntensity:
Should be all done now, thanks!

ZeroIntensity:
I think all the places where an instance was used in `MODES` or `SCREENS` in the docs have been fixed.

src/textual/app.py
```diff
@@ -2833,8 +2867,10 @@ async def _close_all(self) -> None:
 
         # Close pre-defined screens.
         for screen in self.SCREENS.values():
-            if isinstance(screen, Screen) and screen._running:
-                await self._prune(screen)
+            # This is now dead code, since instances cannot be in SCREENS.
```
willmcgugan, line 2870: I think we can remove this loop entirely, since we prevent instances going in to SCREENS.

src/textual/app.py
```diff
@@ -650,6 +652,30 @@ def __init__(
         # Size of previous inline update
         self._previous_inline_height: int | None = None
 
+    def __init_subclass__(cls, *args, **kwargs) -> None:
+        for variable_name, screen_collection in (
+            ("SCREENS", cls.SCREENS),
+            ("MODES", cls.MODES),
+        ):
+            for screen_name, screen_object in screen_collection.items():
+                if not isinstance(screen_name, str):
```
willmcgugan, line 661: I think we can do without this check. It's not likely the dev would make this error. If we checked all typing at runtime, then the codebase would be full of such checks.

Kudos for thinking of it tthough!

willmcgugan (changes requested):
Couple of requests then we should be good to go...

willmcgugan (approved):
LGTM Thanks!

ZeroIntensity:
Sorry; I'll fix the merge conflicts sometime later today.

ZeroIntensity:
Merge conflicts have been fixed!

willmcgugan:
Are you really 16?!

I figured you were already a pro.

ZeroIntensity:
Yup! I've been doing this since I was ~9.

PR #4905, opened 2024-08-20
Added waring when message handling take long time

### LINKED ISSUE #4895: Log slow message processing
If processing messages take longer than a fraction of a second, it can cause the UI to freeze.

This is usually the result blocking operations or cpu intensive work in a message handler. The solution is typically to do the work in a worker, but devs may not know to do that.

Given its a gotcha, I think Textual can do a little more to help.

I think what we should do is time how long a message has taken to process. If it is longer than, say 500ms, it should log a warning.

This should only be enabled in debug mode. The maximum time should be set in an environment variable.

I suspect, the best place to do this would be https://github.com/Textualize/textual/blob/main/src/textual/message_pump.py#L637

## Requirements:

- Display warning for message handers that take a long time
- Enable in debug mode
- Add env var to set the time that produces a warning
- Should not add to much cpu work to message processing

## Bonus points

Strictly following the above, would not produce a warning until the handler returns. If it takes 30 seconds, the dev would not see anything.

It would be great if long running message handlers produce a warning even before the handler returns. This would be a little trickier to implement. And it should be done in a way that doesn't slow down non-debug runs.

Zimzozaur:
### This PR resolves #4895 
cc @willmcgugan

**Implemented Features:**

- Warning for Slow Handlers: Logs a warning if a message handler takes longer than 500ms to process.
- Debug Mode Only: Warnings are enabled only in debug mode to avoid unnecessary overhead in production.
- Configurable Threshold: The time threshold before a warning is issued can be set via the `TEXTUAL_SLOW_THRESHOLD_MS` environment variable.

**Known Limitations:**

Currently, the warning is logged only after the message handler completes. This means that if a handler takes an extended time (e.g., 30 seconds), the warning won’t appear until it finishes.
I have attempted to implement a solution that handles this case in real-time but haven't been able to get it working as intended yet.

**Request for Feedback:**

If the current implementation is satisfactory, I would appreciate feedback on how to improve the async version for real-time warnings.


src/textual/message_pump.py
```diff
@@ -655,6 +658,16 @@ async def _dispatch_message(self, message: Message) -> None:
             # Allow apps to treat events and messages separately
             if isinstance(message, Event):
                 await self.on_event(message)
+            elif "debug" in self.app.features:
+                start = perf_counter()
+                await self._on_message(message)
+                if perf_counter() - start > SLOW_THRESHOLD_MS / 1000:
+                    log.warning.verbosity(message.verbose)(
```
willmcgugan, line 665: This is only in debug mode, so I don't think we need to worry about verbosity.

src/textual/message_pump.py
```diff
@@ -655,6 +658,16 @@ async def _dispatch_message(self, message: Message) -> None:
...
+                        f"method=<{self.__class__.__name__}.{message.handler_name}>",
+                        ">>>",
```
willmcgugan, line 667: I don't think we need the ">>>", that was just to indicate a message being sent.

src/textual/message_pump.py
```diff
@@ -655,6 +658,16 @@ async def _dispatch_message(self, message: Message) -> None:
...
+                        f"Took over {SLOW_THRESHOLD_MS}ms to process.",
+                        "\nTo avoid screen freezes use the [work decorator](/guide/workers/#workers)"
```
willmcgugan, line 669: This goes to the textual console, which doesn't support Markdown syntax. Suggest simply "Consider using a worker."

willmcgugan (changes requested):
A few requests...

Could you add a test?

Put a pause in the message handler, and check the message is written as expected.

Zimzozaur:
I’ve updated the code based on your suggestions. However, I’m still unsure how to test the logging. I’ve reviewed the documentation and other tests but could use some guidance.

darrenburns:
Hey @Zimzozaur, could you run `make format` to format the code? If you want to do this automatically you can run `pre-commit install` and it'll run everything you need (not including the tests) automatically when you `git commit`.

Zimzozaur:
Hey @darrenburns after few tries and some learning, I ran it. Now Black is happy 🙂

darrenburns:
Thanks @Zimzozaur!

To be honest this is very difficult to test, but here's some guidance:

- You'll probably want to reduce the slow threshold env var value for the test - `monkeypatch.setenv("TEXTUAL_SLOW_THRESHOLD", "50")`. Search the codebase for `monkeypatch.setenv` for examples.
- You'll need to create a subclass of `App` - there are many examples of this in tests.
- Inside `App.on_mount`, you'll need to post a message e.g. `self.post_message(MySlowMessage)`.
- Inside the handler for that message (an `async` method `App.on_my_slow_message`), add an `asyncio.sleep` call. This will allow us to simulate what happens when a handler is slow.
- At this point it becomes a bit tricky as we need to swap out Textual's logger for our own one which we can query to inspect what gets logged. Looking at the code here - I think `message_pump.py` probably needs a small change to make this easier, but for now you could set the `_logger` attribute on your `App` to a `Mock` object, the check the values `call_args_list` to ensure that the "warning" log was called with the expected string.

Zimzozaur:
@willmcgugan, @darrenburns I didn't respond because I got sick. I do not mind if you add the tests. For the first good issue, this part is quite hard. 

When you merge changes, can I write/try to implement async from the bonus points section?

willmcgugan:
No worries. Hope you are feeling better. Feel free to tackle the bonus points!

PR #4907, opened 2024-08-20
refactor(switch): make slider_pos private

TomJGooding:
Spotted that the internal slider position for the `Switch` is not private, which also means it appears in the docs. This renames to `_slider_position` per Textual's no abbreviation policy.


willmcgugan (approved):
I feel seen! Thanks.

PR #4915, opened 2024-08-21
context sensitive help

willmcgugan:
- Adds widget help to the help panel (previously the "key panel").
  DOMNodes now have a `HELP` classvar which may be set to Markdown that renders in the help panel.
- Removes some jankiness when summoning the command palette.

Screenshot 2024-08-21 at 13 33 27


darrenburns:
Should the sidebar have a title of "Help", or the name of the widget that the help is being displayed for? Or customisable? Or nothing? 😄

darrenburns:
This looks good to me but I think I'd want a title personally (which could always be hidden with `display: none`) and also some way of controlling it with the keyboard.

willmcgugan:
The title of the widget is shown above the top group of keys.

I don't think it looks good with a title. Seems redundant, and it takes up 2 lines.

Agree about keyboard control, but I'll do that in another pass.

Screenshot 2024-08-22 at 10 42 33

willmcgugan:
How about this?

https://github.com/user-attachments/assets/9003594e-ca17-4f9e-99fc-3a04a41f1f13

darrenburns:
I agree it doesn't look great there! Maybe you're right 😄

willmcgugan:
How about I hide it by default. And if a dev asks, they can unhide it?

darrenburns:
Seems like a reasonable compromise.

darrenburns:
Could you add a snapshot test? After that it LGTM.

darrenburns (approved):
LGTM, thanks!

Issue #4914, opened 2024-08-21
ScrollView/RichLog: Page Up / Page Down off by one with horizontal scrollbar


xavierog:
I noticed this issue while using the RichLog widget but it may actually stem from ScrollView.

Assuming a 25-line high RichLog widget:
- When there is a horizontal scrollbar, Page Up and Page Down move e.g. from lines 1-25 to lines 26-50, which is the expected behaviour.
  This can be confirmed by running the MRE below this way: `SUFFIX= ./richlogtest.py`
- When there is a horizontal scrollbar, Page Up and Page Down move e.g. from lines 1-24 to lines 26-49, thus hiding line 25, hence this report.
  This can be confirmed by running the MRE below this way: `./richlogtest.py`

MRE:
```python
#!/usr/bin/env python3
import os
from textual.app import App
from textual.widgets import RichLog

class RichLogTest(App):
	def compose(self):
		ri = RichLog(auto_scroll=False)
		suffix = os.environ.get('SUFFIX', 'A'*1000)
		for i in range(1, 1000):
			ri.write(f"This is line number {i} {suffix}")
		yield ri

if __name__ == "__main__":
	app = RichLogTest()
	app.run()
```

# Textual Diagnostics

## Versions

| Name    | Value  |
|---------|--------|
| Textual | 0.76.0 |
| Rich    | 13.7.1 |

## Python

| Name           | Value                                              |
|----------------|----------------------------------------------------|
| Version        | 3.12.5                                             |
| Implementation | CPython                                            |
| Compiler       | GCC 14.2.0                                         |
| Executable     | /path/to/venv/bin/python3 |

## Operating System

| Name    | Value                                               |
|---------|-----------------------------------------------------|
| System  | Linux                                               |
| Release | 6.10.4-amd64                                        |
| Version | #1 SMP PREEMPT_DYNAMIC Debian 6.10.4-1 (2024-08-12) |

## Terminal

| Name                 | Value         |
|----------------------|---------------|
| Terminal Application | tmux (3.4)    |
| TERM                 | tmux-256color |
| COLORTERM            | truecolor     |
| FORCE_COLOR          | *Not set*     |
| NO_COLOR             | *Not set*     |

## Rich Console options

| Name           | Value                |
|----------------|----------------------|
| size           | width=196, height=25 |
| legacy_windows | False                |
| min_width      | 1                    |
| max_width      | 196                  |
| is_terminal    | False                |
| encoding       | utf-8                |
[...]

PR #4916, opened 2024-08-21
Fix scroll page

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/4914

### DIFF
--- a/tests/snapshot_tests/snapshot_apps/scroll_page.py
+++ b/tests/snapshot_tests/snapshot_apps/scroll_page.py
@@ -6,8 +6,8 @@ from textual.widgets import RichLog
 class RichLogTest(App):
     def compose(self):
         ri = RichLog(auto_scroll=False)
-        suffix = "A" * 1000
-        for i in range(1, 1000):
+        suffix = "A" * 100
+        for i in range(1, 100):
             ri.write(f"This is line number {i} {suffix}")
         yield ri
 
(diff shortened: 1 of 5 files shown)


tests/snapshot_tests/test_snapshots.py
```diff
@@ -1448,3 +1448,11 @@ def test_command_palette_key_change(snap_compare):
 def test_split(snap_compare):
     """Test split rule."""
     assert snap_compare(SNAPSHOT_APPS_DIR / "split.py", terminal_size=(100, 30))
+
+
+def test_scroll_page_down(snap_compare):
+    """Regression test for https://github.com/Textualize/textual/issues/4914"""
+    # Should show 25 at the top
```
darrenburns, line 1455: Just a reminder that if you write the comment in the docstring, it'll  appear in the snapshot report now which is really handy if the test fails.

Also, you can inline the app if you want and just pass in the instance rather than putting it in a separate file.

tests/snapshot_tests/snapshot_apps/scroll_page.py
```diff
@@ -0,0 +1,17 @@
+import os
+from textual.app import App
+from textual.widgets import RichLog
+
+
+class RichLogTest(App):
```
darrenburns, line 6: Could you reduce the numbers here a bit? This test adds 2s to the test suite runtime on my machine.

Perhaps a small 5x10 terminal with smaller strings?

darrenburns (changes requested):

tests/snapshot_tests/snapshot_apps/scroll_page.py
willmcgugan: Ok, reduced the snapshot size.

darrenburns (approved):

PR #4920, opened 2024-08-22
added system commands

willmcgugan:
- Adds mechanism to add system commands that is easier to extend.
- Rename `SystemCommands` to `SystemCommandProvider`


tests/snapshot_tests/test_snapshots.py
```diff
@@ -1450,8 +1450,23 @@ def test_split(snap_compare):
     assert snap_compare(SNAPSHOT_APPS_DIR / "split.py", terminal_size=(100, 30))
 
 
+def test_system_commands(snap_compare):
+    """Test help panel."""
```
darrenburns, line 1454: Remember that these docstrings appear in the snapshot report now, so it's useful to describe what we expect to see in the snapshot. If it fails, it can sometimes be hard to know what we expect the snapshot to look like.

```suggestion
    """The system commands should appear in the command palette."""
```

src/textual/app.py
```diff
@@ -922,6 +928,59 @@ def active_bindings(self) -> dict[str, ActiveBinding]:
...
+        def get_system_commands(self):
+            yield from super().get_system_commands()
+            yield ("Bell", "Ring the bell", self.bell)
+        ```
+
+        !!! note
+            Requires that [`SystemCommandsProvider`][textual.system_commands.SystemCommandsProvider] is in `App.COMMANDS` class variable.
+
+        Yields:
+            tuples of (TITLE, HELP TEXT, CALLBACK)
```
darrenburns, line 950: IMO a NamedTuple or something would be cleaner here as you'd be able to name and document each field. I always use the `Binding` class instead of tuples for example because I cannot remember the order and would have to check the docs each time. Using `Binding` makes it easy to remember because my IDE helps.

An optional 4th bool which determines whether to include/exclude commands from the "discovery" section without needing to create a custom provider would also be super useful.

willmcgugan: Ok, done.

Personally I find atomic types and a docstring to be less taxing. But a NT is more IDE friendly.

darrenburns: Thank you 🙏

darrenburns (approved):

PR #4931, opened 2024-08-24
Feature to maximize widgets

willmcgugan:
Adds functionality to maximize widgets, which will show them full-screen in isolation.

- The screen has a `maximize` method which maximizes the given widget. It is possible to maximize anything, but it seems most appropriate for focused widgets.
- Widget subclasses can add a `ALLOW_MAXIMIZE` classvar to allow containers to be maximized. Handy for forms containing various inputs.
- Adds a system command to maximize widgets.

https://github.com/user-attachments/assets/77459dae-f5bc-4785-81a1-f72eae4d5295


src/textual/containers.py
```diff
@@ -29,6 +29,10 @@ class Container(Widget):
 class ScrollableContainer(Widget, can_focus=True, inherit_bindings=False):
     """A scrollable container with vertical layout, and auto scrollbars on both axis."""
 
+    # We don't typically want to maximize scrollable containers,
+    # since the user can easily navigate the contents
+    ALLOW_MAXIMIZE = False
```
darrenburns, line 34: Not sure about this comment. The areas of Posting I've had people ask most for "maximise" in the past are scrollable. They just want to see more on screen at once.

src/textual/screen.py
```diff
@@ -184,6 +186,11 @@ class Screen(Generic[ScreenResultType], Widget):
 
     Should be a set of [`command.Provider`][textual.command.Provider] classes.
     """
+    ALLOW_IN_MAXIMIZED_VIEW: ClassVar[str] = ".-textual-system,Footer"
```
darrenburns, line 189: Maybe?
```suggestion
    ALLOW_IN_MAXIMIZED_VIEW: ClassVar[str] = ".-textual-system,Header,Footer"
```

darrenburns, line 189: Does this only work with widgets on other layers and docked widgets?

src/textual/screen.py
```diff
@@ -621,6 +674,41 @@ def focus_previous(self, selector: str | type[QueryType] = "*") -> Widget | None
         """
         return self._move_focus(-1, selector)
 
+    def maximize(self, widget: Widget, container: bool = True) -> None:
+        """Maximize a widget, so it fills the screen.
+
+        Args:
+            widget: Widget to maximize.
+            container: Maximize container if possible.
```
darrenburns, line 682: Could you add a bit more detail to the `container` part of this docstring?

src/textual/screen.py
```diff
@@ -621,6 +674,41 @@ def focus_previous(self, selector: str | type[QueryType] = "*") -> Widget | None
...
+        """
+        if container:
+            # If we want to maximize the container, look up the dom to find a suitable widget
+            for maximize_widget in widget.ancestors:
```
darrenburns, line 686: If I pass a container into this method, won't this code mean it's going to maximise the first widget with `allow_maximize = True` instead of the container I passed in?

src/textual/screen.py
```diff
@@ -621,6 +674,41 @@ def focus_previous(self, selector: str | type[QueryType] = "*") -> Widget | None
...
+                self.scroll_to_widget, self.focused, animate=False, center=True
+            )
+
+    def action_maximize(self) -> None:
+        """Action to maximize the currently focused widget."""
+        if self.focused is not None:
+            self.maximize(self.focused)
+
+    def action_minimize(self) -> None:
+        """Action to minimize the currently focused widget."""
```
darrenburns, line 709:
```suggestion
        """Action to minimize the currently maximized widget."""
```

tests/snapshot_tests/test_snapshots.py
```diff
@@ -1488,3 +1489,41 @@ def test_scroll_page_down(snap_compare):
...
+
+        def compose(self) -> ComposeResult:
+            yield Button("Hello")
+            yield Button("World")
+            yield Footer()
+
+    assert snap_compare(MaximizeApp(), press=["m"])
+
+
+def test_maximize_container(snap_compare):
```
darrenburns, line 1508: I'd suggest you rename this to `test_maximize_widget_in_container` and then add another test that maximizes a container. From looking at the code I'm not 100% sure that's going to work as expected.

darrenburns (changes requested):

src/textual/app.py
```diff
@@ -3441,6 +3461,11 @@ async def _on_layout(self, message: messages.Layout) -> None:
         message.stop()
 
     async def _on_key(self, event: events.Key) -> None:
+        # Special case for maximized widgets
+        # If something is maximized, then escape should minimize
+        if self.screen.maximized is not None and event.key == "escape":
```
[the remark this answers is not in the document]
willmcgugan: The widget should minimize.

src/textual/containers.py
willmcgugan: But most containers we wouldn't want to be maximized. Apps tend to have a lot of containers, and we wouldn't want them to maximized for the most part. But you can always set that to `True` when you do.

src/textual/screen.py
willmcgugan: The Header seemed less important. But the Footer contains stuff that may be required for the maximized widget.

I think its best to leave that to the dev.

willmcgugan: Only if they are a direct child of the Screen.

src/textual/screen.py
willmcgugan: k

src/textual/screen.py
darrenburns: In addition to what I said?

src/textual/screen.py
willmcgugan: You don't pass in the container. You maximize a widget, and if that widget is within a maximizeable container, it will maximize that.

Feel free to suggest rewording to docstring to make that clearer.

tests/snapshot_tests/test_snapshots.py
willmcgugan: There are two tests, one that maximizes a widget, and one that maximizes a widget in a container. I think the covers the two cases?

src/textual/containers.py
darrenburns: Isn't your comment here at odds with the current implementation though? Containers currently can be maximised if they're focusable?

src/textual/screen.py
willmcgugan: Not sure what you are asking. Can you elaborate?

darrenburns: > Only if they are a direct child of the Screen.

Do you mean widgets on layers and docked widgets that are also direct children of the screen?

Or _any_ widget that is a direct child of the screen.

src/textual/screen.py
```diff
@@ -621,6 +674,42 @@ def focus_previous(self, selector: str | type[QueryType] = "*") -> Widget | None
         """
         return self._move_focus(-1, selector)
 
+    def maximize(self, widget: Widget, container: bool = True) -> None:
+        """Maximize a widget, so it fills the screen.
+
+        Args:
+            widget: Widget to maximize.
+            container: If one of the widgets ancestors is a maximizeable widget, also maximize that.
```
darrenburns, line 682:
```suggestion
            container: If one of the widgets ancestors is a maximizeable widget, maximize that instead.
```

darrenburns (approved):
One suggestion for the docstring and one open question but LGTM.

src/textual/screen.py
willmcgugan: Any widget. i.e. `Screen > *`

darrenburns: Got it, thanks 👍

PR #4946, opened 2024-08-27
fix(screenshots): simplify segments only for snapshot tests

### LINKED ISSUE #4929: Extra space when saved to SVG
Running the `DirectoryTree` widget [example](https://textual.textualize.io/widgets/directory_tree/#example) gives this:

image

But when saved to SVG with the command paletter, an extra space is added:

directorytreeapp_2024-08-24T19_56_13_434479

```
# Textual Diagnostics

## Versions

| Name    | Value  |
|---------|--------|
| Textual | 0.77.0 |
| Rich    | 13.7.1 |

## Python

| Name           | Value                                                                     |
|----------------|---------------------------------------------------------------------------|
| Version        | 3.12.5                                                                    |
| Implementation | CPython                                                                   |
| Compiler       | GCC 12.4.0                                                                |
| Executable     | /home/david/.local/share/hatch/env/virtual/jpterm/N1y3coPp/dev/bin/python |

## Operating System

| Name    | Value                                                       |
|---------|-------------------------------------------------------------|
| System  | Linux                                                       |
| Release | 6.8.0-40-generic                                            |
| Version | #40-Ubuntu SMP PREEMPT_DYNAMIC Fri Jul  5 10:34:03 UTC 2024 |

## Terminal

| Name                 | Value                              |
|----------------------|------------------------------------|
| Terminal Application | WezTerm (20240203-110809-5046fc22) |
| TERM                 | xterm-256color                     |
| COLORTERM            | truecolor                          |
| FORCE_COLOR          | *Not set*                          |
| NO_COLOR             | *Not set*                          |

## Rich Console options

| size           | width=135, height=31 |
| max_width      | 135                  |
| max_height     | 31                   |
```

TomJGooding:
This draft PR attempts to implement the suggested solution to fix #4929:

> I think the solution may be to offer the normalization on a switch, so that screenshots are exported without it, but tests are still normalized.

However, the problem is that  [pytest-textual-snapshot](https://github.com/Textualize/pytest-textual-snapshot) relies on the `take_svg_screenshot` function used to generate the docs. This shared function makes it difficult to fix the issues in the docs while still "simplifying" the snapshot tests.

- [ ] Updated documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 
 - Input validation of floats no longer accepts NaN (not a number). https://github.com/Textualize/textual/pull/4784
+- Fixed issues with screenshots by simplifying segments only for snapshot tests https://github.com/Textualize/textual/issues/4929
 
 ## [0.79.1] - 2024-08-31
 


src/textual/_doc.py
```diff
@@ -65,6 +65,7 @@ def take_svg_screenshot(
     terminal_size: tuple[int, int] = (80, 24),
     run_before: Callable[[Pilot], Awaitable[None] | None] | None = None,
     wait_for_animation: bool = True,
+    simplify=True,
```
willmcgugan, line 68: Could you add the docstring for this.

src/textual/app.py
```diff
@@ -1379,7 +1379,12 @@ def action_screenshot(self, filename: str | None = None, path: str = "./") -> No
         """
         self.save_screenshot(filename, path)
 
-    def export_screenshot(self, *, title: str | None = None) -> str:
+    def export_screenshot(
+        self,
+        *,
+        title: str | None = None,
+        simplify: bool = False,
```
willmcgugan, line 1386: Arg in docstring please.

willmcgugan (changes requested):
If we update `format_svg` to set `simplify=False`, then I think that gives us what we want.

Issue #4939, opened 2024-08-27
`TextArea` with indent tab behaviour won't minimize on `esc` key


TomJGooding:
I'm not sure what default behaviour is expected in this scenario, but consider this simple example containing a `TextArea.code_editor`. After maximizing this widget, pressing the <kbd>esc</kbd> key won't minimize it again because it is expected this will instead move focus.

```python
from textual.app import App, ComposeResult
from textual.containers import Vertical
from textual.widgets import TextArea

TEXT = """\
def hello(name):
    print("hello" + name)

def goodbye(name):
    print("goodbye" + name)
"""


class TextAreaExample(App):
    CSS = """
    Screen {
        align: center middle;
    }

    #code-container {
        width: 20%;
        height: 20%;
    }
    """

    def compose(self) -> ComposeResult:
        with Vertical(id="code-container"):
            yield TextArea.code_editor(TEXT, language="python")


app = TextAreaExample()
if __name__ == "__main__":
    app.run()
```

darrenburns:
It's intended to minimize: https://github.com/Textualize/textual/pull/4931#discussion_r1731011637

willmcgugan:
Good catch, the minimize should be done earlier.

PR #4947, opened 2024-08-27
Escape to minimize

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/4939


darrenburns (approved):

TomJGooding:
What happens on <kdb>esc</kbd> when the `TextArea` is inside a maximized container with other widgets?

willmcgugan:
It will always minimize, yes. So you may have to press esc twice when in a textarea with tab indent.

PR #4950, opened 2024-08-28
Faster query_one

willmcgugan:
A faster `query_one`.

- Exit when a match is found, without iterating the full DOM
- Caches simple queries (containing just IDs or types)
- Optimizes `get_widget_by_id` by going through `query_one`

### DIFF
--- a/src/textual/dom.py
+++ b/src/textual/dom.py
@@ -78,6 +78,10 @@ WalkMethod: TypeAlias = Literal["depth", "breadth"]
 ReactiveType = TypeVar("ReactiveType")
 
 
+QueryOneCacheKey: TypeAlias = "tuple[int, str, Type[Widget]]"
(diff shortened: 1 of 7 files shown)


src/textual/dom.py
```diff
@@ -213,6 +218,8 @@ def __init__(
             dict[str, tuple[MessagePump, Reactive | object]] | None
         ) = None
         self._pruning = False
+        self._query_one_cache: LRUCache[tuple[object, ...], DOMNode] = LRUCache(1024)
```
darrenburns, line 221: Do you think we could make the typing here more robust by writing out the types that are actually used and including the tuple length?

If the cache key changes it's handy to have the type checker point you to all the places where the code needs to change.

darrenburns (approved):

Issue #4949, opened 2024-08-28
Make "escape to dismiss" optional for maximised widgets


darrenburns:
This will allow users to use a custom keybinding for maximising/minimising, and address the issue of "escape" potentially conflicting with other widgets bindings in maximised mode.

This could be passed in as an argument to the maximise action or method. (`escape_to_dismiss=False/True`)

willmcgugan:
I'm thinking this should be a classvar constant on Screen. You probably know in advance if you want the escape_to_dismiss behavior

PR #4951, opened 2024-08-28
Add classvar ESCAPE_TO_MINIMIZE

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/4949

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 
 - Added `DOMNode.check_consume_key` https://github.com/Textualize/textual/pull/4940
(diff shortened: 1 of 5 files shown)


darrenburns:
As mentioned, maybe this should be union of true/false/none. The value at App takes precedence. If the value is set (not None) at the Screen level, the the value from the Screen overrides.

darrenburns (approved):
Looks good. Just as a general question - is there something about our binding system that means we're preferring to handle this inside `on_key` rather than as a binding (or priority binding)? If we're opting to do this, it might suggest we're missing some functionality in the binding system?

willmcgugan:
It could be done with a binding. But it would place a burden on the dev not to break it, and to do the right thing if they want to bind escape for something else. Maybe we could add something to the binding system to make that a non-issue in the future, but I don't know what that would look like.

PR #4957, opened 2024-08-31
Fix non active screen updates

willmcgugan:
Fixed an issue with no active screen updates.

You can see this on the tutorial app. Start a timer then open the command palette. Note that the command palette list doesn't update properly.


TomJGooding:
@willmcgugan I spotted that v0.79.1 isn't showing as the latest release in the repo (https://github.com/Textualize/textual/releases/latest). Probably not important with how fast Textual is developing, but just thought I should let you know.

willmcgugan:
Thanks. Fixed.

Issue #4959, opened 2024-09-01
`clear_panes` and memory leak


peppedilillo:
Hello! Thanks for this wonderful framework, having a blast with it.
I seem to have a memory leak with my app. I suspect the problem to be caused by my usage of the `TabbedContent`'s method `clear_panes()`. I have a `TabbedContent` which i want to refresh every time a user provides a file input. The widget contains tabs which reference large tables. I was expecting `clear_panes()` to remove all references to the stale tabs, letting the stale tables to the garbage collector.  This doesn't seem to be happening. I have reproduced this behavior with the following MRE:

```python
from textual.app import App
from textual.widgets import Header, TabbedContent, Footer, TabPane
from textual import work


class MyTabPane(TabPane):
    def __init__(self):
        super().__init__("table")
        self.df = [1] * 100_000_000


class Explorer(App):
    BINDINGS = [
        ("ctrl+o", "add_table", "add table"),
    ]

    def __init__(self):
        super().__init__()

    def compose(self):
        yield Header(show_clock=False)
        yield TabbedContent()
        yield Footer()

    @work
    async def action_add_table(self):
        tabs = self.query_one(TabbedContent)
        tabs.loading = True
        await tabs.clear_panes()
        await tabs.add_pane(MyTabPane())
        tabs.loading = False


if __name__ == "__main__":
    app = Explorer()
    app.run()
```

Pressing `ctrl+o` repeatedly I can see an increase in the process's memory usage. I am on MacOS Ventura, iterm2.
What can I do to solve this problem with my program?
Thank you.

willmcgugan:
Will take a look later. In the meantime you could do `del self.df` in `on_unmount`

PR #4964, opened 2024-09-03
fix object leak

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/4959

This turned out to be quite a rabbit hole. I fixed a class of problems where something was holding on to references to DOM nodes, and preventing garbage collection.

One culprit was `count_parameters`. The `@lru_cache` was holding on to a reference chain. I refactored that method to cache in a different way.

Caching or cache-like structures were also holding on to references. I've added clearing of these "caches" to make garbage collection more prompt.

I've made some references weak references, to prevent potential reference cycles.

But the largest part of this work was related to context vars. These weren't being managed well, and it was leaving references around to the app or widgets, which would prevent entire branches from being collected.

The context var issue I fixed by writing a context var context manager, which resets the context var. DOMNodes and App have grown a `_context()` which sets and restores the active app and message pump.

Incidentally, I think this is the cause of some of the weirdness in tests where tests work in a run but fail in isolation. Errors in managing the context vars could result in an app being shared across tests.

Many of the tests had to be updated after the context var changes. For some reason the `pilot` fixture broke. I think because of the way that pytest runs async tests.

I'm pretty sure these changes improve shutdown time. Although I haven't profiled.


src/textual/app.py
```diff
@@ -1355,8 +1362,8 @@ def call_from_thread(
 
         async def run_callback() -> CallThreadReturnType:
             """Run the callback, set the result or error on the future."""
-            self._set_active()
-            return await invoke(callback_with_args)
+            with self._context():
```
kinow, line 1365: An interesting trick, I might use that in some of my code. Thanks for the tip in the tweet!

tests/test_gc.py
```diff
@@ -0,0 +1,87 @@
...
+
+        objgraph.show_backrefs(
+            [
+                obj
+                for obj in gc.get_objects()
+                if any(cls.__name__ == "App" for cls in obj.__class__.__mro__)
+            ],
+            filename="graph.png",
+            max_depth=15,
+        )
```
darrenburns, line 77: Does this need to be removed or commented out?

tests/test_unmount.py
```diff
@@ -50,4 +50,7 @@ async def on_mount(self) -> None:
         "MyScreen#main",
     ]
 
+    print(unmount_ids)
+    print(expected)
```
darrenburns, line 54:
```suggestion
```

src/textual/reactive.py
```diff
@@ -203,7 +200,7 @@ def _reset_object(cls, obj: object) -> None:
         Args:
             obj: A reactive object.
         """
-        getattr(obj, "__watchers", {}).clear()
+        getattr(obj, "_watchers", {}).clear()
         getattr(obj, "__computes", []).clear()
```
darrenburns, line 204: I feel we should either rename both `__watchers` and `__computes` to single underscore prefix or neither.

darrenburns (approved):
Looks good - just a few minor comments. I tried this PR out in Posting to exercise it and could see no issues.

tests/test_gc.py
willmcgugan: It's marked as xfail for now. When not run in isolation it fails because pytest keeps references around, presumably to inspect and report errors.

src/textual/reactive.py
willmcgugan: Watchers was promoted because I needed to clear it from outside of the class. `__computes` didn't need that.

On reflection though, I can add a method to Reactive to do that, and I can keep the `__`

peppedilillo:
Thank you!

PR #4982, opened 2024-09-10
Allowing `none` in some CSS rules

### LINKED ISSUE #4977: Add support for `dock: none`
@davidfokkema pointed out on Discord that `dock: none;` doesn't work in CSS.

We would expect this to "undock" a docked widget.

It seems `widget.styles.dock = ""` works.

darrenburns:
- [x] Closes #4977 
- [x] Adds support for `split: none` to CSS and inline Python.
- [x] Adds support for `hatch: none` to CSS and inline Python.
- [x] Adding docstrings to many of the attributes on `widget.styles` (helpful popups appear in VS Code now when you hover)
- [x] Adds support for `widget.styles.border = "none"`, `widget.style.border_{edge} = "none"`, `widget.style.outline = "none"`, `widget.style.outline_{edge} = "none"`, which worked when using CSS but crashed with an `IndexError` when done inline. (Previously you could make it work by doing something like `("", "red")`.
- [x] Default `scrollbar_size_vertical` value was wrong in `StylesBase` - which was surprisngly affecting the location that notifications are rendered in. There was a subsequent workaround (`margin-right: 1`) in `ToastRack` which I've now removed.

Any others?


src/textual/_arrange.py
```diff
@@ -47,17 +47,28 @@ def arrange(
 
     placements: list[WidgetPlacement] = []
     scroll_spacing = Spacing()
-    get_dock = attrgetter("styles.dock")
-    get_split = attrgetter("styles.split")
+
+    def _attrgetter_convert_none(attribute: str) -> Callable[[Widget], str | None]:
```
willmcgugan, line 51: Lets add `is_docked` and `is_split` methods to Styles.

willmcgugan (changes requested):

willmcgugan (approved):

PR #4985, opened 2024-09-11
Add `can_focus` to guide, mention how `BINDINGS` are checked

### LINKED ISSUE #4918: Document can_focus
I don't think we document can_focus anywhere other than the API reference.

We should add it to the widget guide.

darrenburns:
Closes #4918.

Also noticed that we only seem to give examples of "global"/`App`-level bindings, so I've added a couple of sentences explaining how the focused widget is checked first, followed by a look upwards through the DOM.

@willmcgugan You mentioned in the issue to add it to the "widget" guide, but it seemed like the "input" when I checked as that's where most of the discussion around focus happens. Let me know if you disagree.


docs/guide/input.md
```diff
@@ -113,9 +113,16 @@ The app splits the screen in to quarters, with a `RichLog` widget in each quarte
 
 You can move focus by pressing the ++tab++ key to focus the next widget. Pressing ++shift+tab++ moves the focus in the opposite direction.
 
+### Focusable widgets
+
+Each widget has a boolean `can_focus` attribute which determines if it is capable of receiving focus.
```
willmcgugan, line 118: Could we also document that you can set `can_focus=True` with the subclass args?

darrenburns: Where would you like me to do that? At this point in the guide we haven't described how to create even a basic widget.

willmcgugan: I guess it would have to be in the [widgets](https://textual.textualize.io/guide/widgets/) section of the guide.

It could use a short example as well.

darrenburns: That section of the docs is feeling quite bare - it only shows how to create static widgets that display some text. There's no mention of widget-level bindings at all and so no interactivity (other than the Rich markup links). In fact, I don't think there's a single example of bindings on a widget in the entire guide.

Can I suggest we add an example of a simple interactive widget with `can_focus` demonstrated too? Otherwise I think the `can_focus=True` example would look out of place?

willmcgugan: Yup. Makes sense. Can you run it by me when you have a rough draft please.

darrenburns: Added a section

docs/guide/widgets.md
```diff
@@ -51,6 +51,32 @@ The addition of the CSS has completely transformed our custom widget.
 ```{.textual path="docs/examples/guide/widgets/hello02.py"}
 ```
 
+### Responding to key presses
```
willmcgugan, line 54: I think we need a more elaborate example here.

We've managed to avoid abbreviated code. Virtually everything has a fully working example for any given feature.

Suggest we come up with some kind of input widget that shows focus, and does *something*. How about a widget that displays a 24-hour clock time. Where clicking on a digit increments that digit. Just an idea.

willmcgugan (changes requested):
I think a fully working example is needed here. It is a major deficiency in the docs that we don't cover focusing and bindings.

docs/guide/widgets.md
darrenburns: Maybe I'm misunderstanding but I don't think that idea would really demonstrate bindings or focus? I find examples with that kind of scope to be a major distraction too - 90% of the code is boiler plate and the actual stuff related to the concept we're trying to explain is buried away. I personally end up skipping examples like that due to the verbosity (and my goldfish-like attention span)!

I have a simple keyboard driven `Counter` widget I'm using to test keymaps that might make a good example? You focus it and then use up/right/k/plus to increment, and can decrement too.

darrenburns:
I've added a section to the guide showing how to construct a basic focusable widget with simple bindings defined.

image

docs/guide/widgets.md
```diff
@@ -190,6 +190,67 @@ If the supplied text is too long to fit within the widget, it will be cropped (a
...
+## Focus & keybindings
+
+Widgets can have a list of associated key [bindings](../guide/input.md#bindings),
+which let them call [actions](../guide/actions.md) in response to key presses.
+
+A widget is only able to handle key presses if it or one of its descendants has [focus](../guide/input.md#input-focus).
+
+To demonstrate, let's design a simple interactive counter widget which can be incremented and decremented using the keyboard.
+
+In the following example, we define a simple `Counter` widget with `can_focus=True`, and some CSS to make it stand out when focused.
```
willmcgugan, line 202: the `can_focus=True` is a really important takeaway here. It's what we were lacking, and why we keep getting asked. Here it seems too much of a sidenote.

"Widgets aren't focusable by default. In order for a widget to be focusable we need to set `can_focus=True` when we define a widget subclass. Here's an example of a focusable widget ..."

docs/guide/widgets.md
```diff
@@ -190,6 +190,67 @@ If the supplied text is too long to fit within the widget, it will be cropped (a
 There are a number of styles that influence how titles are displayed (color and alignment).
 See the [style reference](../styles/index.md) for details.
 
+## Focus & keybindings
+
+Widgets can have a list of associated key [bindings](../guide/input.md#bindings),
+which let them call [actions](../guide/actions.md) in response to key presses.
+
+A widget is only able to handle key presses if it or one of its descendants has [focus](../guide/input.md#input-focus).
```
willmcgugan, line 198:
```suggestion
A widget is able to handle key presses if it or one of its descendants has [focus](../guide/input.md#input-focus).
```

omitting needless words. And it probably isn't the *only* way to handle keypresses...

docs/guide/widgets.md
```diff
@@ -190,6 +190,67 @@ If the supplied text is too long to fit within the widget, it will be cropped (a
...
+
+To demonstrate, let's design a simple interactive counter widget which can be incremented and decremented using the keyboard.
```
willmcgugan, line 200: Could omit this sentence if you agree with the following comment.

willmcgugan (changes requested):

docs/guide/widgets.md
```diff
@@ -190,6 +190,67 @@ If the supplied text is too long to fit within the widget, it will be cropped (a
 There are a number of styles that influence how titles are displayed (color and alignment).
 See the [style reference](../styles/index.md) for details.
 
+## Focus & keybindings
```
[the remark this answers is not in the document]
darrenburns: Yeah I changed this around a few times - 

We're on the "Widgets" page, so it felt clear to me that it was in the context of widgets. It also adds "keybindings" to the table of contents for the page, which I felt was important for discoverability.. "how do I add keybindings to my widget" feels like the "question" I'd be looking to an answer for, and having headings suggest that there might be answer contained within is important.

docs/guide/widgets.md
```diff
@@ -190,6 +190,67 @@ If the supplied text is too long to fit within the widget, it will be cropped (a
...
+
+Widgets can have a list of associated key [bindings](../guide/input.md#bindings),
```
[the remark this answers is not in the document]
darrenburns: It certainly is repetition, but our last mention of this stuff was quite some time ago and repetition is sometimes important to build context or refresh memories.

My thought was someone unfamiliar with Textual and reading this for the first time would appreciate a 1-sentence refresher as well as a link back to the "main" explanation.

docs/guide/widgets.md
darrenburns: Removing the "only" changes the nuance of the sentence, so I wouldn't call it needless. You're right that it's not technically correct though - will fix!

docs/guide/widgets.md
darrenburns: I think this is clearer in the rendered docs, as the example app below has a single highlighted line for the `can_focus=True` as well as a popup explaining it. The whole purpose of the example is to draw attention to the `can_focus=True`.

Have reworded to try and emphasise it more.

willmcgugan (approved):

PR #4987, opened 2024-09-11
Add (possibly flaky) test for slow handler logging

darrenburns:
I think this is gonna be flaky but let's see :)


darrenburns:
This is very tricky to test properly and I suspect it's not worth the effort.

PR #4994, opened 2024-09-12
Fix sorting in command palette search hits

### LINKED ISSUE #4927: Add sorting to command palette
The command palette doesn't sort the matches other than alphabetically.

There does seem to be code in there to rate matches, and the docs do say the score is used in sorting. But no sorting occurs.

You can see this with the current system commands. If I type **hel** I get two results: `Quit the application` and `show keys and help panel`, in that order. I would expect the help panel to be on top as it is a better match according to the code in fuzzy.py.

darrenburns:
Fixes #4927.

This will only sort _search_ hits by their match score, and do no sorting for discovery hits.

Discovery hits are left unsorted as we decided to leave that up to the end user - they may wish the most frequent commands to appear at the top for example.

### DIFF
--- a/src/textual/command.py
+++ b/src/textual/command.py
@@ -20,7 +20,6 @@ from asyncio import (
 from dataclasses import dataclass
 from functools import total_ordering
 from inspect import isclass
-from operator import attrgetter
 from time import monotonic
 from typing import TYPE_CHECKING, Any, AsyncGenerator, AsyncIterator, ClassVar, Iterable
 
@@ -893,7 +892,12 @@ class CommandPalette(SystemModalScreen):
             else None
         )
 
-        sorted_commands = sorted(commands, key=attrgetter("hit"), reverse=True)
+        def sort_key(command: Command) -> tuple[float, str]:
+            score = getattr(command.hit, "score", 0)
+            text = command.hit.text
+            return (-score, text or "")
+
+        sorted_commands = sorted(commands, key=sort_key)
         command_list.clear_options().add_options(sorted_commands)
         if highlighted is not None and highlighted.id:
             command_list.highlighted = command_list.get_option_index(highlighted.id)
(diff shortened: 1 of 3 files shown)


tests/snapshot_tests/__snapshots__/test_snapshots/test_command_palette.svg
darrenburns, line 1: I think this has uncovered a bug in the match scoring. The scores of these items (in the "Added" snapshot) are like `0.9`, `0.8`, ... `0.0`. I'd probably expect them to have the same scores?

willmcgugan: Looks like the app is assigning the incremental scores. It's not using the default matcher

darrenburns: Ah ok, thanks 😄

src/textual/command.py
```diff
@@ -891,7 +891,15 @@ def _refresh_command_list(
             if command_list.highlighted is not None and not clear_current
             else None
         )
-        command_list.clear_options().add_options(commands)
+
+        def sort_key(command: Command) -> float:
```
willmcgugan, line 895: Could you add a constant / property on DiscoveryHit that is always 0? The feels a little cleaner than the getattr to me.

I'm actually somewhat surprised that one doesn't inherit from the other. I wonder what the thinking was behind that?

darrenburns: I think because DiscoveryHits don't have a "score", and they were also added later. I can understand why someone might want to not give DiscoveryHits a score hard-coded to 0 from a design POV. That's why I chose the getattr too, but it doesn't particularly matter that much so I'll change it on Monday.

willmcgugan (approved):

Issue #5012, opened 2024-09-17
Quitting the calculator app with the command palette causes escape sequences to be emitted at the prompt


willmcgugan:
Suspect this is something do to with the inline driver. We should compare it to the linux driver, to see if something is being handled differently.

TomJGooding:
I can't seem to reproduce this issue. If this is only in iTerm, perhaps related to these lines?

https://github.com/Textualize/textual/blob/c054af8ab9fcd5957f659d77dc814f4021e0bd29/src/textual/drivers/linux_driver.py#L258-L259

willmcgugan:
It wasn't that. There was some updates to the key reader thread that hadn't been replicated in the inline driver.

PR #5015, opened 2024-09-17
Fix output glitch with inline driver

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/5012


willmcgugan:
#executivedecision

PR #5018, opened 2024-09-18
fix(inline driver): check padding when clearing output on quit

### LINKED ISSUE #5017: Quitting inline app without padding will also clear the line above
inline_app_no_padding

```python


class InlineAppWithoutPadding(App):
    INLINE_PADDING = 0

        yield Static("Quitting the app will also clear the line above me")


if __name__ == "__main__":
    app = InlineAppWithoutPadding()
    app.run(inline=True)
```

TomJGooding:
Fixes #5017 by updating the `\x1b[2A` escape code to check the new `App.INLINE_PADDING`.

(Edit: I don't know why anyone would do this, but just in case I've now accounted for negative padding!)

### DIFF
--- a/src/textual/drivers/linux_inline_driver.py
+++ b/src/textual/drivers/linux_inline_driver.py
@@ -306,7 +306,8 @@ class LinuxInlineDriver(Driver):
         self._disable_bracketed_paste()
         self.disable_input()
 
-        self.write("\x1b[2A\x1b[J")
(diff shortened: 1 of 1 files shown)


TomJGooding:
I wanted to check when `inline_no_clear=True`, but unfortunately it looks like this is currently broken (see #5019) so I've marked this as draft for now.

Issue #4882, opened 2024-08-14
Inline mode doesn't show CSS errors


patrick91:
Just discussed this briefly with Will on discord, turns out that inline mode doesn't show CSS errors, here's a small-ish reproduction:


```python
from textual.app import App
from textual.containers import Grid
from textual.widgets import Placeholder


class PaddingAllApp(App):
    CSS = "crash" # 👈 error here

    def compose(self):
        yield Grid(
            Placeholder("no padding", id="p1"),
        )


if __name__ == "__main__":
    app = PaddingAllApp()
    app.run(inline=True, inline_no_clear=True)
```

Traceback:

```
╭─────────────────────────────────────────────────────────────────────────────────────────────────── Traceback (most recent call last) ───────────────────────────────────────────────────────────────────────────────────────────────────╮
│ /Users/patrick/github/patrick91/textual-demo/.venv/lib/python3.12/site-packages/textual/app.py:2606 in _process_messages                                                                                                                │
│                                                                                                                                                                                                                                         │
│   2603 │   │   │   │   │   │   )                                                                ╭───────────────────────────────────────────────────── locals ──────────────────────────────────────────────────────╮                   │
│   2604 │   │   │   │   │   if inline_no_clear:                                                  │             app_path = '/Users/patrick/github/patrick91/textual-demo/app.py'                                      │                   │
│   2605 │   │   │   │   │   │   console = Console()                                              │              console = <console width=235 ColorSystem.TRUECOLOR>                                                  │                   │
│ ❱ 2606 │   │   │   │   │   │   console.print(self.screen._compositor)                           │                  css = '\n    App {\n        background: $background;\n        color: $text;\n    }\n    *:di'+51 │                   │
│   2607 │   │   │   │   │   │   console.print()                                                  │             cursor_x = 0                                                                                          │                   │
│   2608 │   │   │   │   │                                                                        │             cursor_y = 0                                                                                          │                   │
│   2609 │   │   │   │   │   driver.stop_application_mode()                                       │               driver = <LinuxInlineDriver PaddingAllApp(title='PaddingAllApp', classes={'-dark-mode'})>           │                   │
│                                                                                                 │                error = ScreenStackError('No screens on stack')                                                    │                   │
│                                                                                                 │             headless = False                                                                                      │                   │
│                                                                                                 │               inline = True                                                                                       │                   │
│                                                                                                 │      inline_no_clear = True                                                                                       │                   │
[...]

PR #5022, opened 2024-09-18
fix(app): fix `inline_no_clear`

### LINKED ISSUE #5019: `inline_no_clear` broken in v0.77.0
It looks like the `inline_no_clear` option was broken in v0.77.0, from a quick git bisect the culprit seems to be commit 8c01ef7.

If you run the app below where `inline_no_clear=True`, the app output is cleared on exit:

```python


class InlineNoClearApp(App):
        yield Static("Quitting should not clear the app output")


if __name__ == "__main__":
    app = InlineNoClearApp()
    app.run(inline=True, inline_no_clear=True)

```

TomJGooding:
Fixes #5019. 

I'm afraid I don't fully understand commit 8c01ef7, but this change seems to fix the broken `inline_no_clear` without any regression related to issue #4882 which I believe this commit intended to fix.

- [ ] Updated documentation

### DIFF
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -2909,10 +2909,14 @@ class App(Generic[ReturnType], DOMNode):
                             self._driver.write(
                                 Control.move(-cursor_x, -cursor_y + 1).segment.text
                             )
-                            if inline_no_clear and not not self.app._exit_renderables:
+                            if inline_no_clear:
                                 console = Console()
-                                console.print(self.screen._compositor)
-                                console.print()
+                                try:
+                                    console.print(self.screen._compositor)
+                                except ScreenStackError:
+                                    pass
+                                else:
+                                    console.print()
 
                         driver.stop_application_mode()
             except Exception as error:


src/textual/app.py
```diff
@@ -2909,10 +2909,14 @@ async def invoke_ready_callback() -> None:
                             self._driver.write(
                                 Control.move(-cursor_x, -cursor_y + 1).segment.text
                             )
-                            if inline_no_clear and not not self.app._exit_renderables:
```
willmcgugan, line 2912: The "not not" is a bug. I think the reasoning was that if there are errors (in the exit renderables) they should be printed rather than the last frame.

willmcgugan:
@TomJGooding Sorry for usurping your PR. This seems to give the results we are looking for (from experimentation)

TomJGooding:
No problem. I did notice now running the example in issue #4882 leaves a bit of a gap before the start of the error message, but probably not really an issue..

PR #5038, opened 2024-09-23
Keymaps

darrenburns:
A mechanism for allowing dynamic bindings (e.g. loaded from config).


src/textual/binding.py
```diff
@@ -252,3 +360,61 @@ def get_bindings_for_key(self, key: str) -> list[Binding]:
             return self.key_to_bindings[key]
         except KeyError:
             raise NoBinding(f"No binding for {key}") from None
+
+
+BindingIDString = str
```
willmcgugan, line 365: Should these be marked as `TypeAlias` ?

src/textual/binding.py
```diff
@@ -193,6 +236,71 @@ def merge(cls, bindings: Iterable[BindingsMap]) -> BindingsMap:
...
+        key_to_bindings = list(self.key_to_bindings.items())
+        for key, bindings in key_to_bindings:
+            for binding in bindings:
+                binding_id = binding.id
+                if binding_id is None:
+                    # Bindings without an ID are irrelevant when applying a keymap
+                    continue
+
+                # If the keymap has an override for this binding ID
+                if keymap_key_string := keymap.get(binding_id):
```
willmcgugan, line 262: `:=`!

giphy

src/textual/app.py
```diff
@@ -4314,3 +4314,15 @@ def _on_delivery_failed(self, event: events.DeliveryComplete) -> None:
...
+        """Gets the keymap used to override default keyboard shortcuts.
+
+        This may be overridden by subclasses to change the keyboard
+        shortcuts used by the application (e.g. by loading
+        from a config/keymap file).
+
+        Returns:
+            The Keymap to apply.
+        """
+        return Keymap(mapping={})
```
willmcgugan, line 4328: Maybe `mapping` is redundant here? Given it is a positional argument.

src/textual/binding.py
```diff
@@ -193,6 +236,71 @@ def merge(cls, bindings: Iterable[BindingsMap]) -> BindingsMap:
                 keys.setdefault(key, []).extend(key_bindings)
         return BindingsMap.from_keys(keys)
 
+    def apply_keymap(self, keymap: Keymap) -> KeymapApplyResult:
```
willmcgugan, line 239: Is this a destructive operation? i.e. can I update the keymap on the fly if I wanted to?

Just thinking about the possibility of a keymap editor in the future, or the ability to load keymaps on the fly.

src/textual/binding.py
```diff
@@ -252,3 +360,61 @@ def get_bindings_for_key(self, key: str) -> list[Binding]:
...
+
+    Textual will look for bindings with the same ID as the
+    the keys in this mapping, and replace the key strings in the
+    default binding with the key strings from this mapping.
+
+    The mapping is immutable - it's created once and then should
+    never be mutated at runtime.
+    """
+
+    def __init__(self, mapping: Mapping[BindingIDString, KeyString]):
```
willmcgugan, line 395: Maybe make this positional only?

willmcgugan (commented):
Can't fault the implementation. Just pedantic stuff and questions...

src/textual/binding.py
darrenburns: It's destructive in that it will completely replace the _key_ associated with a binding in the bindings map (although still retrievable via the default `BINDINGS` list). However, the binding would still maintain it's ID, so could be re-assigned to any key at any time.

Updating on the fly should work as it stands, as the keymap is just "overlayed" over the bindings on-demand, but I suspect you'd need to force a refresh of the key panel and footer.

src/textual/binding.py
darrenburns: I can't even remember how to do this 😂 not really sold on them having any benefits though (open for convincing if you care enough).

willmcgugan: You add a forward slash. But you're right, not much benefit now I think about it.

src/textual/app.py
```diff
@@ -4314,3 +4314,15 @@ def _on_delivery_failed(self, event: events.DeliveryComplete) -> None:
             self.notify(
                 "Failed to save screenshot", title="Screenshot", severity="error"
             )
+
+    def get_keymap(self) -> Keymap:
```
willmcgugan, line 4318: Given the keymap is a wrapper around a dict, could be just use a dict?

It also looks like it would not be a problem to modify the dict on the fly. Perhaps we should just expose a dict via an attribute. And the user could replace / update `self.keymap` as needed?

src/textual/screen.py
```diff
@@ -351,9 +351,21 @@ def _binding_chain(self) -> list[tuple[DOMNode, BindingsMap]]:
...
                         del bindings_map.key_to_bindings[key]
+
             filter_namespaces.append(namespace)
 
+        keymap = self.app.get_keymap()
+        for namespace, bindings_map in namespace_bindings:
+            if keymap:
+                result = bindings_map.apply_keymap(keymap)
+                if result.clashed_bindings:
+                    self.app.post_message(
```
willmcgugan, line 365: This could result in sending potentially messages on every keypress. TBH I don't think that is worth it.

willmcgugan (commented):
I did some profiling, and the `apply_keymap` is clearly doing a bit more work, but nothing dramatic. Although it will grow with larger keymaps.

I also suspect there are some optimization wins in that method. So I'm not too concerned that it is doing additional work on each keypress.

Just a few more questions, but otherwise I think it is good to go. Nice work.

src/textual/screen.py
darrenburns: Do you have any suggestions for an alternative? As a developer I really want to be able to handle a user clobbering bindings cleanly - at least informing them of it.

src/textual/app.py
darrenburns: I wasn't really intending this to be mutable at this stage to be honest.

I wrapped it to keep the mapping immutable and also leave flexibility for extension in the future without a breaking change. Any utility methods/helper properties could be encapsulated inside it, etc.

I'm worried if we make it mutable then we run into some of the same problems we get with mutable reactives. If I update a key, in the keymap, would I need to refresh the footer and keypanel? I've no major problem with it being mutable, but would probably prefer if we had some means of doing it that kept the footer/help panel up-to-date when that mutation happens.

What about an reactive attribute `App.keymap` of type `Keymap`, which remains as-is. The `App.watch_keymap` watcher would refresh any key panels and footers as required?

willmcgugan: How about making the attribute private. Then having `update_keymap` and `replace_keymap` methods? That would avoid the issues with mutability.

src/textual/screen.py
willmcgugan: Could you check for clashes outside of a keypress somehow?

Otherwise, how about calling a method on App and letting the dev decide how to proceed?

src/textual/app.py
darrenburns: Are you opposed to having a `Keymap` object? If so I would like to understand why 😄 

Using primitive dicts makes me sadder than when I see a walrus operator inside a complex expression :( - it's harder to document, means any utility methods/properties we want would need to be crammed in `App`, we lose typing benefits if we want to pass it around, it's less readable because we if I see `dict[str, str]` then I don't know what the strings are (and if I alias them to make it clearer then I need to import both aliases everywhere I use the dict), if we wanted to extend the functionality we need to do a breaking change, we can't add custom reprs etc.

src/textual/screen.py
darrenburns: Anything is possible 😄 but I suspect checking outside of a keypress would be quite a bit of work, and not sure if it's worth it at this point.

Calling a method on App seems like a nice approach.

src/textual/app.py
willmcgugan: Yeah, apposed I'm afraid.

If its private, then the `Keymap` would just exist to wrap a dict and make it immutable. I don't think we've wrapped an atomic type anywhere else with the only goal of making it immutable. The closest I can think of is NodeList, but that adds additional functionality on top of a list.

We can still type methods are receiving a TypeAlias. I don't think it is a major hassle to have to import a type. We do that a lot already.

Ultimately, while it is just a dict, I think it should remain as such.

I can't see how it would result in a breaking change if we needed to go from a dict, to a dict-like object in the future.

Can chat about it if you like. Just in case we aren't on the same page.

darrenburns: Let's agree to disagree. I feel like a gave a pretty strong list of reasons why I think an object is better and you didn't really say concretely why you think a dict is better... but it's your call, I'll make the change tomorrow 😄

src/textual/binding.py
```diff
@@ -20,6 +21,22 @@
...
+Corresponds to the `id` parameter of the `Binding` class.
+"""
+
+KeyString: TypeAlias = str
+"""A string that represents a key binding.
+
+For example, "x", "ctrl+i", "ctrl+shift+a", "ctrl+j,space,x", etc.
+"""
+
+Keymap = Mapping[BindingIDString, KeyString]
```
willmcgugan, line 38: Does this not need a `TypeAlias` ?

willmcgugan (approved):
Just one thing.

src/textual/binding.py
darrenburns: I think that's only needed if you're doing a forward reference in the definition by assigning to a string, like this:

```python
Keymap = "Mapping[BindingIDString, KeyString]"
```

Type-checkers have no way to know that's a type-alias unless you tell them.

> TypeAlias is particularly useful on older Python versions for annotating aliases that make use of forward references, as it can be hard for type checkers to distinguish these from normal variable assignments

https://docs.python.org/3/library/typing.html#typing.TypeAlias

PR #5043, opened 2024-09-23
Extending SelectioList to allow swapping of elements

BurnoutDV:
A minor edit of the std SelectionList I originally extended in a small project of mine because I wanted to use it to select and order the displayed columns of a datatable. I am quite sure the code itself is crude and lacks the usual mastery of a seasoned python dev but its all I got in me. I added some docstrings but felt unable to meaningfully extend the widget documentation. I also wrote no test case as I am not familiar with the proper frameworks.

Let me know if its any good or if the whole Idea is just superfluous.

[video.webm](https://github.com/user-attachments/assets/b9cc1368-ad51-453b-86d6-9440b247a1c4)

### DIFF
--- a/src/textual/widgets/_selection_list.py
+++ b/src/textual/widgets/_selection_list.py
@@ -71,11 +71,17 @@ class Selection(Generic[SelectionType], Option):
 class SelectionList(Generic[SelectionType], OptionList):
     """A vertical selection list that allows making multiple selections."""
 
-    BINDINGS = [Binding("space", "select")]
+    BINDINGS = [
+        Binding("space", "select"),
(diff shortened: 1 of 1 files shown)


BurnoutDV:
Oh don't worry about it. I actually did exactly that, brought a new derived widget into being but then I thought, maybe someone is interested at it and created this commit out of the code. I actually made it as an optional parameter but I didn't knew how to make it so that the bindings are only set when the parameter is true.

BurnoutDV:
Oh forgot to close. As I lack the ability to produces packages I will just let this die. It wasn't a big idea anyway.

Issue #5042, opened 2024-09-23
Textualize v0.80.0 Textarea tries to set cursor position before its mounted and throws error textual._context.NoActiveAppError


paulrobello:
# Textual Diagnostics

## Versions

| Name    | Value  |
|---------|--------|
| Textual | 0.80.0 |
| Rich    | 13.8.1 |

## Python

| Name           | Value                                      |
|----------------|--------------------------------------------|
| Version        | 3.11.9                                     |
| Implementation | CPython                                    |
| Compiler       | MSC v.1938 64 bit (AMD64)                  |
| Executable     | E:\Repos\parllama\.venv\Scripts\python.exe |

## Operating System

| Name    | Value      |
|---------|------------|
| System  | Windows    |
| Release | 10         |
| Version | 10.0.22631 |

## Terminal

| Name                 | Value            |
|----------------------|------------------|
| Terminal Application | Windows Terminal |
| TERM                 | xterm-256color   |
| COLORTERM            | *Not set*        |
| FORCE_COLOR          | *Not set*        |
| NO_COLOR             | *Not set*        |

## Rich Console options

| Name           | Value                |
|----------------|----------------------|
| size           | width=138, height=39 |
| legacy_windows | False                |
| min_width      | 1                    |
| max_width      | 138                  |
| is_terminal    | True                 |
| encoding       | utf-8                |
| max_height     | 39                   |
| justify        | None                 |
| overflow       | None                 |
| no_wrap        | False                |
| highlight      | None                 |
| markup         | None                 |
| height         | None                 |

The TextArea is created in a widgets __init__ then mounted in its compose
```python
self._text_area = TextArea(
    id="user_input_textarea",
)
```
Throws Error: textual._context.NoActiveAppError
```
  File ".\.venv\Lib\site-packages\textual\widgets\_text_area.py", line 631, in _watch_selection
    self.app.cursor_position = self.cursor_screen_offset
```

PR #5045, opened 2024-09-24
prevent error in TextArea constructor

willmcgugan:
The TextArea was invoking code in the constructor that expected an active app.

Fixes https://github.com/Textualize/textual/issues/5042


darrenburns (approved):
LGTM assuming the tests pass :)

PR #5046, opened 2024-09-24
Adjust Log.highlighter type to allow for custom highlighters.

xavierog:
This is the exact same as #4247 but for Log instead of RichLog: without an explicit type, tools like mypy assume Log.highlighter must be a ReprHighlighter instead of a Highlighter.

Checklist: not worth it.


willmcgugan (approved):
Thanks

Issue #5050, opened 2024-09-24
tree behavior changes in Textual 0.80


godlygeek:
With Textual 0.79.1, running `python docs/examples/widgets/tree.py` looks like this:

![image](https://github.com/user-attachments/assets/16c50bbd-509e-4de7-b09d-b889aada8d3d)

With Textual 0.80.1, running `python docs/examples/widgets/tree.py` looks like this:

![image](https://github.com/user-attachments/assets/3f84d4db-c894-4d72-a849-578b33794c4c)

Note that the arrows representing whether non-leaf nodes are expanded or collapsed are reversed, and the orange lines indicating the current node and all of its descendants are no longer there in the latest version. Neither of these changes are mentioned in the changelog.

```
# Textual Diagnostics

## Versions

| Name    | Value  |
|---------|--------|
| Textual | 0.80.1 |
| Rich    | 13.7.1 |

## Python

| Name           | Value                                                            |
|----------------|------------------------------------------------------------------|
| Version        | 3.11.10                                                          |
| Implementation | CPython                                                          |
| Compiler       | GCC 12.1.1 20220628 (Red Hat 12.1.1-3)                           |
| Executable     | /home/mwoznisk/.pyenv/versions/3.11/envs/memray-3.11/bin/python3 |

## Operating System

| Name    | Value                               |
|---------|-------------------------------------|
| System  | Linux                               |
| Release | 5.15.153.1-microsoft-standard-WSL2  |
| Version | #1 SMP Fri Mar 29 23:14:13 UTC 2024 |

## Terminal

| Name                 | Value          |
|----------------------|----------------|
| Terminal Application | *Unknown*      |
| TERM                 | putty-256color |
| COLORTERM            | truecolor      |
| FORCE_COLOR          | *Not set*      |
| NO_COLOR             | *Not set*      |

## Rich Console options

| Name           | Value                |
|----------------|----------------------|
| size           | width=212, height=98 |
| legacy_windows | False                |
| min_width      | 1                    |
| max_width      | 212                  |
| is_terminal    | True                 |
| encoding       | utf-8                |
| max_height     | 98                   |
| justify        | None                 |
| overflow       | None                 |
| no_wrap        | False                |
| highlight      | None                 |
| markup         | None                 |
| height         | None                 |
```

godlygeek:
For context, I noticed this while updating the snapshots for Memray. This shows up as an unexpected difference for all of our `memray tree` tests:

[...]

PR #5052, opened 2024-09-25
fix tree issues

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/5050

### DIFF
--- a/src/textual/widgets/_directory_tree.py
+++ b/src/textual/widgets/_directory_tree.py
@@ -364,7 +364,9 @@ class DirectoryTree(Tree[DirEntry]):
         """
         self.reset_node(self.root, str(self.path), DirEntry(self.PATH(self.path)))
         await self.reload()
-        self.move_cursor(self.root, animate=False)
+        if self.cursor_node is not None:
+            self.cursor_line = 0
+            self.move_cursor(self.cursor_node, animate=False)
 
     def process_label(self, label: TextType) -> Text:
         """Process a str or Text into a label. May be overridden in a subclass to modify how labels are rendered.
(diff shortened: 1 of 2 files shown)


src/textual/widgets/_directory_tree.py
```diff
@@ -364,7 +364,9 @@ async def watch_path(self) -> None:
         """
         self.reset_node(self.root, str(self.path), DirEntry(self.PATH(self.path)))
         await self.reload()
-        self.move_cursor(self.root, animate=False)
+        if self.cursor_node is not None:
+            self.cursor_line = 0
```
darrenburns, line 368: I'm surprised that you need to explicitly set this reactive here, as I'd expect that to be done by `move_cursor`? If they're not being kept in sync automatically, is that a bug?

src/textual/widgets/_tree.py
```diff
@@ -930,6 +930,24 @@ def move_cursor(
...
+
+        Args:
+            line: The line number (negative indexes are offsets from the last line).
+            animate: Enable animation.
+
+        Raises:
+            IndexError: If the line doesn't exist.
+        """
+        if line < 0:
+            line = len(self._tree_lines) + line
```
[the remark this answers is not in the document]
willmcgugan: You're right. I think it is superfluous.

src/textual/widgets/_directory_tree.py
darrenburns: So if I, as a user, set `cursor_line`, it _doesn't_ move the cursor? I need to explicitly call `move_cursor`?

willmcgugan: It does move the cursor, but it doesn't change the scroll offset. Setting `cursor_line` will also update `cursor_node`.

darrenburns (approved):

PR #5062, opened 2024-09-26
Updating descriptions in BINDINGS

### LINKED ISSUE #5003: Help panel isn't helpful for bindings without descriptions
The summary of available keys in the help panel perhaps isn't helpful for bindings without a description. 

For example, consider the help panel for the `RadioSet` and `Select` examples in the docs:

image

image

darrenburns:
Fixes #5003, ensuring bindings all have descriptions so don't look weird in the help panel.

Also updates the casing of binding descriptions to be consistent - I went with capitalizing only the first word, since that seems most common and is what has been used recently.


willmcgugan (dismissed):
LGTM Think the failed tests are just snapshots.

willmcgugan (approved):

PR #5066, opened 2024-09-27
fix(help panel): prevent duplicated key displays

### LINKED ISSUE #5037: `Binding.key_display` and help panel interact poorly for comma-separated Binding definitions
If you declare a binding with commas e.g. `Binding("j,down,minus", ..., key_display="foo")`, the key panel will display that same "key display" for each key, resulting in repetition.

I think deduping the key displays might be enough here?

image

TomJGooding:
Fixes #5037

- [ ] Updated documentation


willmcgugan:
I don't think this will de-duplicate the key displays, except where we have identical keys to the same action.

i.e something like this:

```python
BINDINGS = [
    ("foo", "do_foo"),
    ("foo", "do_foo"),
    ("bar", "do_foo")
]
```

Should display `foo bar` for the key display.

willmcgugan:
Consider the following:

```python

from textual.widgets import Footer


class HelpPanelApp(App):
    BINDINGS = [
        Binding("a", "bell", "Ring the bell", key_display="bar"),
        Binding("b,e,l", "bell", "Ring the bell", key_display="foo"),
    ]

        yield Footer()


HelpPanelApp().run()
```

This will only display a single key, when there are multiple that would apply.

A contrived example perhaps, but even more likely now we have keymaps.

TomJGooding:
Thanks Will for clarifying. I see your point about this edge case, but why would you define multiple bindings like this? The key panel also wouldn't account for these bindings having different descriptions, so why should the key display be different?

TomJGooding:
I _think_ I now understand what you meant by this not de-duping the key displays, now that support for keymaps has been released.

```python
from textual.widgets import Footer, Label


class MyApp(App[None]):
    BINDINGS = [
        Binding(
            key="i,up",
            action="increment",
            description="Increment",
            id="app.increment",
            key_display="INCORRECT",
        ),
    ]

        yield Label("Check the footer and help panel")
        yield Footer()

    def on_mount(self) -> None:
        self.action_show_help_panel()
        self.set_keymap({"app.increment": "k,plus,j,l"})

    def get_key_display(self, binding: Binding) -> str:
        if binding.id == "app.increment":
            return "correct"
        return super().get_key_display(binding)


if __name__ == "__main__":
    app = MyApp()
    app.run()
```

TomJGooding:
@willmcgugan Friendly ping as currently the duplication of key displays is a bit of a wart in the help panel.

The updated `test_keymap_bindings_key_display` snapshot shows the difference after this fix. But please let me know if I might have still missed something.

src/textual/widgets/_key_panel.py
```diff
@@ -90,11 +90,14 @@ def render_description(binding: Binding) -> Text:
             get_key_display = self.app.get_key_display
             for multi_bindings in action_to_bindings.values():
                 binding, enabled, tooltip = multi_bindings[0]
-                key_display = " ".join(
-                    get_key_display(binding) for binding, _, _ in multi_bindings
-                )
+                unique_key_displays: list[str] = []
```
willmcgugan, line 93: I'd be tempted to go with this. Faster, but uses an idiom not everyone is familiar with.

```python
keys_display = " ".join(
    dict.fromkeys(
        get_key_display(binding) for binding, _, _ in multi_bindings
    )
)
```

willmcgugan (approved):
Thanks. Just a suggestion, which I'm happy to leave as "your call"!

src/textual/widgets/_key_panel.py
TomJGooding: Thanks, I wasn't aware of `dict.fromkeys`.

PR #5067, opened 2024-09-27
faster walk

willmcgugan:
An optimization to `walk_depth_first`. A little OTT perhaps, but this is a critical path.

### DIFF
--- a/src/textual/walk.py
+++ b/src/textual/walk.py
@@ -56,23 +56,31 @@ def walk_depth_first(
     Returns:
         An iterable of DOMNodes, or the type specified in ``filter_type``.
     """
-    from textual.dom import DOMNode
-
     stack: list[Iterator[DOMNode]] = [iter(root.children)]
     pop = stack.pop
     push = stack.append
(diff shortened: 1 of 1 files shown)


darrenburns (approved):
A bit OTT 😅

darrenburns (approved):

src/textual/walk.py
```diff
@@ -50,29 +50,37 @@ def walk_depth_first(
...
+                if children := node._nodes:
+                    push(iter(children))
+    else:
+        if with_root and isinstance(root, filter_type):
+            yield root
+        while stack:
+            if (node := next(stack[-1], None)) is None:
+                pop()
+            else:
+                if isinstance(node, filter_type):
```
darrenburns, line 80: I feel like writing 'if filter_type and isinstance(...)' here would have the same overall result and would only be a single check of a Boolean value, but would allow us to keep the single loop.

willmcgugan: It would still be a check per-node. Which we can avoid.

I did say it was OTT!

Issue #5064, opened 2024-09-27
Running self.run_worker crashes the app with no error message


mzebrak:
This was observed in: https://github.com/Textualize/textual/issues/5008 but I think it's a separate issue.

MRE: 
```python
 from __future__ import annotations

from typing import cast

from textual import on
from textual.app import App, ComposeResult
from textual.binding import Binding
from textual.screen import Screen
from textual.widgets import Button, Footer, Header, Label


class FirstScreen(Screen):
    def compose(self) -> ComposeResult:
        yield Header()
        yield Label("this is the first screen")
        yield Footer()


class SecondScreen(Screen):
    BINDINGS = [
        Binding("a", "finish", "Here it works"),  # key binding works just fine
    ]
    def compose(self) -> ComposeResult:
        yield Header()
        yield Label("This is the second screen")
        yield Button("Here it doesn't work")
        yield Footer()

    async def action_finish(self) -> None:
        await self._finish()

    @on(Button.Pressed)
    async def finish(self) -> None:
        # Running in worker causes app closure with no error
        # can't run directly also because of a deadlock  https://github.com/Textualize/textual/issues/5008
        self.run_worker(self._finish())

    async def _finish(self) -> None:
        app = cast(MyApp, self.app)
        await app.action_pop_until_first_screen()
        self.app.mount(Label("done"))  # assume something needs to be done after awaiting previous action


class MyApp(App):
    BINDINGS = [
        Binding("d", "push_second_screen", "Push second screen"),
    ]

    async def on_mount(self) -> None:
        await self.push_screen(FirstScreen())

    async def action_pop_until_first_screen(self) -> None:
        while not isinstance(self.screen, FirstScreen):
            await self.pop_screen()

    async def action_push_second_screen(self) -> None:
        await self.push_screen(SecondScreen())


MyApp().run()
```

---

I investigated a bit and found out that when the app closure happens, there is `asyncio.CancelledError` being raised from this line: https://github.com/Textualize/textual/blob/main/src/textual/worker.py#L339

After adding a wrapper to log cancellation inspired by https://stackoverflow.com/a/71356489 and some more logging looks like it comes from await_prune of  AwaitRemove (log from mentioned issue and its MRE, but I hope helps a bit):

```bash
 272 │ 2024-09-26 10:35:07.036 | ❌ ERROR    | textual.worker:_run:380 - In worker run of: action_pop_current_screen
[...]

Issue #5008, opened 2024-09-16
Awaiting pop_screen in @on causes app to freeze


mzebrak:
Version: 0.79.1

Adding `await` to `self.app.pop_screen` causes app to freeze when triggered via button, but works fine when triggered via key binding.

```python
from __future__ import annotations

from typing import cast

from textual import on
from textual.app import App, ComposeResult
from textual.binding import Binding
from textual.screen import Screen
from textual.widgets import Button, Footer, Header, Label


class FirstScreen(Screen):
    def compose(self) -> ComposeResult:
        yield Header()
        yield Label("this is the first screen")
        yield Footer()


class SecondScreen(Screen):
    def compose(self) -> ComposeResult:
        yield Header()
        yield Label("this is the second screen")
        yield Button("Pop back to first screen")
        yield Footer()

    @on(Button.Pressed)
    async def pressed(self) -> None:
        app = cast(MyApp, self.app)
        app.notify("Going back to first screen...")
        await app.action_pop_current_screen()


class MyApp(App):
    BINDINGS = [
        Binding("a", "pop_current_screen", "Pop screen"),  # key binding works just fine with await
        Binding("d", "push", "Push screen"),
    ]

    async def on_mount(self) -> None:
        await self.push_screen(FirstScreen())

    async def action_pop_current_screen(self) -> None:
        while not isinstance(self.screen, FirstScreen):
            await self.pop_screen()  # awaiting causes the bug, removing await fixes it in case of pressing the button

    async def action_push(self) -> None:
        await self.push_screen(SecondScreen())


MyApp().run()
```

willmcgugan:
By awaiting the `pop_screen`, you are waiting for all its messages to be processed. But your button.pressed message handler will never return, because it is waiting for itself to be popped. In other words, a deadlock.

I suspect the best thing for us to do is make it error / warning. To fix it, you can add `@work` to your action.

mzebrak:
Putting `@work` over action_pop_current_screen solves the case here. I thought it would be the same as 
`self.run_worker(app.action_pop_current_screen)` in `def pressed`, however the second one crashes without any error.

mzebrak:
This shows the crash I've been talking about in the previous comment. Am I doing something wrong there?

<details>

<summary> TEXTUAL CONSOLE OUTPUT </summary>

```python
[08:23:50] EVENT                                                                                                                                                                                        message_pump.py:740
[...]

PR #5069, opened 2024-09-27
Add pop_until_active

willmcgugan:
Removes temptation for devs to roll their own version of this, as there are gotchas in the implementation.

Fixes https://github.com/Textualize/textual/issues/5008
Fixes https://github.com/Textualize/textual/issues/5064


mzebrak:
Thanks for your work! But I'm wondering - to me it looks like it should close https://github.com/Textualize/textual/discussions/5071 but https://github.com/Textualize/textual/issues/5008 and https:/ /github.com/Textualize/textual/issues/5064 are about other things. Looks like it's related, but it can still happen.

tests/snapshot_tests/test_snapshots.py
```diff
@@ -2019,3 +2019,38 @@ def action_toggle_console(self) -> None:
...
+            await self.push_screen(BarScreen())
+
+        def action_make_base_active(self) -> None:
+            self.get_screen("base").pop_until_active()
+
+    app = PopApp()
+    # App will push three screens
+    # Pressing "b" will call pop_until_active, and pop two screens
+    # End result should be screen showing "BASE"
+    assert snap_compare(app, press=["b"])
```
darrenburns, line 2056: Please add docstrings to snapshot tests, as when they fail it will appear in the snapshot report and allows us to tell if they're correctly failing without needing to consult the code each time.

The docstring should explain what the expected result looks like.

```suggestion
def test_pop_until_active(snap_compare):
    """End result should be screen showing 'BASE'"""

    class BaseScreen(Screen):
        def compose(self) -> ComposeResult:
            yield Label("BASE")

    class FooScreen(Screen):
        def compose(self) -> ComposeResult:
            yield Label("Foo")

    class BarScreen(Screen):
        BINDINGS = [("b", "app.make_base_active")]

        def compose(self) -> ComposeResult:
            yield Label("Bar")

    class PopApp(App):
        SCREENS = {"base": BaseScreen}

        async def on_mount(self) -> None:
            # Push base
            await self.push_screen("base")
            # Push two screens
            await self.push_screen(FooScreen())
            await self.push_screen(BarScreen())

        def action_make_base_active(self) -> None:
            self.get_screen("base").pop_until_active()

    app = PopApp()
    # App will push three screens
    # Pressing "b" will call pop_until_active, and pop two screens
    # End result should be screen showing "BASE"
    assert snap_compare(app, press=["b"])
```

darrenburns (approved):
LGTM

willmcgugan:
@mzebrak If you feel there are other problems, please file an issue that describes the problem, and an example as short as possible that reproduces it. One issue per problem.

Issue #3935, opened 2023-12-29
`Widget.loading` can cause widgets to disappear


davep:
This might be a case of `Widget.loading` needing some more work, or [the documentation for the property needing expanding](https://textual.textualize.io/api/widget/#textual.widget.Widget.loading), but right now using the widget `loading` property can have results that are counter what you'd expect given the documentation:

> If set to True this widget will temporarily be replaced with a loading indicator.

This would suggest that a loading indicator would appear in a space the same dimension as the widget it's overlaying. With some widgets, that isn't the case. As a sample:

```python
from textual.app import App, ComposeResult
from textual.widgets import Static, Label, Footer, Button

class Fixed(Label):

    DEFAULT_CSS = """
    Fixed {
        width: 1fr;
        height: 10;
        background: teal;
    }
    """

class LabelLoadingTestApp(App[None]):

    BINDINGS = [
        ("space", "toggle", "Toggle the loading indicator"),
    ]

    def compose(self) -> ComposeResult:
        yield Label()
        yield Label("This is a test")
        yield Static()
        yield Static("This is a test")
        yield Footer()
        yield Button("")
        yield Button("Foo")
        yield Fixed("This is fixed")

    def action_toggle(self) -> None:
        for widget in self.query("Screen > *").results():
            widget.loading = not widget.loading

if __name__ == "__main__":
    LabelLoadingTestApp().run()
```

without loading, the screen looks like this:

![Screenshot 2023-12-29 at 16 30 30](https://github.com/Textualize/textual/assets/28237/09300cea-7022-41af-bfbb-dec315dcf786)

when you toggle on loading, it looks like this:

![Screenshot 2023-12-29 at 16 30 34](https://github.com/Textualize/textual/assets/28237/e94df411-d473-44c3-8e90-aa501cedd782)

TomJGooding:
For anyone who might stumble across this issue, you can workaround this by adding a min width/height to your widgets, for example:

```python
class LabelLoadingTestApp(App[None]):
    CSS = """
    Static, Label {
        min-height: 1;
        min-width: 9;
    }

    Button {
        min-height: 3;
    }
    """
```

willmcgugan:
The loading indicator works with containers, but when applied to simple widgets it transforms them into containers that don't render because they have no dimensions.

I don't think there is an easy fix for this one. We likely need some additional mechanism to change how a widget is rendered without changing its dimensions.

Worth doing, but not a small job.

Parking for now.

[...]

PR #5079, opened 2024-10-02
new loading indicator method

willmcgugan:
Replaces the current method of setting a loading indicator, with a newer mechanism that doesn't suffer from the problem that it turns widgets into containers.

This change does mean that you can't query for the loading indicator. I don't think anyone will care. It's more of a problem to accidentally query it. If someone complains, we can add an accessor of some kind.

As an aside, it wasn't too hard to add this mechanism where things can be rendered but aren't in the DOM. We may want to do this for tooltips and notifications.

Fixes https://github.com/Textualize/textual/issues/3935

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 
 - Fixed issue with screen not updating when auto_refresh was enabled https://github.com/Textualize/textual/pull/5063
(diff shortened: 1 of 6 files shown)


darrenburns:
> This change does mean that you can't query for the loading indicator. 

It's still selectable via CSS though right?

willmcgugan:
Yeah, you can still style it.

darrenburns (approved):
Looks good.

PR #5087, opened 2024-10-03
Themes and command palette improvements

darrenburns:
- Removes `App.dark`
- Adds `App.theme: Reactive[str]` which can be used to set the theme (registered themes only).
- Adds `variant` parameter to `Label` (success/primary/accent/error, etc.)
- Adds `InvalidThemeError`, raised when an unregistered theme is set.
- Adds `App.get_theme(theme: str) -> Theme`
- Adds `App.register_theme(theme: Theme)`
- Adds `App.unregister_theme(theme: Theme)`
- Adds `App.theme_changed_signal: Signal[Theme]`.
- Adds `App.available_themes -> dict[str, Theme]`
- Fix Select widget sometimes immediately re-opening when closed with a click of the mouse
- Allows `CommandPalette` to be instantiated and pushed by users, with arbitrary providers and a custom placeholder per-instance.
- Added `SimpleProvider` which allows a caller to pass commands _into_ a CommandPalette rather than having a provider be aware of where it might be invoked (experiment).
- Added `App.search`, a shorthand for the above that lets you quickly invoke a searchable menu with a list of commands, e.g. `self.app.search([("copy", self.copy), ("paste", self.paste)]` (an experiment - creating a separate provider each time feels like a big overhead and the providers need to query the active screen etc.)
- Adds `App.search_themes` which pulls up a command palette containing all registered themes. This is what is called when you select "Change theme" in the command palette.
- Defines some new builtin CSS vars `$hover`, `$cursor-highlight`, `$cursor-highlight-blurred`, `$hover-highlight`, `$hover-highlight-blurred`.
- Tries to make usage of variables a bit more consistent without changing the look and feel of Textual.
- Changes tree and data table cursors to match other cursors, and removes the "green" accent from the tree lines.
- Updates a bunch of widgets to use nested CSS instead of flat CSS as it seems easier to read and keep related rules together.
- I tried to make RadioSet look a bit more like SelectionList - haven't quit achieved what I intended there though.
- All widgets with a "highlight cursor" (the blue rectangle behind the highlighted option) now have a "blurred" version of that cursor, so when the widget loses focus, the cursor fades slightly.
- Some widgets had an underline indicating the "highlighted" option (e.g. RadioSet) - I replaced this with a cursor (just changing the styles). It's more coherent/consistent with other widgets, and also gains the "blurred" cursor effect when it loses focus. I'm not sure it looks better in isolation 🤷 
- Added consistent block cursor styling.

https://github.com/user-attachments/assets/732b331a-91d2-4137-beb7-9e4340939835

### Widget borders

I tried removing the widget borders and although I do think they look better in isolation, when they touch each other there's no way to distinguish between them:

image

I opted to keep the border but make it very subtle:

image

### Background tint

The majority of builtin widgets now get a subtle background tint when focused e.g:

image

This is particularly useful for users who prefer compact styles and remove borders and therefore don't receive the "focus styling" that gets applied to borders.


src/textual/design.py
```diff
@@ -190,6 +190,21 @@ def luminosity_range(spread) -> Iterable[tuple[str, float]]:
             colors["text-muted"] = "ansi_default"
             colors["text-disabled"] = "ansi_default"
 
+        # The cursor color for widgets such as OptionList, DataTable, etc.
+        colors["cursor"] = accent.hex
```
darrenburns, line 194: Note to self - this probably would need a more unique name is it could interact poorly with user defined variables?

willmcgugan (approved):

TomJGooding:
I spotted that the unselected crosses in the `SelectionList` aren't visible in the `textual-light` theme, was this intentional? I just thought I'd check as the similar widgets look okay.

image

darrenburns:
Thanks Tom, we're doing another pass now before releasing this. Tweaking scrollbars now, but didn't notice the issue with the SelectionList - will get that sorted too. Cheers!

darrenburns:
@TomJGooding cheers - working on SelectionList, RadioSet and ToggleButton now! Will try to make them consistent on light themes.

TomJGooding:
Sorry will leave this now in your capable hands! This is a fantastic feature and I can see the work that's gone into it, hopefully you don't think I'm just nitpicking.

darrenburns:
No no, not at all! We appreciate it. This was so much ground to cover and I definitely missed a few things. If you find anything else that you think looks off please let us know :)

zzzeek:
is there a document on how to control the wide range of focus-oriented color changes that seem to occur with this change?  trying to have a white background.   I can set it in a custom theme, or add it to CSS, but there still seem to be all kinds of "shaded" variations of white that are popping in when the focus is given to say a ListView.    Focus goes away, then my white comes back.  This all seems way more complicated and elaborate and I dont see how any of this is controlled; I tried setting all of "background", "surface", "panel", "boost" to the same color, and still, list panels are changing colors when focus is granted /lost.  The docs talk about "lighten", "darken", "muted", and I assume those colors are what's happening here but what if we dont want those ?

zzzeek:
example of listview color changing when focus is gained / lost from another window

image
image

willmcgugan:
Hi Michael. What you are seeing there is the focus style, which adds emphasis when the widget is focused. You can modify that, and pretty much anything display related with CSS.

You can see where the default CSS is set [here](https://github.com/Textualize/textual/blob/main/src/textual/widgets/_list_view.py#L51).

You could disable the background effect with something like this in your App's CSS:

```
ListView:focus { background-tint: transparent; }
```

If you don't want the UI to update when the terminal doesn't have focus, you could disable app focus entirely with something like the following:

```python
@on(events.AppFocus)
@on(events.AppBlur)
def on_app_focus(self, event):
    event.prevent_default()
```

Issue #5083, opened 2024-10-03
tooltips stop appearing when a widget is maximized


ByTheSeaL:
Before I maximize any widget tooltips work fine within the Header of my app:

![Screenshot 2024-10-03 130021](https://github.com/user-attachments/assets/7c11ca94-81c5-4758-b5d3-911c97c714c6)

And within my main Datatable (these are assigned dynamically by watching the Datatable's 'hover_coordinate'):

![Screenshot 2024-10-03 130033](https://github.com/user-attachments/assets/e53d3da7-f361-459d-883c-e0d80dc6c81c)

I have a link which maximizes the Container that contains the Datatable ("ALLOW_MAXIMIZE" is set on the link's Label and on the Container), and I make the app's header appear in maximized view still by setting:

```self.app.screen.ALLOW_IN_MAXIMIZED_VIEW = "MyHeader"```

When I'm in maximized view the tooltips just stop appearing:

![Screenshot 2024-10-03 130451](https://github.com/user-attachments/assets/b6ce6673-e837-4f16-abc9-3242a4c49a59)
![Screenshot 2024-10-03 130459](https://github.com/user-attachments/assets/55f78927-e38e-4afc-b71a-189e134879bb)

(if it's useful I can see that the watch method watching hover_coordinate is still being triggered in maximized view. There's no interruption in other events - for eg changing the value in an Input still works etc)

willmcgugan:
Add `.textual-system` to your ALLOW_IN_MAXIMIZED_VIEW. i.e. `self.app.screen.ALLOW_IN_MAXIMIZED_VIEW = ".textual-system,MyHeader"`

willmcgugan:
There will be an update in the next version, so this won't be neccesary.

PR #5088, opened 2024-10-03
Allow tooltips in maximize view

willmcgugan:
- Include `textual-system` widgets (such as tooltips) in maximize view by default.
- Allow default `ALLOW_IN_MAXIMIZED_VIEW` to be set on the App.

Fixes https://github.com/Textualize/textual/issues/5083


darrenburns (approved):
LGTM

PR #5094, opened 2024-10-05
Added hex chars to digits

willmcgugan:
- Adds A-F to digits
- Uses a thin variation by default


willmcgugan:
#executivedecision

PR #5109, opened 2024-10-11
fix(radioset): make radio set scrollable

### LINKED ISSUE #5100: RadioSets are not scrollable
All of the "similar" widgets (those which involve moving a cursor through a collection of options) scroll when you move the cursor through the children (SelectionList, OptionList, Tree, DataTable, and so on).

It seems like a RadioSet is not scrollable at all though. If you set a fixed height on it, you can move the cursor completely out of view, with no way of scrolling to make it visible.

TomJGooding:
Fixes #5100

- [ ] Updated documentation


darrenburns (approved):
Looks good - thanks!

PR #5135, opened 2024-10-17
fix(listview): update index after items removed

### LINKED ISSUE #5114: Calling `remove_items` or `pop` on a ListView does not update index or highlighting
It seems that calling `pop` (or `remove_items`, though that's not what I'm using for my use case) does not correctly update highlighting (presumably because calling `pop` does not update the index).

A small reproduction is as follows:

```python
from textual.widgets import ListItem, ListView, Footer


class TestApp(App):
    BINDINGS = [
        ('a', 'add', 'Adds a blank item'),
        ('d', 'delete_highlighted', 'Deletes the highlighted cell'),
    ]

    _listview: ListView

        self._listview = ListView(ListItem())

        yield Footer()
        yield self._listview

    def action_add(self):
        self._listview.append(ListItem())

    def action_delete_highlighted(self):
        index = self._listview.index
        self._listview.pop(index)


if __name__ == '__main__':
    TestApp().run()
```

Spawning a few items and trying to delete them does not set any highlighting, only manually updating the index will do that.

# Textual Diagnostics

## Versions

| Name    | Value  |
|---------|--------|
| Textual | 0.83.0 |
| Rich    | 13.9.2 |

## Python

| Name           | Value                                       |
|----------------|---------------------------------------------|
| Version        | 3.12.1                                      |
| Implementation | CPython                                     |
| Compiler       | Clang 17.0.6                                |
| Executable     | /home/king/Projects/private/.venv/bin/python |

## Operating System

| Name    | Value                                                |
|---------|------------------------------------------------------|
| System  | Linux                                                |
| Release | 6.10.11-amd64                                        |
| Version | #1 SMP PREEMPT_DYNAMIC Debian 6.10.11-1 (2024-09-22) |

## Terminal

## Rich Console options

| size           | width=223, height=76 |
| max_width      | 223                  |
| max_height     | 76                   |

TomJGooding:
Fixes #5114. 

- The `pop` and `remove_items`  methods now return  `AwaitComplete` rather than `AwaitRemove` to account for the index update
- Previously trying to `pop()` from an empty list would raise a `NoMatches` error, this will now raise an `IndexError` for consistency.

- [ ] Updated documentation


TomJGooding:
I've now also updated the `remove_items` method, so hopefully I was on the right track!

TomJGooding:
This now also reinstates `always_update` to the `index` reactive that was removed in recent changes. When items are removed, the `index` may not change but the highlighted item should be updated.

willmcgugan:
I'd recommend against the `always_update`. Since it calls watchers you can get unintended side effects, such as re-sending messages.

Better to factor out the specific functionality you need and call it directly.

TBH I regret adding `always_update`.  I know we've recommended it in the past, but this will rightly confuse anything not intimately familiar with reactives:

```python
self.index = self.index
```

src/textual/widgets/_list_view.py
```diff
@@ -286,13 +287,30 @@ def pop(self, index: Optional[int] = None) -> AwaitRemove:
...
-
-    def remove_items(self, indices: Iterable[int]) -> AwaitRemove:
+        if len(self) == 0:
+            raise IndexError("pop from empty list")
+
+        index = index if index is not None else -1
+        item_to_remove = self.query("ListItem")[index]
+        normalized_index = index if index >= 0 else index + len(self)
+
+        async def do_pop():
```
willmcgugan, line 297: Docstring and typing  por favor. Since its internal, and essentially private, the docstrings can be written for future Textual core-devs.

src/textual/widgets/_list_view.py
```diff
@@ -303,8 +321,28 @@ def remove_items(self, indices: Iterable[int]) -> AwaitRemove:
         """
         items = self.query("ListItem")
         items_to_remove = [items[index] for index in indices]
-        await_remove = self.remove_children(items_to_remove)
-        return await_remove
+        normalized_indices = set(
+            index if index >= 0 else index + len(self) for index in indices
+        )
+
+        async def do_remove_items():
```
willmcgugan, line 328: docstrings please

willmcgugan (changes requested):
Just a request for docstrings...

src/textual/widgets/_list_view.py
TomJGooding: I've added docstrings in 9314cde, but feel free to tweak!

willmcgugan:
Hi Tom, is this ready to go?

darrenburns:
@willmcgugan Was going to merge, but it's blocked until you approve.

willmcgugan (approved):

PR #5139, opened 2024-10-18
Pseudo update

willmcgugan:
Improves the mechanism to expose pseudo classes.

Previously a widget would return all its pseudo classes, even if Textual just needed to know if a single one was present. This made it expensive, and there was a reluctance to add more pseudo classes. This new mechanism can check for the presence of pseudo classes without generating all of them. It's faster, and we can add new ones without slowing down.

- Added `first-of-type`, `last-of-type`, `odd`, and `even` pseudo classes.
- Optimizations to caching stylesheet.apply


willmcgugan:
#executivedecision

@darrenburns Feel free to do a retroactive review for this one.

PR #5148, opened 2024-10-21
Mark `label`/`title` as `RenderableType`

thejcannon:
👋 Howdy!

I spent an embarrassingly long time figuring out how to get a `Static` to be the `title` of a `Collapsable`, so much so I forked this repo and prepared a change to support it (which, I might add, didn't work).

And then after string at my own code for long enough it hit me. The title is already a `Static`, so `label`/`title` must be a `RenderableType`.

### DIFF
--- a/src/textual/widgets/_collapsible.py
+++ b/src/textual/widgets/_collapsible.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+from rich.console import RenderableType
+
 from textual import events
 from textual.app import ComposeResult
 from textual.binding import Binding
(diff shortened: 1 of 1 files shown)


darrenburns:
Hey, this is marked as a draft - is it ready for review?

thejcannon:
Oh yeah! My bad. I usually mark draft and make she CI is passing before opening for review, but forgot to mark after CI.

willmcgugan (changes requested):
I think we would need to type the `label` reactive with this change.

Issue #5151, opened 2024-10-22
TextArea code editor freezes the app


Klavionik:
Hey again. :)

Mounting the TextArea code editor with any code other than the comment freezes my terminal. I tried several versions of Textual, I tried running it both on Mac and Linux and using Python 3.10 and 3.12. It freezes every time. I then tried an example from the docs and it freezes too.

An MRE:
```python
from textual.app import App, ComposeResult
from textual.widgets import TextArea

# OK_TEXT = "# OK"
FAIL_TEXT = "x = 1"


class CodeApp(App):
    def compose(self) -> ComposeResult:
        yield TextArea.code_editor(FAIL_TEXT, language="python")


if __name__ == "__main__":
    app = CodeApp()
    app.run()
```

If you replace usage of `FAIL_TEXT` with `OK_TEXT`, it doesn't freeze. Somehow it doesn't freeze with `FAIL_TEXT` too if you set `show_line_numbers=False`.

I tried investigating it a bit myself and I found out that the freeze happens because of the `while True` loop in `rich.segment.Segment._split_cells`: it keeps counting `pos` down and never stops. I'm not sure why, though. And I still don't know why the comment works alright even with the line numbers. 

# Textual Diagnostics

## Versions

| Name    | Value  |
|---------|--------|
| Textual | 0.83.0 |
| Rich    | 13.9.2 |

## Python

| Name           | Value                                                  |
|----------------|--------------------------------------------------------|
| Version        | 3.12.3                                                 |
| Implementation | CPython                                                |
| Compiler       | GCC 11.4.0                                             |
| Executable     | /home/jediroman/Desktop/textarea-mre/.venv/bin/python |

## Operating System

| Name    | Value                                        |
|---------|----------------------------------------------|
| System  | Linux                                        |
| Release | 5.15.0-122-generic                           |
| Version | #132-Ubuntu SMP Thu Aug 29 13:45:52 UTC 2024 |

## Terminal

| Name                 | Value                              |
|----------------------|------------------------------------|
| Terminal Application | WezTerm (20230712-072601-f4abf8fd) |
| TERM                 | xterm-256color                     |
| COLORTERM            | truecolor                          |
| FORCE_COLOR          | *Not set*                          |
| NO_COLOR             | *Not set*                          |

## Rich Console options

| Name           | Value                |
|----------------|----------------------|
| size           | width=105, height=25 |
| legacy_windows | False                |
| min_width      | 1                    |
| max_width      | 105                  |
[...]

PR #5154, opened 2024-10-22
fix infinite loop in cropping

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/5151

I think the issue was in `TextArea.render_line`.  The previous version of `Segment.split_cells` was buggy, but may have coicidentally done compensated for this bug.

`Segment.split_cells` should probably also use an assert. Nobody likes an infinite loop.

@darrenburns Could you take a quick look at this to make sure I'm on the right track? Can you suggest a test?


darrenburns:
I'm confused about why the tests weren't already catching this!

TomJGooding:
> I'm confused about why the tests weren't already catching this!

Probably because the tests weren't running against the latest rich version?

darrenburns:
I ran them locally and they all passed, even when the example failed when run manually 🤷

darrenburns (approved):

Issue #5163, opened 2024-10-23
Implement long list of scrollable and selectable widgets


willmcgugan:
### Discussed in https://github.com/Textualize/textual/discussions/5160

<div type='discussions-op-text'>

<sup>Originally posted by **anlar** October 23, 2024</sup>
Hello,

I'm working on application that has vertical list of items which should be scrollable by cursor. Each item is a widget itself and contains about 10 sub-widgets. Here is a screenshot for better understanding: https://github.com/anlar/tewi/blob/master/docs/images/tewi-screenshot-1.png .

Problem is that with large number of items (500+) list become unresponsive. Container scroll itself works fine, but if I want to use cursor (selecting item, changing its style and manually scroll to widget location) it is slow.

I have tried both ScrollableContainer and ListView (former works worse). All obvious (for me) optimization are made:
1. Widgets mounted/installed together.
2. There are no widgets re-composition during scroll.

Here is example app using scroll and list showing what app is doing:

```python
from textual.app import App, ComposeResult                                                                                                                                   
from textual.binding import Binding                                                                                                                                          
from textual.containers import Grid, ScrollableContainer, Horizontal, Container                                                                                              
from textual.widgets import Static, Label, ProgressBar, DataTable, ContentSwitcher, TabbedContent, TabPane, TextArea, \                                                      
        ListView, ListItem, Header, Footer                                                                                                                                   
                                                                                                                                                                             
class LabelItem(Label):                                                                                                                                                      
    w_next = None                                                                                                                                                            
    w_prev = None                                                                                                                                                            
                                                                                                                                                                             
class ScrollablePanel(ScrollableContainer):                                                                                                                                  
                                                                                                                                                                             
    BINDINGS = [                                                                                                                                                             
            Binding("k", "move_up", "Move up"),                                                                                                                              
            Binding("j", "move_down", "Move down"),                                                                                                                          
            ]                                                                                                                                                                
                                                                                                                                                                             
    selected_item = None                                                                                                                                                     
                                                                                                                                                                             
    def on_mount(self):                                                                                                                                                      
        widgets = []                                                                                                                                                         
                                                                                                                                                                             
        w_prev = None                                                                                                                                                        
        for x in range(1000):                                                                                                                                                
            item = LabelItem("test-scroll-" + str(x))                                                                                                                        
                                                                                                                                                                             
            if w_prev:                                                                                                                                                       
                w_prev.w_next = item                                                                                                                                         
[...]

PR #5164, opened 2024-10-23
Fix list view flicker

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/5163

- Added `immediate` bool to scroll methods. If you are making changes to the DOM (without adding new nodes), this will allow the scroll and those changes to be done in a single frame -- and reducing flicker.


darrenburns:
@willmcgugan LGTM - I swapped out the implementation of `find_next_enabled` to use the new function you wrote and it's significantly faster (two orders of magnitude) for scrolling in an OptionList of 1k items.

darrenburns (approved):

PR #5183, opened 2024-10-27
reduce jank

willmcgugan:
- Avoid a superfluous refresh when an app starts up. This will make apps with a lot of widgets display something faster. YMMV
- Refreshes the footer bindings early when switching screens, to avoid a frame of no bindings


willmcgugan:
#executivedecision

PR #5186, opened 2024-10-28
micro ops

willmcgugan:
I few micro-optimizations for layout.

These do reduce readability, but this is critical code in vertical layouts.

This could be done to the horizontal layouts -- but it does tend to be the vertical layouts which have a tonne of widgets.

- Replace `max()` with equivalent ternary operator, sometimes with a walrus operator
- Replaced a few `int(fraction)` with `fraction.__init__()`
- Pulled Fraction in to local scope

### DIFF
--- a/src/textual/_resolve.py
+++ b/src/textual/_resolve.py
@@ -104,8 +104,9 @@ def resolve_fraction_unit(
     Returns:
         The value of 1fr.
     """
+    F = Fraction
     if not remaining_space or not widget_styles:
-        return Fraction(1)
+        return F(1)
 
     initial_space = remaining_space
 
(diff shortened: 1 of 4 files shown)


src/textual/_resolve.py
```diff
@@ -104,8 +104,9 @@ def resolve_fraction_unit(
     Returns:
         The value of 1fr.
     """
+    F = Fraction
     if not remaining_space or not widget_styles:
-        return Fraction(1)
+        return F(1)
```
darrenburns, line 109: `_Fraction`?

darrenburns (approved):

Issue #5184, opened 2024-10-27
focus-within pseudo-class: caching-induced rendering issues 


xavierog:
## Context

My application uses Collapsible widgets that change the color of their left border when focused:

```css
AbstractStep:focus, AbstractStep:focus-within {
    border-left: thick $accent-lighten-3;
}
```

## Observed behaviour

Since Textual v0.84.0, more specifically since commit `9e0b04712 snapshot test`, focusing these widgets triggers rendering issues:

![moulti-diff-looking-for-textual-bug](https://github.com/user-attachments/assets/36de7846-3bed-4cb0-843f-60e2f4883731)

## Workaround

A workaround is to reinsert `"focus-within"` inside the `EXPENSIVE_PSEUDO_CLASSES` set in `css/stylesheet.py`, thus disabling pseudo-classes caching, which is presumably the root cause here.

### Code

```python
"""
MRE for https://github.com/Textualize/textual/issues/5184
Run with Textual >= 0.84.0 or post-9e0b04712 ("snapshot test").
Click the placeholders or just hit Tab to trigger rendering issues.
"""
from textual.app import App
from textual.containers import VerticalScroll
from textual.widgets import Placeholder, Collapsible

class FocusablePlaceHolder(Placeholder, can_focus=True):
	pass

class MRE(App):
	CSS = """
	Collapsible, PlaceHolder { border-left: blank; }
	FocusablePlaceHolder:focus, Collapsible:focus-within, Collapsible:focus { border-left: thick $accent-lighten-3; }
	Placeholder { height: 5; }
	"""
	def compose(self):
		with VerticalScroll():
			for i in range(10):
				with Collapsible(title=f'{i}', collapsed=False):
					yield FocusablePlaceHolder()

if __name__ == '__main__':
	app = MRE()
	app.run()
```

### Screenshot

On the left: expected behaviour with Textual 0.83.0: no rendering issues
On the right: observed behaviour with Textual 0.84.0: rendering issues

![textual-issue-5184-mre-screenshot](https://github.com/user-attachments/assets/9be9ec19-a88e-40f5-9e5f-1575a009b208)

PR #5190, opened 2024-10-28
fixed focus-within

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/5184

### DIFF
--- a/src/textual/css/stylesheet.py
+++ b/src/textual/css/stylesheet.py
@@ -433,13 +433,14 @@ class Stylesheet:
             if _check_selectors(selector_set.selectors, css_path_nodes):
                 yield selector_set.specificity
 
-    # pseudo classes which iterate over many nodes
-    # these have the potential to be slow, and shouldn't be used in a cache key
-    EXPENSIVE_PSEUDO_CLASSES = {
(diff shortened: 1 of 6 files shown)


src/textual/css/stylesheet.py
```diff
@@ -433,13 +433,14 @@ def _check_rule(
             if _check_selectors(selector_set.selectors, css_path_nodes):
                 yield selector_set.specificity
 
-    # pseudo classes which iterate over many nodes
-    # these have the potential to be slow, and shouldn't be used in a cache key
-    EXPENSIVE_PSEUDO_CLASSES = {
+    # pseudo classes which iterate over multiple nodes
+    # These shouldn't be used in a cache key
+    EXCLUDE_PSEUDO_CLASSES_FROM_CACHE = {
```
darrenburns, line 438: Does this need to be a classvar if it's never going to change?

willmcgugan: Good point. I think actually `Final` would be appropriate, and making it private. Since its not something the dev should know / care about / change.

darrenburns (approved):

PR #5197, opened 2024-10-29
New render protocol

willmcgugan:
This adds a new system for rendering content. Previously Textual uses Rich Renderables.

Now we have a `textual.visual.Visual` which is a replacement for the renderable protocol, which allows for layered alpha colors via the `Content` class.

This is backwards compatible in that Textual you can still return a renderable via the `render()` method, which will be wrapped in a `textual.visual.RichVisual` class.

This is a large change, but only the OptionList actively uses Visuals currently.

In the future, there will be a Visual markup language, like console markup, but supporting alpha transparency and CSS variables. For instance, you could do this:

```python
return "[$accent on blue %20]Hello, World![/]"
```

We don't have that yet, and there are still gaps in functionality, so this will remain undocumented for now.


src/textual/demo/projects.py
```diff
@@ -36,28 +36,28 @@ class ProjectInfo:
         "Darren Burns",
         "https://posting.sh/",
         """Posting is an HTTP client, not unlike Postman and Insomnia. As a TUI application, it can be used over SSH and enables efficient keyboard-centric workflows. """,
-        "4.7k",
+        "5.0k",
```
darrenburns, line 39: 😆 
```suggestion
        "6.0k",
```

src/textual/strip.py
```diff
@@ -157,6 +158,101 @@ def from_lines(
...
+
+        if horizontal == "left":
+            for strip in strips:
+                if strip.cell_length == width:
+                    yield strip
+                else:
+                    yield Strip(
+                        line_pad(strip._segments, 0, width - strip.cell_length, style),
+                        width,
+                    )
```
darrenburns, line 226: This code looks like there's an assumption that the cell length of the strip is less than the width. Is this accounted for elsewhere?

darrenburns (approved):

willmcgugan:
Will merge and look at that align issue

Issue #5215, opened 2024-11-06
DuplicateIds exception raised when removing a tab and adding a new tab


mathman1618:
I have an app in which multiple tabs of a `TabbedContent` widget may be created on startup.  These tabs can also be closed/removed by the user.  If two (or more) tabs are created at startup, closing a tab *other than the last* followed by creating a new tab crashes the app with an exception like the following:

`DuplicateIds: Tried to insert a widget with ID '--content-tab-tab-2', but a widget ContentTab(id='--content-tab-tab-2') already exists with that ID in this list of children. The children of a widget must have unique IDs.`

My code adds new tabs via `TabbedContent.add_pane` and removes them via `TabbedContent.remove_pane`.

I think the exception is due to how new `TabPane` ids are generated by `TabbedContent._set_id` here:
https://github.com/Textualize/textual/blob/09003dfbc0ff2879811cdc0b33dfdb26731c2cb1/src/textual/widgets/_tabbed_content.py#L427

# Textual Diagnostics

## Versions

| Name    | Value  |
|---------|--------|
| Textual | 0.85.2 |
| Rich    | 13.7.1 |

## Python

| Name           | Value                                              |
|----------------|----------------------------------------------------|
| Version        | 3.12.7                                             |
| Implementation | CPython                                            |
| Compiler       | GCC 11.4.0                                         |
| Executable     | /home/justin/.pyenv/versions/3.12.7/bin/python3.12 |

## Operating System

| Name    | Value                                                            |
|---------|------------------------------------------------------------------|
| System  | Linux                                                            |
| Release | 6.8.0-47-generic                                                 |
| Version | #&NoBreak;47~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Wed Oct  2 16:16:55 UTC 2 |

## Terminal

| Name                 | Value        |
|----------------------|--------------|
| Terminal Application | tmux (3.2a)  |
| TERM                 | xterm-direct |
| COLORTERM            | truecolor    |
| FORCE_COLOR          | *Not set*    |
| NO_COLOR             | *Not set*    |

## Rich Console options

| Name           | Value                |
|----------------|----------------------|
| size           | width=135, height=72 |
| legacy_windows | False                |
| min_width      | 1                    |
| max_width      | 135                  |
| is_terminal    | False                |
| encoding       | utf-8                |
| max_height     | 72                   |
| justify        | None                 |
| overflow       | None                 |
| no_wrap        | False                |
| highlight      | None                 |
[...]

PR #5216, opened 2024-11-07
fix(tabbed content): prevent duplicate ids error

TomJGooding:
Fixes https://github.com/Textualize/textual/issues/5215 where removing then adding a pane could crash with a `DuplicateIds` exception.

Rather than assigning the ID based on the *current* tab count, this adds a `_cumulative_tab_count` to ensure added panes have a unique ID.

- [ ] Updated documentation


TomJGooding:
This doesn't fix the highlighting issue when tabs are removed (https://github.com/Textualize/textual/issues/5215#issuecomment-2462555201), but that should probably be a separate issue/PR.

PR #5226, opened 2024-11-11
lazy reveal

willmcgugan:
Adds lazy.Reveal which displays widgets sequentially.

This can make large UIs feel more responsive, even though they are technically slower.

https://github.com/user-attachments/assets/27011595-f47f-49b3-aa4e-82cae8c55015

### DIFF
--- a/tests/test_lazy.py
+++ b/tests/test_lazy.py
@@ -45,8 +45,10 @@ async def test_lazy_reveal():
         assert not app.query_one("#baz").display
 
         # All children should be visible after a pause
-        await pilot.pause(3 / 60)
         await pilot.pause()
+        for n in range(3):
+            await pilot.pause(1 / 60)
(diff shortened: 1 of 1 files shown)


src/textual/containers.py
```diff
@@ -73,6 +73,40 @@ class ScrollableContainer(Widget, can_focus=True, inherit_bindings=False):
...
 
+    def __init__(
+        self,
+        *children: Widget,
+        name: str | None = None,
+        id: str | None = None,
+        classes: str | None = None,
+        disabled: bool = False,
+        can_focus: bool | None = None,
+        can_focus_children: bool | None = None,
```
darrenburns, line 84: Very useful.

willmcgugan: it was becoming a pattern!

darrenburns (approved):

PR #5233, opened 2024-11-13
fix(tabs): update highlighting when tab removed

### LINKED ISSUE #5218: `Tabs.remove_tab` does not update the highlighting
Originally posted in https://github.com/Textualize/textual/issues/5215#issuecomment-2462555201, but this is actually an issue with `Tabs` where `remove_tab` does not update the highlighting.

```python
from textual.widgets import Footer, Tab, Tabs


class TabsApp(App):
    BINDINGS = [("r", "remove_foo", "Remove foo")]

        yield Tabs(
            Tab("foo", id="foo"),
            Tab("bar", id="bar"),
            active="bar",
        )
        yield Footer()

    def action_remove_foo(self) -> None:
        tabs = self.query_one(Tabs)
        tabs.remove_tab("foo")


if __name__ == "__main__":
    app = TabsApp()
    app.run()
```

TomJGooding:
Fixes #5218.

Ideally this added call to `_highlight_active` should only happen if the removed tab is *before* the active tab, but I'm not sure there's a simple way of doing that?

- [ ] Updated documentation


darrenburns (approved):
LGTM 👍 Thanks!

TomJGooding:
Oops, sorry didn't spot that there's was a tweak to the Tab styling yesterday. Updating the snapshot now...

PR #5238, opened 2024-11-15
Demo polish

willmcgugan:
- Added many more widgets to demo.
- Some style tweaks. Made the background of the command palette $panel. I find it looks better.
- Added `TEXTUAL_THEME` env var.


willmcgugan:
#executivedecision

Issue #5249, opened 2024-11-17
Offset Widgets Popping In And Out While Scrolling


ddkasa:
I am working on a timeline widget and have an issue with disappearing/popping widgets if they have an offset set while scrolling off the original region/viewport. I am wondering if this is an intended behaviour or if I will have to change my approach to this functionality in order to prevent popping? Expected behaviour for me would be that the widgets stay in view without disappearing.

From my limited exploration of the textual source code I am thinking that this might be a compositor issue where the viewport logic doesn't properly validate the widgets location with the offset included.

I have added a MRE and video to demonstrate the issue, so please let me know if this needs more information.

## Video 

[offset_bug.webm](https://github.com/user-attachments/assets/e220b501-c1d6-4d10-bb7f-3ccdcacf57b8)

# Working Example
<details>
  <summary>main.py</summary>

```python
from __future__ import annotations

from dataclasses import dataclass

from textual import on
from textual.app import App, ComposeResult
from textual.containers import ScrollableContainer
from textual.events import MouseDown, MouseEvent, MouseMove, MouseUp, Resize
from textual.geometry import Offset
from textual.message import Message
from textual.widget import Widget
from textual.widgets import Label


class Adjustable(Widget):
    @dataclass
    class Resize(Message):
        adjustable: Adjustable
        delta: int

    def __init__(self, id: str) -> None:
        super().__init__(id=id)
        self.clicked: Offset | None = None
        self.styles.width = 10

    def compose(self) -> ComposeResult:
        yield Label("Test Content")

    async def on_mouse_down(self, event: MouseDown) -> None:
        if self.app.mouse_captured is None:
            self.capture_mouse()
        await self.is_focused(event)

    async def on_mouse_up(self, event: MouseUp) -> None:
        if self.app.mouse_captured:
            self.capture_mouse(False)
        await self.is_unfocused()

    async def is_focused(self, event: MouseEvent) -> None:
        self.clicked = event.offset
        self.moving = 2 < event.offset.x < (self.size.width - 2)

    async def is_unfocused(self) -> None:
        self.clicked = None

    async def on_mouse_move(self, event: MouseMove) -> None:
        if self.clicked is not None and event.button != 0:
            if hasattr(self, "moving") and self.moving:
                await self._move(event)
            else:
                await self._resize(event)

    async def _resize(self, event: MouseMove) -> None:
        delta = event.delta_x
        if self.clicked and self.clicked.x < 2:
            self.offset += Offset(delta, 0)
[...]

PR #5253, opened 2024-11-18
Fix scroll placements

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/5249

Bump to 0.86.2

A Widget's `offset` wasn't being taken in to account when scrolling. A widget with an offset could be hidden when it should be visible.

This bug won't show up on the snapshot tests, as they prompt a full refresh, and this issue occurs with the fast path scrolling.

### DIFF
--- a/tests/test_compositor.py
+++ b/tests/test_compositor.py
@@ -4,7 +4,9 @@ from textual.widgets import Static
 
 
 async def test_compositor_scroll_placements():
-    """Regression test for https://github.com/Textualize/textual/issues/5249"""
+    """Regression test for https://github.com/Textualize/textual/issues/5249
+    The Static should remain visible.
+    """
 
     class ScrollApp(App):
         CSS = """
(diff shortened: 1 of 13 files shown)


tests/test_compositor.py
```diff
@@ -0,0 +1,40 @@
+from textual.app import App, ComposeResult
+from textual.containers import Container
+from textual.widgets import Static
+
+
+async def test_compositor_scroll_placements():
+    """Regression test for https://github.com/Textualize/textual/issues/5249"""
```
darrenburns, line 7:
```suggestion
    """Regression test for https://github.com/Textualize/textual/issues/5249
    The Static should remain visible.
    """
```

darrenburns: The docstrings are visible in the snapshot report. If a snapshot fails, it's not always obvious whether what you're looking at is expected or not. A little comment like this can make it immediately obvious, and beats having to review and understand the code in the test.

willmcgugan: That's not a snapshot test, but I'm not apposed to the extra text.

darrenburns (approved):

PR #5257, opened 2024-11-19
Tutorial update

willmcgugan:
Updated the tutorial to use Digits display, improve styling with new themes, ands improve wording.


docs/tutorial.md
```diff
@@ -306,30 +314,31 @@ Button {
...
-We want our `Stopwatch` widget to have two states: a default state with a Start and Reset button; and a _started_ state with a Stop button. When a stopwatch is started it should also have a green background and bold text.
+We want our `Stopwatch` widget to have two states: a default state with a Start and Reset button; and a _started_ state with a Stop button. When a stopwatch is started it should also have a green background to indicate it is currently active.
 
 <div class="excalidraw">
 --8<-- "docs/images/css_stopwatch.excalidraw.svg"
 </div>
 
 
-We can accomplish this with a CSS _class_. Not to be confused with a Python class, a CSS class is like a tag you can add to a widget to modify its styles.
+We can accomplish this with a CSS _class_. Not to be confused with a Python class, a CSS class is like a tag you can add to a widget to modify its styles. A widget may have any number of CSS classes, which may be added and removes to change the style of your widget.
```
darrenburns, line 337: Typo and seems repetitive
```suggestion
We can accomplish this with a CSS _class_. Not to be confused with a Python class, a CSS class is like a tag you can add to a widget to modify its styles. A widget may have any number of CSS classes, which may be added and removed to change its appearance.
```

docs/tutorial.md
```diff
@@ -343,46 +352,60 @@ Some of the new styles have more than one selector separated by a space. The spa
 }
 ```
 
-The `.started` selector matches any widget with a `"started"` CSS class. While `#start` matches a child widget with an ID of `"start"`. So it matches the Start button only for Stopwatches in a started state.
+The `.started` selector matches any widget with a `"started"` CSS class. While `#start` matches a child widget with an ID of `"start"`.
+When combined with a space, the the selector will match the start button *only* if it is inside a container with a CSS class of "started".
```
darrenburns, line 356: I had to re-read this 2-3 times. Might want to reiterate that "the selector" is referring to `.started #start`.

```suggestion
When combined with a space, the selector will match the start button *only* if it is inside a container with a CSS class of "started".
```

docs/tutorial.md
```diff
@@ -343,46 +352,60 @@ Some of the new styles have more than one selector separated by a space. The spa
...
+The following code will start or stop the stopwatches in response to clicking a button:
 
 ```python title="stopwatch04.py" hl_lines="13-18"
 --8<-- "docs/examples/tutorial/stopwatch04.py"
 ```
 
-The `on_button_pressed` method is an *event handler*. Event handlers are methods called by Textual in response to an *event* such as a key press, mouse click, etc. Event handlers begin with `on_` followed by the name of the event they will handle. Hence `on_button_pressed` will handle the button pressed event.
+The `on_button_pressed` method is an *event handler*. Event handlers are methods called by Textual in response to an *event* such as a key press, mouse click, etc.
+Event handlers begin with `on_` followed by the name of the event they will handle.
+Hence `on_button_pressed` will handle the button pressed event.
```
darrenburns, line 377: This felt a bit abrupt relative to the detail elsewhere in the tutorial. I'd assumed after following the tutorial folks could write an event handler, but I don't think there's enough info here to actually learn about event handlers other than just copy-pasting. Not sure if just a link is needed to help direct people to the relevant page in the guide or something else.

docs/tutorial.md
```diff
@@ -343,46 +352,60 @@ Some of the new styles have more than one selector separated by a space. The spa
...
 
+When the button event handler adds or removes the `"started"` CSS class, Textual re-applies the CSS and updates the visuals.
+
+
 ## Reactive attributes
 
-A recurring theme in Textual is that you rarely need to explicitly update a widget. It is possible: you can call [refresh()][textual.widget.Widget.refresh] to display new data. However, Textual prefers to do this automatically via _reactive_ attributes.
+A recurring theme in Textual is that you rarely need to explicitly update a widget's visuals.
+It is possible: you can call [refresh()][textual.widget.Widget.refresh] to display new data.
+However, Textual prefers to do this automatically via _reactive_ attributes.
```
darrenburns, line 391: Feels like we're describing how to declare and giving an example of using "reactive" before defining what it actually _is_. It's tricky to follow examples that contain a new concept before you've been given at least a high-level/simplified definition.

Maybe add something like (probably improve the wording):

```suggestion
It is possible: you can call [refresh()][textual.widget.Widget.refresh] to display new data.
However, Textual prefers to do this automatically via _reactive_ attributes. A reactive attribute is an attribute that, when set, will automatically refresh the widget.
```

darrenburns (changes requested):
Looks good 👍 great usecase for Digits!

Just a few typos and 1 or 2 other things.

darrenburns (approved):

PR #5265, opened 2024-11-20
tweak to select

willmcgugan:
Trying to fix the issue where the select drop-down sometimes appears blank.

I think it may be because of race condition in OptionList, where the order of messages may cause a failure to render.

This *may* work around it. Rather than display:none I've used visibility:hidden. This is because visibility:hidden will still go through a layout procedure and get dimensions even if it can't be seen. 

Also prevented the expanded select from scrolling the container. It is not really needed as the drop-down has the constrain rules which will make it visible.

This is a workaround, until we can look at OptionList in more detail.

If anyone sees the blank Select after this has gone in, let me know!


willmcgugan:
This doesn't fix anything.

Issue #5266, opened 2024-11-20
Setting a widget to loading still allows interaction


bclemmey:
Looking at closed issues it looks like this has come up before: https://github.com/Textualize/textual/issues/3516 - This suggests it was possibly resolved, but I am definitely seeing it on the current version (`0.86.3`).

Widgets remain interactable when set to loading.

```python
from textual import on
from textual.app import App, ComposeResult
from textual.containers import Container
from textual.widgets import Button, Footer


class ButtonApp(App):
    BINDINGS = [
        ("c", "loading('c')", "Toggle Container"),
        ("b", "loading('b')", "Toggle Button")
    ]

    def compose(self) -> ComposeResult:
        yield Footer()
        with Container():
            yield Button("Kick me")

    @on(Button.Pressed)
    def pressed(self):
        self.notify("Ouch!")

    def action_loading(self, item: str):
        widget = self.query_one(Container if item == "c" else Button)
        widget.loading = not widget.loading


if __name__ == '__main__':
    ButtonApp().run()

```

In this example you can toggle loading on the button or container using bindings. Normally, the button can be interacted with by mouse or keyboard. If you toggle loading on for the container, the button can be interacted with still via keyboard, but not by mouse. 

If you toggle loading on the button, it can be clicked on with the mouse, but not pressed using the keyboard.

It feels to me that if the toggle loading feature is there, the widget should be fully disabled when loading is set. The current wording [of the docs](https://textual.textualize.io/api/widget/#textual.widget.Widget.loading) also supports this I believe:
"If set to True this widget will temporarily be replaced with a loading indicator." - Replaced suggests the widget is not present while loading is set to True.

PR #5267, opened 2024-11-20
disable loading widgets

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/5266


src/textual/widgets/_loading_indicator.py
```diff
@@ -56,6 +58,12 @@ def _on_mount(self, _: Mount) -> None:
         self._start_time = time()
         self.auto_refresh = 1 / 16
 
+    @on(InputEvent)
+    def on_input(self, event: InputEvent) -> None:
+        """Prevent all input events from bubbling, thus disabling widgets in a loading state."""
+        event.stop()
```
willmcgugan, line 64: @darrenburns Should we do app.bell here? Seems just the error condition one might use it for.

TomJGooding: I'm not Darren, but does Textual use the terminal bell in any other contexts? Personally I don't think trying to interact with a loading widget should make a fuss about it.

willmcgugan: We play the bell when clicking on disabled footer keys, and when input is restricted. So there is precedence.

I suspect bell makes more sense in disabled widgets.

willmcgugan:
#executivedecision

PR #5270, opened 2024-11-21
Select constrain

willmcgugan:
Fixed an issue where the constrain wasn't applied to the Select. This ensures that if you expand the select when it is at the bottom of the screen, it will always be visible. See snapshot if that isn't clear.

Also may have fixed the erratic invisible select issue. The OptionList updates caches on idle, these caches influence the height returned by `get_content_height`. So the auto height returned may not reflect the actual items, depending on when idle events are generated.

If Select issue is fixed (still not certain), I can at least postpone the OptionList refactor.


darrenburns (approved):
Tested it out in Posting and looks good.

Issue #5276, opened 2024-11-24
CommandPalette no longer working with markup


charles-001:
Hello,

After upgrading to `0.86.3` from `0.85.2`, `CommandPalette`  is no longer supporting markup as seen in my screenshot. I think it has to do with this [commit](https://github.com/Textualize/textual/commit/1a5fbb3b357395c6cb8d1ae377395f37bf86ee56#diff-831dc295b9160343bfc9c3409863065d235dbb76d1d1625a5ecbf2b6f4c430b8R961). Can we please get this back to its previous behavior? You can see how I build my `CommandPalette` [here](https://github.com/charles-001/dolphie/blob/353e65cc051d2b3ccb0ac8474b8abf3e9de33601/dolphie/app.py#L85C7-L85C20).

<img width="2476" alt="Screenshot_2024-11-22_at_5 58 04_PM" src="https://github.com/user-attachments/assets/f5bf748a-432c-43b0-a9ac-102ad5ee1993">

Thanks!

PR #5294, opened 2024-11-27
Allow console markup in commands

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/5276


tests/snapshot_tests/__snapshots__/test_snapshots/test_markup_command_list.svg
darrenburns, line 1: Shouldn't these be Monokai red and green given the default value of `App.ansi_theme_dark`?

willmcgugan: Better?

PR #5308, opened 2024-11-28
Correct a typo in Tree.add

davep:
Typo fix for `Tree.add`; looks like it was supposed to say "user" rather than "use".


willmcgugan (approved):
Just like old times

PR #5312, opened 2024-11-30
ci: run tests with Python 3.13

TomJGooding:
Update CI to run tests with Python 3.13. All jobs look like they ran successfully, but hopefully I haven't missed something!

- Excludes the `syntax` tests for Python 3.13. as the tree-sitter version used by Textual fails to build
- Update job for Python 3.12 to include the `syntax` tests, now that those dependencies can be installed


willmcgugan (approved):
Thanks!

PR #5323, opened 2024-12-02
no clear of footer

willmcgugan:
This remove a single frame where the footer is blank, that occurs when you push a modal.

Had to make a screen recording to see it, but it was there.

The fix was to fire the signal immediately and not on-idle, so there is no time to render a blank frame.


willmcgugan:
Replaced by https://github.com/Textualize/textual/pull/5330

Issue #5359, opened 2024-12-08
Alignment breaking auto size container


willmcgugan:
When alignment is applied to an auto-sizing container, it adjusts the size in strange ways.

If parent and child are the same size, as should be the case in this code, then alignment shouldn't have any effect.

```python
from textual.app import App, ComposeResult
from textual import containers
from textual.widgets import Button


class Sidebar(containers.Vertical):
    DEFAULT_CSS = """
    Sidebar {       
        width: auto;
        height: auto;
        background: blue;
        align-vertical: middle;

        #contents {
            width: auto;
            height: auto;
            background: red;
            border: white;
        }        
    }
    """

    def compose(self) -> ComposeResult:
        with containers.Vertical(id="contents"):
            yield Button("Start")
            yield Button("Stop")


class FloatSidebarApp(App):
    def compose(self) -> ComposeResult:
        yield Sidebar()


FloatSidebarApp().run()
```

PR #5360, opened 2024-12-08
Fix alignment in auto container

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/5359


willmcgugan:
#excecutivedecision

PR #5361, opened 2024-12-08
highlight top item

willmcgugan:
Selects the first item in the command palette when the search is updated.

Previously the first item would be selected if you hit return, but there was no visual indicator of this. Even knowing this, I would find myself hitting down then return, when return would suffice.

I chose the blurred style for the option list since the Input is where the action happens.

https://github.com/user-attachments/assets/1dba7ee7-4188-4ce6-8d2b-75b698b851e6


willmcgugan:
@darrenburns I think we have a conversation about this a while back. This feels way more intuitive to me now. WDYT?

willmcgugan:
#executivedecision

PR #5365, opened 2024-12-09
boost for first letter matches

willmcgugan:
Tweak to the command palette matcher to boost matches for the first word(s).

Consider the system commands. If I were to type "SS" I would get "**S**how Key**s** and Help Panel". Now I get "**S**ave **S**creenshot" because I'm matching the first letters.

Knowing this makes it easier to get a precise match.

### DIFF
--- a/src/textual/fuzzy.py
+++ b/src/textual/fuzzy.py
@@ -7,9 +7,8 @@ This class is used by the [command palette](/guide/command_palette) to match sea
 
 from __future__ import annotations
 
-from dataclasses import dataclass, field
 from re import IGNORECASE, compile, escape, finditer
-from typing import Iterable
+from typing import Iterable, NamedTuple
 
 import rich.repr
 from rich.style import Style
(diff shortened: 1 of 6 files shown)


src/textual/fuzzy.py
```diff
@@ -40,6 +40,10 @@ def __init__(
             ".*?".join(f"({escape(character)})" for character in query),
             flags=0 if case_sensitive else IGNORECASE,
         )
+        self._first_word_regex = compile(
```
darrenburns, line 43: Is a new `Matcher` created on each key press in the command palette？

willmcgugan: Yeah

willmcgugan:
Actually, I've nerd sniped myself. We can do so much better with the matching, but regexes aren't going to be enough.

willmcgugan:
Putting this on hold. I'm going to show great strength of character and not work on this, because there are higher priorities right now.

darrenburns:
@willmcgugan Do you want to merge this to get the improvement in or stick with what we have?

willmcgugan:
Going to hold off merging this. It's not quite a 100% improvement.

willmcgugan:
@darrenburns slept on it, and came up with this.

You may want to check it against Posting.

darrenburns:
Cool - I'll try it against Posting soon.

darrenburns:
Minor but these look the wrong way around:

image

Should the first `s` in `response` not match here, and mean that these results should be in the opposite order?

Also I'm not sure about underlining the spaces.

willmcgugan:
Wait, I see what you mean. It could match `res` and `sec`.

I may be able to tweak the heuristic. ATM it gives a boost to single words. It could boost based on the least number of groups.

willmcgugan:
@darrenburns tweaked the heuristic so it favors matches with fewer groups.

darrenburns:
I'm not sure that was the problem as in the original example wasn't there 4 matching groups in both results?

It still seems like the first `s` in `response` is being skipped here, and the results aren't what I'd expect.

image

willmcgugan:
My estimate for a reasonable maximum on the number of loops was too low.

Seems to work well with your example now. "exp res" puts response at the top.

darrenburns (approved):

PR #5368, opened 2024-12-09
docs(timer): remove outdated return header in stop method docstring

ddkasa:
Removed outdated timer stop method docstring return header.


willmcgugan (approved):

willmcgugan:
Thanks!

PR #5379, opened 2024-12-11
Don't apply Input "select on focus" behaviour when app is focused

darrenburns:
In the browser for example, an input element will only select its content if I tab to it. If I alt+tab to another app then alt+tab back, it doesn't select everything.

This updates Textual to match this behaviour.

### Added

- Added `from_app_focus` to `Focus` event to indicate if a widget is being focused because the app itself has regained focus or not https://github.com/Textualize/textual/pull/5379

### Changed

- The content of an `Input` will now only be automatically selected when the widget is focused by the user, not when the app itself has regained focus (similar to web browsers). https://github.com/Textualize/textual/pull/5379


willmcgugan:
I don't think so. At least not currently. But you can find the login in `App._watch_app_focus`

willmcgugan (approved):

PR #5381, opened 2024-12-11
Release post

willmcgugan:
Adds a blog post.

@darrenburns  I'm happy for this to be in my voice, so I'm not too concerned about grammar issues unless they are real clangers. Looking for more general opinions...


docs/blog/posts/release1.0.0.md
```diff
@@ -0,0 +1,276 @@
+---
+draft: true
+date: 2024-12-05
```
willmcgugan, line 3: Will update the date on release

docs/blog/posts/release1.0.0.md
```diff
@@ -0,0 +1,276 @@
...
+    (0, 1): [widget4, widget5, widget6],
+    (1, 1): [widget4, widget5, widget6],
+    (0, 2): [widget7, widget8],
+    (1, 2): [Widget7, widget8]
+}
+```
+
+This data is cacheable.
+If the widgets don't change their position or size such as when user is *scrolling*, then we can reuse the information.
+
```
darrenburns: Does this mean we can reuse the information while scrolling or we can't? I can read it both ways

src/textual/widget.py
```diff
@@ -3733,12 +3733,12 @@ def _scroll_update(self, virtual_size: Size) -> None:
                 height - self.scrollbar_size_horizontal
             )
             if self.vertical_scrollbar._repaint_required:
-                self.call_later(self.vertical_scrollbar.refresh)
+                self.vertical_scrollbar.refresh()
```
darrenburns, line 3736: There should be a policy that every time `call_later` is used in Textual we must write a comment explaining why 😄 

I've certainly had a number of issues in the past where the scrollbars wouldn't update (size/position of them) until they were interacted with. Not sure if the `call_later` was resolving that.

docs/blog/posts/release1.0.0.md
darrenburns: New version is great, cheers 👍

darrenburns (approved):

src/textual/widget.py
willmcgugan: Yeah, probably.

The refresh defers the update anyway, so I don't see what the call_later was doing. And I couldn't get anything to break without it.

PR #5405, opened 2024-12-17
Implement double/triple/quadruple click in TextArea

darrenburns:
Note to self: Fix the behaviour when you click on whitespace.

- [ ] Updated documentation

### DIFF
--- a/src/textual/widgets/_text_area.py
+++ b/src/textual/widgets/_text_area.py
@@ -1781,9 +1781,11 @@ TextArea {
         # Search for the start and end of a word from the current location.
         # If we want the search to be inclusive of the current location, so start
         # the search for the left boundary from one character to the right.
-        left = self.get_word_left_location(self.navigator.get_location_right(location))
-        right = self.get_word_right_location(location)
-        self.selection = Selection(left, right)
+        left = self.get_word_right_location(
+            self.get_word_left_location(self.navigator.get_location_right(location))
+        )
+        right = self.get_word_left_location(self.get_word_right_location(location))
+        self.selection = Selection(*sorted((left, right)))
         self.record_cursor_width()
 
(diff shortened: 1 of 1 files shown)


src/textual/widgets/_text_area.py
```diff
@@ -1612,6 +1642,17 @@ async def _on_hide(self, event: events.Hide) -> None:
         """Finalize the selection that has been made using the mouse when the widget is hidden."""
         self._end_mouse_selection()
 
+    async def on_click(self, event: events.Click) -> None:
```
willmcgugan, line 1645: Maybe this shouldn't use modulus. It doesn't look like vscode wraps the click chain in this way.

willmcgugan (approved):
Just one thing for your consideration

src/textual/widgets/_text_area.py
darrenburns: Experimented with this. For now wrapping works better for us for reasons™.

Without the modulus here, when we get to `chain == 5` and above, the selection goes back to being a cursor and it appears like it has stopped working.

I _think_ VSCode does chaining on the "mouse down" event. This would let us do a better implementation of this and remove a subtle flicker (between chain 2 and 3, the selection disappears briefly as we set the location in the mouse down handler, then the click handler fires later and selects the line). While we only have it on the click event, it's very difficult to replicate their behaviour exactly.

willmcgugan:
@darrenburns Was this good to go?

darrenburns:
I think I wasn't 100% satisfied with the behaviour without support for chaining on the mouse down event (there's a brief moment where the cursor jumps when going from a double to triple click, as it always moves on the mouse down event).

I think it may be good enough- but I recommend checking it out and having a play first.

PR #5408, opened 2024-12-18
Fix calculator example

BrorSebastianSjovald:
Just a NIT that the example of the calculator is wrong

- [x] Updated documentation


willmcgugan:
Why is it wrong?

BrorSebastianSjovald:
If it is PI, then it should be 3.141592

https://www.piday.org/million/

PR #5414, opened 2024-12-20
Added Blur message to Input widget

edward-jazzhands:
I realized working with the Input widget that its options for `validate_on` are "submitted", "changed", and "blur". However, there's only messages for Submitted and Changed. This makes it trickier to synchronize a function with the blur validator than it is with the other two validators, which I believe in this situation does not make sense.

Now I had figured a Blur event most likely does not exist for a reason (i.e. perhaps if one widget has it then why wouldn't all of them). But in this situation I can't help but feel like its existence is warranted in order to make the `validate_on=['blur']` parameter be able to be used in a manner that's identical to "submitted" and "changed" without needing to subclass the Input widget.

### DIFF
--- a/src/textual/widgets/_input.py
+++ b/src/textual/widgets/_input.py
@@ -284,6 +284,29 @@ class Input(ScrollView):
             """Alias for self.input."""
             return self.input
 
+    @dataclass
+    class Blur(Message):
+        """Posted when the widget is blurred (loses focus).
+
+        Can be handled using `on_input_blur` in a subclass of `Input` or in a parent
+        widget in the DOM.
+        """
+
+        input: Input
+        """The `Input` widget that was changed."""
+
+        value: str
+        """The value that the input was changed to."""
+
+        validation_result: ValidationResult | None = None
+        """The result of validating the value (formed by combining the results from each validator), or None
+            if validation was not performed (for example when no validators are specified in the `Input`s init)"""
+
+        @property
+        def control(self) -> Input:
+            """Alias for self.input."""
+            return self.input
+
     def __init__(
         self,
         value: str | None = None,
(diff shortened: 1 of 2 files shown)


davep:
> Now I had figured a Blur event most likely does not exist for a reason

The reason is most likely that the [`DescendentBlur`](https://textual.textualize.io/events/descendant_blur/) message is the thing to use. For example:

```python
from textual import on
from textual.events import DescendantBlur
from textual.widgets import Input, Log

class InputBlurTest(App[None]):

        with Horizontal():
            with Vertical():
                for n in range(10):
                    yield Input(placeholder=str(n), id=f"input-{n}")
            yield Log()

    @on(DescendantBlur, "Input")
    def log_input_blur(self, message: DescendantBlur) -> None:
        self.query_one(Log).write_line(f"This input lost focus: {message.control!r}")

if __name__ == "__main__":
    InputBlurTest().run()
```

TomJGooding:
@davep I think the problem being addressed here is that while...

> You can check whether the validation succeeded or failed inside an [Input.Changed](https://textual.textualize.io/widgets/input/#textual.widgets.Input.Changed) or [Input.Submitted](https://textual.textualize.io/widgets/input/#textual.widgets.Input.Submitted) handler by looking at the `validation_result` attribute on these events.

...you can't do the same when the `Input` is validated on blur.

davep:
I was offering a possible explanation as to why that particular message was missing; not suggesting that this PR shouldn't have been PRd.

src/textual/widgets/_input.py
```diff
@@ -284,6 +284,29 @@ def control(self) -> Input:
             """Alias for self.input."""
             return self.input
 
+    @dataclass
+    class Blur(Message):
```
willmcgugan, line 288: Let's call it `Blurred`, which is more consistent with existing messages, and avoids confusion with the existing `Blur` message.

willmcgugan (changes requested):
Good idea, just a request.

willmcgugan:
@davep I believe @edward-jazzhands was talking about a hypothetical `Input.Blur` message, and not the standard `Blur` event.

PR #5417, opened 2024-12-21
Bash alias command substitution window resize bug

lap1nou:

- [ ] Updated documentation

Greetings,

I'm making this PR following a conversation with @xavierog (thanks to him) on Discord. I asked him for help regarding my TUI application.

My application was not resized correctly, the display was bugged and for example the Footer was disappearing, the thing is that I was running the Python application from a bash alias using command substitution (`alias cmd='$(python3 tui-app.py)'`), @xavierog discovered that the Textual driver was using `shutil.get_terminal_size()` to get the window size, but in that specific case, the function is failling because the command is ran from a bash command substitution.

This PR aim to fix that specific (and probably uncommon) situation, I ran all tests and here is the results:
```
================================== short test summary info ==================================
FAILED tests/snapshot_tests/test_snapshots.py::test_text_area_themes[css] - AssertionError: assert False
FAILED tests/snapshot_tests/test_snapshots.py::test_text_area_themes[monokai] - AssertionError: assert False
FAILED tests/snapshot_tests/test_snapshots.py::test_text_area_themes[dracula] - AssertionError: assert False
FAILED tests/snapshot_tests/test_snapshots.py::test_text_area_themes[vscode_dark] - AssertionError: assert False
FAILED tests/snapshot_tests/test_snapshots.py::test_text_area_themes[github_light] - AssertionError: assert False
FAILED tests/snapshot_tests/test_snapshots.py::test_text_area_wrapping_and_folding - AssertionError: assert False
FAILED tests/text_area/test_languages.py::test_setting_unknown_language - Failed: DID NOT RAISE <class 'textual.widgets._text_area.LanguageDoesNotExist'>
FAILED tests/text_area/test_languages.py::test_register_language - ImportError: cannot import name 'get_language' from 'textual._tree_sitter' (/root/textua...
FAILED tests/text_area/test_languages.py::test_update_highlight_query - AssertionError: assert 0 > 0
======== 9 failed, 3104 passed, 2 skipped, 4 xfailed, 1 warning in 259.55s (0:04:19) ========
```

I had the same results with a fresh `Textual` clone so I don't think my changes broke any tests.

### DIFF
--- a/src/textual/drivers/linux_driver.py
+++ b/src/textual/drivers/linux_driver.py
@@ -102,19 +102,26 @@ class LinuxDriver(Driver):
         Returns:
             The size of the terminal as a tuple of (WIDTH, HEIGHT).
         """
-        width: int | None = 80
-        height: int | None = 25
-        import shutil
+        try:
+            width = int(os.environ['COLUMNS'])
+        except (KeyError, ValueError):
+            width: int | None = 0
 
         try:
-            width, height = shutil.get_terminal_size()
-        except (AttributeError, ValueError, OSError):
+            height = int(os.environ['LINES'])
+        except (KeyError, ValueError):
+            height: int | None = 0
+
+        if width <= 0 or height <= 0:
             try:
-                width, height = shutil.get_terminal_size()
+                width, height = os.get_terminal_size(self._file.fileno())
             except (AttributeError, ValueError, OSError):
-                pass
-        width = width or 80
-        height = height or 25
+                try:
+                    width, height = os.get_terminal_size(self._file.fileno())
+                except (AttributeError, ValueError, OSError):
+                    pass
+            width = width or 80
+            height = height or 25
         return width, height
 
     def _enable_mouse_support(self) -> None:
--- a/src/textual/drivers/linux_inline_driver.py
+++ b/src/textual/drivers/linux_inline_driver.py
@@ -61,19 +61,26 @@ class LinuxInlineDriver(Driver):
         Returns:
             The size of the terminal as a tuple of (WIDTH, HEIGHT).
         """
-        width: int | None = 80
-        height: int | None = 25
-        import shutil
+        try:
+            width = int(os.environ['COLUMNS'])
+        except (KeyError, ValueError):
+            width: int | None = 0
 
         try:
-            width, height = shutil.get_terminal_size()
-        except (AttributeError, ValueError, OSError):
+            height = int(os.environ['LINES'])
+        except (KeyError, ValueError):
+            height: int | None = 0
+
+        if width <= 0 or height <= 0:
             try:
-                width, height = shutil.get_terminal_size()
+                width, height = os.get_terminal_size(self._file.fileno())
             except (AttributeError, ValueError, OSError):
-                pass
-        width = width or 80
-        height = height or 25
+                try:
+                    width, height = os.get_terminal_size(self._file.fileno())
+                except (AttributeError, ValueError, OSError):
+                    pass
+            width = width or 80
+            height = height or 25
         return width, height
 
     def _enable_mouse_support(self) -> None:
--- a/src/textual/drivers/headless_driver.py
+++ b/src/textual/drivers/headless_driver.py
@@ -18,19 +18,26 @@ class HeadlessDriver(Driver):
     def _get_terminal_size(self) -> tuple[int, int]:
         if self._size is not None:
             return self._size
-        width: int | None = 80
-        height: int | None = 25
-        import shutil
+        try:
+            width = int(os.environ['COLUMNS'])
+        except (KeyError, ValueError):
+            width: int | None = 0
 
         try:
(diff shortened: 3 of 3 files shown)


src/textual/drivers/linux_inline_driver.py
xavierog, line 1: Good job thinking about the inline driver too!

Questions for @willmcgugan and @darrenburns:
1. Here, we have clear code duplication, calling for refactoring: either introducing an AbstractLinuxDriver class or turning that method into a standalone function with an `fd` argument that would go into e.g. `drivers/_tty.py` -- is this worth changing here?
2. What about HeadlessDriver? I think it could be stripped down to:
```python
    def _get_terminal_size(self) -> tuple[int, int]:
        return (80, 25) if self._size is None else self._size
```

src/textual/drivers/linux_driver.py
```diff
@@ -102,19 +102,26 @@ def _get_terminal_size(self) -> tuple[int, int]:
...
-            width, height = shutil.get_terminal_size()
-        except (AttributeError, ValueError, OSError):
+            height = int(os.environ['LINES'])
+        except (KeyError, ValueError):
+            height: int | None = 0
+
+        if width <= 0 or height <= 0:
             try:
-                width, height = shutil.get_terminal_size()
+                width, height = os.get_terminal_size(self._file.fileno())
```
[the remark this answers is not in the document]
lap1nou: Doesn't this condition prevent that ? https://github.com/lap1nou/textual/blob/91d02b18ec3dbf1ab66738933d4b7bd83986f592/src/textual/drivers/linux_driver.py#L115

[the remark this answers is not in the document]
xavierog: Consider the case when `COLUMNS` is set to e.g. 120 and `LINES` is not set: width is 120, height is 0, this condition is True, both width and height end up with the actual terminal size and `COLUMNS` is thus ignored.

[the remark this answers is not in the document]
lap1nou: Got it @xavierog, this should do the trick

src/textual/drivers/linux_driver.py
```diff
@@ -102,19 +102,29 @@ def _get_terminal_size(self) -> tuple[int, int]:
...
-        height = height or 25
+                try:
+                    size = os.get_terminal_size(self._file.fileno())
+                except (AttributeError, ValueError, OSError):
+                    pass
+
+            if columns <= 0: 
+                width = size.columns or 80
+            if lines <= 0:
+                height = size.lines or 25
```
xavierog, line 127: `columns` and `lines`? I think you meant `width` and `height`.

xavierog: Additionally, if we hit `pass` at runtime, size may be unset, and trying to reach `size.columns` or `size.lines` would yield a NameError.

xavierog: I went digging inside Rich and Textual history: that double try-except we were trying to preserve here was initially meant to try `os.get_terminal_size(sys.stdin.fileno())` then `os.get_terminal_size(sys.stdout.fileno())`... but that no longer makes sense here:
- stdout may already be a pipe
- stdin is currently used to read terminal events but it could become a pipe in the future too
- stderr (i.e. `self._file`) is where Textual renders applications

Otherly put: we can **discard that double try-except**, which should make things easier.

src/textual/drivers/linux_driver.py
```diff
@@ -102,19 +102,28 @@ def _get_terminal_size(self) -> tuple[int, int]:
...
                 pass
-        width = width or 80
-        height = height or 25
+
+            try:
+                width = size.columns
+                height = size.lines
+            except NameError:
+                width = 80
+                height = 25
```
xavierog, line 126: We lost the width/height separation again -- tricky, eh?
May I suggest this?
```python
if width <= 0 or height <= 0:
    # LINES and COLUMNS do not fully define terminal size: query the tty to get its actual size:
    try:
        size = os.get_terminal_size(self._file.fileno())
        if width <= 0:
            width = size.columns
        if height <= 0:
            height = size.lines
    except (AttributeError, ValueError, OSError):
        pass
    
    # Default to the de facto standard terminal size:
    if width <= 0:
        width = 80
    if height <= 0:
        height = 25

return width, height
```

xavierog:
@lap1nou: Looks good to me.

Questions for @willmcgugan and @darrenburns:
1. What do you think re: the issue itself and the suggested fix?
1. Beyond that fix: here, we have clear code duplication, calling for refactoring: either introducing an AbstractLinuxDriver class or turning that method into a standalone function with an `fd` argument that would go into e.g. `drivers/_tty.py` -- is this worth changing here?
2. What about HeadlessDriver? I think it could be stripped down to:
```python
    def _get_terminal_size(self) -> tuple[int, int]:
        return (80, 25) if self._size is None else self._size
```

PR #5439, opened 2024-12-27
Tweak command palette tooltip

davep:
Simply makes the tooltip for the binding for the command palette a wee bit more friendly in how it reads.


willmcgugan:
https://github.com/Textualize/textual/commit/160f7784146e60759b3175e5c6adc8d385677610

PR #5444, opened 2024-12-29
fix grid sizing

willmcgugan:
Fix an issue with grid. If a grid cell contained a label, or other widget with text, the optimal width calculation could be larger than a grid cell width. Since grid cells can't scroll, the actual width was restricted -- which would leave an erroneous value for height.

The fix was to add a flag to constrain the width to the cell size.

I'm surprised that no tests broke here. I suspect because we always put content inside a container, which would have negated this issue.


willmcgugan:
@darrenburns Tagging you out of habit. Let me know if you'd rather I didn't.

PR #5462, opened 2025-01-08
fix(scrollbar): fix scrollbar background opacity

### LINKED ISSUE #5458: Scrollbars appear to ignore the background colour percentage
Each of the scrollbar styles say that they take a percentage that is the opacity to use. However, it appears that this value gets ignored. With this code:

```python

class ScrollBarPercentageApp(App[None]):

    CSS = """
    Screen {
        layout: horizontal;
    }

    VerticalScroll {
        scrollbar-size-vertical: 10;

        &#no-pcent {
            background: $surface;
            scrollbar-background: $surface;
            scrollbar-background-active: $surface;
            scrollbar-background-hover: $surface;
        }

        &#pcent {
            background: $surface 10%;
            scrollbar-background: $surface 10%;
            scrollbar-background-active: $surface 10%;
            scrollbar-background-hover: $surface 10%;
        }
    }
    """

        with VerticalScroll(id="no-pcent"):
            yield Static(
                "\n".join(
                    f"This is some text {n}" for n in range(100)
                )
            )
        with VerticalScroll(id="pcent"):
            yield Static(
                "\n".join(
                    f"This is some text {n}" for n in range(100)
                )
            )

if __name__ == "__main__":
    ScrollBarPercentageApp().run()
```

the result is this:

Screenshot 2025-01-06 at 11 30 51

Note that on the left the scrollbar background and the container are set to `$surface` and they match; on the right they're set to `$surface 10%` and they don't match, which the scrollbar background appearing to still be just `$surface`.

TomJGooding:
Fixes #5458.

This assumes that when the scrollbar background has opacity, it should blend with the 'base' background and not the background of its widget.

- [ ] Updated documentation

### DIFF
--- a/src/textual/scrollbar.py
+++ b/src/textual/scrollbar.py
@@ -286,7 +286,7 @@ class ScrollBar(Widget):
         else:
             background = styles.scrollbar_background
             color = styles.scrollbar_color
-        if background.a > 0:
+        if background.a < 1:
             base_background, _ = self.parent._opacity_background_colors
             background = base_background + background
         color = background + color
--- a/src/textual/scrollbar.py
+++ b/src/textual/scrollbar.py
@@ -286,6 +286,8 @@ class ScrollBar(Widget):
         else:
             background = styles.scrollbar_background
             color = styles.scrollbar_color
+        base_background, _ = self.parent._opacity_background_colors
+        background = base_background + background
         color = background + color
         scrollbar_style = Style.from_color(color.rich_color, background.rich_color)
         if self.screen.styles.scrollbar_color.a == 0:
--- a/src/textual/scrollbar.py
+++ b/src/textual/scrollbar.py
@@ -286,8 +286,9 @@ class ScrollBar(Widget):
         else:
             background = styles.scrollbar_background
             color = styles.scrollbar_color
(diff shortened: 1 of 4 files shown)


willmcgugan:
Yeah, this looks like the right fix. Could you humor me and make a small optimization. If the scrollbar background has zero alpha, then don't call the slightly expensive `_opacity_background_colors` method...

TomJGooding:
@willmcgugan Glad I was on the right track but good point about the optimisation. Updated in a35d48a.

willmcgugan (approved):
lgtm

src/textual/scrollbar.py
```diff
@@ -286,6 +286,9 @@ def render(self) -> RenderableType:
         else:
             background = styles.scrollbar_background
             color = styles.scrollbar_color
+        if background.a > 0:
```
willmcgugan, line 289: Should this not be `background.a < 1` ?

TomJGooding: Sorry yes, more haste less speed!

willmcgugan:
Gracias

davep:
Thanks @TomJGooding

PR #5467, opened 2025-01-09
Log selection

willmcgugan:
Add selection to Log widget


willmcgugan:
Enables arbitrary selection in the Log widget

PR #5478, opened 2025-01-16
fix(option list): fix render cache option indexing

### LINKED ISSUE #5431: Updating an unfocused `OptionList` that contains seperater lines can result in a garbled display (since 0.86.0)
Take this code:

```python
from textual import on
from textual.widgets import OptionList, Button

class FocusTest(App[None]):

    CSS = """
    OptionList {
        height: 1fr;
    }
    """

    counter: var[int] = var(0)

        with Horizontal():
            yield Button("Add")
            yield OptionList()

    @on(Button.Pressed)
    def add_more_stuff(self) -> None:
        self.counter += 1
            (f"This is option {self.counter}" if self.counter % 2 else None)
        )

if __name__ == "__main__":
    FocusTest().run()
```

if you use the `Add` button, and keep focus on the `Add` button, the display within the `OptionList` starts to make no sense; only when it receives focus does it show correctly:

https://github.com/user-attachments/assets/15c4abed-2335-477a-9122-a38b2cfb0d8e

Going back through recent Textual releases, this effect first appears in [v0.86](https://github.com/Textualize/textual/releases/tag/v0.86.0).

TomJGooding:
Fixes #5431

- [ ] Updated documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Fixed special case with calculating the height of a container where all children have dynamic heights https://github.com/Textualize/textual/pull/5463
 - Fixed scrollbars ignoring background opacity https://github.com/Textualize/textual/issues/5458
 - Fixed `Header` icon showing command palette tooltip when disabled https://github.com/Textualize/textual/pull/5427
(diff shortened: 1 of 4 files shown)


TomJGooding:
I'm _fairly_ confident this is the correct fix (and all tests pass!), but there's a few moving parts in the `OptionList` rendering so please let me know if I've missed something!

TomJGooding:
Closed in favour of #5510.

PR #5494, opened 2025-01-24
Fix MaskedInput selection replace with invalid value

### LINKED ISSUE #5493: MaskedInput ValueError Value does not match template

Yes, no issue mentioned problems with MaskedInput validation.

Have you checked against the most recent version of Textual? https://pypi.org/search/?q=textual

Yes

1. Run the app
  ```sh
  uv run --script masked_input_value_error.py
  ```
2. Press the key `e` or any other non digit.
3. Watch the app crash with a ValueError.

`masked_input_value_error.py`
```python
# /// script
# requires-python = ">=3.13"
# dependencies = [
#     "textual==1.0.0",
# ]
# ///


class BugApp(App[None]):
        yield MaskedInput("9", "9")


app = BugApp()
app.run()
```

```
textual diagnose
```

If you don't have the `textual` command on your path, you may have forgotten to install the `textual-dev` package.

https://github.com/user-attachments/assets/a5bdb6e1-89c8-44a4-b1f2-c67cc7d1c843

Image

Image

# Textual Diagnostics

## Versions

| Name    | Value  |
|---------|--------|
| Textual | 1.0.0  |
| Rich    | 13.9.4 |

## Python

| Version        | 3.13.0                                                               |
| Compiler       | Clang 18.1.8                                                         |
| Executable     | /home/jonathan/.cache/uv/archive-v0/uhCuU97F8DMiIyfN6OiqV/bin/python |

## Operating System

| Release | 6.12.9-100.fc40.x86_64                              |
| Version | #1 SMP PREEMPT_DYNAMIC Thu Jan  9 16:04:00 UTC 2025 |

## Terminal

| Terminal Application | vscode (1.96.4) |

## Rich Console options

| size           | width=142, height=23 |
| max_width      | 142                  |
| max_height     | 23                   |

JonathanPlasse:

- Fix #5493

This PR implements the `MaskedInput.replace()` method by using the already existing `MaskedInput.insert_text_at_cursor()` which replace the characters following the cursor position.

- [ ] Updated documentation


TomJGooding:
The`MaskedInput` will select all text on focus by default, but this isn't obvious until you [add the highlighting](https://github.com/Textualize/textual/pull/5496).

Wouldn't this PR only replace a single character, rather than the whole selection as you would expect?

JonathanPlasse:
You are right

JonathanPlasse:
> Wouldn't this PR only replace a single character, rather than the whole selection as you would expect?
Fixed

TomJGooding:
Unfortunately, I think fixing the `MaskedInput` to handle text selection is going to be more complicated.

For example, try running [the credit card example in the docs](https://textual.textualize.io/widgets/masked_input/#example) with the changes in this PR. After entering some numbers to complete the input, try selecting some text using  the mouse (it might help [adding the highlighting](https://github.com/Textualize/textual/pull/5496)) then enter a number. This will crash if the selection included a separator as the value does not match the template.

JonathanPlasse:
This is now fixed.
The replace now behave like this:
- if the replace text is the same length as the selection (without separator), it replace exactly the selection
- if the replace text is greater, the replace text will be cut to fit within the selection
- if the replace text is lesser, the remaining selection text will be deleted
- if the selection starts on a separator, the replace text start after it

willmcgugan:
Hi @JonathanPlasse Thanks for this. Could we please have a test?

JonathanPlasse:
I am sorry, I do not have the bandwidth to add tests at the moment.

PR #5496, opened 2025-01-25
fix(masked input): highlight selected text

### LINKED ISSUE #5495: `MaskedInput` does not highlight the selected text
Following on from #5493, I think this also reveals another issue with the `MaskedInput`:

```python
from textual.widgets import Input, Label, MaskedInput


class InputSelectionApp(App):
    CSS = """
    Label {
        margin: 1;
    }
    """

        yield Label("Input text is now selected on focus by default.")

        yield Label(
            "This is obvious to the user for the Input widget "
            "where the selection is highlighted:"
        )
        yield Input("text")

        yield Label(
            "But not the MaskedInput, "
            "which is confusing when typing will overwrite the previous text:"
        )
        yield MaskedInput(template="AAAAAAAA;-", value="text")


if __name__ == "__main__":
    app = InputSelectionApp()
    app.run()
```

TomJGooding:
Fixes #5495.

- [ ] Updated documentation

### DIFF
--- a/src/textual/widgets/_masked_input.py
+++ b/src/textual/widgets/_masked_input.py
@@ -569,12 +569,19 @@ class MaskedInput(Input, can_focus=True):
             if char == " ":
                 result.stylize(style, index, index + 1)
 
-        if self._cursor_visible and self.has_focus:
-            if self._cursor_at_end:
-                result.pad_right(1)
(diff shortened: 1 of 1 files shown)


TomJGooding:
Superseded by #6287

Issue #5498, opened 2025-01-29
Large `OptionList` widgets are slow to load (since 0.86.0)


TomJGooding:
I've noticed that `OptionList` widgets with a large number of options take significantly longer to load since Textual 0.86.0.

The example below is an extreme scenario, but it demonstrates the difference in loading times.

**Textual 0.85.2:**
Loading OptionList with 10,000 options took 4.02s

**Textual 0.86.0:**
Loading OptionList with 10,000 options took 9.17s

**Textual 1.0.0:**
Loading OptionList with 10,000 options took 13.28s

```python
from time import perf_counter

from textual import __version__
from textual.app import App, ComposeResult
from textual.widgets import OptionList

TOTAL_OPTIONS = 10_000


class ExampleApp(App):
    def compose(self) -> ComposeResult:
        content = [f"Option {i}/{TOTAL_OPTIONS}" for i in range(TOTAL_OPTIONS)]
        yield OptionList(*content)

    def on_ready(self) -> None:
        self.call_after_refresh(self.exit)


if __name__ == "__main__":
    start = perf_counter()

    app = ExampleApp()
    app.run()

    elapsed = perf_counter() - start
    print(f"**Textual {__version__}:**")
    print(f"Loading OptionList with {TOTAL_OPTIONS:,} options took {elapsed:.2f}s")

```

willmcgugan:
Previous iteration of the OptionList were lazy in calculating dimensions, which cause all sorts of flakiness. Now it does more work up-front. I suspect the timings look better in older versions because it is exiting before that lazy work is done.

It takes 6.45s on my machine currently with a feature branch that should speed up rendering.

The OptionList is also due for a re-write.

Closing this for now, will revisit after these changes are in.

TomJGooding:
> It takes 6.45s on my machine currently with a feature branch that should speed up rendering.

@willmcgugan Can I ask which feature branch so I can give it a go? That's quite a difference, though I suspect my old Thinkpad won’t keep up with your machine!

willmcgugan:
It's in main now

TomJGooding:
Ah, that's what I feared. It takes ~16s on main on my machine, so actually 3s longer than previously.

TomJGooding:
Thanks for reopening this issue - that's quite an impressive difference on the new option list branch!

willmcgugan:
It's a wee bit faster now!

TomJGooding:
Closed by #5510.

Issue #5489, opened 2025-01-20
An empty `OptionList` with `width: auto` causes a crash in `get_content_width`


davep:
This code:

```python
from textual.app import App, ComposeResult
from textual.widgets import OptionList

class OptionListAutoCrash(App[None]):

    CSS = """
    OptionList {
        width: auto;
    }
    """

    def compose(self) -> ComposeResult:
        yield OptionList()

if __name__ == "__main__":
    OptionListAutoCrash().run()
```

results in this crash:

```
│ /Users/davep/temp/x/.venv/lib/python3.13/site-packages/textual/widgets/_option_list.py:396 in get_content_width                                       │
│                                                                                                                                                       │
│    393 │   │   """Get maximum width of options."""                                                                                                    │
│    394 │   │   console = self.app.console                                                                                                             │
│    395 │   │   options = console.options                                                                                                              │
│ ❱  396 │   │   return max(                                                                                                                            │
│    397 │   │   │   Measurement.get(console, options, option.prompt).maximum                                                                           │
│    398 │   │   │   for option in self._options                                                                                                        │
│    399 │   │   )                                                                                                                                      │
│                                                                                                                                                       │
│ ╭─────────────────────────── locals ────────────────────────────╮                                                                                     │
│ │   console = <console width=153 ColorSystem.TRUECOLOR>         │                                                                                     │
│ │ container = Size(width=149, height=66)                        │                                                                                     │
│ │   options = ConsoleOptions(                                   │                                                                                     │
│ │             │   size=ConsoleDimensions(                       │                                                                                     │
│ │             │   │   width=153,                                │                                                                                     │
│ │             │   │   height=68                                 │                                                                                     │
│ │             │   ),                                            │                                                                                     │
│ │             │   legacy_windows=False,                         │                                                                                     │
│ │             │   min_width=1,                                  │                                                                                     │
│ │             │   max_width=153,                                │                                                                                     │
│ │             │   is_terminal=True,                             │                                                                                     │
│ │             │   encoding='utf-8',                             │                                                                                     │
│ │             │   max_height=68,                                │                                                                                     │
│ │             │   justify=None,                                 │                                                                                     │
│ │             │   overflow=None,                                │                                                                                     │
│ │             │   no_wrap=False,                                │                                                                                     │
│ │             │   highlight=None,                               │                                                                                     │
│ │             │   markup=None,                                  │                                                                                     │
│ │             │   height=None                                   │                                                                                     │
[...]

Issue #5488, opened 2025-01-20
When an `OptionList` has a `transparent` background option colours can get out of sync (since 0.86.0)


davep:
Given this code:

```python
from textual.app import App, ComposeResult
from textual.containers import Vertical
from textual.widgets import OptionList, Input

class Panel(Vertical, can_focus=True):
    pass

class FocusWithinTransparentApp(App[None]):

    CSS = """
    Screen {
        layout: horizontal;
    }

    Input {
        width: 1fr;
        height: 1fr;
    }

    Panel {
        padding: 5 10;
        background: red;
        &:focus, &:focus-within {
            background: yellow;
        }

        OptionList, OptionList:focus {
            height: 1fr;
            background: transparent;
        }
    }
    """

    def compose(self) -> ComposeResult:
        yield Input(placeholder="This is here to escape to")
        with Panel():
            yield OptionList(*["This is an option" for _ in range(30)])
            yield Input(placeholder="Escape out via here for the bug")

if __name__ == "__main__":
    FocusWithinTransparentApp().run()
```

If you tab into the right panel the background of the `OptionList` updates as you'd expect (goes from red to yellow, as per its background), but if you keep tabbing through and land back in the left panel, the `OptionList`'s options' backgrounds will be out of sync with its parent (they will remain with a yellow background when it should be red). Actions that cause a refresh of some sort (resize of window, some mouse movements) cause the background to refresh to the correct colour.

https://github.com/user-attachments/assets/16610c05-7a17-4bbc-afdc-47970ee2e313

# Textual Diagnostics

## Versions

| Name    | Value  |
|---------|--------|
| Textual | 1.0.0  |
| Rich    | 13.9.4 |

## Python

| Name           | Value                                |
|----------------|--------------------------------------|
| Version        | 3.13.1                               |
| Implementation | CPython                              |
| Compiler       | Clang 16.0.0 (clang-1600.0.26.4)     |
| Executable     | /Users/davep/temp/x/.venv/bin/python |

## Operating System

| Name    | Value                                                                                                 |
|---------|-------------------------------------------------------------------------------------------------------|
| System  | Darwin                                                                                                |
| Release | 24.2.0                                                                                                |
| Version | Darwin Kernel Version 24.2.0: Fri Dec  6 18:51:28 PST 2024; root:xnu-11215.61.5~2/RELEASE_ARM64_T8112 |

## Terminal

| Name                 | Value           |
|----------------------|-----------------|
[...]

PR #5510, opened 2025-02-08
WIP new option list

willmcgugan:
A rewrite of OptionList internals, *mostly* preserving the original interface. The original was brittle, small changes would break everything. There was also some misguided attempts at doing things lazily, which resulted in doing work multiple times, particularly when first adding the widget. The original would also not be able to report its size accurately, until two or three frames later. An earlier pass partially fixed that, but it still took a frame before the OL could report its size.

- Simplify internal data structures.
- Fix "lazy" operations.
- Integrate better with the new Visual API.
- Replace `Separator` with `None`

- [x] Fixes https://github.com/Textualize/textual/issues/5489
- [x] Fixes https://github.com/Textualize/textual/issues/5488
- [x] Fixes https://github.com/Textualize/textual/issues/5418
- [x] Fixes https://github.com/Textualize/textual/issues/5431
- [x] Fixes https://github.com/Textualize/textual/issues/5498


TomJGooding:
I hope you don't mind me sticking my nose in (as usual!), especially since this is still a WiP.

Is it necessary to remove the `Separator` for this improved version of the `OptionList, or is this a design choice? Specifying a separator with None is obviously less intuitive, and it would be a shame to introduce a breaking change if it can be avoided.

willmcgugan:
I would disagree it is less intuitive. The OptionList intuitively takes a list of options. But a "Separator" is not an option, nor anything that is present in the option list. It is just a way of marking the end of a logical group. I think `None` does a better job of dividing the options into such groups.

Internally, it does simplify things. There is no need to distinguish between a real option and this special Separator case.

tests/snapshot_tests/test_snapshots.py
```diff
@@ -3421,3 +3426,126 @@ def compose(self) -> ComposeResult:
...
+            yield Input(placeholder="This is here to escape to")
+            with Panel():
+                yield OptionList(*["This is an option" for _ in range(30)])
+                yield Input(placeholder="Escape out via here for the bug")
+
+    snap_compare(FocusWithinTransparentApp(), press=["tab"])
+
+
+def test_option_list_wrapping(snap_compare):
+    """You should see a 40 cell wide Option list with a single line, ending in an ellipsis."""
```
TomJGooding, line 3497: Sorry I didn't spot this before it was merged, from the snapshot it looks like the OptionList height is based on the full height of the prompt rather than the truncated version with `nowrap`. Is that expected?

willmcgugan: Good catch, thanks.

PR #5519, opened 2025-02-12
test: add missing assertions to snapshot tests

TomJGooding:
I noticed that several recent snapshot tests simply call `snap_compare` without any `assert`. Without the assertion, these tests will not fail even if there is no snapshot history, as confirmed in #5518.


willmcgugan:
Good catch. Tests are failing, but they probably always were. Going to merge and fix the broken test.

PR #5538, opened 2025-02-16
Updated documentation with regard to poetry shell #5509

### LINKED ISSUE #5509: CONTRIBUTE.md references deprecated 'poetry shell'
In CONTRIBUTE.md's Setup section, users are instructed to use `poetry shell` to activate the virtual environment.

Poetry released a major upgrade to [2.0.0](https://python-poetry.org/blog/announcing-poetry-2.0.0/) in early January. This introduced a breaking change where the `poetry shell` command was removed completely.  

The [official documentation](https://python-poetry.org/docs/managing-environments/#activating-the-environment) now recommends either passing the command `poetry env activate` to the shell, or installing the [poetry-plugin-shell](https://github.com/python-poetry/poetry-plugin-shell) to add this command back to Poetry.

Poetry itself is a bit vague on the issue when it mentions none of these directly in response to `poetry shell`.

```
~/s/textual (main)> poetry shell

The command "shell" does not exist.

Did you mean one of these?
    self show
    self sync
    show
    self add
    self install
    install
    self lock
    help
```   

This will need to be added to the documentation rather than replacing the old command, as not all OS have added the new version of Poetry to their stable repos yet (Debian Bookworm naturally being one that I've found at least).

dominikgehl:
Fixes #5509 

- [x] Updated documentation

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,10 +19,13 @@ You can follow these steps:
 
  1. Make sure you have Poetry installed ([see instructions here](https://python-poetry.org))
  2. Clone the Textual repository
- 3. Run `poetry shell` to create a virtual environment for the dependencies
- 4. Run `make setup` to install all dependencies
- 5. Make sure the latest version of Textual was installed by running the command `textual --version`
- 6. Install the pre-commit hooks with the command `pre-commit install`
+ 3. Run `cd textual` to make textual your current working directory
+ 4. Run `python -m venv .venv` to create a new python virtual environment
+ 5. Run `source .venv/bin/activate` to activate the virtual environment
+ 6. Run `python -m pip install poetry` to install poetry
+ 7. Run `make setup` to install all dependencies
+ 8. Make sure the latest version of Textual was installed by running the command `textual --version`
+ 9. Install the pre-commit hooks with the command `pre-commit install`
 
 ([Read this](#makefile-commands) if the command `make` doesn't work for you.)
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,13 +19,10 @@ You can follow these steps:
 
  1. Make sure you have Poetry installed ([see instructions here](https://python-poetry.org))
  2. Clone the Textual repository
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -19,10 +19,13 @@ You can follow these steps:
...
  1. Make sure you have Poetry installed ([see instructions here](https://python-poetry.org))
  2. Clone the Textual repository
- 3. Run `poetry shell` to create a virtual environment for the dependencies
- 4. Run `make setup` to install all dependencies
- 5. Make sure the latest version of Textual was installed by running the command `textual --version`
- 6. Install the pre-commit hooks with the command `pre-commit install`
+ 3. Run `cd textual` to make textual your current working directory
+ 4. Run `python -m venv .venv` to create a new python virtual environment
+ 5. Run `source .venv/bin/activate` to activate the virtual environment
+ 6. Run `python -m pip install poetry` to install poetry
```
TomJGooding, line 25: I don't think it's a good idea to install Poetry in the environment of the project like this.

From the [Poetry installation docs:](https://python-poetry.org/docs/#installation):

> Poetry should always be installed in a dedicated virtual environment to isolate it from the rest of your system. Each of the above described installation methods ensures that. It should in no case be installed in the environment of the project that is to be managed by Poetry. This ensures that Poetry’s own dependencies will not be accidentally upgraded or uninstalled. In addition, the isolated virtual environment in which poetry is installed should not be activated for running poetry commands.

dominikgehl: Thanks for the comment. Change made

SYK17:
Hello!

Was going to make this very change myself as I ran into this exact issue.
This change is verified to work with the rest of the setup already in place.

sasaki0129:
Since `poetry shell` is no longer included by default in Poetry 2,
would it make sense to recommend the native workflow instead of installing the optional shell plugin?

For new contributors, something like this may be more future-proof:

```bash
poetry install
eval "$(poetry env activate)"
```

This avoids relying on an additional plugin and uses the workflow currently documented by Poetry itself.

Issue #5550, opened 2025-02-17
Textual 2.0 Regression - Auto no longer making widgets / containers as small as possible


driscollis:
Have you checked closed issues? (https://github.com/Textualize/textual/issues?q=is%3Aissue+is%3Aclosed) - YES

Have you checked against the most recent version of Textual? (https://pypi.org/search/?q=textual)  - YES

## Feature requests

Please post feature requests to Ideas. (https://github.com/Textualize/textual/discussions/categories/ideas)

## The bug

Please give a brief but clear explanation of the issue. If you can, include a complete working example that demonstrates the bug. **Check it can run without modifications.**

Using `auto` to set the height of a `Horizontal` container used to make a row of widgets shrink to the smallest size possible while showing the widgets. This worked in Textual 1.0 and earlier. In 2.0.0, the `Horizontal` container is taking up around 50% of the height.

Python code:

```python
import datetime

from textual.app import App, ComposeResult
from textual.containers import Horizontal, Vertical
from textual.widgets import Button, DataTable, Header, Label

class MyApp(App):

    CSS_PATH = "myapp.tcss"

    def compose(self) -> ComposeResult:
        """
        Create the user interface
        """
        now =  datetime.datetime.now().isoformat()
        self.last_updated = Label(f"Last Updated: {now[0:-7]}",
                                  id="last_updated",
                                  )

        yield Header()
        yield Vertical(
                    DataTable(id="datatable"),
                    Horizontal(
                        Button("OK", variant="primary", id="ok"),
                        Button("Cancel", variant="error", id="cancel"),
                        self.last_updated,
                        id="button_row",
                    ),
                    id="main_tab",
                )

    def on_mount(self) -> None:
        rows = [("Name", "Occupation", "Country",),
                ("Mike", "Python Wrangler", "USA",),
                ("Bill", "Engineer", "UK",),
                ("Dee", "Manager", "Germany"),
                ]
        table = self.query_one(DataTable)
        table.clear(columns=True)
        table.add_columns(*rows[0])
        table.add_rows(rows[1:])


if __name__ == "__main__":
    app = MyApp()
    app.run()
```

CSS code:

```css
MyApp {
    align: center middle;
    #datatable {
        height: 1fr;
    }

    Horizontal {
        height: auto;
    }

    #button_row {
        align: center middle;
    }

    Button {
        margin: 1;
    }

    #last_updated {
        dock: right;
        offset: 0 2;
    }
}
```

Screenshot using Textual 1.0:

[...]

PR #5552, opened 2025-02-17
auto height logic

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/5550


willmcgugan:
The markdown snapshot is due to Pygments being bumped, but otherwise unrelated to this PR.

PR #5553, opened 2025-02-17
fix(border): fix flipped title colors in panel border

### LINKED ISSUE #5548: Border Title Color TCSS Style Regression
## The bug

Border title colors are being displayed as backgrounds instead of changing the font color in Textual 2.0.3.

### 1.0.0
Image

### 2.0.3
Image

## MRE
```python
from textual.app import App


class BugReportApp(App[None]):
    CSS = """
    Vertical {
        border-top: panel $surface-lighten-1;
        border-title-color: yellow;
    }
    """

    def compose(self):
        with Vertical() as c:
            c.border_title = "Panel Title"
            yield Static("Content")


if __name__ == "__main__":
    BugReportApp().run()
```

Not 100% sure if  this is an unintended change, as it might relate to #5389.

# Textual Diagnostics

## Versions

| Name    | Value  |
|---------|--------|
| Textual | 2.0.3  |
| Rich    | 13.9.4 |

## Python

| Version        | 3.12.5                                             |
| Compiler       | GCC 14.2.1 20240801 (Red Hat 14.2.1-1)             |
| Executable     | /home/dk/dev/tmp/textual-sandbox/.venv/bin/python3 |

## Operating System

| Release | 6.12.13-100.fc40.x86_64                             |
| Version | #1 SMP PREEMPT_DYNAMIC Sat Feb  8 17:10:01 UTC 2025 |

## Terminal

| Name                 | Value       |
|----------------------|-------------|
| Terminal Application | Kitty       |
| TERM                 | xterm-kitty |
| COLORTERM            | truecolor   |
| FORCE_COLOR          | *Not set*   |
| NO_COLOR             | *Not set*   |

## Rich Console options

| size           | width=137, height=54 |
| max_width      | 137                  |
| max_height     | 54                   |

TomJGooding:
Revert the change to the color flipping logic for `panel` border titles.

The issue with the new method using `Style(reverse=True)` is that this will also flip any border title colors set with Textual CSS.

Fixes #5548.

- [ ] Updated documentation


src/textual/_styles_cache.py
```diff
@@ -379,7 +379,11 @@ def post(segments: Iterable[Segment]) -> Iterable[Segment]:
                         if label_background.a
                         else TRANSPARENT
                     ),
-                    (base_label_background + border_color + label_color),
+                    (
+                        (base_label_background + label_color)
+                        if label_color.a
+                        else TRANSPARENT
+                    ),
```
TomJGooding, line 386: This change effectively reverts to the original logic, which is necessary since the foreground/background is flipped for the panel border.

I wanted to flag this change in case there is something I've missed. While this did cause a couple of snapshot tests to fail,  visually they looked identical.

willmcgugan: It's encouraging that no tests broke

The previous version added the border color, this version doesn't. I don't recall what, if anything, that was there for. I have a suspicion it is something to do with transparent borders. i.e. if the border color has alpha.

TomJGooding: Thanks for checking, I wasn't sure what the border color was there for either.

If the border title actually on sits on the border color for the panel/tab styles, maybe this isn't needed....?

willmcgugan: Wish I could go back in time and ask myself why I did that. If I didn't write a test for it, then it's probably safe.

willmcgugan:
Thanks

PR #5562, opened 2025-02-19
App query change

willmcgugan:
This hopefully fixes a common gotcha.

Previously when you do `app.query` it will query the active screen. This is completely reasonable from a definitional POV. But it is a frequent source of confusion because when you push a screen, previously working queries now fail.

This change ensures that `app.query` queries the screen that was composed by the App itself, i.e. the default screen. So it always queries the same screen. If you do want to query the current screen it is as simple as `app.screen.query`.

This shouldn't impact many apps, because as soon as devs run in to the query broken by push_screen gotcha, we advise them to be explicit about screens anyway.


willmcgugan:
@darrenburns Any thoughts on this one?

darrenburns:
I think this is a better approach!

willmcgugan:
On hold for a future major bump. I'm expecting at least one more bug fix release in the interim

Issue #5568, opened 2025-02-21
`SelectionList` -> `OptionList.render_line` Unhandled IndexError


ddkasa:
## The Bug

When using a `SelectionList` the `render_line` method of the new option list is throwing an `IndexError` exception when the strips are empty.

## Traceback

```sh
│                                                                                                                                       │
│ /home/dk/dev/tmp/textual-sandbox/.venv/lib/python3.12/site-packages/textual/widgets/_selection_list.py:514 in render_line             │
│                                                                                                                                       │
│   511 │   │   # TODO: This is rather crufty and hard to fathom. Candidate for a rewrite.       ╭──────── locals ────────╮             │
│   512 │   │                                                                                    │ self = SelectionList() │             │
│   513 │   │   # First off, get the underlying prompt from OptionList.                          │    y = 0               │             │
│ ❱ 514 │   │   line = super().render_line(y)                                                    ╰────────────────────────╯             │
│   515 │   │                                                                                                                           │
│   516 │   │   # We know the prompt we're going to display, what we're going to do                                                     │
│   517 │   │   # is place a CheckBox-a-like button next to it. So to start with                                                        │
│                                                                                                                                       │
│ /home/dk/dev/tmp/textual-sandbox/.venv/lib/python3.12/site-packages/textual/widgets/_option_list.py:860 in render_line                │
│                                                                                                                                       │
│   857 │   │                                                                                                                           │
│   858 │   │   strips = self._get_option_render(option, style)                                                                         │
│   859 │   │                                                                                                                           │
│ ❱ 860 │   │   return strips[line_offset]                                                                                              │
│   861 │                                                                                                                               │
│   862 │                                                                                                                               │
│   863 │   def validate_highlighted(self, highlighted: int | None) -> int | None:                                                      │
│                                                                                                                                       │
│ ╭──────────────────────────────────────── locals ─────────────────────────────────────────╮                                           │
│ │ component_class = 'option-list--option-highlighted'                                     │                                           │
│ │     line_number = 0                                                                     │                                           │
│ │     line_offset = 0                                                                     │                                           │
│ │      mouse_over = False                                                                 │                                           │
│ │          option = Selection(<text 'Professional nor' [] ''>)                            │                                           │
│ │    option_index = 0                                                                     │                                           │
│ │            self = SelectionList()                                                       │                                           │
│ │          strips = []                                                                    │                                           │
[...]

PR #5574, opened 2025-02-22
Fix ol render line

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/5568

Not entirely sure about the conditions which lead up to the IndexError. I suspect its a race condition with the resize event. The only way I can reproduce it is to resize the terminal until it occurs.

### DIFF
--- a/src/textual/widgets/_option_list.py
+++ b/src/textual/widgets/_option_list.py
@@ -853,7 +853,10 @@ class OptionList(ScrollView, can_focus=True):
             style = self.get_visual_style("option-list--option")
 
         strips = self._get_option_render(option, style)
-        strip = strips[line_offset]
+        try:
+            strip = strips[line_offset]
+        except IndexError:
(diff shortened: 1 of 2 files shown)


src/textual/widgets/_option_list.py
```diff
@@ -853,7 +853,10 @@ def render_line(self, y: int) -> Strip:
             style = self.get_visual_style("option-list--option")
 
         strips = self._get_option_render(option, style)
-        strip = strips[line_offset]
+        try:
+            strip = strips[line_offset]
+        except IndexError:
```
ddkasa, line 858: One suggestion. When I was trying to find the source of the bug and reading the previous `OptionList`. It would have been nice if there was comment here stating, why this exception handler was added in the first place.

willmcgugan: I don't think a comment here could say more than the code could.

PR #5594, opened 2025-02-27
docs(button): update note about removing spacing

TomJGooding:
Update the note in the `Button` docs about removing the spacing. The default styling changed at some point from `height: 3` to `auto`, but this note should also mention that you also need to update the pseudo classes.

- [x] Updated documentation

### DIFF
--- a/docs/widgets/button.md
+++ b/docs/widgets/button.md
@@ -51,7 +51,10 @@ This widget has no component classes.
 
 ## Additional Notes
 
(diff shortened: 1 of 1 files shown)


TomJGooding:
You actually don't need to worry about the pseudo classes (or classes for the button variants) if you set `border: none`. Sorry for the noise, I'm not sure if I just imagined this or perhaps this it necessary in a previous version...

PR #5598, opened 2025-03-01
Update containers.py

friscorose:
Using None as default for min_column_width breaks automatic horizontal flow of children into columns per the API at https://textual.textualize.io/api/containers/#textual.containers.ItemGrid

Setting to the minimum positive integer that maintains documented behavior. Though this may actually be a deeper issue in layout, I leave that decision to someone more knowledgeable. Other positive integers may also be more desirable and it has been suggested to pick a value that is a power of 2.

- [ ] Updated documentation

### DIFF
--- a/src/textual/containers.py
+++ b/src/textual/containers.py
@@ -254,7 +254,7 @@ class ItemGrid(Widget, inherit_bindings=False):
     """
 
     stretch_height: reactive[bool] = reactive(True)
-    min_column_width: reactive[int | None] = reactive(None, layout=True)
+    min_column_width: reactive[int | None] = reactive(1, layout=True)
     regular: reactive[bool] = reactive(False)
 
     def __init__(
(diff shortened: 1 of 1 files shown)


friscorose:
> 
> A min column width of 1 is likely always wrong. Most widgets would be unusable at that size.
> 
This PR is not really me - it was a discussion in @help-wanted to help @Xotembotz, but I think the solution hoped for was something more along the lines of auto resizing columns based on dynamic item input.

That said, I can imagine a single character item column being useful in a few odd corner cases and since 1 is technically the minimum possible for the item to actually be present, it seemed a suitable stop gap solution until a more robust solution exists.

Please feel free to close this PR or ask me to - it was just a breadcrumb to mark a user issue.

willmcgugan:
At some point there will be a more flexible ItemGrid, or something like it.

Issue #5611, opened 2025-03-04
Flaky `test_markdown_theme_switching` snapshot test


TomJGooding:
I haven't dived into this yet, but recently the `test_markdown_theme_switching` snapshot test seems a bit flaky.

This snapshot test fails when running test locally on main, but looks identical in the snapshot report.

TomJGooding:
It looks like this test was failing locally because my dependencies weren't in sync - see 9769743 when this snapshot was last updated. After running `poetry install` (via `make setup`), this test now passes.

PR #5612, opened 2025-03-04
fix: fix alignment with auto and min height/width

### LINKED ISSUE #5608: Align doesn't work with height and min-height combination in widget
This example shows that align property is not working well. Static widget should be positioned vertically in the middle of `CustomWidget`. It was broken between 0.83.0 version and 2.1.2.
```python

from textual.widgets import Footer, Static


class CustomWidget(Widget):
    DEFAULT_CSS = """
    CustomWidget {
      width: 1fr;
      background: grey;
      height: auto;
      min-height: 3;
      align: center middle; /* align doesn't work with height and min-height combination */
      
      Static {
        background: red;
      }
    }
    """

class MyApp(App):
        with CustomWidget():
            yield Static("Some text")
        yield Footer()


MyApp().run()
```
With deleted `min-height ` and static `height: 3` static widget is positioned properly:
```python

from textual.widgets import Footer, Static


class CustomWidget(Widget):
    DEFAULT_CSS = """
    CustomWidget {
      width: 1fr;
      background: grey;
      height: 3;
      align: center middle; /* now align works properly */
      
      Static {
        background: red;
      }
    }
    """

class MyApp(App):
        with CustomWidget():
            yield Static("Some text")
        yield Footer()


MyApp().run()

```

TomJGooding:
Fix issue with alignment in containers with auto _and_ min height/width. Currently the size of the parent passed to `_align_size` doesn't account for any min height/width when the container has auto dimensions.

Fixes #5608

- [ ] Updated documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+### Fixed
+
+- Fixed issue with alignment in containers with auto and min height/width https://github.com/Textualize/textual/issues/5608
+
 ### Added
 
 - Added Widget.preflight_checks to perform some debug checks after a widget is instantiated, to catch common errors. https://github.com/Textualize/textual/pull/5588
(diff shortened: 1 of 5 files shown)


TomJGooding:
I confess I don't _fully_ understand all the inner workings of sizing and alignment in Textual, so there might be a better way of fixing this. All the tests seem to pass[^1] after this change so hopefully I'm on the right track, but I'd appreciate any feedback.

[^1]: Apart from an unrelated snapshot test - see  #5611

PR #5613, opened 2025-03-05
Input widget: Validation for initial value

recsater:

- [ ] Updated documentation

In the existing Input implementation, validation is not performed in the initial state. Therefore, the styling based on validation results is only applied when the Input value changes or receives focus.

If this is not the intended behavior, performing validation during initialization would ensure the correct validation state for the initial value.

### DIFF
--- a/src/textual/widgets/_input.py
+++ b/src/textual/widgets/_input.py
@@ -431,6 +431,7 @@ class Input(ScrollView):
             self.tooltip = tooltip
 
         self.select_on_focus = select_on_focus
+        self.validate(self.value)
 
     def _position_to_cell(self, position: int) -> int:
         """Convert an index within the value to cell position.


TomJGooding:
This example shows the invalid styling - have I misunderstood?

```python
from textual import on
from textual.validation import Number
from textual.widgets import Input, Label


class InputApp(App):
        yield Label("Enter a number")
        yield Input(
            value="This is not a number",
            validators=[Number()],
        )


if __name__ == "__main__":
    app = InputApp()
    app.run()
```

recsater:
@TomJGooding 

I tested your example and it seems the invalid styling only occurs when the Input's value is empty.

Try testing with this example:

```python3
from textual.validation import Function
from textual.widgets import Input, Label


class InputApp(App):
        yield Label("Enter a number")
        yield Input(
            value="",
            validators=[Function(lambda v: False)],
        )
        yield Input(
            placeholder="Focus to here",
            validators=[Function(lambda v: False)],
        )


if __name__ == "__main__":
    app = InputApp()
    app.run()
```

recsater:
This seems to be because when the value passed to the constructor is empty, it doesn't change the default reactive value, thus not triggering the `_watch_value` function.

Therefore, modifying the PR with `self._watch_value(value)` also should apply the correct styling.

recsater:
ok, I agree with you.

PR #5618, opened 2025-03-06
Accept the new VisualType in Static and Label

davidfokkema:
This change keeps mypy happy.

- [x] Updated documentation

### DIFF
--- a/src/textual/widgets/_label.py
+++ b/src/textual/widgets/_label.py
@@ -4,9 +4,7 @@ from __future__ import annotations
 
 from typing import Literal
 
-from rich.console import RenderableType
-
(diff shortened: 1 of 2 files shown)


davidfokkema:
Thanks to @TomJGooding pointing out the solution to https://github.com/Textualize/textual/discussions/5617.

willmcgugan (approved):
I think this is right.

That setter should probably be renamed to `content`. Since that's a breaking change, it probably makes sense to put it off until the next major release.

Would you mind adding a TODO for that in this PR?

davidfokkema:
I added TODO items for the Label parameter `renderable` and the setter on Static.

PR #5638, opened 2025-03-11
fix(tabs): update mismatched types

### LINKED ISSUE #5635: Tabs widget does not support Rich
Rich theme (and Rich in general) is not being supported with `Tabs` widget using `v2.0.0`. Also, emojis are not working anymore.

I tried to use `from_markup`, but it isn't working:
```python
def on_mount(self) -> None:
        self.query_one(Tabs).add_tab(Text.from_markup("[special_red]Special Red Again"))
```

Error:
`AttributeError: 'Text' object has no attribute 'translate'`

MRE:
```python
from rich.text import Text
from rich.theme import Theme as RichTheme
from textual.widgets import Tabs

NAMES = [
    "[special_red]Special Red",
    "[red]Red",
    "[yellow]Yellow",
    ":skull: Skull",
]


class MREApp(App):

    def __init__(self):
        super().__init__()

        theme = RichTheme(
            {
                "special_red": "#ff5e5e",
            }
        )
        self.console.push_theme(theme)

        yield Tabs(*NAMES)


if __name__ == "__main__":
    app = MREApp()
    app.run()

```

TomJGooding:
Revert the type of the `Tab` label to `TextType`.

Currently there's a mismatch between the expected types for `Tabs` and `Tab`, which can cause the app to crash. While `Tabs` still expects a `TextType`, the `Tab` was recently updated to expect a `ContentType`.

Since the tabs documentation is based on using `Text` objects, I think reverting this changes makes sense, especially given that `Tab` seems to be the only widget that now exclusively expects `ContentType`.

This might _technically_ be a breaking change, but the current incompatibility of types means that the tabs were broken anyway.

Fixes #5635.

- [ ] Updated documentation

### DIFF
--- a/src/textual/widgets/_tabs.py
+++ b/src/textual/widgets/_tabs.py
@@ -23,6 +23,7 @@ from textual.widgets import Static
 
 if TYPE_CHECKING:
     from textual.content import Content, ContentType
+    from textual.visual import VisualType
 
 
 class Underline(Widget):
@@ -179,9 +180,9 @@ class Tab(Static):
         self._label = self.render_str(label)
         self.update(self._label)
 
-    def update(self, content: ContentType = "") -> None:
+    def update(self, content: VisualType = "") -> None:
         self.post_message(self.Relabelled(self))
-        return super().update(self.render_str(content))
+        return super().update(content)
 
     @property
     def label_text(self) -> str:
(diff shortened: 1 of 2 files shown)


willmcgugan:
This seems to be going in the wrong direction. Content should be preferred now. What docs refer to Text? Seems like the fix should be updating the docs to refer to Content.

charles-001:
I actually got it working... I just had to specify the console:
```python
self.host_tabs.query(TabWidget).filter("#" + tab.id)[0].label = Content.from_rich_text(
     new_name, console=self.app.console
)
```

PR #5639, opened 2025-03-12
TextArea improvements - lazy language import and allow installation of only required languages

darrenburns:
Launching TextArea with the `syntax` extras has a large upfront cost the first time it's run in a new virtualenv.

A one-off compilation step is performed when the language is imported, and since all languages were being imported at the module level, it was adding a 1-2 second overhead at startup on the first run.

Subsequent launches within the same venv had little overhead (a few milliseconds).

image

### Lazy importing of tree-sitter languages

This PR updates the TextArea to lazily import the languages, only when they're used, trimming 1-2 seconds off of these "cold starts".

### Allowing users to install only what they need

With this PR, users don't need to install the `syntax` extras which includes a bunch of languages. They can instead just install `tree-sitter` plus the languages they require.

For example, with Posting I uninstalled the syntax extras then done this:

```bash
uv add tree-sitter tree-sitter-json tree-sitter-html tree-sitter-xml
```

...and it just worked without any code changes. Apps no longer need to pull in a bunch of transitive dependencies that they don't need.

- [ ] Updated documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Added Widget.preflight_checks to perform some debug checks after a widget is instantiated, to catch common errors. https://github.com/Textualize/textual/pull/5588
 
+### Changed
+
+- tree-sitter languages are now loaded lazily, improving cold-start time https://github.com/Textualize/textual/pull/5639
+
 ## [2.1.2] - 2025-02-26
 
 ### Fixed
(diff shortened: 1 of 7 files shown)


TomJGooding:
> With this PR, users don't need to install the `syntax` extras which includes a bunch of languages. They can instead just install `tree-sitter` plus the languages they require.

You obviously have a better understanding of tree-sitter than I do, but could this not be quite brittle if the language package version isn't compatible with the highlight query?

darrenburns:
It's more brittle than the syntax extras yeah, but you can still supply your own highlight query that aligns with the parser when you register the language. If you're just doing "pip install tree-sitter-python" then relying on the built-in highlight query, then that'd be quite brittle. You should probably be installing a specific version, and registering the language and matching highlight query pair with TextArea.register_language.

PR #5650, opened 2025-03-14
Bugfix Select.selection with no value selected

Zaloog:
Based on the docs `Select.selection` should return `None` if no value is selected, i.e. the value is `Select.Blank`. It just raised an assertion error though. Fixed the issue and added a test.

Already present, just fixed behavior to represent actual description
- [x] Updated documentation
Already present, just fixed behavior to represent actual description


willmcgugan (approved):
Thanks

Issue #5559, opened 2025-02-19
Textual 2.0.4 plus ghostty "fakes" escape press when mouse crosses window border


davep:
With this code:

```python
from textual.app import App

class WeirdEscapeApp(App[None]):

    BINDINGS = [("escape", "escape")]

    def action_escape(self) -> None:
        self.notify("Escape was pressed!")

if __name__ == "__main__":
    WeirdEscapeApp().run()
```

I see this effect:

https://github.com/user-attachments/assets/ac15548c-3381-43ed-beeb-1d53a7e2fbb0

Testing with Textual 1.0.0 I don't see this. On the other hand I can't recreate it with iTerm and Textual 2.0.4 (haven't had the chance to test other terminals).

While I appreciate this might be specific to a particular pretty new terminal, I thought it worth a heads-up at least.

# Textual Diagnostics

## Versions

| Name    | Value  |
|---------|--------|
| Textual | 2.0.4  |
| Rich    | 13.9.4 |

## Python

| Name           | Value                                |
|----------------|--------------------------------------|
| Version        | 3.13.2                               |
| Implementation | CPython                              |
| Compiler       | Clang 16.0.0 (clang-1600.0.26.6)     |
| Executable     | /Users/davep/temp/x/.venv/bin/python |

## Operating System

| Name    | Value                                                                                                 |
|---------|-------------------------------------------------------------------------------------------------------|
| System  | Darwin                                                                                                |
| Release | 24.3.0                                                                                                |
| Version | Darwin Kernel Version 24.3.0: Thu Jan  2 20:23:36 PST 2025; root:xnu-11215.81.4~3/RELEASE_ARM64_T8112 |

## Terminal

| Name                 | Value           |
|----------------------|-----------------|
| Terminal Application | ghostty (1.1.2) |
| TERM                 | xterm-256color  |
| COLORTERM            | truecolor       |
| FORCE_COLOR          | *Not set*       |
| NO_COLOR             | *Not set*       |

## Rich Console options

| Name           | Value               |
|----------------|---------------------|
| size           | width=90, height=37 |
| legacy_windows | False               |
| min_width      | 1                   |
| max_width      | 90                  |
| is_terminal    | False               |
| encoding       | utf-8               |
| max_height     | 37                  |
| justify        | None                |
| overflow       | None                |
| no_wrap        | False               |
| highlight      | None                |
| markup         | None                |
| height         | None                |

willmcgugan:
[...]

PR #5688, opened 2025-03-29
fix: handle negative mouse coordinates

TomJGooding:
It is possible that terminals will report negative mouse coordinates when using SGR-Pixels (1016). While this arguably violates the standard for CSI sequences (see https://github.com/Textualize/textual/issues/5559), perhaps Textual should be more resilient to handle these sequences.

- [ ] Updated documentation

### DIFF
--- a/src/textual/_xterm_parser.py
+++ b/src/textual/_xterm_parser.py
@@ -18,7 +18,7 @@ from textual.message import Message
 # to be unsuccessful?
 _MAX_SEQUENCE_SEARCH_THRESHOLD = 32
 
-_re_mouse_event = re.compile("^" + re.escape("\x1b[") + r"(<?[\d;]+[mM]|M...)\Z")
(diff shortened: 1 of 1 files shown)


willmcgugan:
I'm all for a pragmatic approach. It's certainly better than breaking. But there are a couple of issues here.

What happens when a negative coordinate arrives? Presumably Textual will send that negative coordinate to the app. But then we have different behavior compared to more compliant terminals. I think we should ignore negative coordinates to maintain consistency.

I'd also like to tighten up the regexes a bit. My regex-fu isn't great, but if I have read the code correctly, it will allow nonsense strings like "12-1234-234-234-23". Which isn't likely to occur, but I'd feel better about it if the hyphen could occur in the first position only.

TomJGooding:
Agreed, I'm very wary to change the regex which is why I left this in draft!

Should the negative coordinates just be ignored? I think clicking and dragging outside the terminal window still reports mouse events in other emulators.

TomJGooding:
Apparently superseded by #5926

PR #5700, opened 2025-04-02
fix(binding): fix comma-space separated keymaps

### LINKED ISSUE #5694: A `Keymap` binding that is the same as a `BINDING` binding will break when there's a space
With this code:

```python
from textual.widgets import Button, Footer

class BadMapApp(App[None]):

    BINDINGS = [
        Binding("f1, f2", "good", description="Good", id="good"),
        Binding("f3, f4", "bad", description="Bad", id="bad"),
    ]

        yield Button("Break bad")
        yield Footer()

    def action_good(self) -> None:
        self.notify("Good")

    def action_bad(self) -> None:
        self.notify("Bad")

    def on_button_pressed(self) -> None:
        self.update_keymap({"bad": "f3, f4"})
        self.notify("f4 will no longer work now")

if __name__ == "__main__":
    BadMapApp().run()
```

if you press the button, setting the binding for `action_bad` to be the same as it was before, but via `update_keymap`, <kbd>f4</kbd> no longer works. If you remove the space after the comma in the call to `update_keymap` it works. Presumably the code that consumes the keymap bindings needs to `strip` the keys in the same way I seem to remember `BINDINGS` processing does.

### LINKED ISSUE #5704: A `Keymap` binding that is the same as a `BINDING` binding will break with symbol keys
> I have a similar problem that is probably related: applying a keymap containing a symbol key (like "?" or "/") will cause the keybinding to fail.  Logging the value of `app.active_bindings` showed that before applying a keymap the key's value was `"question"` and after applying the keymap it was `"?"`.  
> 
> Here is a modified version of @davep 's code above to illustrate:
> 
> ```python
> from textual.app import App, ComposeResult
> from textual.binding import Binding
> from textual.widgets import Button, Footer
> 
> class BadMapApp(App[None]):
>     BINDINGS = [
>         Binding("f1", "good", description="Good", id="good"),
>         Binding("?", "bad", description="Bad", id="bad"),
>     ]
> 
>     def compose(self) -> ComposeResult:
>         yield Button("Break bad")
>         yield Footer()
> 
>     def action_good(self) -> None:
>         self.notify("Good")
> 
>     def action_bad(self) -> None:
>         self.notify("Bad")
> 
>     def on_button_pressed(self) -> None:
>         self.update_keymap({"bad": "?"})
>         self.notify("? will no longer work now")
> 
> if __name__ == "__main__":
>     BadMapApp().run()
> ``` 

 _Originally posted by @mathman1618 in [#5694](https://github.com/Textualize/textual/issues/5694#issuecomment-2776018302)_

TomJGooding:
This PR initially just added `strip` for the keymap keys, but hopefully now ensures consistent key handling when applying keymaps.

Fixes #5694
Fixes #5704

- [ ] Updated documentation


willmcgugan (approved):
LGTM Thanks!

PR #5713, opened 2025-04-06
docs: fix camel_to_snake docstring

TomJGooding:
Spotted in passing and decided just to fix it after wondering if it was even worth a pull request.


willmcgugan (approved):
This is a contender for the longest standing doc bug in the project!

PR #5722, opened 2025-04-08
remove RadioButton glyph edges to prevent rendering artifacts in term…

### LINKED ISSUE #5721: RadioButton unicode circle display has a line through it
## The bug

Issue: When using RadioButton in a RadioSet, the toggle glyph (●) renders with a visible horizontal line through it, making it appear broken or misaligned.

Image

Environment: Ubuntu 22.04.5 LTS using the default GNOME Terminal.

```
textual diagnose
```
# Textual Diagnostics

## Versions

| Name    | Value  |
|---------|--------|
| Textual | 2.1.2  |
| Rich    | 13.9.4 |

## Python

| Name           | Value            |
|----------------|------------------|
| Version        | 3.10.12          |
| Implementation | CPython          |
| Compiler       | GCC 11.4.0       |
| Executable     | /usr/bin/python3 |

## Operating System

| Release | 6.8.0-52-generic                                                 |
| Version | #53~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Wed Jan 15 19:18:46 UTC 2 |

## Terminal

| Terminal Application | GNOME Terminal |

## Rich Console options

| size           | width=123, height=50 |
| max_width      | 123                  |
| max_height     | 50                   |

presstab:
closes #5721 

🛠️ Summary

This PR fixes a rendering issue in the RadioButton widget where an extra line or visual artifact appears around the toggle glyph (●) in certain terminal environments (e.g. Ubuntu 22.04.5 LTS with GNOME Terminal).
✅ Fix

The issue was caused by inherited BUTTON_LEFT and BUTTON_RIGHT characters from the ToggleButton base class. These characters, while visually subtle in some environments, render poorly or as visible lines in others.

To resolve this, BUTTON_LEFT and BUTTON_RIGHT are explicitly set to empty strings in RadioButton:

BUTTON_LEFT = ""
BUTTON_RIGHT = ""

This ensures only the intended center glyph (●) is rendered, avoiding any extra characters or alignment issues.

Bug screenshot:
Screenshot from 2025-04-08 13-48-03

Fixed screenshot:
Screenshot from 2025-04-08 13-11-50

💻 Environment

Tested on:

    Ubuntu 22.04.5 LTS

    GNOME Terminal (default)

    Textual version: [insert version if needed]

- [X] Updated documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Added missing `scroll_end` parameter to the `Log.write_line` method https://github.com/Textualize/textual/pull/5672
 - Restored support for blink https://github.com/Textualize/textual/pull/5675
 - Fixed scrolling breaking on DataTable with `overflow: hidden` https://github.com/Textualize/textual/pull/5681
+- Fixed ToggleButton glyph edges drawing lines through RadioButton https://github.com/Textualize/textual/pull/5722
 
 ### Added
 
--- a/src/textual/widgets/_radio_button.py
+++ b/src/textual/widgets/_radio_button.py
@@ -15,6 +15,12 @@ class RadioButton(ToggleButton):
     BUTTON_INNER = "\u25cf"
     """The character used for the inside of the button."""
 
+    BUTTON_LEFT = ""
+    """Unused left character from toggle button (empty to avoid unwanted rendering artifacts in some terminals)."""
+
+    BUTTON_RIGHT = ""
+    """Unused right character from toggle button (empty to avoid unwanted rendering artifacts in some terminals)."""
+
     class Changed(ToggleButton.Changed):
         """Posted when the value of the radio button changes.
 


TomJGooding:
These characters aren't "unused", as you can clearly see in the `RadioSet` example: https://textual.textualize.io/widgets/radioset/#example

But I guess you wouldn't know that, as it looks like this was LLM-generated based on the placeholder for the Textual version you forgot to change.

You're proposing breaking the style of the radio button based on the circle glyph not working with only in a specific terminal emulator and only with a specific font.

presstab:
> These characters aren't "unused", as you can clearly see in the `RadioSet` example: https://textual.textualize.io/widgets/radioset/#example
> 
> But I guess you wouldn't know that, as it looks like this was LLM-generated based on the placeholder for the Textual version you forgot to change.
> 
> You're proposing breaking the style of the radio button based on the circle glyph not rendering correctly only in a specific terminal emulator and only with a specific font.

Not sure what the hostility is for, I just saw an issue, and thought I could help. I used chatgpt to generate the pr summary, sometimes I don't have the best wording. I also noticed that other PR's here from large contributors use LLM generated PR summaries. I was excited to get this PR together and posted, and forgot to change that tag.

I tracked down this manually (not that it matters), as chatgpt kept trying to change the CSS I didn't feed it Textual source code, so it had limited guesses. I don't see the side buttons used in that example you linked, perhaps you are suggesting they are spacers, I don't know... I am just a first time contributor and I obviously don't have any expertise with the codebase (I am not sure what breaks what, etc).

I would disagree that the default terminal on Ubuntu is some specific use that shouldn't be considered, but don't really care enough to have an argument. I'll just override it in my own project and move on from this topic :)

willmcgugan:
@TomJGooding That was rather ungenerous. I think @presstab 's intentions entirely honorable.

@presstab The issue seems to be that Gnome terminal is rendering that glyph too wide. Having one cell overlap with a subsequent cell just shouldn't happen. So I think the fault lies with GT.

But I don't like to pass the buck with such things. Even if Gnome fix it, it will be a long time before every user gets the update. So finding a workaround is desirable.

Unfortunately we can't make that change, as it would break the output on other terminals. Here are some alternative options as I see it.

- Find an alternative glyph that happens to work across all terminals, and looks good.
- Sniff the terminal and change only affected terminals. I generally don't like to do that, as it can be brittle. But it is an option.

I'm going to open the original issue, as it is still valid. We can move the discussion there.

TomJGooding:
@presstab I'm sorry for jumping to conclusions. Reading this back in the cold light of day, my review was definitely more hostile than I intended. I would hate to think that I'm putting off first time contributors, especially as I'm only a very minor contributor myself.

PR #5730, opened 2025-04-10
fix(option list): fix size when options removed or cleared

### LINKED ISSUE #5728: `OptionList` not fully updating on options being cleared (since v2.0.0)
Given this code:

```python

class BadScrollBarApp(App[None]):

    BINDINGS = [
        ("space", "nuke")
    ]

        yield OptionList(*[f"Option {n}" for n in range(500)])

    def action_nuke(self) -> None:
        self.query_one(OptionList).clear_options()

if __name__ == "__main__":
    BadScrollBarApp().run()
```

With textual v3.x, on startup, it looks as you'd expect:

Image

but then after pressing <kbd>space</kbd> it looks like this:

Image

note how the widget itself is still the full height of the screen despite `OptionList` being `height: auto`; and also note how the scrollbar still seems to think there is the full range of options.

Going back through Textual releases, the last time I see the expected behaviour is with Textual v1.0.0; where after pressing <kbd>space</kbd> we see this:

Image

Installing Textual v2.0.0 introduces the problem.

TomJGooding:
Fix the `OptionList` size not updating correctly after removing or clearing options.

Fixes #5728

- [ ] Updated documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Fixed incorrect auto height in Collapsible https://github.com/Textualize/textual/pull/5703
 - Fixed issue with keymaps and single-letter keys https://github.com/Textualize/textual/pull/5726
 - Fixed `OptionList` size after removing or clearing options https://github.com/Textualize/textual/issues/5728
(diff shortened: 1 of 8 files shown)


willmcgugan (approved):
LGTM

davep:
Thanks @TomJGooding.

PR #5732, opened 2025-04-10
Extrema

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/5608


willmcgugan:
Didn't notice https://github.com/Textualize/textual/pull/5612 before I started this, which addresses the same issue!

TomJGooding:
@willmcgugan This is definitively more 'extreme' than my attempt to fix this!

Thanks for the acknowledgment in the comment, but can I make a suggestion? When including someone else's code, [adding them as a co-author to the commit ](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors) is a nice way to give credit. I realise this involves a bit more work, but it might mean a lot to someone to showcase they've contributed to a project like this.

PR #5738, opened 2025-04-11
make it so TEXTUAL_THEME="" is default

Commandcracker:
Before, it would crash when setting TEXTUAL_THEME=""


willmcgugan:
It doesn't crash for me. I will need an issue before I can accept a PR.

Issue #5729, opened 2025-04-10
`OptionList` with non-trivial options causes UI lag since v2.0.0


davep:
I have a few applications that make great use of `OptionList` and Rich's tables and groups -- it makes for some wonderful layouts and data presentation. However, when updating these applications to use Textual later than v1.0.0 I'm noticing some very obvious UI lag when moving focus around. I've managed to whittle the effect down to this:

```python
from rich.console import Group
from rich.table import Table
from rich.rule import Rule

from textual.app import App, ComposeResult
from textual.widgets import OptionList
from textual.widgets.option_list import Option

class ComplexOption(Option):

    def __init__(self, number: int) -> None:
        second_line = Table.grid(expand=True)
        second_line.add_column(justify="left", ratio=2)
        second_line.add_column(justify="right", width=28)
        second_line.add_row(
            f"[bold]{number}[/]",
            "This is just some text here",
            f"[dim]And some more: {number}[/]",
        )
        info = Table.grid(expand=True)
        info.add_column(width=6)
        info.add_column(ratio=1)
        info.add_column(width=11, justify="right")
        info.add_row(
            "", f"[dim]Stuff[/]", f"[dim]Things[/]"
        )

        super().__init__(Group(
            f"{number}. This is a simple line",
            second_line,
            info,
            Rule()
        ))

class OptionListLagApp(App[None]):

    CSS = """
    Screen {
        layout: horizontal;
    }
    """

    @property
    def simple_content(self) -> list[str]:
        return [f"{n}. This is simple option" for n in range(1000)]

    @property
    def complex_content(self) -> list[ComplexOption]:
        return [ComplexOption(n) for n in range(1000)]

    def compose(self) -> ComposeResult:
        yield OptionList(*self.simple_content, id="simple")
        yield OptionList(*self.complex_content, id="complex")

if __name__ == "__main__":
    OptionListLagApp().run()
```

This is best experienced by running the above code in an environment that has Textual v1.0.0 backing it, vs Textual v3.0.1.

Once the application is run up and ready, start spamming the <kbd>tab</kbd> key. You'll notice that focus movement within the v1.0.0 application is smooth and instant; in the v3.0.1 application it's at best laggy and often just gets "stuck" for a while.

By way of illustration of what I'm seeing, here is the above code running side-by-side, with Textual v3 on the left and v1 on the right: https://www.youtube.com/watch?v=l1m43-OcrAM

[...]

PR #5740, opened 2025-04-12
Optimize Option List style change

willmcgugan:
Avoid clearing all caches when focus style changes.

Fixes https://github.com/Textualize/textual/issues/5729


davep:
Thanks, I'll look forward to giving this a spin.

PR #5748, opened 2025-04-15
refactor: reduce dependency on typing-extensions

TomJGooding:
Remove imports from typing-extensions that are no longer needed since Textual now requires Python 3.8+.

(Inspired by https://github.com/Textualize/rich/pull/3700)

- [ ] Updated documentation

### DIFF
--- a/examples/five_by_five.py
+++ b/examples/five_by_five.py
@@ -15,7 +15,7 @@ from textual.widget import Widget
 from textual.widgets import Button, Footer, Label, Markdown
 
 if TYPE_CHECKING:
-    from typing_extensions import Final
+    from typing import Final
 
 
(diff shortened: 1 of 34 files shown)


TomJGooding:
Currently these imports are still required from typing-extensions:

- Self (added in Python 3.11)
- TypeAlias (added in Python 3.10)
- TypeGuard (added in Python 3.10)
- ParamSpec (added in Python 3.10)

airtower-luna:
Could you maybe wrap the imports in
```py
try:
    from typing import ...
except ImportError:
    from typing_extensions import ...
```
and then limit the typing-extensions dependency to older Python versions? That'd allow systems with newer Python to avoid the dependency entirely.

PR #5751, opened 2025-04-17
Update truecolor_style to use background.rich_color in Tint class

kevinjterry:
Fixed an issue in the Tint.process_segments method where passing a Textual Color object to truecolor_style() instead of a Rich color value was causing the demo game to crash on startup.

✅ pytest tests/renderables/test_tint.py -v


willmcgugan:
Thanks. Fixed in https://github.com/Textualize/textual/pull/5757.

PR #5753, opened 2025-04-20
Create new event loop as fallback behaviour.

paul-ollis:
The method App.push_screen tries to create a Future using `asyncio.get_running_loop()`. If this raises a RuntimeError it simply invokes `Future()`, as a fallback, order to support some of the tests.

When running tests using pytest-xdist it is possible for this fallback mechanism to fail and instead raise another RuntimeError when `Future()` tries to get the 'standard' event loop. This is a rare occurrence with the current tests' structure, but I have seen it a number of times.

It is causes by a test that invokes App.run (which invokes asyncio.run) being executed before a test that depends on the aforementioned fallback mechanism. Both tests must be executed by the same pytest-xdist worker process.

- [n/a ] Docstrings on all new or modified functions / classes 
- [n/a ] Updated documentation
- [n/a ] Updated CHANGELOG.md (where appropriate)

### DIFF
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -314,8 +314,8 @@ class App(Generic[ReturnType], DOMNode):
                 scrollbar-background-active: ansi_default;
                 scrollbar-color: ansi_blue;
                 scrollbar-color-active: ansi_bright_blue;
-                scrollbar-color-hover: ansi_bright_blue;    
-                scrollbar-corner-color: ansi_default;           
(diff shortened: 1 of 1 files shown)


graingert:
This is resolved by https://github.com/Textualize/textual/pull/5765

paul-ollis:
I have converted this to a draft on the assumption that by #5765 will get merged in due course.

PR #5765, opened 2025-04-24
avoid exposing asyncio.Future directly to api consumers

graingert:
This avoids exposing an 'asyncio.Future' to api consumers - instead returning an awaitable object wrapping an Event and a Result.

* This should fix an issue where cancellation would result in an invalid state error
* This work is needed for https://github.com/Textualize/textual/discussions/3#discussioncomment-806536 where asyncio Futures do not work with the trio backend.

I didn't see any documentation for this feature outside docstrings, it's probably worth a note in the changelog

- [ ] Updated documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,11 +5,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
-## [Unreleased]
+## [6.9.0] - 2025-12-14
 
 ### Added
 
+- Added Solarized Dark theme https://github.com/Textualize/textual/pull/6278
 - Added Rosé Pine themes https://github.com/Textualize/textual/pull/6277
+
+### Fixed
+
+- Fixed fuzzy matcher displaying wrong matched characters with simple substring match https://github.com/Textualize/textual/pull/6282
 
 ## [6.8.0] - 2025-12-07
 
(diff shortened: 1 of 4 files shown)


src/textual/app.py
```diff
@@ -2695,6 +2688,7 @@ def push_screen(
         except LookupError:
             message_pump = self.app
 
+        future: AwaitScreen[ScreenResultType] = AwaitScreen()
```
willmcgugan, line 2691: The `future` is no longer a Future. We should rename accordingly...

src/textual/screen.py
```diff
@@ -91,7 +115,7 @@ def __init__(
         self,
         requester: MessagePump,
         callback: ScreenResultCallbackType[ScreenResultType] | None,
-        future: asyncio.Future[ScreenResultType] | None = None,
+        future: AwaitScreen[ScreenResultType] | None = None,
```
willmcgugan, line 118: Same here. Since it is no longer a future, it deserves a better name.

My bad. `future` was a terrible name, since it presumed the implementation.

src/textual/screen.py
```diff
@@ -82,6 +85,27 @@
 ]
 """Type of a screen result callback function."""
 
+class _Unset(enum.Enum):
+    UNSET = enum.auto()
+
+class AwaitScreen(Generic[ScreenResultType]):
```
willmcgugan, line 91: Docstrings please.

willmcgugan (changes requested):
Fine in principle. Just some requests.

src/textual/app.py
```diff
@@ -2695,6 +2688,7 @@ def push_screen(
...
         next_screen._push_result_callback(message_pump, callback, future)
```
graingert, line 2692:
```suggestion
        await_screen: AwaitScreen[ScreenResultType] = AwaitScreen()
        next_screen._push_result_callback(message_pump, callback, await_screen)
```

src/textual/screen.py
graingert, line 91:
```suggestion
class AwaitScreen(Generic[ScreenResultType]):
    """
    An Awaitable object that resumes with the result of a Screen.
    """
```

src/textual/screen.py
graingert: do you have an idea for the name? I was going to go for `await_screen` ?

PR #5772, opened 2025-04-27
start fixing mypy issues

graingert:

- [ ] Updated documentation


src/textual/dom.py
```diff
@@ -1206,12 +1206,11 @@ def ancestors_with_self(self) -> list[DOMNode]:
         Returns:
             A list of nodes.
         """
-        nodes: list[MessagePump | None] = [self]
-        add_node = nodes.append
-        node: MessagePump | None = self
-        while (node := node._parent) is not None:
-            add_node(node)
-        return cast("list[DOMNode]", nodes)
+        nodes = [self]
```
willmcgugan, line 1209: This change will make this method slower. If it is to satisfy Mypy, can we not have both fast and correct?

src/textual/document/_syntax_aware_document.py
```diff
@@ -82,14 +82,16 @@ def query_syntax_tree(
...
         """
-        captures_kwargs = {}
-        if start_point is not None:
-            captures_kwargs["start_point"] = start_point
-        if end_point is not None:
-            captures_kwargs["end_point"] = end_point
-
-        captures = query.captures(self._syntax_tree.root_node, **captures_kwargs)
-        return captures
+
```
willmcgugan, line 85: This change is more than typing. Can you explain why this change was necessary?

src/textual/dom.py
```diff
@@ -1220,12 +1219,11 @@ def ancestors(self) -> list[DOMNode]:
         Returns:
             A list of nodes.
         """
-        nodes: list[MessagePump | None] = []
-        add_node = nodes.append
-        node: MessagePump | None = self
-        while (node := node._parent) is not None:
-            add_node(node)
-        return cast("list[DOMNode]", nodes)
+        nodes = []
```
willmcgugan, line 1222: Same comment as above.

src/textual/drivers/_input_reader.py
```diff
@@ -1,10 +1,11 @@
 import sys
+from typing import Final
 
 __all__ = ["InputReader"]
 
-WINDOWS = sys.platform == "win32"
+WINDOWS: Final = sys.platform == "win32"
 
-if WINDOWS:
+if sys.platform == "win32":
```
willmcgugan, line 8: If we are going to do this, do we need the `WINDOWS` constant?

src/textual/drivers/win32.py
```diff
@@ -162,6 +163,8 @@ def enable_application_mode() -> Callable[[], None]:
         A callable that will restore terminal to previous state.
     """
 
+    assert sys.__stdin__ is not None
```
willmcgugan, line 166: Let's add a message to these asserts. It is possible for `sys.__stdin__` to be `None`, although unlikely for Textual apps.

src/textual/drivers/win32.py
```diff
@@ -217,7 +220,7 @@ def __init__(
         loop: AbstractEventLoop,
         app: App,
         exit_event: threading.Event,
-        process_event: Callable[[Event], None],
+        process_event: Callable[[Message], None],
```
willmcgugan, line 223: Perhaps we should update the name of the parameter to reflect the change in signature, possibly the class as well. i.e. `process_message`, and `MessageMonitor`.

src/textual/message_pump.py
```diff
@@ -243,7 +243,7 @@ def is_attached(self) -> bool:
         except NoActiveAppError:
             return False
         node: MessagePump | None = self
-        while (node := node._parent) is not None:
+        while node is not None and (node := node._parent) is not None:
```
willmcgugan, line 246: This adds an additional check per loop, which is never going to be truthy. I think we should find another way to satisfy Mypy if that is why this was added.

src/textual/reactive.py
```diff
@@ -150,7 +151,7 @@ def __rich_repr__(self) -> rich.repr.Result:
         yield "name", getattr(self, "name", None), None
 
     @classmethod
-    def _clear_watchers(cls, obj: Reactable) -> None:
+    def _clear_watchers(cls, obj: MessagePump) -> None:
```
willmcgugan, line 154: Given that a Reactable is a DOMNode, why is this change neccesary?

src/textual/walk.py
```diff
@@ -17,22 +17,20 @@
     WalkType = TypeVar("WalkType", bound=DOMNode)
 
 
-if TYPE_CHECKING:
```
willmcgugan, line 20: Same comment re overloads. They just add to startup time.

src/textual/widget.py
```diff
@@ -979,15 +986,14 @@ def get_child_by_id(
             )
         return child
 
-    if TYPE_CHECKING:
```
willmcgugan, line 982: Overloads again

src/textual/worker.py
```diff
@@ -306,22 +306,20 @@ def run_callable(work: Callable[[], ResultType]) -> ResultType:
...
             and inspect.iscoroutinefunction(self._work.func)
         ):
-            runner = run_coroutine
-        elif inspect.isawaitable(self._work):
-            runner = run_awaitable
-        elif callable(self._work):
-            runner = run_callable
-        else:
-            raise WorkerError("Unsupported attempt to run a thread worker")
+            return await loop.run_in_executor(None, run_coroutine, self._work)  # type: ignore[arg-type]
```
willmcgugan, line 315: Is this really better? Can't we modify the original in a simpler way to satisfy Mypy?

willmcgugan (commented):
I'm not keen on changes which make things slower purely to satisfy the type checker.

One-off asserts or casts I can accept. But I'd prefer to avoid code within loops that isn't required for runtime.

src/textual/_work_decorator.py
```diff
@@ -33,42 +33,41 @@ class WorkerDeclarationError(Exception):
     """An error in the declaration of a worker method."""
 
 
-if TYPE_CHECKING:
```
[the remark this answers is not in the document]
graingert: This is not supported by mypy, it needs the definition in the same block as the overrides

src/textual/document/_syntax_aware_document.py
graingert: This is needed for the new interface in tree-sitter, it's a breaking change in v24

src/textual/drivers/_input_reader.py
graingert: I suspect the WINDOWS constant is redundant

src/textual/dom.py
graingert: the add_node = node.append trick is actually slower now

src/textual/reactive.py
graingert: _clear_watchers is called with a less specific type than Reactable or DOMNode elsewhere

src/textual/worker.py
graingert: The type for the `runner` local is very complicated and would need to be explicitly stated here

src/textual/_work_decorator.py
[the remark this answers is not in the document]
willmcgugan: Just tried, and Mypy seems happy with the condition. Do we have different versions of Mypy?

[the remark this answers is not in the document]
graingert: I assumed I had the one from the poetry lock file, I'll check

src/textual/drivers/_input_reader.py
willmcgugan: Suggest we remove it.

src/textual/_work_decorator.py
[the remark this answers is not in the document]
graingert: I have 1.14.1

for example when I run mypy I get no-overload-impl for this patten.

```
mypy src/textual/walk.py
src/textual/walk.py:88: error: An overloaded function outside a stub file must have an implementation  [no-overload-impl]
src/textual/walk.py:104: error: Name "walk_breadth_first" already defined on line 88  [no-redef]
Found 2 errors in 1 file (checked 1 source file)
```

however not for this case:

```
mypy src/textual/_work_decorator.py
src/textual/_work_decorator.py:71: error: Type variable "textual._work_decorator.ReturnType" is unbound  [valid-type]
src/textual/_work_decorator.py:71: note: (Hint: Use "Generic[ReturnType]" or "Protocol[ReturnType]" base class to bind "ReturnType" inside a class)
src/textual/_work_decorator.py:71: note: (Hint: Use "ReturnType" in function signature to bind "ReturnType" inside a function)
```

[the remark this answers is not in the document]
graingert: as discussed in Discord - I'll keep the TYPE_CHECKING removals as the import speed difference is negligible

src/textual/dom.py
graingert: I've looked into this and have benchmarked some options:

(there's not all that much in it, even if you really crank up the number of nodes)

https://gist.github.com/graingert/23d43b7fb7c8d5663cc355e8390f19d9

the option I want to go for is:

```python
    nodes = []
    node: DOMNode | None = self

    while True:
        if node is None:
            return nodes

        nodes.append(node)
        node = node._parent
```

some remarks:  repeatedly looking up .append is actually much faster, doing the condition in the body of a while loop is faster this is probably a bug!

PR #5776, opened 2025-04-29
Add support for `first-child` and `last-child` pseudo-classes

sponsfreixes:
They work almost identically than the existing `first-of-type` and `last-of-type`, with the difference that they don't distinguish between widget type.

Based on https://developer.mozilla.org/en-US/docs/Web/CSS/:first-child.

- [x] Updated documentation
- [x] Updated CHANGELOG.md


willmcgugan (approved):
LGTM

willmcgugan:
Thanks

PR #5779, opened 2025-04-30
Breakpoints

willmcgugan:
Adds responsive breakpoints.

### DIFF
--- a/src/textual/screen.py
+++ b/src/textual/screen.py
@@ -196,6 +196,11 @@ class Screen(Generic[ScreenResultType], Widget):
     you can set the [sub_title][textual.screen.Screen.sub_title] attribute.
     """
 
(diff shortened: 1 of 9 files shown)


src/textual/screen.py
```diff
@@ -1354,6 +1359,41 @@ async def _on_resize(self, event: events.Resize) -> None:
...
+            self.app.VERTICAL_BREAKPOINTS
+            if self.VERTICAL_BREAKPOINTS is None
+            else self.VERTICAL_BREAKPOINTS
+        ) or []
+
+        width, height = event.size
+        if horizontal_breakpoints:
+            self._set_breakpoints(width, horizontal_breakpoints)
+        if vertical_breakpoints:
+            self._set_breakpoints(height, horizontal_breakpoints)
```
python-and-novella, line 1378: Should it be `self._set_breakpoints(height, vertical_breakpoints)`?

PR #5780, opened 2025-04-30
Some clarification on smart refresh

davidfokkema:
A bit of clarification on how smart refresh works. It just calls the render() method of the widget where the reactive attribute is defined. It does nothing more. If you have a compound widget, that will not work and you'll need watch methods or use recompose.

- [X] Updated documentation

I did not update the changelog since nothing changed in the code.

### DIFF
--- a/docs/guide/reactivity.md
+++ b/docs/guide/reactivity.md
@@ -65,7 +65,7 @@ You may want to add explicit type hints if the attribute type is a superset of t
 
 ## Smart refresh
 
-The first superpower we will look at is "smart refresh". When you modify a reactive attribute, Textual will make note of the fact that it has changed and refresh automatically.
+The first superpower we will look at is "smart refresh". When you modify a reactive attribute, Textual will make note of the fact that it has changed and refresh automatically. It does this by calling the `render()` method of the widget that contains the reactive. A typical render method will make use of the reactive's value although that is not strictly necessary.
 
 !!! information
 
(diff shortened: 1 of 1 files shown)


docs/guide/reactivity.md
```diff
@@ -96,6 +96,10 @@ The `Name` widget has a reactive `who` attribute. When the app modifies that att
 
     Textual will check if a value has really changed, so assigning the same value wont prompt an unnecessary refresh.
 
+!!! information
+
+    [Compound widgets](widgets.md/#compound-widgets) do not contain a `render()` method, but rather a `compose()` method. Smart refresh will do nothing, but you can use a [watch method](#watch-methods) to run code to update child widgets if needed. Alternatively, you can use [recompose](#recompose) to remove all child widgets and call `compose()` again.
```
willmcgugan, line 101: This isn't quite accurate. Compound widgets do have render methods, and they are called. Although for most compound widgets, the child widgets cover the parent fully so you never see that.

I've updated the initial paragraph, but I feel this may be too much information at this point. I don't like to have two admonitions next to each other. I also prefer to keep things as linear as possible within a single page, so we aren't encouraging the reader to jump down before they have read the initial text.

davidfokkema: Didn't know that the render method of compound widgets was actually called. Makes sense, but in my head, writing widgets, it was always either a render() method _or_ a compose() method. But the one does not exclude the other.

I agree that too much information is not good, and two admonitions are a bit ugly. If you don't want users to jump ahead, I feel that a remark should be made in the relevant sections. I had a bit of trouble with smart refresh as well, and understanding that I should use a watch method for updating child widgets. The word "child" is only first mentioned in the recompose section. But, for most uses, you don't want to use recompose since that is not efficient for simply updating child widgets.

PR #5784, opened 2025-05-02
Fix vertical_breakpoints won't work

python-and-novella:
It should be `self._set_breakpoints(height, vertical_breakpoints)`.


python-and-novella (commented):
Just a small fix.

willmcgugan:
Can we have a test, please?

PR #5785, opened 2025-05-03
Fix breakpoints

python-and-novella:
Fix `VERTICAL_BREAKPOINTS` won't work.


TomJGooding (commented):
It might be better to add a separate snapshot test for vertical breakpoints?

The updated snapshots don't really demonstrate the change in layout, but also don't match the description in the docstring:

https://github.com/Textualize/textual/blob/e5fd661dbe8171443633adf6d195a86be6b16189/tests/snapshot_tests/test_snapshots.py#L4067-L4073

TomJGooding (commented):
LGTM, thanks for adding the snapshot test! But this still will need Will's approval and an update to the CHANGELOG.

willmcgugan:
Please add an entry to the CHANGELOG, and we should be good to go!

python-and-novella:
If someone is waiting for this pr, put following code in the head of source code, that will fix it:
```python3
# patch is here
from textual.app import App
from textual.screen import Screen
from textual import events

class Screen(Screen):
    async def _on_resize(self, event: events.Resize) -> None:
        event.stop()
        self._screen_resized(event.size)
        for screen in self.app._background_screens:
            screen._screen_resized(event.size)

        horizontal_breakpoints = (
            self.app.HORIZONTAL_BREAKPOINTS
            if self.HORIZONTAL_BREAKPOINTS is None
            else self.HORIZONTAL_BREAKPOINTS
        ) or []

        vertical_breakpoints = (
            self.app.VERTICAL_BREAKPOINTS
            if self.VERTICAL_BREAKPOINTS is None
            else self.VERTICAL_BREAKPOINTS
        ) or []

        width, height = event.size
        if horizontal_breakpoints:
            self._set_breakpoints(width, horizontal_breakpoints)
        if vertical_breakpoints:
            self._set_breakpoints(height, vertical_breakpoints)
            
class App(App):
    def get_default_screen(self) -> Screen:
        """Get the default screen.

        This is called when the App is first composed. The returned screen instance
        will be the first screen on the stack.

        Implement this method if you would like to use a custom Screen as the default screen.

        Returns:
            A screen instance.
        """
        return Screen(id="_default")
# patch is over
```

PR #5792, opened 2025-05-10
docs(color): fix hsl order in docstrings

TomJGooding:
Just some minor fixes spotted in passing, where docstrings refer to HLS rather than HSL.


willmcgugan:
thanks

PR #5842, opened 2025-06-04
fix(text area): fix initial flicker in rendering

### LINKED ISSUE #5841: `TextArea` has initial flicker with each character on a new line
When I mount a TextArea, for a split second it has a larger size before snapping to the intended size. Here is a code to illustrate the behaviour:

```python
from textual.app import App
from textual.widgets import Button, TextArea

TEXT = "Lorem ipsum dolor sit amet"


class TestApp(App):
    CSS = """
TextArea {
    height: auto;
}

    """

    def on_button_pressed(self):
        self.mount(TextArea(TEXT), after=0)

    def compose(self):
        yield Button()


TestApp().run()
```

When pressing the button you can see that for a split second every character or `TEXT` is rendered in a new line before it is collapsed to a single line.

How can this behaviour be avoided?

_Originally posted by @lyovushka in https://github.com/Textualize/textual/discussions/5838_

TomJGooding:
Fix initial flicker in the `TextArea`, where for a brief moment each character is rendered on a new line before the size is settled.

The width for the `WrappedDocument` is _slightly_ confusing where zero means no wrapping, but the current negative wrap width looks like the bug.

Fixes #5841

- [ ] Updated documentation

### DIFF
--- a/src/textual/widgets/_text_area.py
+++ b/src/textual/widgets/_text_area.py
@@ -1021,7 +1021,7 @@ TextArea {
         width, _ = self.scrollable_content_region.size
         cursor_width = 1
         if self.soft_wrap:
-            return width - self.gutter_width - cursor_width
+            return max(0, width - self.gutter_width - cursor_width)
         return 0
 
     def _rewrap_and_refresh_virtual_size(self) -> None:


TomJGooding:
I've tested manually obviously with the examples in the linked issue, but I'm not sure it's possible to add a regression test for this flicker.

lyovushka:
I think there is still a flickering even when `soft_wrap=False`.

TomJGooding:
> I think there is still a flickering even when `soft_wrap=False`.

Strange, I don't see the same flicker with this change. Do you mean a different kind of flicker, or do you have an example to demonstrate?

Running your example with some debug logs, here's extracts from the console before and after this change. 

## Before

```
[15:02:19] DEBUG                                              screen.py:1027
Button() was focused
────────────────────────────────────────────────────────────────────────────
[15:02:20] DEBUG                                          _text_area.py:1309
render_line(0): L                                                           
[15:02:20] DEBUG                                          _text_area.py:1309
render_line(1): o                                                           
[15:02:20] DEBUG                                          _text_area.py:1309
render_line(2): r                                                           

...

[15:02:20] DEBUG                                          _text_area.py:1309
render_line(27): m                                                          
[15:02:20] DEBUG                                          _text_area.py:1309
render_line(28): e                                                          
[15:02:20] DEBUG                                          _text_area.py:1309
render_line(29): t                                                          
[15:02:20] DEBUG                                          _text_area.py:1309
render_line(0): Lorem ipsum dolor sit amet                                  
────────────────────────────────────────────────────────────────────────────
```

## After

```
[15:12:00] DEBUG                                              screen.py:1027
Button() was focused
────────────────────────────────────────────────────────────────────────────
[15:12:02] DEBUG                                          _text_area.py:1309
render_line(0): Lorem ipsum dolor sit amet                                  
────────────────────────────────────────────────────────────────────────────
```

lyovushka:
No the letters don't appear to be rendered individually with `soft_wrap=False`. But there is still a flicker. I've attached a screencast (this is without `soft_wrap`). The `TextArea` initially has a larger height before snapping to a single line. It also makes the vertical scroll visible momentarily.

https://github.com/user-attachments/assets/2d4bf280-95ea-457e-a73f-f6cb59942d15

TomJGooding:
Sorry I think I understand what you mean now - you've not tested with this change but just setting `soft_wrap` to False?

Good catch - it looks like the `wrap_width` is being called before this is set!

TomJGooding:
Fixed in 4653f30 by ensuring the attribute is assigned *before* trying to wrap the document.

Again tested manually but don't think it's possible to add a regression test for this.

willmcgugan (approved):
LGTM.

PR #5891, opened 2025-06-25
Simple valid empty fix

holdenweb:

- [n/a] Updated documentation
- [?] Updated CHANGELOG.md (where appropriate)

This PR fixes the error that validates an empty input with no validators as valid even though `valid_empty` is False.

### DIFF
--- a/src/textual/widgets/_input.py
+++ b/src/textual/widgets/_input.py
@@ -149,7 +149,7 @@ class Input(ScrollView):
     | ctrl+k | Delete everything to the right of the cursor. |
     | ctrl+x | Cut selected text. |
     | ctrl+c | Copy selected text. |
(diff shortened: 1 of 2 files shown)


src/textual/widgets/_input.py
```diff
@@ -571,13 +571,14 @@ def set_classes() -> None:
...
-        if not self.validators:
+        # Empty inputs are always OK if valid_empty is set.
+        if self.valid_empty and not value:
             self._valid = True
             set_classes()
             return None
 
-        if self.valid_empty and not value:
+        # If no validators are supplied, and therefore no validation occurs, we return None.
+        if not self.validators:
```
holdenweb, line 581: Since we here have two conditions with the same action, would it overcomplicate the code to elide them? At present I suspect the sequence improves readability.

PR #5905, opened 2025-06-27
docs: add installation guide and Chinese localization

oddune7:
docs(en/zh): update installation guide and add localization

•Added technical feature icons documentation

•Improved prerequisites section in installation guide

•Introduced Chinese (zh) version of the docs


docs.md
```diff
@@ -1,16 +1,16 @@
-# Documentation Workflow
+# Documentation Workflow <!-- by oddune7 -->
```
TomJGooding, line 1: ?!?!

TomJGooding (commented):
Was this generated using an LLM? There's a number of issues in this PR.

If you're actually serious about adding a translation for the README, you should probably check first and start by translating it verbatim.

fancidev:
The Chinese version of the README looks fine to me. But is there really a need to include a Chinese version of the README, given that this library is developer (rather than end-user) oriented?

PR #5909, opened 2025-06-28
Fix valid_empty

holdenweb:

- [✅] Docstrings on all new or modified functions / classes 
- [✅] Updated documentation
- [?] Updated CHANGELOG.md (where appropriate)

This fixes a problem where an `Input` with `valid_empty=False` validates successfully in the absence of any validators.

At present the correct handling of this condition breaks one existing test: `test_focus_within_transparent` fails because the Input box on the left is no longer incorrectly validated, and so has a red outline that was previously absent. I'd like to discuss how best to proceed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -128,3 +128,7 @@ sandbox/
 
 # Used by mkdocs-material social plugin
 .cache
+
+# Wing Pro configuration files
+*.wp[ru]
+
--- a/docs/widgets/input.md
+++ b/docs/widgets/input.md
@@ -114,7 +114,7 @@ as seen for `Palindrome` in the example above.
 
 #### Validate Empty
 
-If you set `valid_empty=True` then empty values will bypass any validators, and empty values will be considered valid.
+Empty values _always_ bypass validation, and are considered valid only when `valid_empty=True`.
 
 ## Reactive Attributes
 
--- a/src/textual/widgets/_input.py
+++ b/src/textual/widgets/_input.py
@@ -149,7 +149,7 @@ class Input(ScrollView):
     | ctrl+k | Delete everything to the right of the cursor. |
     | ctrl+x | Cut selected text. |
     | ctrl+c | Copy selected text. |
-    | ctrl+v | Paste text from the clipboard. | 
+    | ctrl+v | Paste text from the clipboard. |
     """
 
     COMPONENT_CLASSES: ClassVar[set[str]] = {
@@ -187,9 +187,9 @@ class Input(ScrollView):
         }
 
         &:focus {
-            border: tall $border;            
+            border: tall $border;
             background-tint: $foreground 5%;
-            
+
         }
         &>.input--cursor {
             background: $input-cursor-background;
@@ -207,12 +207,12 @@ class Input(ScrollView):
         }
         &.-invalid:focus {
             border: tall $error;
-        }    
+        }
 
         &:ansi {
             background: ansi_default;
             color: ansi_default;
-            &>.input--cursor {     
+            &>.input--cursor {
                 text-style: reverse;
             }
             &>.input--placeholder, &>.input--suggestion {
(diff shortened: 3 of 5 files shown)


holdenweb:
Fixes #5917. Now ready for review. Apologies for the whitespace amendments - I'll try to avoid this in future.

TomJGooding:
Sorry if this has already been discussed over on Discord, but I'm just concerned about the impact of this change.

Currently `valid_empty` is only intended for any input `validators`. This is _implied_ in the docs, but I agree is maybe a bit confusing.

My concern is that the impact of this change will be more surprising than the problem it intended to solve. This would mean all Inputs created with just the defaults are suddenly going to be invalid when empty and display the red border.

holdenweb:
I see your point about the change of behaviour, but would argue that this should be seen as the textual equivalent of the ubiquitous web red asterisk to indicate a required field. I also feel this change would introduce a consistency that is currently lacking: it makes little sense to me that the test added in [the first commit](https://github.com/Textualize/textual/pull/5909/commits/51434b4c1534dac531e6e7a09ec4752c3796e797) should fail.

I don't really understand "Currently valid_empty is only intended for any input validators;" it seems I completely missed the implication, and since explicit is in any case preferable to implicit this would imply at least the need for a docs change. Nor do I appear to understand what use validators are supposed to make of the valid_empty flag, or quite whose benefit it is intended for.

The fact that just one single test broke, and only in a non-functional way, surely implies some measure of safety for this. Alternatively, can this change be tabled for a 4.0 revision as a breaking change, in which valid_empty is replaced by a `required` reactive, with the semantics this change implies for `not valid_empty`? In short, I suppose I'm asking a) if this really should be considered a breaking change, and b) if so, when and whether can it be introduced?

holdenweb:
@willmcgugan I wonder if you have anything to add here?

TomJGooding:
Why don't you consider this a breaking change? Currently an `Input` without any validators is never invalid - that's why I wanted to flag the impact of this change.

holdenweb:
I believe I understand your objections, but would suggest that the current behaviour is a bug rather than a feature, essentially allowing something specifically tagged as `valid_empty=False` to be empty. In which case it would seem that change of some sort is advisable, whether to the documentation or the code.

brent-lidstone:
Id just like to document here that this was indeed discussed on the Discord a couple weeks ago, Will saw it and said this:

> Will McGugan — 2025-06-24 1:50 PM
> I'm wondering if we should ditch valid_empty in favor of required. Even if that would be a breaking change.
> And required would just prepopulate the validators with a Required validators.
>
> Edward Jazzhands — 2025-06-24 1:52 PM
> could you just add in the docstring that valid_empty is being deprecated in favor of required and then make required override valid_empty if it is used? just to make the transition a bit easier
>
> Will McGugan — 2025-06-24 2:03 PM
> You could, but it would still catch people when we do eventually drop valid_empty. I generally prefer to bump the major version and just take the flak.

https://discord.com/channels/1026214085173461072/1033754296224841768/1387112695169417247

willmcgugan:
Sorry I haven't kept up. Bit distracted atm.

You know, I don't think this is desirable behavior either.

If `valid_empty==True` and the input is empty, I think that means set it as valid, and don't run the validators. 

But the converse isn't true. If `valid_empty==False` and the input is empty, you want it to drop through to the validators. So it can say "Number between 1 and 10 required" or whatever.

brent-lidstone:
Yeah that makes perfect sense. `valid_empty==False` should simply defer to the validators. That sounds fairly intuitive to me: "If `valid_empty` is `True`, an empty input will bypass any validators and is considered valid automatically".

holdenweb:
Just before I close this PR, might I suggest that an explanation equivalent to 

> If `valid_empty` is `True`, an empty input will bypass any validators and is considered valid automatically.

be added to the documentation.

TomJGooding:
Like here? https://textual.textualize.io/widgets/input/#validate-empty

holdenweb:
[facepalm]

PR #5925, opened 2025-07-05
Textarea optimizations

willmcgugan:
A few optimizations to TextArea.

Started out as a modest optimization to avoid calculating cell lengths. But turned in to implementing a line cache, which results in a very good win, but potential for breakage. I'd appreciate an extra set of eyes on this.


CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Widget.release_mouse will now only release the mouse, if it was captured by self https://github.com/Textualize/textual/pull/5900
 - Some optimizations to TextArea, which may be noticeable during scrolling (note: may break snapshots with a TextArea) https://github.com/Textualize/textual/pull/5925
 - Selecting in the TextArea now hides the cursor until you release the mouse https://github.com/Textualize/textual/pull/5925
+- Read only TextAreas will no longer display a cursor https://github.com/Textualize/textual/pull/5925
```
TomJGooding, line 20: You can still navigate and select in read-only mode, so don't we need the cursor?

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Widget.release_mouse will now only release the mouse, if it was captured by self https://github.com/Textualize/textual/pull/5900
 - Some optimizations to TextArea, which may be noticeable during scrolling (note: may break snapshots with a TextArea) https://github.com/Textualize/textual/pull/5925
+- Selecting in the TextArea now hides the cursor until you release the mouse https://github.com/Textualize/textual/pull/5925
```
TomJGooding, line 19: I think we also still want the cursor here, as it helps show what exactly is being selected?

For example, try selecting a line with the mouse, starting at the end of the line. If you select only until the start of that line and hit backspace, that leaves a blank line. Whereas if you select until the end of the previous line, the line is removed entirely.

When you're quickly trying to edit, after selecting the text you're probably almost immediately going to hit the next key, which is why I think the cursor context while selecting is important.

willmcgugan: I don't follow. The cursor is always directly under the mouse, which is why it feels redundant. As soon as you release the mouse the cursor reappears.

https://github.com/user-attachments/assets/2cb03500-5781-4e68-9f47-6af5fd87f463

I'm going to merge this shortly. But if I've missed something, I'm happy to revisit it.

PR #5926, opened 2025-07-06
allow negative mouse coordinates

willmcgugan:
Pretty sure this is a bug in Ghostty.

### DIFF
--- a/src/textual/_xterm_parser.py
+++ b/src/textual/_xterm_parser.py
@@ -78,6 +78,9 @@ class XTermParser(Parser[Message]):
             buttons = int(_buttons)
             x = float(int(_x) - 1)
             y = float(int(_y) - 1)
+            if x < 0 or y < 0:
+                # TODO: Workaround for Ghostty erroneous negative coordinate bug
+                return None
             if (
                 self.mouse_pixels
                 and self.terminal_pixel_size is not None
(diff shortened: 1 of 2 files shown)


src/textual/_xterm_parser.py
```diff
@@ -18,7 +18,7 @@
 # to be unsuccessful?
 _MAX_SEQUENCE_SEARCH_THRESHOLD = 32
 
-_re_mouse_event = re.compile("^" + re.escape("\x1b[") + r"(<?[\d;]+[mM]|M...)\Z")
+_re_mouse_event = re.compile("^" + re.escape("\x1b[") + r"(<?[-?\d;]+[mM]|M...)\Z")
```
TomJGooding, line 21: Doesn't this include `?` in the character class?

src/textual/_xterm_parser.py
```diff
@@ -50,7 +50,7 @@
 
 
 class XTermParser(Parser[Message]):
-    _re_sgr_mouse = re.compile(r"\x1b\[<(\d+);(\d+);(\d+)([Mm])")
+    _re_sgr_mouse = re.compile(r"\x1b\[<(-?\d+);(-?\d+);(-?\d+)([Mm])")
```
TomJGooding, line 53: Is the negative button value just in case?

PR #5930, opened 2025-07-06
added textual.getters

willmcgugan:
Good for DRY

- Adds `getters.query_one`
- Adds `getters.child_by_id`

These create nicely typed properties which return a Widget.

```python
class MyWidget(Widget):
    @property
    def my_input(self) -> Input:
        return self.query_one("#my_input", Input)
```

Can be replaced with:

```python
class MyWidget(Widget):
    my_input = getters.query_one("#my_input", Input)
```

Works nicely, just don't look behind the curtain.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Added textual.getters 
 
+### Changed
+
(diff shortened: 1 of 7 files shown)


python-and-novella:
`query_one`'s `__get__` is using `obj.query_one`,
I think it is maybe better for `child_by_id`'s `__get__` to use `obj.get_child_by_id`.

Issue #5928, opened 2025-07-06
Cursor removed in read-only TextArea


TomJGooding:
You can still navigate and select the text in read-only mode, so don't we need the cursor?

_Originally posted by @TomJGooding in https://github.com/Textualize/textual/pull/5925#discussion_r2187634303_

Here's an example to demonstrate:

```python
from textual.app import App, ComposeResult
from textual.widgets import TextArea


class ExampleApp(App):
    def compose(self) -> ComposeResult:
        yield TextArea(
            "Try navigating by holding down the arrow keys. This is a "
            + "very, " * 10
            + "very long line. "
            + "While you're here, try pressing 'ctrl+shift+right'"
            + "Are we nearly there yet? This line seems endless!",
            read_only=True,
            soft_wrap=False,
        )


if __name__ == "__main__":
    app = ExampleApp()
    app.run()
```

darrenburns:
I agree - most times when I use the TextArea in read-only mode I want there to be a cursor so that I can select and copy text (and move the cursor around using keybindings to make selection easier).

davep:
Count me in too; I have one application that makes a point of using `TextArea` in read-only mode but with a visible cursor so the user can select and copy specific text.

willmcgugan:
I was replicating the browser's TextArea, which has no cursor when disabled. For me, cursor means insertion point, but since a read only TextArea can't insert anything, a cursor may be misleading for people who think like me.

So you can still select and copy with the mouse. But I recognize that Textual should be keyboard friendly, and there should ideally be no actions which are mouse only.

For something I was working on, I really don't want a read only cursor. So how about we have `read_only` and an independent `show_cursor` boolean?

I'd also like for a read only cursor to be different somehow. So there is a at least some visual indication that you can't actually insert anything. Different color maybe?

willmcgugan:
Tom's issue was orthogonal. If there is no cursor, then cursor actions and keys should be disabled. And the TextArea should default to the standard ScrollView behaviors with regards to scrolling with the arrow keys.

davep:
> For something I was working on, I really don't want a read only cursor. So how about we have read_only and an independent show_cursor boolean?

> I'd also like for a read only cursor to be different somehow. So there is a at least some visual indication that you can't actually insert anything. Different color maybe?

[...]

PR #5934, opened 2025-07-07
Add show cursor boolean

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/5928

A recent change to hide the cursor in read-only mode, revealed plenty of edge cases.

- Restores cursor to read_only mode.
- Added a `show_cursor` which applies when `read_only==True`. Note that `show_cursor` is ignored when not in read only mode, as being able to make edits without a cursor just seems like sadism.
- When in read only mode the cursor doesn't blink, to give a subtle indication that you can't insert anything.
- When in read only mode with `show_cursor==False` then the cursor keys will scroll the window (i.e. the default behavior for scrollable containers).


src/textual/widgets/_text_area.py
```diff
@@ -373,6 +373,15 @@ class TextArea(ScrollView):
     The document can still be edited programmatically via the API.
     """
 
+    show_cursor: Reactive[bool] = reactive(True)
+    """Show the cursor in read only mode?
+
+    If `False`, the cursor will be visible when `read_only==True`.
+    If `True`, the cursor will not be visible when `read_only==True`, and the TextArea may be
+    scroll like other containers.
```
TomJGooding, line 381: I think you might have mixed up the boolean here?

```suggestion
    show_cursor: Reactive[bool] = reactive(True)
    """Show the cursor in read only mode?

    If `True`, the cursor will be visible when `read_only==True`.
    If `False`, the cursor will not be visible when `read_only==True`, and the TextArea may be
    scrolled like other containers.
```

TomJGooding (commented):
Thanks for revisiting this. I haven't tested the code yet but spotted a possible mistake in the docstring.

Issue #5918, opened 2025-07-02
Keyline rendering appears intermittent under scrolling.


holdenweb:
Have you checked closed issues? (https://github.com/Textualize/textual/issues?q=is%3Aissue+is%3Aclosed) _yes_

Have you checked against the most recent version of Textual? (https://pypi.org/search/?q=textual) _yes_

## The bug

Please give a brief but clear explanation of the issue. If you can, include a complete working example that demonstrates the bug. **Check it can run without modifications.**

The following simple program demonstrates oddities in rendering as the container scrolls.
```
from textual.app import App, ComposeResult
from textual.widgets import Input
from textual.containers import Horizontal, Vertical, VerticalScroll

class MyApp(App):

    DEFAULT_CSS = """\
#my-container {
    keyline: heavy blue;
    Vertical {
        margin: 1;
        width: 1fr;
    height: 3;
    }
}
"""
    def compose(self) -> ComposeResult:
        with VerticalScroll(id="my-container"):
            for i in range(10):
                yield Vertical(Input(valid_empty=False, id=f"test-{i}", value="x"))

if __name__ == '__main__':
    MyApp().run()
```

The initial display is sane.
<img width="405" alt="Image" src="https://github.com/user-attachments/assets/07f58e85-9d15-4762-aa8b-d71c943035bf" />

Pressing the \<down\> key, however, does not redraw correctly.
<img width="405" alt="Image" src="https://github.com/user-attachments/assets/179969e5-273a-4403-b474-010967d463b1" />

A mouse click in the window at that stage triggers a correct redraw. Also please note that with a Vertical height of 3, the window paints correctly every 4 keypresses. Adding one to the height means 5 keypresses are required, and so on.
```
textual diagnose
<!-- This is valid Markdown, please paste the following directly in to a GitHub issue -->
# Textual Diagnostics

## Versions

| Name    | Value  |
|---------|--------|
| Textual | 3.5.0  |
| Rich    | 14.0.0 |

## Python

| Name           | Value                                                         |
|----------------|---------------------------------------------------------------|
| Version        | 3.12.8                                                        |
| Implementation | CPython                                                       |
| Compiler       | Clang 19.1.6                                                  |
| Executable     | /Users/sholden/Projects/Python/textual-forms/.venv/bin/python |

## Operating System

| Name    | Value                                                                                                  |
|---------|--------------------------------------------------------------------------------------------------------|
| System  | Darwin                                                                                                 |
[...]

PR #5936, opened 2025-07-07
Fix keyline glitch

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/5918


willmcgugan:
Can still reproduce it, occasionally.

PR #5943, opened 2025-07-09
fix: fix regexes for mouse escape sequences

TomJGooding:
Fix potential issues with the regexes for mouse escape sequences, which were recently updated to allow negative coordinates.

- Fix the updated character class `[-?\d;]`. This will match the literal character `?` which I assume was not the intention.
- Fix the regex for SGR mouse sequences, where the first parameter is the button value and should not be negative.

_(Note for reviewers: These regexes are the same as my original PR #5688 which was superseded, so if I've just misunderstood why the updated regexes are more correct please let me know!)_

- [ ] Updated documentation

### DIFF
--- a/src/textual/_xterm_parser.py
+++ b/src/textual/_xterm_parser.py
@@ -18,7 +18,7 @@ from textual.message import Message
 # to be unsuccessful?
 _MAX_SEQUENCE_SEARCH_THRESHOLD = 32
 
-_re_mouse_event = re.compile("^" + re.escape("\x1b[") + r"(<?[-?\d;]+[mM]|M...)\Z")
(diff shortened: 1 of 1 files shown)


src/textual/_xterm_parser.py
```diff
@@ -50,7 +50,7 @@
 
 
 class XTermParser(Parser[Message]):
-    _re_sgr_mouse = re.compile(r"\x1b\[<(-?\d+);(-?\d+);(-?\d+)([Mm])")
+    _re_sgr_mouse = re.compile(r"\x1b\[<(\d+);(-?\d+);(-?\d+)([Mm])")
```
willmcgugan, line 53: The irony is that the terminal shouldn't send hyphens in any of those fields. But I guess we haven't seen negative mouse buttons in the wild, so this makes sense.

PR #5945, opened 2025-07-10
Fix query wrong type

willmcgugan:
Thanks to @TomJGooding for spotting a test wasn't running.

Enable the test revealed a few bugs. Namely `WrongType` wasn't being thrown by the original implementation, contrary to the docs.


TomJGooding:
Wasn't expecting this to open a can of worms, sorry!

willmcgugan:
> Wasn't expecting this to open a can of worms, sorry!

I wonder if I subconsciously sabotaged that test, to avoid fixing this?

PR #5948, opened 2025-07-10
Optimize query_one

willmcgugan:
A special case optimization for `query_one` with a simple id selector.

Previously `widget.query_one("#foo")` would search each node in turn until it found a matching ID. In most cases this is quick enough, but here are potential pathological cases where it could be slow.

This update uses the index on the node list to make this much faster.


willmcgugan:
@TomJGooding Want to take a gander?

PR #5950, opened 2025-07-10
WIP Markdown streaming

willmcgugan:
- Adds a Markdown.update method which appends to the Markdown widget
- Change to the anchor API (breaking change)


TomJGooding:
I'd be so tempted to name these methods `drop_anchor` and `weigh_anchor`, so kudos for resisting that urge!

Issue #5955, opened 2025-07-11
screenshot omits Footer text but keeps Foot bar background


friscorose:
## The bug

Screenshot and more pointedly app.screen_stack does not capture Footer text. See discussion with MRE and screenshots in #help-wanted 
https://discord.com/channels/1026214085173461072/1393356222266081441

It will be helpful if you run the following command and paste the results:

```
# Textual Diagnostics

## Versions

| Name    | Value  |
|---------|--------|
| Textual | 2.1.2  |
| Rich    | 13.9.4 |

## Python

| Name           | Value                                                     |
|----------------|-----------------------------------------------------------|
| Version        | 3.13.3                                                    |
| Implementation | CPython                                                   |
| Compiler       | GCC 14.2.0                                                |
| Executable     | /home/frisco/.local/share/uv/tools/textual-dev/bin/python |

## Operating System

| Name    | Value                                               |
|---------|-----------------------------------------------------|
| System  | Linux                                               |
| Release | 6.12.5-amd64                                        |
| Version | #1 SMP PREEMPT_DYNAMIC Debian 6.12.5-1 (2024-12-15) |

## Terminal

| Name                 | Value         |
|----------------------|---------------|
| Terminal Application | tmux (3.5a)   |
| TERM                 | tmux-256color |
| COLORTERM            | truecolor     |
| FORCE_COLOR          | *Not set*     |
| NO_COLOR             | *Not set*     |

## Rich Console options

| Name           | Value                |
|----------------|----------------------|
| size           | width=110, height=43 |
| legacy_windows | False                |
| min_width      | 1                    |
| max_width      | 110                  |
| is_terminal    | True                 |
| encoding       | utf-8                |
| max_height     | 43                   |
| justify        | None                 |
| overflow       | None                 |
| no_wrap        | False                |
| highlight      | None                 |
| markup         | None                 |
| height         | None                 |

```

If you don't have the `textual` command on your path, you may have forgotten to install the `textual-dev` package.

Feel free to add screenshots and / or videos. These can be very helpful!

brent-lidstone:
can confirm I see the same issue. I recreated this entirely on my own so I'm gonna assume there's no need to post any code.

my own screenshot Sample app:
<img width="727" height="464" alt="Image" src="https://github.com/user-attachments/assets/60cb7ee3-ab5d-4c7b-96fe-cf7d6ae51664" />

SVG saved through command palette:
[...]

PR #5956, opened 2025-07-12
add delay to screenshot

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/5955

There is a frame where the footer hasn't been built. This is usually invisible but it is captured by the screenshot. The fix is to add a short delay to allow the full frame to be rendered.


matmaer:
Thank you.

PR #5959, opened 2025-07-12
Add caching for DataTable._get_row_renderables()

fancidev:
The `DataTable` widget appears slow to load and scroll when the number of columns is large. This PR improves the situation by adding a cache for the `_get_row_renderables()` method, which is called $O(m^2)$ times for each rendered row where $m$ is the number of columns. This speeds up the execution of certain benchmarks by 50x.

More details (including a benchmark script) can be found at Discussion #5953.


willmcgugan:
Thanks

PR #5965, opened 2025-07-15
chore(text area): prepare for tree-sitter breaking changes

TomJGooding:
> [!IMPORTANT]  
> DO NOT MERGE!

Temporarily install tree-sitter from GitHub to test against breaking changes before v0.25.0 is released on PyPI.

- [ ] Updated documentation


TomJGooding:
This apparently causes some (strange?) changes in the syntax highlighting, which is why the snapshot tests are failing:

snapshot report; https://github.com/Textualize/textual/actions/runs/16304923238/artifacts/3540275570

TomJGooding:
This might have revealed a possible bug in the current Python syntax highlighting. It looks like any variables are currently being highlighted as `variable.builtin`, whereas I _think_ this should be `variable`.

TomJGooding:
This actually revealed another bug, where the query predicates were being ignored so highlighting much more they should!

After the fix in b85d4f3, the failing snapshots look much better: https://github.com/Textualize/textual/actions/runs/16349424666/artifacts/3556076660

PR #5970, opened 2025-07-17
feat(highlight): use theme text colors

TomJGooding:
The new `highlight` module for syntax highlighting currently uses the 'base' theme colors. Perhaps this was intentional, but to my eye this makes the colors in `Markdown` code blocks so prominent that they feel a bit out of place.

This updates the highlight theme to instead use the dedicated theme text colors, which I think looks more consistent.

- [ ] Updated documentation


willmcgugan:
I'd forgotten about those. Yeah, that looks a little kinder on the eyes.

TomJGooding:
Here's a side-by-side comparison of the Textual demo to demonstrate:

| Before | After |
|--------|-------|
| image       | image      |

TomJGooding:
> I'd forgotten about those. Yeah, that looks a little kinder on the eyes.

I'm glad you agree as honestly I was a bit hesitant to make this PR. Probably I'm just a bit fixated on theme colors after working on my theme editor!

I'm not really a fan of the new inline code colors either, but that's obviously just personal taste.

willmcgugan:
How would you improve the inline code style?

TomJGooding:
@willmcgugan What was wrong with the old colors? I'm used to `inline code looking like this` so the change to warning colors just looks strange to me.

willmcgugan:
Previous version was bold and dim, IIRC. I didn't feel it had enough emphasis. And obviously it already has a monospace font, so there was nothing to say this was code.

PR #5977, opened 2025-07-21
fix: bump tree sitter to 0.25.0

TomJGooding:

- [ ] Updated documentation


TomJGooding:
This causes a change in the snapshot test for Python syntax highlighting which I'm struggling to understand.

### Before

snapshot before

### After

snapshot after

It looks like this is highlighting 'range' as `type.builtin` rather than `function.call`. 

But if you comment out some of the code above, now the highlighting is the same:

snapshot with commented out code

TomJGooding:
I think this `query_syntax_tree` method hasn't been working as intended for a while. It looks like the 'range' arguments were removed from `Query.captures` back in tree-sitter v0.23.0.

[EDIT: Looking closer, it doesn't look like Textual ever actually used the range arguments in this method. ~I'm not sure whether to just remove them or try to fix this?~ I decided worth fixing anyway in case it might help future improvements]

https://github.com/Textualize/textual/blob/c9439ff36d3810641e2cff35f318571e51f5ff2d/src/textual/document/_syntax_aware_document.py#L85-L91

TomJGooding:
For reference here's the latest snapshot report from CI: https://github.com/Textualize/textual/actions/runs/16452663173/artifacts/3590409485

The failed snapshots look fine for the most part (and actually fixed a few bugs), but unfortunately there seems to be some quirks [as mentioned above](https://github.com/Textualize/textual/pull/5977#issuecomment-3101803684).

TomJGooding:
It looks like the syntax highlighting was possibly already broken...

```python
PYTHON_EXAMPLE = """\
# Uncomment the line below
# foo = 69

print("Hello world")
x = range(10)
"""


class TextAreaSnapshot(App):
        yield TextArea.code_editor(
            PYTHON_EXAMPLE,
            language="python",
        )


app = TextAreaSnapshot()
if __name__ == "__main__":
    app.run()
```

image

image

TomJGooding:
I _think_ I'm finally starting to understand the issues with the highlighting.

Consider this Python code:

```python
print("Hello world")
x = range(10)
```

The tree-sitter query captures (on this branch) look something like this.[^1] Notice that `function.call` is *before* `type.builtin`, since 'print' was matched first.

```
[('variable', ['print', 'range', 'x']),
 ('function.call', ['print', 'range']),
 ('function.builtin', ['print', 'range']),
 ('punctuation.bracket', ['(', '(', ')', ')']),
 ('string', ['"Hello world"']),
 ('operator', ['=']),
 ('type.builtin', ['range']),
 ('number', ['10'])]
```

After removing the print function:

```python
x = range(10)
```

Notice how the order of the captures has changed, where now `function.call` is *after* `type.builtin`

```
[('variable', ['x', 'range']),
 ('operator', ['=']),
 ('type.builtin', ['range']),
 ('function.call', ['range']),
 ('function.builtin', ['range']),
 ('punctuation.bracket', ['(', ')']),
 ('number', ['10'])]
```

The `TextArea` loops over the captures to build the 'highlight map':

https://github.com/Textualize/textual/blob/c9439ff36d3810641e2cff35f318571e51f5ff2d/src/textual/widgets/_text_area.py#L665-L673

When the `TextArea` lines are rendered, it loops over the line highlights to apply the syntax highlighting. This explains why 'range' will be highlighted differently depending on whether there's another function call before it!

https://github.com/Textualize/textual/blob/c9439ff36d3810641e2cff35f318571e51f5ff2d/src/textual/widgets/_text_area.py#L1296-L1304

[^1]: `QueryCursor.captures` actually returns "A dict where the keys are the names of the captures and the values are lists of the captured nodes". I've formatted the captures to make them simpler to understand.

TomJGooding:
I've updated the snapshots and marked this as ready for review.

Since it turns out the syntax highlighting is already broken, I think better to prioritise fixing the tree-sitter crash and revisit these issues later. Bumping tree-sitter actually fixes some issues with incorrect highlighting.

Note I still need to update the CHANGELOG if this is approved. This might be considered a breaking change since the syntax extras would now require Python >=3.10,

willmcgugan:
Thanks for putting in the leg work on this one. Will get this in the new release (out soon).

willmcgugan:
Thanks, Tom

TomJGooding:
No problem. Sorry I didn't manage to get back to the CHANGELOG before this was merged - is it worth adding something now (and maybe also the release notes)?

PR #5985, opened 2025-07-25
console options optimization

willmcgugan:
According to @fancidev auto-detecting the console size is slow.

This change avoid that by using Textual's cached size.

Probably not a big win, as the new Content class doesn't call `console.size`. But might help anywhere we render Rich renderables.


fancidev:
Thanks for the quick fix!

Issue #5998, opened 2025-07-28
[BUG] Emojis in Markdown headers cause a crash


bootloop-chris:
## The bug

Including an emoji (specifically "🤖") in a markdown header causes a crash.

## Minimal Repro

Tested against v5.0.1 and main@e3bae0071311d30a89b756919f1fe4800bb6a7cf

```python
from textual.app import App, ComposeResult
from textual.widgets import Markdown


class TestApp(App):
    """Markdown append bug repro"""

    def compose(self) -> ComposeResult:
        yield Markdown("## 🤖 This heading is broken\n")

if __name__ == "__main__":
    app = TestApp()
    app.run()
```

## Traceback

(captured in the console using markdown append, doesn't fail as loudly as the above repro but is easier to copy/paste the exception which is the same)

```
future: <_GatheringFuture finished exception=BadIdentifier("'heading-%F0%9F%A4%96-this-heading-is-broken-4381065296' is an invalid id; identifiers must contain only letters,
numbers, underscores, or hyphens, and must not begin with a number.")>
Traceback (most recent call last):
  File "/Users/chris/bootloop/textual/src/textual/widgets/_markdown.py", line 1416, in await_append
    new_blocks = list(self._parse_markdown(tokens))
  File "/Users/chris/bootloop/textual/src/textual/widgets/_markdown.py", line 1274, in _parse_markdown
    block.id = f"heading-{slug(block._content.plain)}-{id(block)}"
    ^^^^^^^^
  File "/Users/chris/bootloop/textual/src/textual/dom.py", line 783, in id
    check_identifiers("id", new_id)
    ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^
  File "/Users/chris/bootloop/textual/src/textual/dom.py", line 98, in check_identifiers
    raise BadIdentifier(
    ...<2 lines>...
    )
textual.dom.BadIdentifier: 'heading-%F0%9F%A4%96-this-heading-is-broken-4381065296' is an invalid id; identifiers must contain only letters, numbers, underscores, or hyphens, and
must not begin with a number.
```

## Textual Diagnose Output

# Textual Diagnostics

## Versions

| Name    | Value  |
|---------|--------|
| Textual | 5.0.1  |
| Rich    | 14.1.0 |

## Python

| Name           | Value                                                    |
|----------------|----------------------------------------------------------|
| Version        | 3.13.5                                                   |
| Implementation | CPython                                                  |
[...]

PR #6006, opened 2025-07-31
Address two crashing markdown rendering bugs

corrylc:
This is still in progress, and I am not entirely sure how the project would want to address these bugs.

Both of these fixes address crashes for certain markdown renders in my app.

The two issues spotted are:
* Markdown headers can have illegal characters, as `quote()` will return `%` strings which are not permitted by the dom.
* The `_get_style()` function has a code path where `visual_style` is undefined, causing a crash.

The `_get_style()` fix is straightforward enough, I think, though I don't understand the larger context enough to say if this is ideal.

The Markdown header fix is kinda over the top, and I suspect there is a simpler solution that y'all would be comfortable with.

### DIFF
--- a/src/textual/_slug.py
+++ b/src/textual/_slug.py
@@ -50,6 +50,34 @@ WHITESPACE_RE: Final[Pattern] = compile(r"\s")
 """A regular expression for finding all the whitespace and turning it into `REPLACEMENT`."""
 
 
+SLUG_ALLOWED = compile(r"[a-zA-Z0-9_\-]")
+def slugify_hex(s: str, *, max_len: int | None = None) -> str:  # py312+
+    """
+    Convert *s* to only contain A-Z a-z 0-9 _ -.
+
+    • Allowed chars are kept verbatim.
+    • Every other byte (after UTF-8 encoding) becomes '_' + two upper-case hex digits.
+      e.g. space (0x20) → '_20', '❤' (0xE2 0x9D 0xA4) → '_E2_9D_A4'.
+
+    If *max_len* is set and the result would exceed it,
+    the tail is replaced by '-' + 8-char SHA-256 digest to keep it unique.
+    """
+    out_parts: list[str] = []
+    for b in s.encode():
+        ch = chr(b)
+        if SLUG_ALLOWED.fullmatch(ch):
+            out_parts.append(ch)
+        else:
+            out_parts.append(f"_{b:02X}")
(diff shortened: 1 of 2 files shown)


src/textual/widget.py
```diff
@@ -1185,6 +1185,7 @@ def _get_style(self, style: VisualStyle | str) -> VisualStyle:
         if style.startswith("."):
             for node in self.ancestors_with_self:
                 if not isinstance(node, Widget):
+                    visual_style = None
```
willmcgugan, line 1188: This was fixed in another PR

willmcgugan (changes requested):

src/textual/_slug.py
```diff
@@ -50,6 +50,34 @@
 """A regular expression for finding all the whitespace and turning it into `REPLACEMENT`."""
 
 
+SLUG_ALLOWED = compile(r"[a-zA-Z0-9_\-]")
+def slugify_hex(s: str, *, max_len: int | None = None) -> str:  # py312+
```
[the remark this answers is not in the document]
corrylc: Nod, agreed, this was definitely a case of an LLM going to town. Will simplify it a bit

src/textual/_slug.py
```diff
@@ -68,7 +96,8 @@ def slug(text: str) -> str:
         (WHITESPACE_RE, WHITESPACE_REPLACEMENT),
     ):
         result = rule.sub(replacement, result)
-    return quote(result)
+    hexed_slug = slugify_hex(result)
```
davep, line 99: This breaks some of the tests for `slug`.

It's worth keeping in mind that `slug` was designed to (and is documented as being designed to) work as close as possible to the way that GitHub flavoured Markdown generates its slugs. You can see some of the more extreme examples that were kept in mind at https://github.com/davep/markdown-header-slug-explorer -- any change to the code that breaks that design aim will in turn potentially break applications that relied on this design promise in Textual.

TomJGooding: I think the problem is that these slugs were never designed to be used for widget IDs. This bug was introduced recently in Textual v5.0.0 with the markdown streaming for LLMs.

I've drafted PR #6019 to remove emojis like ROBOT FACE :robot: as reported in issue #5998, but this doesn't address the underlying problem with widget IDs.

willmcgugan:
Thanks. This has been fixed in a recent release.

PR #6019, opened 2025-08-05
fix(slug): remove all Supplemental Symbols in slug

TomJGooding:
Fix the `slug` function to remove *all* emojis in the Supplemental Symbols and Pictographs Unicode block.

I'm not sure why originally only a partial range was removed, but this change seems to align with GitHub-flavored Markdown.

While this will effectively fix issue #5998, it does not address the underlying problem where slugs were never designed for widget IDs.

- [ ] Updated documentation

### DIFF
--- a/src/textual/_slug.py
+++ b/src/textual/_slug.py
@@ -34,10 +34,10 @@ NONLINGUAL: Final[str] = (
     r"\U000024C2-\U0001F251"
     r"\U00002702-\U000027B0"
     r"\U0001F1E0-\U0001F1FF"
-    r"\U0001F300-\U0001F5FF"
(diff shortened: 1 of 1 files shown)


davep:
To answer the question about a "partial range": no intentional reason. It's a moment since the code was written https://github.com/Textualize/textual/commit/edc0420a5aa64cd7b9467b02d09c44cc8fcb47f7 but I seem to recall the list as was was the most comprehensive version we could find at the time. It was always envisaged (and is documented) that it approach would be refined over time as any exceptions were discovered.

TomJGooding:
Thanks @davep. I'm really not trying to blame anyone and I realise this code was written almost two years ago!

Issue #5969, opened 2025-07-17
On Windows 11, pressing the 'escape' key in browser via textual-serve doesn't register until the mouse is moved or the 'escape' key is pressed again


QuantumForced:
On Windows 11 (and possibly other versions too), pressing the 'escape' key in browser via textual-serve doesn't register until the mouse is moved or until the 'escape' key is pressed again. Also, after pressing `escape` once, all further key strokes are "ignored" until the mouse is moved or the 'escape' key is pressed again, only then do all the keystrokes register. I'm not sure what is ultimately responsible for this: textual, the browser, or textual-serve.

Here's a tiny MRE to demonstrate this bug:
```python
from textual.app import App
from textual import events

class TestApp(App):
    def on_key(self, event: events.Key) -> None:
        self.notify(str(event.key))

if __name__ == "__main__":
    TestApp().run()
```
`textual serve mre.py`

The following sequence can be followed: `Esc, a, b, c, <mouse move> OR Esc`. The `a`, `b`, `c` notifications will not show till the end of the sequence.

At the time of this report, I'm running: textual-dev 1.7.0, textual-serve 1.1.2 and textual 4.0.0

`textual diagnose`:
### Versions

| Name    | Value  |
|---------|--------|
| Textual | 4.0.0  |
| Rich    | 14.0.0 |

### Python

| Name           | Value                                                                |
|----------------|----------------------------------------------------------------------|
| Version        | 3.11.0                                                               |
| Implementation | CPython                                                              |
| Compiler       | MSC v.1933 64 bit (AMD64)                                            |
| Executable     | C:\Users\_hidden_\PycharmProjects\_hidden_\venv\Scripts\python.exe |

### Operating System

| Name    | Value      |
|---------|------------|
| System  | Windows    |
| Release | 10         |
| Version | 10.0.22631 |

### Terminal

| Name                 | Value     |
|----------------------|-----------|
| Terminal Application | *Unknown* |
| TERM                 | *Not set* |
| COLORTERM            | *Not set* |
| FORCE_COLOR          | *Not set* |
| NO_COLOR             | *Not set* |

### Rich Console options

| Name           | Value                |
|----------------|----------------------|
| size           | width=135, height=32 |
| legacy_windows | False                |
| min_width      | 1                    |
| max_width      | 135                  |
| is_terminal    | True                 |
| encoding       | utf-8                |
| max_height     | 32                   |
| justify        | None                 |
| overflow       | None                 |
| no_wrap        | False                |
[...]

PR #6031, opened 2025-08-07
Fix ESCAPE key processing with web_driver under Windows

fancidev:
Closes gh-5969.

gh-5969 reports that running ESC key presses are not processed immediately when running textual in the web under Windows. This is due to the `InputReader` under Windows [performing a blocking read](https://github.com/Textualize/textual/blob/da3f608fd7dab10be67f5a4cdf18c57b37b7ca6f/src/textual/drivers/_input_reader_windows.py#L28) without honoring the timeout. This PR fixes the `InputReader`, which in turn fixes gh-5969.

- [ ] Updated documentation

### DIFF
--- a/src/textual/drivers/_input_reader_windows.py
+++ b/src/textual/drivers/_input_reader_windows.py
@@ -1,6 +1,7 @@
 import os
 import sys
-from threading import Event
+from queue import Empty, Queue
+from threading import Event, Thread
 from typing import Iterator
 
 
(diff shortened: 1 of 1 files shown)


fancidev:
I made a new commit that properly implements cancellation for blocking read of anonymous pipe under Windows. The code follows the approach of [DotNET](https://github.com/dotnet/runtime/blob/1209ddec456f1d7871847b07b46f00d44012d687/src/libraries/Common/src/System/Threading/AsyncOverSyncWithIoCancellation.cs), where the key is to use the `CancelSynchronuousIo` function. This seems to be the only way to support asynchronous operation on anonymous pipe properly. (An alternative is busy-polling using `PeekNamedPipe`, which wastes CPU time.)

Issue #6030, opened 2025-08-07
[BUG] Pretty widget not visible with `width:auto` when in a wrapper that also has `width:auto`


MatejManczal:
I tried to make a snug wrapper for multiple vertically laid out Pretty widgets but upon setting its and its containers widths to auto, the pretty widgets disappeared. In the attached image the left column showcases mentioned behavior, the right column showcases the same code with Labels instead of Pretty widgets for comparison

App: top-left
```py
from textual.containers import Container, Vertical, Horizontal
from textual.widgets import Label, Pretty
from textual.app import App, ComposeResult


class Demo(App):
    CSS = """
    Vertical {
        background: blue;
        height: auto;
    }

    Pretty {
        width: auto;
        height: auto;
        background: red;
    }
    """

    def compose(self) -> ComposeResult:
        with Vertical():
            yield Pretty('hello world')
            yield Pretty('hello world')
            yield Pretty('hello world')




if __name__ == "__main__":
    Demo().run()
```

App: bottom-left
```py
from textual.containers import Container, Vertical, Horizontal
from textual.widgets import Label, Pretty
from textual.app import App, ComposeResult
from textual import events


class Demo(App):
    CSS = """
    Vertical {
        background: blue;
        height: auto;
        width: auto;
    }

    Pretty {
        width: auto;
        height: auto;
        background: red;
    }
    """

    def compose(self) -> ComposeResult:
        with Vertical():
            yield Pretty('hello world')
            yield Pretty('hello world')
            yield Pretty('hello world')




if __name__ == "__main__":
    Demo().run()
```

App: top-right
```py
from textual.containers import Container, Vertical, Horizontal
from textual.widgets import Label, Pretty
from textual.app import App, ComposeResult
from textual import events


class Demo(App):
    CSS = """
    Vertical {
        background: blue;
        height: auto;
    }

    Label {
        width: auto;
        height: auto;
        background: red;
    }
    """

    def compose(self) -> ComposeResult:
        with Vertical():
            yield Label('hello world')
            yield Label('hello world')
            yield Label('hello world')




if __name__ == "__main__":
    Demo().run()
```

App: bottom-right
```py
from textual.containers import Container, Vertical, Horizontal
from textual.widgets import Label, Pretty
from textual.app import App, ComposeResult
from textual import events


class Demo(App):
    CSS = """
    Vertical {
        background: blue;
        height: auto;
        width: auto;
    }

    Label {
        width: auto;
        height: auto;
        background: red;
    }
[...]

PR #6040, opened 2025-08-12
fix for pretty

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/6030

### DIFF
--- a/src/textual/widgets/_pretty.py
+++ b/src/textual/widgets/_pretty.py
@@ -38,4 +38,12 @@ class Pretty(Static):
             classes: The CSS classes of the pretty.
         """
         super().__init__(name=name, id=id, classes=classes)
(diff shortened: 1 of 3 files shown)


TomJGooding:
Yikes, are you sure this is the proper solution? https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides

willmcgugan:
Yeah, its not great. It would be better to pull out the relevant functionality from Static rather than subclass. Don't suppose you are up for a PR?

TomJGooding:
> Don't suppose you are up for a PR?

I almost did make a PR when I saw this issue to simply add `shrink=False`, but it depends what you mean by other "relevant functionality".

Besides, I don't think the author of the issue will agree this is resolved as they have different expectations for how wrapping should work when the width is auto.

PR #6041, opened 2025-08-13
don't ineherit from Static

willmcgugan:
- Don't inherit Pretty from Static
- Renamed `Static.renderable` to `Static.content`


TomJGooding:
The `Static` widget has changed a lot recently so maybe I just haven't kept up, but I don't understand why there's an `update` method and now a `content` setter? Is this just for backward compatibility?

You can also get weird results like this:

```python
static = Static("Old content")
static.content = "New content"
assert static.content == "New content"
# AssertionError: assert 'Old content' == 'New content'
```

PR #6052, opened 2025-08-18
Update _progress_bar.py

cartland:
This library crashes when total_imaginary_width == 0.0 because the modulo cannot divide by zero. Although it's probably a bug if the a user of the library is trying to render something with width == 0, the library doesn't need to crash here.

We can use start = 0.0 as a reasonable value for a width of 0.

I came across this issue because I am using a collection of libraries that ultimately depend on this library. Since I'm not an owner of the libraries using this API, I thought it would be easiest to address this crash instead of changing the usage of libraries that depend on this one.

- [?] Docstrings on all new or modified functions / classes: I don't know enough about the context for render_indeterminate() or RenderResult to provide a useful docstring. I am just trying to fix a crash.
- [?] Updated documentation: I do not know enough about the documentation to determine if this needs additional documentation.
- [?] Updated CHANGELOG.md (where appropriate): I do not know enough about the prevalence of this issue to know if it's worth a CHANGELOG.md update.

### DIFF
--- a/src/textual/widgets/_progress_bar.py
+++ b/src/textual/widgets/_progress_bar.py
@@ -127,7 +127,11 @@ class Bar(Widget, can_focus=False):
         else:
             speed = 30  # Cells per second.
             # Compute the position of the bar.
-            start = (speed * self._clock.time) % (2 * total_imaginary_width)
+            start = (
+                (speed * self._clock.time) % (2 * total_imaginary_width)
+                if total_imaginary_width != 0.0
+                else 0.0
+            )
             if start > total_imaginary_width:
                 # If the bar is to the right of its width, wrap it back from right to left.
                 start = 2 * total_imaginary_width - start  # = (tiw - (start - tiw))


willmcgugan:
I'll need a regression test for this.

cartland:
Makes sense. I'm not sure I'll have time to create a regression test. I'm unblocked in my workflow and wanted to share in case it is helpful for others. Since my work is a few steps removed from this library, I am not very familiar with the conditions to trigger this issue or how to write tests. I also don't see others reporting the same issue in the issue tracker, so maybe it only happens under a rare condition. Feel free to close this if you decide not to maintain it, or we can leave it pending here in case I have some time in the coming weeks to figure out how to write some tests.

TomJGooding:
I have no idea what the library using Textual is doing, but here's a simple MRE if it helps:

```python
from textual.widgets import ProgressBar


class ProgressBarApp(App):
    CSS = """
    ProgressBar > Bar { width: 0; }
    """

        yield ProgressBar()


if __name__ == "__main__":
    ProgressBarApp().run()
```

willmcgugan:
As far as I can tell, @TomJGooding 's example is the only way to invoke the crash. And that requires some inside knowledge to do. Its not enough to set the ProgressBar to width 0, as Textual will just not render it. So I don't know how this crash can occur in a non-contrived example.

@cartland It is very frustrating if you tell us Textual crashes, but don't give us any information how to reproduce it--or even it what way it crashes.

cartland:
I suggest closing this pull request. If this particular divide by zero crash only happens in contrived examples then maybe something in our configuration of multiple libraries is doing something weird. I'm sorry I don't have more details to share. The specific configuration is confidential and I didn't know how to create a minimal steps to reproduce. I created this pull request to show a potential divide by zero crash, but if this doesn't appear to be an issue affecting other users of the library then you're probably better off closing this  until someone has a more specific reason to make the change.

PR #6080, opened 2025-08-28
fix(app): fix garbled inline app output

### LINKED ISSUE #6064: Inline app output after exit is garbled for Input widget
For an inline app that I'm building I want to keep the content on screen after exiting, so I'm using `inline_no_clear=True`.

For the `Input` widget this flag garbles the terminal output.
For example, the program below:

```python
from textual.widgets import Button, Input


class ExampleApp(App):
        yield Input("Press ctrl+q to quit")


if __name__ == "__main__":
    app = ExampleApp()
    app.run(inline=True, inline_no_clear=True)
```

produces:

Image

after pressing `ctrl+q`

# Textual Diagnostics

## Versions

| Name    | Value  |
|---------|--------|
| Textual | 5.3.0  |
| Rich    | 14.1.0 |

## Python

| Version        | 3.13.7                                                               |
| Compiler       | Clang 17.0.0 (clang-1700.0.13.3)                                     |
| Executable     | /Users/rob/projects/robvanderleek/inquirer-textual/.venv/bin/python3 |

## Operating System

| Name    | Value                                                                                                   |
|---------|---------------------------------------------------------------------------------------------------------|
| System  | Darwin                                                                                                  |
| Release | 24.6.0                                                                                                  |
| Version | Darwin Kernel Version 24.6.0: Mon Jul 14 11:30:29 PDT 2025; root:xnu-11417.140.69~1/RELEASE_ARM64_T6000 |

## Terminal

| Name                 | Value              |
|----------------------|--------------------|
| Terminal Application | iTerm.app (3.5.14) |
| TERM                 | xterm-256color     |
| COLORTERM            | truecolor          |
| FORCE_COLOR          | *Not set*          |
| NO_COLOR             | *Not set*          |

## Rich Console options

| size           | width=80, height=25 |
| max_width      | 80                  |
| is_terminal    | False               |
| max_height     | 25                  |

TomJGooding:
Fix garbled inline app output when `inline_no_clear=True`.

Currently the terminal cursor isn't actually "reset" before printing the screen contents. This doesn't matter for most inline apps as the cursor is already offset at (0, 0). But if the app contains an `Input` widget, this will update the cursor position so the output will be garbled.

Fixes #6064

- [ ] Updated documentation

### DIFF
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -3294,8 +3294,9 @@ class App(Generic[ReturnType], DOMNode):
                         if self._driver.is_inline:
                             cursor_x, cursor_y = self._previous_cursor_position
                             self._driver.write(
-                                Control.move(-cursor_x, -cursor_y + 1).segment.text
+                                Control.move(-cursor_x, -cursor_y).segment.text
                             )
+                            self._driver.flush()
                             if inline_no_clear and not self.app._exit_renderables:
                                 console = Console()
                                 try:


TomJGooding:
Example app for testing:

```python


class ExampleApp(App):
        yield Input("Press ctrl+q to quit")

    def on_mount(self) -> None:
        # Workaround possible bug with the initial input cursor position
        # https://github.com/Textualize/textual/issues/6064#issuecomment-3217177954
        self.simulate_key("right")

        # Show the terminal cursor for debugging
        self._driver.write("\x1b[?25h")


if __name__ == "__main__":
    app = ExampleApp()
    app.run(
        inline=True,
        inline_no_clear=True,
    )
```

TomJGooding:
Currently there is an empty line after the app output, which is removed with this change. But I'm not sure if this bottom 'padding' was intentional or not, since the [`INLINE_PADDING`](https://textual.textualize.io/api/app/#textual.app.App.INLINE_PADDING) attribute is the "number of blank lines **above** an inline app".

I've marked as ready for review - I'm annoyed that I spent so long debugging this for what might be a simple 4 line change!

willmcgugan:
> I've marked as ready for review - I'm annoyed that I spent so long debugging this for what might be a simple 4 line change!

It's not the total number of lines that counts, but knowing which 4 lines to change!

PR #6109, opened 2025-09-07
fix issue rendering strips with missing style

willmcgugan:
Background https://github.com/Textualize/textual/discussions/6095


davidfokkema:
This is _great_! I haven't had the time to file an issue and now it is already fixed! This bug broke textual-hires-canvas and thus also textual-plot. They both work again with 6.2.0 🎉

PR #6124, opened 2025-09-18
fix(text area): fix rendering with a multi-line placeholder

### LINKED ISSUE #6122: `TextArea.placeholder` renders incorrectly if it contains newline characters
I was trying to add a little bit of help text to a TextArea in Posting, as it's prime real estate for it :)

### Example

```python


class TextAreaPlaceholderMultipleLines(App[None]):
        yield TextArea(placeholder="Line 1\nLine 2\nLine 3")


if __name__ == "__main__":
    app = TextAreaPlaceholderMultipleLines()
    app.run()
```

### Result

Image

TomJGooding:
Fixes #6122

- [ ] Updated documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 
 - Fixed issue where Segments with a style of `None` aren't rendered https://github.com/Textualize/textual/pull/6109
+- Fixed visual glitches and crash when changing `DataTable.header_height` https://github.com/Textualize/textual/pull/6128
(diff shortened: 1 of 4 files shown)


willmcgugan:
@TomJGooding Would I be correct in thinking that this won't wrap the placeholder? I.e. if it is wider than the text area, it will crop the lines.

The original didn't. But I think if we're going to support multi-line placeholders we may as well go the whole hog and implement wrapping. Hope you don't mind, but I think I might tackle that now. If it isn't too challenging, I should have another PR shortly.

willmcgugan:
Closing in favor of https://github.com/Textualize/textual/pull/6138

PR #6131, opened 2025-09-22
test(modal): change click to mouse down in flaky test

TomJGooding:
`test_modal_pop_screen()` is apparently still flaky in CI even after increasing the pauses in #6116.

Try changing the Footer `click` to `mouse_down` instead to avoid the `NoWidget` error in the pilot mouse events.

- [ ] Updated documentation

### DIFF
--- a/tests/test_modal.py
+++ b/tests/test_modal.py
@@ -92,7 +92,7 @@ async def test_modal_pop_screen():
         await pilot.pause(0.4)
         await app.wait_for_refresh()
         # Check clicking the footer brings up the quit screen
-        await pilot.click(Footer)
+        await pilot.mouse_down(Footer)
         await pilot.pause()
         assert isinstance(pilot.app.screen, QuitScreen)
         # Check activating the quit button exits the app


tests/test_modal.py
```diff
@@ -92,7 +92,7 @@ async def test_modal_pop_screen():
         await pilot.pause(0.4)
         await app.wait_for_refresh()
         # Check clicking the footer brings up the quit screen
-        await pilot.click(Footer)
+        await pilot.mouse_down(Footer)
```
willmcgugan, line 95: Doesn't this change the nature of the test significantly? IIRC a click occurs on mouse up.

willmcgugan:
I'm hoping that https://github.com/Textualize/textual/pull/6137 will provide a better solution to avoiding jank in the footer that doesn't introduce flakiness to the tests.

tests/test_modal.py
willmcgugan: Scratch that. I see that the Footer reacts to MouseDown, not Click.

willmcgugan:
I *think* https://github.com/Textualize/textual/pull/6137 resolves this.

TomJGooding:
My thinking is that the `NoWidget` error is actually on the mouse *up*, rather than the mouse *down* of the click, since now the screen is the modal screen.

TomJGooding:
@willmcgugan Since this still doesn't look fixed, can I suggest at least trying this?

[EDIT: Exploring this in new PR #6142]

```python
        # Pause to ensure the footer is fully composed to avoid flakiness in CI
        await pilot.pause(0.4)
        await app.wait_for_refresh()
        # Check clicking the footer brings up the quit screen
        footer_key_clicked = await pilot.mouse_down("FooterKey")
        assert footer_key_clicked is True # Sanity check
        await pilot.pause()
        assert isinstance(pilot.app.screen, QuitScreen)
```

Issue #6127, opened 2025-09-19
`ProgressBar` can be caused to crash with CSS since 5.0.0


davep:
Given this code:

```python
from textual.app import App, ComposeResult
from textual.widgets import ProgressBar

class WideBarApp(App[None]):

    CSS = """
    ProgressBar Bar {
        width: 1fr;
    }
    """

    def compose(self) -> ComposeResult:
        yield ProgressBar()

if __name__ == "__main__":
    WideBarApp().run()
```

Up until Textual 4.x, it ran fine. As of Textual 5.0 it causes the following crash (trimmed to the last bit, because the traceback ends up being way too big for a GitHub issue; to the point that it refuses to accept it):

```
│ /Users/davep/temp/wide-bar/.venv/lib/python3.11/site-packages/textual/widgets/_progress_bar.py:100 in render        │
│                                                                                                                     │
│    97 │   def render(self) -> RenderResult:                                                                         │
│    98 │   │   """Render the bar with the correct portion filled."""                                                 │
│    99 │   │   if self.percentage is None:                                                                           │
│ ❱ 100 │   │   │   return self.render_indeterminate()                                                                │
│   101 │   │   else:                                                                                                 │
│   102 │   │   │   bar_style = (                                                                                     │
│   103 │   │   │   │   self.get_component_rich_style("bar--bar")                                                     │
│                                                                                                                     │
│ ╭─────── locals ───────╮                                                                                            │
│ │ self = Bar(id='bar') │                                                                                            │
│ ╰──────────────────────╯                                                                                            │
│                                                                                                                     │
│ /Users/davep/temp/wide-bar/.venv/lib/python3.11/site-packages/textual/widgets/_progress_bar.py:128 in               │
│ render_indeterminate                                                                                                │
│                                                                                                                     │
│   125 │   │   else:                                                                                                 │
│   126 │   │   │   speed = 30  # Cells per second.                                                                   │
│   127 │   │   │   # Compute the position of the bar.                                                                │
│ ❱ 128 │   │   │   start = (speed * self._clock.time) % (2 * total_imaginary_width)                                  │
│   129 │   │   │   if start > total_imaginary_width:                                                                 │
│   130 │   │   │   │   # If the bar is to the right of its width, wrap it back from right to le                      │
│   131 │   │   │   │   start = 2 * total_imaginary_width - start  # = (tiw - (start - tiw))                          │
│                                                                                                                     │
│ ╭─────────────── locals ────────────────╮                                                                           │
[...]

PR #6139, opened 2025-09-28
fix for progress bar crash

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/6127

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -19,8 +19,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        #os: [ubuntu-latest, windows-latest, macos-latest]
-        os: [windows-latest]
+        os: [ubuntu-latest, windows-latest, macos-latest]
         python-version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
     defaults:
       run:


TomJGooding:
> Pilot.click will return True if the initial mouse down is on the specified target

Doesn't this risk false positives? For example, if start clicking a button but it somehow moves before mouse up, this wouldn't actually produce a click in the app?

willmcgugan:
> > Pilot.click will return True if the initial mouse down is on the specified target
> 
> Doesn't this risk false positives? For example, if start clicking a button but it somehow moves before mouse up, this wouldn't actually produce a click in the app?

The purpose of that parameter is just to tell the pilot where to click. The return value is a sanity check to confirm that the widget was at least under the "mouse" when the call started.

TomJGooding:
> The purpose of that parameter is just to tell the pilot where to click.

But a click is actually more than just a mouse down, right?

Here's an (admittedly contrived!) example. I would expect this to fail since the button is never actually clicked, but if I've understood this change correctly, the check would now confusingly pass?

```python
from textual.widgets import Button, Static


class Example(Static, can_focus=True):
    DEFAULT_CSS = """
    Example {
        color: $text-accent;
        margin: 1 0;
        &:focus { text-style: bold underline; }
    }
    """

    def on_blur(self) -> None:
        self.display = False


class ExampleApp(App):
    click_count = 0

        yield Example("Click the button")
        yield Button()

    def on_button_clicked(self) -> None:
        self.click_count += 1


async def test_example_app():
    app = ExampleApp()
        # This SHOULD fail
        assert await pilot.click(Button)
        assert app.click_count == 1


if __name__ == "__main__":
    app = ExampleApp()
    app.run()
```

willmcgugan:
> But a click is actually more than just a mouse down, right?

Yes.

The `pilot.click` call returns `True` because there was a Button there to be clicked. That's all the return value confirms.

Don't think of it as delivering a Click event to a widget. It's more that the user clicked that button. Even if the button moves or disappears after the mouse down, from the user's POV they still clicked it.

When you write a test, you still have to confirm it had the intended effect.

It's always failing on Windows btw. Do you have any insight why? I want to solve it without an arbitrary pause.

TomJGooding:
Okay, perhaps I've misunderstood the purpose then, But my reason for questioning this change is because the current behaviour is what helped me understand these flaky tests recently involving the footer.

Sorry, no idea why it always seems to fail on Windows.

PR #6142, opened 2025-09-29
test(modal): add sanity check for flaky test in CI

TomJGooding:
Note that the test will even fail locally with this change instead:

```diff
-        footer_key_clicked = await pilot.mouse_down("FooterKey")
+        footer_key_clicked = await pilot.click("FooterKey")
         assert footer_key_clicked is True  # Sanity check
         await pilot.pause()
```

```
>           assert footer_key_clicked is True  # Sanity check
E           assert False is True
```

Since the footer reacts to mouse *down*, when the click has finished on mouse *up* the screen is now the modal screen.

### DIFF
--- a/tests/test_modal.py
+++ b/tests/test_modal.py
@@ -89,7 +89,7 @@ async def test_modal_pop_screen():
     app = ModalApp()
     async with app.run_test() as pilot:
         # Pause to ensure the footer is fully composed to avoid flakiness in CI
-        await pilot.pause()
+        await pilot.pause(0.4)
         await app.wait_for_refresh()
         # Check clicking the footer brings up the quit screen
         footer_key_clicked = await pilot.mouse_down("FooterKey")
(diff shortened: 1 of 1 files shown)


TomJGooding:
And fails in CI with this change:

```diff
-        await pilot.pause(0.4)
+        await pilot.pause()
```

```
FAILED tests/test_modal.py::test_modal_pop_screen - textual.css.query.NoMatches: No nodes match 'FooterKey' on Screen(id='_default')
```

TomJGooding:
Maybe there is a better fix, but this should hopefully unblock for the time being and at least the sanity check will help debug if the footer key is _actually_ clicked.

willmcgugan:
Going with the other branch. Although the fix isn't all that satisfactory, it avoids the arbitrary sleep which is often a red flag and addresses some genuine issues.

PR #6152, opened 2025-10-02
 Use alias _App for App in type hints under TYPE_CHECKING

uriyyo:
Hi,

First of all, thanks for such a great library, really enjoy using it 🙌 

I have found small issue. `App` imported in `__init__` of `textual`, but it imported only in `if TYPE_CHECKING` block, because of that mypy and other tools think that `from textual import App` is valid code.

```py
from textual import App
```
```
$ mypy example.py
Success: no issues found in 1 source file
```

This PR will import `App` as `_App`, so IDEs and other tools will show an error for:
```py
from textual import App
```

- [x] Updated documentation


willmcgugan:
Thanks. This feels like something the linters should understand. It's also a common pattern.

PR #6153, opened 2025-10-02
Use `os.get_terminal_size` as the primary

wsxrdv:
The function `shutil.get_terminal_size` returns environment variables `COLUMNS` and `LINES` when those exist. But this information is stale after resizing the terminal. First try to retrieve from `os.get_terminal_size` and only after fall back to `shutil`.

- [ ] Updated documentation

### DIFF
--- a/src/textual/drivers/linux_driver.py
+++ b/src/textual/drivers/linux_driver.py
@@ -108,7 +108,7 @@ class LinuxDriver(Driver):
         import shutil
 
         try:
-            width, height = shutil.get_terminal_size()
+            width, height = os.get_terminal_size()
         except (AttributeError, ValueError, OSError):
             try:
                 width, height = shutil.get_terminal_size()
--- a/src/textual/drivers/linux_inline_driver.py
+++ b/src/textual/drivers/linux_inline_driver.py
@@ -66,7 +66,7 @@ class LinuxInlineDriver(Driver):
         import shutil
 
         try:
-            width, height = shutil.get_terminal_size()
+            width, height = os.get_terminal_size()
         except (AttributeError, ValueError, OSError):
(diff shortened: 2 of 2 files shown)


wsxrdv:
Previous version of `src/textual/drivers/linux_driver.py` had `os.get_terminal_size` but was changed by https://github.com/Textualize/textual/commit/d1235e0d97f71120b7ca9cdeeb6443045fcd4dc3 All other changes I found were directly adding the `shutil` version.

wsxrdv:
This change allows proper reactivity to resize on Ubuntu inside tmux. One can also test these changes using the following script while resizing:

```python
import os
import shutil
import time

while True:
  # os.get_terminal_size changes with resize
  # shutil.get_terminal_size uses env variables (is constant)
  print(f"{os.get_terminal_size() = } {shutil.get_terminal_size() = }")
  time.sleep(1)
```

wsxrdv:
The Rich project also uses `os.get_terminal_size`: https://github.com/Textualize/rich/blob/ea9d4db5d84b4e834979304e3053bf757daae322/rich/console.py#L1030

wsxrdv:
I tried to dig deeper:

- Debian 13 seems that neither COLUMNS nor LINES are present in Python os.environ and thus shutil is giving always the correct result.
- Jupyterlab shell with tmux run inside from a tensorflow/tensorflow:latest-gpu docker seems to have both COLUMNS and LINES in the os.environ and thus shutil gives the wrong result.
- Possible workaround (without changing textual) is removing COLUMNS and LINES from os.environ before starting the app:
```python
for d in ["LINES", "COLUMNS"]:
    if d in os.environ:
        del os.environ[d]
```

PR #6160, opened 2025-10-05
Fix str() + Content()

gvx:
`Content.__radd__` does not currently behave as expected. For example:

```
>>> from textual.content import Content
>>> Content('foo') + 'bar'  # works as expected
Content('foobar')
>>> 'foo' + Content('bar') # expected: Content('foobar')
Content('barfoo')
```

The problem is that `Content.__radd__` (which implements  `other + self`) returns `self + other`, which is incorrect for concatenation. This pull requests fixes that and adds tests for it. (Existing tests only test for the empty string, for which `a + b == b + a` happens to hold, unlike most strings.)


willmcgugan:
thanks

PR #6179, opened 2025-10-18
fix: monkey-patch breakpoint builtin function

Pablito2020:
When you run the App, calling the builtin function breakpoint() causes the interface to freeze since it waits for user input that can’t be handled properly in this context (Ctrl-C doesn't work, Ctr-Z doesn't work, etc...)

To prevent this, I monkey-patched the breakpoint() behavior with a function that raises a RuntimeError instead. This makes the issue explicit and avoids the confusing situation of an unresponsive UI.

This is specially useful in cases where a leftover breakpoint() call remains in code that runs within an "obscure and rarely used" part of the UI. Without this, the app would silently hang, making it hard to understand what went wrong.

- [x] Updated documentation

### DIFF
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -8,6 +8,7 @@ See [app basics](/guide/app) for how to build Textual apps.
 from __future__ import annotations
 
 import asyncio
+import builtins
 import importlib
 import inspect
 import io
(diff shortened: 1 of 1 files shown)


willmcgugan:
This will break non-cli debuggers.

Suggest you add a linting rule if you are concerned about forgotten breakpoints.

PR #6224, opened 2025-11-11
fix: add set_options to SelectionList

NSPC911:
threw me off when this happened to rovr, because the self._selected never got cleared when `set_options` was used instead of `clear_options` + `add_options`

- [ ] Updated documentation

### DIFF
--- a/src/textual/widgets/_selection_list.py
+++ b/src/textual/widgets/_selection_list.py
@@ -713,3 +713,16 @@ class SelectionList(Generic[SelectionType], OptionList):
         self._selected.clear()
         self._values.clear()
         return super().clear_options()
+
+    def set_options(self, options: Iterable[SelectionType]) -> Self:
+        """Set options, potentially clearing existing options.
+
+        Args:
+            options: Options to set.
+
+        Returns:
+            The `SelectionList` instance
+        """
+        self._selected.clear()
+        self._values.clear()
+        return super().set_options(options)


TomJGooding (commented):
Are you sure this is correct? [You can build a `SelectionList` with tuples or Selection objects](https://textual.textualize.io/widgets/selection_list/#examples), but here it looks like you're trying to use [SelectionType](https://textual.textualize.io/widgets/selection_list/#textual.widgets.selection_list.SelectionType)?

NSPC911:
> Are you sure this is correct? [You can build a `SelectionList` with tuples or Selection objects](https://textual.textualize.io/widgets/selection_list/#examples), but here it looks like you're trying to use [SelectionType](https://textual.textualize.io/widgets/selection_list/#textual.widgets.selection_list.SelectionType)?

not really sure which one to use, given that `OptionList`'s `set_option` method only takes in an Iterable `OptionListContent`

TomJGooding:
I'm confused, Did you check your code actually worked, or just throw the PR over the wall?

NSPC911:
The general code worked, but I'm not sure about which class to use as the type hint for the function

TomJGooding:
Sorry, I think I understand the problem now.

I thought the signature would just be similar to the `SelectionList.add_options` method:

https://github.com/Textualize/textual/blob/422852a1b74f97fa7770afedd8ae917cb0bde66c/src/textual/widgets/_selection_list.py#L629-L637

But of course this doesn't satisfy mypy and friends for `super().set_options(options)`.

I realise now that I overlooked that this actually ends up calling `SelectionList.add_options`. Repeating the same runtime checks in both `set_option` and `add_options` would definitely be "sort of sub-optimal"!

NSPC911:
@willmcgugan could you take a look at this?

NSPC911:
@willmcgugan really sorry to ping you but can you take a look at this PR? Textual currently assumes that when you call `set_options`, the `values` or `selected` options' values are not reset, which means it is using an outdated dictionary, and hence will cause issues when parsing the `selected` values.

willmcgugan:
> @willmcgugan really sorry to ping you but can you take a look at this PR? Textual currently assumes that when you call `set_options`, the `values` or `selected` options' values are not reset, which means it is using an outdated dictionary, and hence will cause issues when parsing the `selected` values.

I'll need regression test.

Also please add a concrete explanation of the problem. You say "cause issues". What issues exactly?

I will overlook GitHub issues where after reading I'm still not clear on the problem or the fix.

NSPC911:
@willmcgugan I made a super simple POC at https://github.com/NSPC911/textual-trials/blob/master/selection-set-options.py

If you run it, you can press some options and view the values of the selected options. If a `set_options` is triggered, calling `selected` results in an outdated list. However, this gets fixed when a Selection is toggled.

PR #6223, opened 2025-11-11
test: fix flaky test_focus_on_click snapshot test

TomJGooding:
Remove the button animation in the `test_focus_on_click` snapshot test causing flakiness in CI.

**Please review the following checklist.**

- [ ] Docstrings on all new or modified functions / classes 
- [ ] Updated documentation
- [ ] Updated CHANGELOG.md (where appropriate)

PR #6226, opened 2025-11-12
Fix double mention of `DeliveryComplete`

davep:
I should have noticed this when quickly doing #6225. The documentation for `name` in `Driver.deliver_binary` calls out `DeliveryComplete` twice; I would imagine this is really meant to be a call out to `DeliveryComplete` and `DeliveryFailed`.


davep:
Given this was a docstring tweak... looks like the fail might be a flakey test?

TomJGooding:
> Given this was a docstring tweak... looks like the fail might be a flakey test?

This flaky test is fixed by #6223 when Will does his next round of reviewing PRs.

PR #6240, opened 2025-11-20
docs: fix miscellaneous typos

TomJGooding:
I spotted a few typos in passing and grepped to check for similar ones.

(Sorry if this seems a bit pedantic, but I thought it better to address these typos all at once rather than potentially having multiple PRs when others might notice them.)


willmcgugan (approved):
Thanks. Always happy to merge typo PRs.

PR #6251, opened 2025-11-27
TextArea: remove right padding to align vertical scrollbar with other and add testing

### LINKED ISSUE #5954: Inconsistent scrollbar location
Especially given the work that was done a wee while back to harmonise focus/cursor/etc look/feel across various widgets, I'm raising this as as "should this also be consistent too?" query.

Given this code:

```python
from textual.widgets import TextArea, Tree, OptionList, Label

class ScrollBarLocationApp(App[None]):

    CSS = """
    Screen > * {
        width: 1fr;
        height: 1fr;
        border: none;
        background: $surface;
        &:focus {
            background: $panel;
        }
    }
    """

        yield TextArea()
        yield Tree("")
        yield OptionList()
        yield VerticalScroll(id="scroller")

    def on_mount(self) -> None:
        self.query_one(TextArea).load_text("\n".join(f"This is line {n}" for n in range(1000)))
        self.query_one(Tree).root.expand()
        for n in range(1000):
            self.query_one(Tree).root.add(f"This is node {n}")
        self.query_one(OptionList).add_options(f"This is option {n}" for n in range(1000))
        self.query_one("#scroller").mount_all(
            Label(f"This is label {n}") for n in range(1000)
        )

if __name__ == "__main__":
    ScrollBarLocationApp().run()
```

we get this display:

Image

Out of the box I think folk would normally expect all of the scrollbars to line up. This can, of course, be fixed by tweaking the `padding` on the offending widgets, but perhaps it would make sense to have all widgets that can potentially add a scrollbar have them line up under most reasonable circumstances?

frankhan19991001:
… widgets (fixes #5954)

- [ ] Updated documentation

### DIFF
--- a/tests/text_area/test_scrollbar_alignment.py
+++ b/tests/text_area/test_scrollbar_alignment.py
@@ -1,0 +1,72 @@
+from __future__ import annotations
+
+import pytest
+
+from textual.app import App, ComposeResult
+from textual.widgets import TextArea, Tree, OptionList
+from textual.containers import VerticalScroll
+
+
+class ScrollBarLocationApp(App[None]):
+    CSS = """
+    Screen > * {
+        width: 1fr;
+        height: 1fr;
+        border: none;
+        background: $surface;
+        &:focus {
+            background: $panel;
+        }
+    }
+    """
+
+    def compose(self) -> ComposeResult:
+        yield TextArea(id="ta")
+        yield Tree("", id="tree")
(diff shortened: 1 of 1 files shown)


TomJGooding:
Did you generate this PR using AI? The test code is peculiar to say the least.

I think best to wait for feedback from a maintainer on this issue, rather than try to make this sort of PR.

PR #6274, opened 2025-12-08
fix(syntax): remove aggressive python type inference rules

nikhil2004-blip:
### Problem

The current python.scm syntax highlighting queries are too aggressive. They use broad regular expressions to match any identifier starting with a capital letter as a @type or @constructor.

This results in incorrect syntax highlighting behavior, such as:

Variables like Final_Score

Function calls

Even commented text

…being highlighted as Types or Constructors simply because they start with a capital letter.
This leads to inconsistent "rainbow coloring" and reduces readability for Python code displayed inside TextArea.

### Solution

The heuristic rules depending solely on regex matching for types and constructors have been disabled.
These were removed in:

src/textual/tree-sitter/highlights/python.scm

Removed queries:

((identifier) @type
  (#match? @type "^[A-Z].*[a-z]"))

((call
  function: (identifier) @constructor)
  (#match? @constructor "^[A-Z]"))

This ensures Tree-sitter relies more on proper semantic analysis instead of unreliable capitalization guessing.

The result:
✔ Cleaner syntax highlighting
✔ More accurate code representation
✔ No more distracting false-type coloring

- [ ] Updated documentation

### DIFF
--- a/src/textual/tree-sitter/highlights/python.scm
+++ b/src/textual/tree-sitter/highlights/python.scm
@@ -10,14 +10,14 @@
 (interpolation) @none
 
 ;; Identifier naming conventions
-((identifier) @type
- (#match? @type "^[A-Z].*[a-z]"))
-((identifier) @constant
- (#match? @constant "^[A-Z][A-Z_0-9]*$"))
-
-((attribute
-    attribute: (identifier) @field)
- (#match? @field "^([A-Z])@!.*$"))
+;; ((identifier) @type
+;;  (#match? @type "^[A-Z].*[a-z]"))
+;; ((identifier) @constant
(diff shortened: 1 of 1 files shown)


TomJGooding:
> Even commented text [is incorrectly] highlighted as Types or Constructors simply because they start with a capital letter.

I can't reproduce this?

image


```python

CODE = """\
# lower

# Title

# UPPER\
"""


class ExampleApp(App):
    CSS = """
    TextArea { 
        width: 40; 
        height: auto;
    }
    """

        yield TextArea(CODE, language="python")


if __name__ == "__main__":
    app = ExampleApp()
    app.run()
```

TomJGooding:
Closing as this isn't the correct fix and likely just drive-by AI slop.

PR #6278, opened 2025-12-10
added solarized dark theme

akhenakh:
tested with vibe-mistral

Screenshot 2025-12-09 at 21 01 58


akhenakh:
note that I had to set: export COLORTERM=truecolor to get the real colors

willmcgugan:
Looks great. Thanks!

> note that I had to set: export COLORTERM=truecolor to get the real colors

This is the only way to know if truecolor is supported. Most shells should set this automatically.

PR #6284, opened 2025-12-15
Docs: Clarify CSS variable usage scope (Fixes #3951)

### LINKED ISSUE #3951: The documentation for CSS variables should clearly state the scope of variables
The [documentation for CSS variables](https://textual.textualize.io/guide/CSS/#css-variables) doesn't appear to say what the scope of those variables is. For example, it would seem that if you're using `CSS_PATH`, you can't defined a series of variables in one file earlier in the path, and then use those variables in files later on in the path. If this is working as intended we should document this.

Hari-varshan:
Fixes #3951

### Summary
Clarified the documentation regarding CSS variables. While variables are globally defined in the Theme, this update adds a note explaining that their **usage** follows standard CSS inheritance rules:
- Values are inherited by descendants (waterfall).
- Values are isolated from siblings.

*Note: Skipped `CHANGELOG.md` update as this is a documentation-only clarification.*

### DIFF
--- a/docs/guide/CSS.md
+++ b/docs/guide/CSS.md
@@ -465,6 +465,15 @@ If we decide we want to change some aspect of our design in the future, we only 
 
     Variables can only be used in the _values_ of a CSS declaration. You cannot, for example, refer to a variable inside a selector.
 
+!!! note "Variable Scope & Inheritance"
+
+    When you **use** a variable in a CSS rule (e.g., `color: $text-muted;`), it behaves like any other CSS property:
+    
+    1.  **Inheritance:** The value is applied to the widget and inherited by its children (descendants), unless the children override it.
+    2.  **Isolation:** The value does **not** affect siblings or parents.
+    
+    Variables themselves are global (defined by the Theme), but their *application* is scoped to the widget tree where they are used.
+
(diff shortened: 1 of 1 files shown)


TomJGooding:
I don't think this is the 'variable scope' referred to in the issue:

> For example, it would seem that if you're using `CSS_PATH`, you can't define a series of variables in one file earlier in the path, and then use those variables in files later on in the path.

Explaining basic CSS rules seems pretty redundant. But I strongly suspect this PR was just generated with AI.

Hari-varshan:
Hi, thanks for your feedback. I'm new to contributions. I picked this issue because I felt it was easy and thought the requirement was just a basic CSS rule. Yes, I used AI to draft the sentences for the documentation.

After reading your comment carefully, I realized the issue is about file scope. I ran some local tests and confirmed that CSS_PATH does not act as a collective store that inherits variables from one file to the next; instead, it isolates the file variables separately.

Is my understanding correct now? Should I close this PR and open a new one?

TomJGooding:
If you're serious about making your first contribution to open source. you should start by actually understanding the issue rather than simply using AI for a quick win.

PR #5340, opened 2024-12-04
Input selection

darrenburns:
**Please review the following checklist.**

- [x] Docstrings on all new or modified functions / classes 
- [ ] Updated documentation
- [x] Updated CHANGELOG.md (where appropriate)

Issue #6280, opened 2025-12-11
MaskedInput doesn't skip the first character if it's a separator


James-San:
I was trying to create a MaskedInput instance to record the phone number, since I eventually want app to include multiple countries phone numbers, I've added the country number in front. Since then mask would look like +99 999-999-999 for my country, I've entered that as a mask for MaskedInput. It turns out if it's not focused as first widget on the screen, when you drop into it by tabbing - it replaces the + in the mask once you start to type instead of skipping over it and replacing first character that should be modified. I've tried escaping the + even though it shouldn't be any special character for MaskedInput according to the doc's but it doesn't help. 
I attach the example python code to run it: 
``` python
from textual.app import App, ComposeResult
from textual.widgets import Input, MaskedInput

class ExampleApp(App):
    def compose(self) -> ComposeResult:
        yield Input(placeholder="Just something to grab focus first")
        yield MaskedInput("+99 999-999-999")


def main():
    app = ExampleApp()
    app.run()


if __name__ == "__main__":
    main()
```
as well as screenshots of what happens:

<img width="807" height="164" alt="Image" src="https://github.com/user-attachments/assets/256256d8-f499-48a3-a3d6-597bd0b7a573" />
So far so good:

<img width="807" height="164" alt="Image" src="https://github.com/user-attachments/assets/94ea72c4-53b0-41ef-85af-20ca02405514" />

As soon as I type - 9 overrides + which should be skipped 

<img width="807" height="164" alt="Image" src="https://github.com/user-attachments/assets/76639c0f-0034-4c3b-8c59-c3fa65c90293" />

And as I backspace to it twice, it actually shows up as it should by replacing 9 with a +:

<img width="807" height="164" alt="Image" src="https://github.com/user-attachments/assets/7a60c433-c8b2-4083-9ea0-c035d45740db" />

TomJGooding:
You could try adding `select_on_focus=False` to your `MaskedInput` as a sort of workaround. 

There's a few related issues with this widget:

- #5493
- #5495

PR #6287, opened 2025-12-15
fix(masked input): fix selection

### LINKED ISSUE #5493: MaskedInput ValueError Value does not match template

### LINKED ISSUE #5495: `MaskedInput` does not highlight the selected text

TomJGooding:
When text selection was added to the `Input` widget in #5340, it was unfortunately overlooked that `MaskedInput` inherits from this widget.

- Fixes #5495
- Fixes #5493
- Fixes bindings to move and select

- [ ] Updated documentation

### DIFF
--- a/src/textual/widgets/_masked_input.py
+++ b/src/textual/widgets/_masked_input.py
@@ -18,7 +18,7 @@ if TYPE_CHECKING:
 
 from textual.reactive import Reactive, var
 from textual.validation import ValidationResult, Validator
-from textual.widgets._input import Input
+from textual.widgets._input import Input, Selection
 
 InputValidationOn = Literal["blur", "changed", "submitted"]
 """Possible messages that trigger input validation."""
@@ -703,8 +703,15 @@ class MaskedInput(Input, can_focus=True):
         Args:
             select: If `True`, select the text to the left of the cursor.
         """
+        start, end = self.selection
         cursor_position = self._template.move_cursor(-1)
-        self.cursor_position = cursor_position
+        if select:
+            self.selection = Selection(start, cursor_position)
+        else:
+            if self.selection.is_empty:
+                self.cursor_position = cursor_position
+            else:
+                self.cursor_position = min(start, end)
 
     def action_cursor_right(self, select: bool = False) -> None:
         """Move the cursor one position to the right; separators are skipped.
@@ -712,8 +719,15 @@ class MaskedInput(Input, can_focus=True):
         Args:
             select: If `True`, select the text to the right of the cursor.
         """
+        start, end = self.selection
         cursor_position = self._template.move_cursor(1)
-        self.cursor_position = cursor_position
+        if select:
+            self.selection = Selection(start, cursor_position)
+        else:
+            if self.selection.is_empty:
+                self.cursor_position = cursor_position
+            else:
+                self.cursor_position = max(start, end)
 
     def action_home(self, select: bool = False) -> None:
         """Move the cursor to the start of the input.
@@ -722,7 +736,10 @@ class MaskedInput(Input, can_focus=True):
             select: If `True`, select the text between the old and new cursor positions.
         """
         cursor_position = self._template.move_cursor(-len(self.template))
-        self.cursor_position = cursor_position
+        if select:
+            self.selection = Selection(self.cursor_position, cursor_position)
+        else:
+            self.cursor_position = cursor_position
 
     def action_cursor_left_word(self, select: bool = False) -> None:
         """Move the cursor left next to the previous separator. If no previous
@@ -732,12 +749,22 @@ class MaskedInput(Input, can_focus=True):
             select: If `True`, select the text between the old and new cursor positions.
         """
         if self._template.at_separator(self.cursor_position - 1):
-            position = self._template.prev_separator_position(self.cursor_position - 1)
+            separator_position = self._template.prev_separator_position(
+                self.cursor_position - 1
+            )
         else:
-            position = self._template.prev_separator_position()
-        if position:
-            position += 1
-        self.cursor_position = position or 0
+            separator_position = self._template.prev_separator_position()
+
+        if separator_position is None:
+            cursor_position = 0
+        else:
+            cursor_position = separator_position + 1
+
+        if select:
+            start, _ = self.selection
+            self.selection = Selection(start, cursor_position)
+        else:
+            self.cursor_position = cursor_position
 
     def action_cursor_right_word(self, select: bool = False) -> None:
         """Move the cursor right next to the next separator. If no next
@@ -746,11 +773,17 @@ class MaskedInput(Input, can_focus=True):
         Args:
             select: If `True`, select the text between the old and new cursor positions.
         """
-        position = self._template.next_separator_position()
-        if position is None:
-            self.cursor_position = len(self._template.mask)
+        separator_position = self._template.next_separator_position()
+        if separator_position is None:
+            cursor_position = len(self._template.mask)
         else:
-            self.cursor_position = position + 1
+            cursor_position = separator_position + 1
+
+        if select:
+            start, _ = self.selection
+            self.selection = Selection(start, cursor_position)
+        else:
+            self.cursor_position = cursor_position
 
     def action_delete_right(self) -> None:
         """Delete one character at the current cursor position."""
--- a/src/textual/widgets/_masked_input.py
+++ b/src/textual/widgets/_masked_input.py
@@ -572,12 +572,19 @@ class MaskedInput(Input, can_focus=True):
             if char == " ":
                 result.stylize(style, index, index + 1)
 
-        if self._cursor_visible and self.has_focus:
-            if self.cursor_at_end:
-                result.pad_right(1)
-            cursor_style = self.get_component_rich_style("input--cursor")
-            cursor = self.cursor_position
-            result.stylize(cursor_style, cursor, cursor + 1)
+        if self.has_focus:
+            if not self.selection.is_empty:
+                start, end = self.selection
+                start, end = sorted((start, end))
+                selection_style = self.get_component_rich_style("input--selection")
+                result.stylize_before(selection_style, start, end)
+
+            if self._cursor_visible:
+                cursor_style = self.get_component_rich_style("input--cursor")
+                cursor = self.cursor_position
+                if self.cursor_at_end:
+                    result.pad_right(1)
+                result.stylize(cursor_style, cursor, cursor + 1)
 
         segments = list(result.render(self.app.console))
         line_length = Segment.get_line_length(segments)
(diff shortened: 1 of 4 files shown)


TomJGooding:
A `MaskedInput` can end up something like this after deleting sections:

image

After selecting all text - for example on focus or pressing <kbd>shift</kbd>+<kbd>end</kbd> - currently the empty text isn't highlighted:

image

This is confusing since it suggests the selection starts at 'E' rather than the start of the input. But after pressing a key to replace the selected text:

image

I think the empty text should also be highlighted to accurately reflect the current selection.

---

**EDIT:** I'm a bit rusty on how styles are applied with component classes so finding this harder than I expected!

This simple change will highlight the empty text:

```diff
diff --git a/src/textual/widgets/_masked_input.py b/src/textual/widgets/_masked_input.py
index 6013f775e..4ea315d3a 100644
--- a/src/textual/widgets/_masked_input.py
+++ b/src/textual/widgets/_masked_input.py
@@ -637,3 +637,3 @@ class MaskedInput(Input, can_focus=True):
                 selection_style = self.get_component_rich_style("input--selection")
-                result.stylize_before(selection_style, start, end)
+                result.stylize(selection_style, start, end)

```

image

But I don't think that's the correct solution? 

Ideally we still want the indicate where the text is empty, so the selection looks something like this (where 'X' is the placeholder character to make it more obvious):

image

I created this mockup using Rich, but I'm struggling to replicate similar styling in Textual.

<details>
<summary>Mockup code</summary>

```python
from rich import print
from rich.box import HEAVY
from rich.panel import Panel
from rich.style import Style
from rich.text import Text

base_style = Style(color="#e0e0e0", bgcolor="#272727")
selection_style = Style(bgcolor="#2d4e74")
placeholder_style = Style(color="#797979")

input_text = Text("XXXXE-XXXXJ-XXXXO-XXXXT")

for index, char in enumerate(input_text.plain):
    if char == "X":
        input_text.stylize(placeholder_style, index, index + 1)

selection = (0, len(input_text))
start, end = selection
input_text.stylize_before(selection_style, start, end)

cursor = end
input_text.pad_right(1)
input_text.stylize("reverse", cursor, cursor + 1)

input_widget = Panel(
    input_text,
    box=HEAVY,
    style=base_style,
    border_style="#ba3c5b",
)

print(input_widget)
```
</details>

TomJGooding:
There's still an issue to tackle when the `MaskedInput` starts with a separator, as reported in #6280.

The problem is that the design of this widget never accounted for this. Even before the problems when text selection was added, pressing <kbd>home</kbd> then typing would crash with [an assertion error](https://github.com/Textualize/textual/blob/676045381b7178c3bc94b86901f20764e08aca49/src/textual/widgets/_masked_input.py#L246-L247).

Here's an example `MaskedInput` for a US phone number on the current branch now the selection is highlighted:

image


<details>
<summary>Example code</summary>

```python


class ExampleApp(App):
        yield MaskedInput("(999) 999-9999;_")


if __name__ == "__main__":
    app = ExampleApp()
    app.run()
```

</details>

The problems start because, unlike other masked inputs where the initial value is empty, the value starts with the first separator. Before the user even begins interacting with the input, it is already flagged as invalid and there's this strange selection. Trying to start typing will now crash with the assertion error mentioned above.

The reason is that the `MaskedInput` is designed to [automatically insert separators](https://github.com/Textualize/textual/blob/676045381b7178c3bc94b86901f20764e08aca49/src/textual/widgets/_masked_input.py#L181). This works in most cases because the widget takes care to ensure the cursor is never on a separator, but obviously never accounted for when the template mask begins with a separator such as a phone number.

I'm really not sure what the correct solution is here.

I tried checking other examples of masked inputs for comparison, but Textual's widget seems very different where it restricts editing based on the current value length.

For example, here's a similar masked input in Qt (which I believe Textual's widget took some inspiration from). This must have some special handling when the mask begins with a separator, but I don't think would work in the `MaskedInput` since Qt's widget allows navigating anywhere in the template. 

image

image

image

<details>
<summary>Example code</summary>

```python
import sys

from PySide6.QtWidgets import QApplication, QDialog, QLabel, QLineEdit, QVBoxLayout


class PhoneDialog(QDialog):
    def __init__(self, parent=None):
        super().__init__(parent)

        layout = QVBoxLayout()

        phone_label = QLabel("Phone Number")
        layout.addWidget(phone_label)

        self.phone_input = QLineEdit()
        self.phone_input.setInputMask("(999) 999-9999;_")
        layout.addWidget(self.phone_input)

        self.setLayout(layout)


if __name__ == "__main__":
    app = QApplication(sys.argv)
    dialog = PhoneDialog()
    dialog.show()
    sys.exit(app.exec())
```

</details>

Any suggestions for resolving this issue would be welcome!

James-San:
Hi since this appears to be one I found, 
I am no expert on inner workings of textual, but I assume normally it would check when adding character whether next one is mask and then skip over it. Maybe as widget is initialized (or focused) it should perform similar check hardcoded and just advance cursor ? Granted this is just a loose idea, based on how I think it works not that I checked inner code. Hope this helps. If I find some ways to overcome this myself I'll let you know. 

Edit1: 
Okay did a bit of digging around - cusor positions seem to always start at 1 for some strange reason in case template is starting with masked character. That may be an issue. Going to look into regular input now. 

Edit2:
Yes, for normal input cursor starts of with cursor at 0 and advances to 1. For the problem template it starts from 1 and advances nowhere if you put + there. It seems wrong since it should start from 1 but have + rendered or start from 0 and require pressing +. 

Edit 3: 
Checked for non-problematic template - there it starts from 0. So template vs value interaction is miscounting something in case of template starting from separator. 

Edit 4:
Seems to be the cursor_position is set to 1 by constructor of MaskedInput ( as it should ) but then somehow, insert_text_at_cursor is ommited from being called first time a character is typed. 

Edit 5: 
Seems that the issue is that right as the field is focused first time - it has a deafult selection of first field which makes selection.is_empty evaluate to False ? if you press anything, then character that mask expects and then backspace, it ends up with proper separator, and selection.is_empty evaluating to true. That seems to be cause of the problem 

Edit 6:
Okay so I think least invasive to current selection model would be if replace method took into account that it shouldn't replace the separator at start ( which it currently obviously ignores ).

TomJGooding:
@James-San Thanks for your comment, your input on this issue would be really helpful!

> cusor positions seem to always start at 1 for some strange reason in case template is starting with masked character.

When the template mask begins with a separator, you would expect the cursor position to start at 1?

The problem is that this separator is automatically inserted to advance the cursor position (see above). 

Currently you don't realise the [`select_on_focus`](https://textual.textualize.io/widgets/input/#textual.widgets.Input(select_on_focus)) behaviour as `MaskedInput` doesn't highlight the selected text. But when you start typing, you're actually replacing the first separator, hence the issue.

Since the initial separator means the value isn't empty, this also results in the input already flagged as invalid before the user even begins interacting with the widget.

This is why I'm struggling to find a solution: when the template mask begins with a separator, what's the input value and the resulting selection/cursor position?

James-San:
Yes I can see why that might be an issue. I'm playing around with the replace algorithm from your commit to make it work but I think it needs to keep multiple index. Basically you have to scan replacement text char by char and then: 
1. replacement char is separator and template has same separator - consume from both streams 
2. replacement char is not a separator and template has separator - lookahead until not-separator in template and patternmatch, consume one char from replacement stream and as many as you  had to lookahead from template
3. both replacement char and template  aren't separators - pattern match and consume from both 
4. replacement and template separators differ - return None 

when patternmatching you of course return None as well. 

Something like that logic. It's kinda lexer-like in a way. Or finite state automaton.

TomJGooding:
Sorry but I'm not sure I follow, Putting aside the replace algorithm for the moment, how would this resolve an empty value when it starts with a separator?

James-San:
So the issue is that by default there is selection on focus but field is empty. Thus when you don't clear selection before trying input - it will throw assertion error in this version right? But that's because it tries to replace that first character of string and finds separator where it wouldn't expect it. 

Now with the idea above, since it tries to check if user didn't enter valid character omitting the separator - it will no longer crash.

TomJGooding:
Looking at the original implementation again after a break, unfortunately there's something else I missed which further complicates this.

I realise now that [the original code](https://github.com/Textualize/textual/blob/main/src/textual/widgets/_masked_input.py#L203) not only automatically inserts separators, but can also jump to the next separator.

For example, pressing a digit followed immediately by a separator skips over the empty positions:

masked-input-separator-jump

<details>
<summary>Example code</summary>

```python


class MaskedInputApp(App):
        yield MaskedInput(
            template="9999-9999-9999-9999;_",
        )


if __name__ == "__main__":
    app = MaskedInputApp()
    app.run()
```

</details>

I'll need to check my changes haven't broken this behaviour (as there don't seem to be any existing tests for this), but also consider how to potentially handle jumps when text is selected.

TomJGooding:
Unfortunately I found another bug while investigating this jump to separator behaviour.

- #6315

This is proving to be much harder to fix than anticipated!

Issue #6298, opened 2025-12-31
action_suspend_process leaves the terminal the raw state


candidtim:
`App.action_suspend_process` suspends the app but the terminal appears to be in a raw state: all input is echoed but nothing else happens, there is no shell prompt, input is not processed in any other way.

The example code is taken directly from the documentation: https://textual.textualize.io/guide/app/#suspending-from-foreground . Running this code reproduces the issue. A copy of this code is below, but this is really taken from the doc.

I have also tried resetting the terminal settings with `termios.tcsetattr` to no avail, so I am not sure if this is really just the raw mode. There is no shell prompt, so looks like the terminal is owned by python process.

```python
from textual.app import App, ComposeResult
from textual.binding import Binding
from textual.widgets import Label


class SuspendKeysApp(App[None]):

    BINDINGS = [Binding("ctrl+z", "suspend_process")]

    def compose(self) -> ComposeResult:
        yield Label("Press Ctrl+Z to suspend!")


if __name__ == "__main__":
    SuspendKeysApp().run()
```

TomJGooding:
I can't reproduce this? Running the app then  pressing <kbd>ctrl</kbd>+<kbd>z</kbd>:

```
$ python3 repro_6298.py

[1]+  Stopped                    python3 repro_6298.py
$ echo "Hello, world!"
Hello, world!
````

Could you paste the results from `textual diagnose`?

davep:
@TomJGooding For what it's worth I *can* reproduce the issue on macOS (Darwin Kernel Version 25.2.0) and Textual 7.5.0. To get to a prompt I have to <kbd>ctrl</kbd>+<kbd>z</kbd> twice, and then exiting the sell results in a blank screen (albeit it must be a running app again because hitting <kbd>ctrl</kbd>+<kbd>c</kbd> for example gets the usual "this ain't no normal Unix app" warning notification).

TomJGooding:
Thanks Dave - sorry I should have clarified that I can also reproduce the blank display when the app is resumed. I've created #6369 which hopefully resolves that issue at least.

I wonder why <kbd>ctrl</kbd>+<kbd>z</kbd> just works for me first time to get the shell prompt. I'm on Linux so maybe that's the difference...?

PR #6333, opened 2026-01-17
Fix action_suspend_process deadlock in signal handler

pasky:
Root cause: action_suspend_process sent SIGTSTP, whose handler called suspend_application_mode() which does _key_thread.join(). This blocks waiting for the input thread, but the input thread may be posting messages via asyncio which needs the event loop - but the event loop can't run while blocked in the signal handler. This causes a deadlock.

Fix:
- action_suspend_process now calls suspend_application_mode() directly in normal context, then sends SIGSTOP (not SIGTSTP)
- SIGTSTP handler (for external kill -TSTP) schedules the action via call_soon_threadsafe to run in event loop context
- _resume_signal now calls refresh(layout=True) to redraw after resume

Fixes https://github.com/Textualize/textual/issues/6298

### DIFF
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -4555,6 +4555,7 @@ class App(Generic[ReturnType], DOMNode):
     def _resume_signal(self) -> None:
         """Signal that the application is being resumed from a suspension."""
         self.app_resume_signal.publish(self)
+        self.refresh(layout=True)
 
     @contextmanager
     def suspend(self) -> Iterator[None]:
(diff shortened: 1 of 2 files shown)


pasky:
(Note that I decided to ultimately switch from textual to urwid for my project since I couldn't figure out how to avoid textual's input lag [w/o setting insane FPS], so I'm unlikely to make any followups on this PR - hopefully it still serves as a good base for a fix!)

Issue #4968, opened 2024-09-05
`@on` decorator matches widget subclass unexpectedly


darrenburns:
The code below fires a notification when you click the button.

```python
from __future__ import annotations

from textual import on
from textual.app import App, ComposeResult
from textual.widgets import Button


class MyButton(Button):
    pass


class MyApp(App[None]):
    def compose(self) -> ComposeResult:
        yield Button("Normal Button")

    @on(MyButton.Pressed)
    def notify_on_press_my_button(self) -> None:
        self.notify("MyButton.Pressed matched!")


if __name__ == "__main__":
    MyApp().run()

```

On clicking a `Button`, I would not expect a handler decorated with `@on(MyButton.Pressed)` to fire.

If I click a `MyButton`, I'd expect a `@on(Button.Pressed)` handler to fire, and I've confirmed this works as I expect.

willmcgugan:
@darrenburns The current behavior fits my understanding, but I am not apposed to changing it.

If you subclass `Button`, you don't also subclass `Button.Pressed`

```python
>>> from textual.widgets import Button
>>> class MyButton(Button):
...     pass
...
>>> MyButton.Pressed is Button.Pressed
True
```

So as far as the `on` decorator is concerned, they are the same event.

Feel free to propose a solution...

darrenburns:
Yeah, I understand the mechanics, but if the identity check `is` is how the matching is happening, then I reckon we can do better. It's just such a huge footgun and so unintuitive that I don't think it could be considered anything other than a bug.

If I create a subclass like `class DeleteAllFilesOnMyHardDriveButton(Button)` with a handler like `@on(DeleteAllFilesOnMyHardDriveButton.Pressed)`, I think it's reasonable to expect that would only fire when I click that specific type of button and not the "back up all my files safely" button on the other side of the screen. (Example intended to frighten 😂)

I have some ideas for how we can fix it but will wait until the theme stuff is out of the way.

HarperZ9:
I traced this on current `main` because the failure mode looked like it might be a small `@on` dispatch issue. The current behavior is coming from two separate facts:

- `MyButton.Pressed is Button.Pressed` is `True`, so `@on(MyButton.Pressed)` stores the same message class as `@on(Button.Pressed)`.
[...]

PR #6342, opened 2026-01-26
fix: on-decorator-control-type-filter

wingding12:

(https://github.com/Textualize/textual/issues/4968)

## Summary

adds a `control_type` parameter to the `@on` decorator that allows filtering message handlers by the widget class of the message's `control` property.

```
class MyButton(Button):
    pass

class MyApp(App):
    @on(Button.Pressed, control_type=MyButton)
    def handle_my_button_only(self) -> None:
        # Only fires for MyButton instances, not regular Button
```

tests:
- test_on_control_type_filters_widget_class
- test_on_control_type_with_selector
- test_on_control_type_no_control_raises

### DIFF
--- a/src/textual/_on.py
+++ b/src/textual/_on.py
@@ -1,11 +1,14 @@
 from __future__ import annotations
 
-from typing import Callable, TypeVar
+from typing import TYPE_CHECKING, Callable, TypeVar
 
 from textual.css.model import SelectorSet
 from textual.css.parse import parse_selectors
 from textual.css.tokenizer import TokenError
 from textual.message import Message
+
+if TYPE_CHECKING:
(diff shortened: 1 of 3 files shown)


TomJGooding:
Isn't this already possible using the selector?

```python
    @on(Button.Pressed, "MyButton")
    def handle_my_button_only(self) -> None:
        # Only fires for MyButton instances, not regular Button
```

This also doesn't really seem to address the problem in the linked issue?

> On clicking a Button, I would not expect a handler decorated with `@on(MyButton.Pressed)` to fire.

willmcgugan:
I'm assuming this is AI slop. Does fix the quoted issue, or any issue, AFAICT.

Issue #6315, opened 2026-01-08
Typing separator deletes existing text in `MaskedInput`


TomJGooding:
While trying to fix some bugs in the `MaskedInput`, I discovered that typing a separator is expected to jump to the next separator, skipping over any empty positions.

For example, notice the jumps in the example below after typing a digit then a separator:

![masked-input-separator-jump](https://github.com/user-attachments/assets/16c2f416-985c-4d3d-9f3d-727c9d03c64a)

<details>
<summary>Example code</summary>

```python
from textual.app import App, ComposeResult
from textual.widgets import MaskedInput


class MaskedInputApp(App):
    def compose(self) -> ComposeResult:
        yield MaskedInput(
            template="9999-9999-9999-9999;_",
        )


if __name__ == "__main__":
    app = MaskedInputApp()
    app.run()
```

</details>

This behaviour doesn't appear to be documented anywhere (or covered in tests), but it seems to emulate masked inputs in Qt, which the Textual widget is based on.

The problem is that the [current implementation](https://github.com/Textualize/textual/blob/main/src/textual/widgets/_masked_input.py#L225-L242) will also delete any existing text when jumping to the next separator.

For example, pressing the right arrow then the separator key with existing text:

![Image](https://github.com/user-attachments/assets/584a8681-f167-4515-ac09-73d39030e20d)

Unfortunately this looks like another bug in the `MaskedInput`, especially when compared to the Qt widget.

PR #6343, opened 2026-01-26
fix: preserve existing text when typing separator in MaskedInput

wingding12:

https://github.com/Textualize/textual/issues/6315

when typing a separator character in MaskedInput, the cursor should jump to the position after the next separator without overwriting any existing text. the code would fill non-separator positions with spaces, deleting any text that was already there.

the fix preserves existing text at non-separator positions by only inserting the separator characters themselves, not spaces.

### DIFF
--- a/src/textual/widgets/_masked_input.py
+++ b/src/textual/widgets/_masked_input.py
@@ -231,14 +231,15 @@ class _Template(Validator):
                                 _CharFlags.SEPARATOR
                                 in self.template[cursor_position].flags
                             ):
-                                char = self.template[cursor_position].char
-                            else:
-                                char = " "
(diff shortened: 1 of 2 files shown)


TomJGooding:
Sorry but this doesn't look like the correct fix. Run the example app below and try typing a digit followed by a seperator -  it will crash with `ValueError: Value does not match template!`

```python


class ExampleApp(App):
        yield MaskedInput("9999-9999-9999-9999;0")


if __name__ == "__main__":
    app = ExampleApp()
    app.run()
```

TomJGooding:
Closing as this isn't the correct fix and likely just drive-by AI slop.

PR #6364, opened 2026-02-13
Fix DataTable auto-width columns not shrinking after row removal

### LINKED ISSUE #3449: DataTable auto-width columns don't shrink back down
Removing rows won't recompute the width of auto-width columns.

To reproduce, run the app below and press <kbd>R</kbd>

<details>
<summary> Repro app </summary>

```py
from textual.widgets import DataTable

ROWS = [
    (1, "Aleksandr Sadovnikov =============", "Russia", 51.84),
]


class TableApp(App):
        yield DataTable()

    def on_mount(self) -> None:
        table.add_columns(*ROWS[0])
        *_, self.last_row = table.add_rows(ROWS[1:])

    def key_r(self):
        self.query_one(DataTable).remove_row(self.last_row)


app = TableApp()
if __name__ == "__main__":
    app.run()
```

Mr-Neutr0n:
## Summary

Fixes #3449

When rows are removed from a `DataTable` with auto-width columns, the columns stay as wide as the longest value that ever existed, even though that row is gone. This happens because column widths are only recalculated based on `_new_rows` or `_updated_cells` in `_on_idle`, and both are empty during deletion.

**Changes:**

- Added a `_rows_removed` flag that gets set in `remove_row`
- In `_on_idle`, when this flag is set, all auto-width columns have their `content_width` recalculated by scanning every remaining cell value and comparing against the column header width
- The label column width is also recalculated in the same pass

This is intentionally scoped to auto-width columns only, since fixed-width columns aren't affected by content changes.

## Test plan

- Run the reproduction app from #3449
- Add rows with varying lengths, press R to remove the longest row
- Verify that the column shrinks back to fit the remaining content
- All 87 existing `test_data_table.py` tests pass

### DIFF
--- a/src/textual/widgets/_data_table.py
+++ b/src/textual/widgets/_data_table.py
@@ -782,6 +782,8 @@ class DataTable(ScrollView, Generic[CellType], can_focus=True):
         """Tracking newly added rows to be used in calculation of dimensions on idle."""
         self._updated_cells: set[CellKey] = set()
         """Track which cells were updated, so that we can refresh them once on idle."""
+        self._rows_removed: bool = False
+        """Flag to indicate rows were removed, requiring column width recalculation."""
 
         self._show_hover_cursor = False
         """Used to hide the mouse hover cursor when the user uses the keyboard."""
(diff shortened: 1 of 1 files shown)


willmcgugan:

PR #6365, opened 2026-02-13
Fix component class background blending against wrong colour

### LINKED ISSUE #5155: Transparent component class background blends with wrong colour
When using component classes, if you set a transparent background, the blend is done against the parent of the widget the component class is instantiated in, rather than the parent of the widget where the component class is actually used.

This means if you use a component class inside a widget which contains nested children, the blend will be done against the wrong ancestor background colour.

In the example below, the parent of the widget the component class is applied to is red, however the style is blended against the green background of the `Screen`. I would expect the background of the label to appear a pinkish colour (white 50% blended with red) instead of the light green colour.

image

```python
from rich.text import Text


class MyWidget(Widget):
    COMPONENT_CLASSES = {"foo"}
    DEFAULT_CSS = """
    MyWidget {
        background: white 50%;
    }
    """

        with Horizontal():  # this has a red background
            style = self.get_component_rich_style("foo")

            # the label below has a white 50% background,
            # so we would expect it to blend with the red background
            # of its parent.
            # it instead blends with the green background of the screen.
            yield Label(Text("Hello, world!", style=style))


class MyApp(App[None]):
    CSS = """
    Screen {
        background: green;
    }
    Horizontal {
        background: red;
    }
    """

        yield MyWidget(id="widget")


app = MyApp()
if __name__ == "__main__":
    app.run()

```

Mr-Neutr0n:
Fixes #5155

## The problem

`get_component_rich_style` delegated to the virtual node's `rich_style` property to compute the full (non-partial) style. The virtual node is attached to the defining widget for CSS selector matching, so its ancestor chain reflects the *definition-time* position rather than the position where the style is actually rendered.

This caused two issues:

1. **Component classes without a background** — the returned `Style` still had `bgcolor` populated from the ancestor chain, which overwrote the natural background at the render site. In the reported example, a component class with no background of its own leaked the defining widget's blended background (green-ish) onto a label whose parent was red.

2. **Component classes with transparent backgrounds** — the alpha blending was performed against the defining widget's ancestor backgrounds instead of the rendering context.

## The fix

Instead of delegating to `component_styles.rich_style` (which walks the virtual node's ancestor chain), the method now:

- Walks `self.ancestors_with_self` to resolve inherited text properties (color, text-style, auto-color, opacity) — using the widget's real position in the render tree.
- Only includes a `bgcolor` when the component class explicitly sets a `background` rule with non-zero alpha.
- When no component background is set, leaves `bgcolor` unset so the rendering context's own background shows through.

The existing `partial_rich_style` (returned when `partial=True`) is unchanged.

## Testing

Ran the full test suite (excluding snapshots) — all pre-existing tests pass. The only failures are in unrelated modules (`test_languages.py` for tree-sitter and `test_features.py` for devtools).

### DIFF
--- a/src/textual/widget.py
+++ b/src/textual/widget.py
@@ -1167,7 +1167,74 @@ class Widget(DOMNode):
                 except KeyError:
                     return default
 
-            style = component_styles.rich_style
+            # Build the full style using the widget's own position in
+            # the render tree rather than the virtual node used for CSS
+            # matching.  The virtual node is attached to the defining
+            # widget for selector matching, but its ancestor chain does
+            # not reflect where the resulting style will actually be
+            # rendered.  By walking the real widget's ancestor chain and
+            # only including a background when the component class
+            # explicitly sets one, transparent backgrounds will blend
+            # correctly against the rendering context instead of the
+            # definition-time position.
+            background = Color(0, 0, 0, 0)
+            color = Color(255, 255, 255, 0)
+            style = Style()
+            opacity = 1.0
+
+            for node in reversed(self.ancestors_with_self):
+                styles = node.styles
+                has_rule = styles.has_rule
+                opacity *= styles.opacity
+                if has_rule("background"):
+                    text_background = background + styles.background.tint(
+                        styles.background_tint
+                    )
+                    background += (
+                        styles.background.tint(styles.background_tint)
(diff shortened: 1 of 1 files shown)


willmcgugan:

PR #6367, opened 2026-02-13
fix(widget): fix visual not updated after style changes

### LINKED ISSUE #6235: [BUG] Changing Background style via code does not propagate properly into Markdown Widget
## The bug
Changing the background of a Markdown Widget dynamically via code is not changing the background of all Markdown Segments accordingly. Also see example screenshot, where the code segment is still the previous color.

Changing the background via tcss works properly.

## MRE:
```python
from textual.widgets import Markdown, Select

EXAMPLE_MARKDOWN = """\
# Markdown Viewer

This is an example of Textual's `MarkdownViewer` widget.


## Features

Markdown syntax and extensions are supported.

- Typography *emphasis*, **strong**, `inline code` etc.
- Headers

## Tables

Tables are displayed in a DataTable widget.

| Name            | Type   | Default | Description                        |
| --------------- | ------ | ------- | ---------------------------------- |
| `show_header`   | `bool` | `True`  | Show the table header              |
| `fixed_rows`    | `int`  | `0`     | Number of fixed rows               |


## Code Blocks

Code blocks are syntax highlighted.


"""


class MarkdownExampleApp(App):
        yield Select.from_values(["red", "green", "blue"])
        markdown_viewer = Markdown(EXAMPLE_MARKDOWN)
        yield markdown_viewer

    def on_select_changed(self, event: Select.Changed):
        self.query_one(Markdown).styles.background = event.value

if __name__ == "__main__":
    app = MarkdownExampleApp()
    app.run()
```

Image

### LINKED ISSUE #6289: Color of Static not changing when updated
Using textual 6.8.0 and python 3.14 (GIL and no GIL), I am finding unexpected behavior when trying to set the color of a Static widget in code. I have a Checkbox and when a check/uncheck event occurs, I use styles.color to change the color but nothing changes in the terminal. If the terminal window loses focus and then I bring it back and hover the mouse over the text label, the correct color appears. Here is a simple test app that shows the behavior.

```
from textual.widgets import Checkbox, Static

class CheckboxApp(App):
        yield Checkbox("Change text color", id="my_checkbox")
        yield Static("I am static text.", id="my_text")

    def on_checkbox_changed(self, event: Checkbox.Changed) -> None:
        """An event handler called when a checkbox is changed."""
        text_widget = self.query_one("#my_text", Static)
        if event.value:
            text_widget.styles.color = "green"
        else:
            text_widget.styles.color = "red"
        self.refresh()

if __name__ == "__main__":#	print(os.get_terminal_size())
	app = CheckboxApp()
	app.run()
```

I have tried using reactives, calling refresh(), different shells (default terminal on Almalinux 9, xterm, Kitty) but the results are always the same: checking/unchecking the box leads to no change in the appearance of the app but losing window focus, getting it back and then hovering over the Static does cause the color to change. The problem sounds a bit like issue #6280.

### LINKED ISSUE #6322: Reactivity examples broken in docs
Try running the [Watch methods](https://textual.textualize.io/guide/reactivity/#watch-methods) or [Compute methods](https://textual.textualize.io/guide/reactivity/#compute-methods) examples from the Reactivity docs. These should update the background color of the widgets, but now the background stays "blank" and only the border cells are updated.

TomJGooding:
Fix a significant regression in Textual v6.0.0 where the widget visual is not updated after style changes.

This effectively reverts the changes in dca9439, which intended to optimize styles caching but clearly had unintended consequences.

Fixes https://github.com/Textualize/textual/issues/6235 
Fixes https://github.com/Textualize/textual/issues/6289 
Fixes https://github.com/Textualize/textual/issues/6322

### DIFF
--- a/src/textual/widget.py
+++ b/src/textual/widget.py
@@ -455,8 +455,6 @@ class Widget(DOMNode):
         self._layout_cache: dict[str, object] = {}
         """A dict that is refreshed when the widget is resized / refreshed."""
 
-        self._visual_style: VisualStyle | None = None
-
         self._render_cache = _RenderCache(_null_size, [])
         # Regions which need to be updated (in Widget)
         self._dirty_regions: set[Region] = set()
@@ -720,14 +718,6 @@ class Widget(DOMNode):
                 self.log.warning(
                     f"'{self.__class__.__name__}.CSS' will be ignored (use 'DEFAULT_CSS' class variable for widgets)"
                 )
-
-    def pre_render(self) -> None:
-        """Called prior to rendering.
-
-        If you implement this in a subclass, be sure to call the base class method via super.
-
-        """
-        self._visual_style = None
 
     def _cover(self, widget: Widget) -> None:
         """Set a widget used to replace the visuals of this widget (used for loading indicator).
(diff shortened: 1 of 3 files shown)


willmcgugan:
Thanks. Going with https://github.com/Textualize/textual/pull/6372

TomJGooding:
@willmcgugan The fix in #6372 doesn't seem to account for the other issues related to the visual style?

For example, notice the label background isn't updated here:

```python
from textual.containers import CenterMiddle
from textual.widgets import Label, Select


class ExampleApp(App):
        yield Select.from_values(["red", "green", "blue"])
        with CenterMiddle(id="color-container"):
            yield Label("Try changing the background color")

    def on_select_changed(self, event: Select.Changed):
        self.query_one("#color-container").styles.background = event.value


if __name__ == "__main__":
    app = ExampleApp()
    app.run()
```

shadeyg56:
> @willmcgugan The fix in https://github.com/Textualize/textual/pull/6372 doesn't seem to account for the other issues related to the visual style?
For example, notice the label background isn't updated here:

Still seeing this issue months later. It seems a workaround is to call `.notify_style_update()` on each container child after programmatically changing the container background color, but I really feel like this should be done inherently, Not sure if this is a bug or intended behavior, so some clarification would be helpful.

```py
from textual.containers import CenterMiddle
from textual.widgets import Label, Select

class ExampleApp(App):

        yield Select.from_values(["red", "green", "blue"])
        with CenterMiddle(id="color-container"):
            yield Label("Try changing the background color")

    def on_select_changed(self, event: Select.Changed):
        container = self.query_one("#color-container")
        container.styles.background = event.value
        for descendant in container.walk_children():
            descendant.notify_style_update()

app = ExampleApp()
app.run()
```

Issue #5528, opened 2025-02-15
Display not being refreshed after App.suspend (since 0.85)


davep:
Using this code:

```python
from os import system

from textual.app import App, ComposeResult
from textual.containers import Horizontal, Vertical
from textual.widgets import Footer, Header, Input


class SuspendingApp(App[None]):

    CSS = """
    Screen {
        Vertical {
            border: solid;
        }
    }
    """

    BINDINGS = [("f1", "edit")]

    def compose(self) -> ComposeResult:
        yield Header()
        with Horizontal():
            yield Vertical()
            yield Vertical()
        yield Input()
        yield Footer()

    def action_edit(self) -> None:
        with self.suspend():
            system("vim")


if __name__ == "__main__":
    SuspendingApp().run()
```

if you press <kbd>f1</kbd> and then exit `vim` the screen of the Textual application doesn't fully restore:

https://github.com/user-attachments/assets/0c8009f8-fc60-4c8e-b65b-1511fc989a81

Working backwards through Textual releases, the last version where `App.suspend` worked as intended was [v0.84](https://github.com/Textualize/textual/releases/tag/v0.84.0); it very much appears that [v0.85.0](https://github.com/Textualize/textual/releases/tag/v0.85.0) introduced this issue.

TomJGooding:
```python
system("vim")
```

Joining the dark side Dave?

git bisect points to 6298acb, presumably this change in screen:

```diff
@@ -1205,8 +1205,8 @@ def _get_inline_height(self, size: Size) -> int:
 
     def _screen_resized(self, size: Size):
         """Called by App when the screen is resized."""
+        self._compositor_refresh()
         self._refresh_layout(size)
-        self.refresh()
 
     def _on_screen_resume(self) -> None:
         """Screen has resumed."""
```

I did try adding a `refresh()` in `_on_screen_resume` but that doesn't seem to work. I'm not really familar enough with how the refreshing works to say this is the correct fix, but just sharing my findings in case it helps someone...

PR #6369, opened 2026-02-15
fix(app): refresh after suspend

TomJGooding:
Fix the app display not refreshing after `action_suspend_process`.

A similar issue was reported in https://github.com/Textualize/textual/issues/5528 but specifically when using the `App.suspend` context manager. This was fixed in 81392ff by adding `refresh(layout=True)`, however this doesn't cover when the app was suspended with `action_suspend_process`. This simply moves that call to `refresh` to account for both suspend methods.

Related issue: https://github.com/Textualize/textual/issues/6298

### DIFF
--- a/src/textual/app.py
+++ b/src/textual/app.py
@@ -4594,6 +4594,7 @@ class App(Generic[ReturnType], DOMNode):
     def _resume_signal(self) -> None:
         """Signal that the application is being resumed from a suspension."""
         self.app_resume_signal.publish(self)
+        self.refresh(layout=True)
 
     @contextmanager
     def suspend(self) -> Iterator[None]:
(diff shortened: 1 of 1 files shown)


willmcgugan:
Can we have a test for this?

TomJGooding:
Is it actually possible to test for this that involves suspending and resuming the app?

willmcgugan:
There is `test_suspend.py`. Looks like it mocks out the suspend functionality.

Maybe that can be extended with a simple check that refresh was called at the expected point?

TomJGooding:
I had a look at the existing tests but they don't cover `action_suspend_process` which involves SIGSTOP/SIGCONT - sorry if I've misunderstood or missing something obvious?

Issue #6370, opened 2026-02-15
Zero division error


DubiousCactus:
I checked the related issues and couldn't find the same issue.
I am using the latest Textual version (7.5.0).

## The bug
I get a "zero division error" when I run my Textual app. The error is raised in: site-packages/textual/timer.py:163 in _run.

<img width="1228" height="214" alt="Image" src="https://github.com/user-attachments/assets/b71d5c08-f178-438f-9992-47ecdce4e5b4" />

# Textual Diagnostics

## Versions

| Name    | Value  |
|---------|--------|
| Textual | 7.5.0  |
| Rich    | 14.3.2 |

## Python

| Name           | Value                                                                       |
|----------------|-----------------------------------------------------------------------------|
| Version        | 3.14.2                                                                      |
| Implementation | CPython                                                                     |
| Compiler       | Clang 17.0.0 (clang-1700.6.3.2)                                             |
| Executable     | /Users/cactus/Library/Caches/uv/archive-v0/-HobocSzp8mrunkklOkv2/bin/python |

## Operating System

| Name    | Value                                                                                                  |
|---------|--------------------------------------------------------------------------------------------------------|
| System  | Darwin                                                                                                 |
| Release | 24.5.0                                                                                                 |
| Version | Darwin Kernel Version 24.5.0: Tue Apr 22 19:54:26 PDT 2025; root:xnu-11417.121.6~2/RELEASE_ARM64_T8112 |

## Terminal

| Name                 | Value          |
|----------------------|----------------|
| Terminal Application | tmux (3.6a)    |
| TERM                 | xterm-256color |
| COLORTERM            | truecolor      |
| FORCE_COLOR          | *Not set*      |
| NO_COLOR             | *Not set*      |

## Rich Console options

| Name           | Value                |
|----------------|----------------------|
| size           | width=317, height=90 |
| legacy_windows | False                |
| min_width      | 1                    |
| max_width      | 317                  |
| is_terminal    | True                 |
| encoding       | utf-8                |
| max_height     | 90                   |
| justify        | None                 |
| overflow       | None                 |
| no_wrap        | False                |
| highlight      | None                 |
| markup         | None                 |
| height         | None                 |

willmcgugan:
An MRE is require to diagnose your issue.

An MRE is a Minimally Reproducible Example. It should be the minimal amount of code required to reproduce the issue, which should run without changes. Run the code yourself, to confirm this is the case.

[...]

PR #6371, opened 2026-02-15
Prevent zero division error in timer.py

DubiousCactus:
Address issue #6370


willmcgugan:
This PR requires further discussion and will be closed.

The maintainer would like to consider the solution more carefully before merging any changes.

Please return to the issue or discussion, and explain your solution in more detail. Feel free to create a new PR once @willmcgugan has approved your solution.

Thank you for understanding.

PR #6377, opened 2026-02-17
Add Containers reference page and link from reference index

### LINKED ISSUE #2032: Add a reference section for containers.
Other than a couple of random mentions in the Layout guide and the barebones API page, there isn't a great deal of info/examples for the containers.

A short reference might make sense, especially now that we have `Center`/`Middle` as of #2030.

Mahajan-Sachin:
Closes #2032

Adds a dedicated reference page for containers under docs/reference/containers.

Includes:
- Container
- Vertical
- Horizontal
- Default behavior
- Default CSS

Also fixes navigation link so the Containers reference page loads correctly.

### DIFF
--- a/docs/reference/index.md
+++ b/docs/reference/index.md
@@ -46,7 +46,7 @@ Welcome to the Textual Reference.
 
     Layout and alignment containers for arranging widgets.
 
-    :octicons-arrow-right-24: [Containers Reference](../containers/index.md)
+    :octicons-arrow-right-24: [Containers Reference](containers/index.md)
 
   
 


willmcgugan:
Thanks. This is needed, but I would want to do a more comprehensive reference with diagrams and examples.

if you want to go to that effort, I wold be happy to assist. But I can't merge this as is.

Mahajan-Sachin:
Thanks for the feedback!

I'd be happy to expand this into a more comprehensive reference with diagrams and practical examples.

To make sure I align with your expectations, would you prefer the focus to be on:
- Real-world usage examples?
- Visual layout diagrams explaining container behavior?
- Comparisons between container types?
- Or a combination of these?

Happy to iterate in the direction you think would be most valuable.

PR #6380, opened 2026-02-18
fixed ListView (src/textual/widgets/_list_view.py): clear stuck hover state during scrollbar drag

### LINKED ISSUE #6259: Scrolling ListView with mouse leaves highlighted rows from hovering
- [x] Have you checked closed issues? (https://github.com/Textualize/textual/issues?q=is%3Aissue+is%3Aclosed)

- [x] Have you checked against the most recent version of Textual? (https://pypi.org/search/?q=textual)

## Consider discussions!

Issues are for actionable items only.
If Textual crashes or behaves differently from the docs, then submit an issue.
If you want to know how to do something, or you have a suggestion for a new feature, then open a discussion (https://github.com/Textualize/textual/discussions/).

For realtime help, join our discord server (https://discord.gg/Enf6Z3qhVr)

## The bug

If you make a `ListView` which is long enough to need a scroll, and you use your mouse to scroll (as in, click and hold on the scrollbar and drag up and down), and your mouse happens to move off the scrollbar sideways into the `ListView` as you are holding and scrolling, it will highlight rows based on where your cursor hovers, however some of these will not un-highlight, resulting in multiple highlighted `ListItem` rows. Here is a minimum PoC:

```
from textual.widgets import ListView, ListItem, Static

class MyApp(App):
        yield ListView()

    def on_mount(self):
        lv = self.query_one(ListView)
        for _ in range(200):
            item = ListItem(Static(str(_)))
            lv.append(item)

if __name__ == "__main__":
    MyApp().run()
```

And here is a screenshot showing the issue:

Image

Hoteira:
Fixes #6259 

## Cause
When dragging the scrollbar, it'd call capture_mouse() which prevented Leave events from firing on ListItem widgets. This causes 
items to remain in the -hovered state even after the mouse moves away.

## Fix
Override _on_scroll_to in ListView to clear the -hovered class from 
any items the mouse is no longer over whenever a scrollbar drag causes a scroll.

## Reproduction
Run the MRE from the issue and drag the scrollbar sideways into the list.

Before this fix, items would remain highlighted after the mouse moved away.
After this fix, highlighted items return to their normal state when not hovered.

### DIFF
--- a/src/textual/widgets/_list_item.py
+++ b/src/textual/widgets/_list_item.py
@@ -35,5 +35,7 @@ class ListItem(Widget, can_focus=False):
     @on(events.Enter)
     @on(events.Leave)
     def on_enter_or_leave(self, event: events.Enter | events.Leave) -> None:
+        """Remove mouse hover when dragging the scrollbar"""
         event.stop()
-        self.set_class(self.is_mouse_over, "-hovered")
+        mouse_captured = self.app.mouse_captured is not None
+        self.set_class(self.is_mouse_over and not mouse_captured, "-hovered")
--- a/src/textual/widgets/_list_view.py
+++ b/src/textual/widgets/_list_view.py
@@ -394,12 +394,3 @@ class ListView(VerticalScroll, can_focus=True, can_focus_children=False):
     def __len__(self) -> int:
         """Compute the length (in number of items) of the list view."""
         return len(self._nodes)
-
-    def _on_scroll_to(self, message) -> None:
-        """Clear any stuck hovered items that the mouse is no longer over"""
-        hovered = self.query(".-hovered")
-        if hovered:
-            for item in hovered:
-                if not item.is_mouse_over:
-                    item.set_class(False, "-hovered")
-        super()._on_scroll_to(message)


TomJGooding:
Thanks for your interest in contributing!

This doesn't seem like the correct fix? When dragging a scrollbar, I wouldn't expect any mouse interactions like the hover effect in _any_ widget, not just this issue with the `ListView`?

Hoteira:
Probably a misunderstanding on my part, I thought the hover (while dragging the scrollbar) effect was intentional since it looked nice and that the bug lied in the residual highlighting not getting cleaned up (as the issue's title suggested).
If you confirm the effect should not happen at all I have no problem updating my fix.

TomJGooding:
What if the scrollbar being dragged is the screen or an outer container? If the mouse wanders into a `ListView` while scrolling, you see the same issues with the hover effect. 

I don't think this is the correct fix but I'm not a maintainer so this isn't really my call.

Hoteira:
Thanks for the feedback, I've updated the PR to include a more generic fix that simply disables hover interactions when dragging the scrollbar. This also covers the edge case you mentioned.

PR #6410, opened 2026-03-08
pause gc on scroll

willmcgugan:
- Replace DOM circular references with weakref
- Pause GC on scroll (behind switch)


falbrechtskirchinger:
@willmcgugan Did you miss `test_arrange_dock_left` by chance? The test is failing for me with `NoActiveAppError`, and I stumbled over this PR while investigating.

bmorphism:
is this... the quantum? 😳

PR #6415, opened 2026-03-12
docs: Fix ToastRack location example in documentation

Jah-yee:
The documentation showed `align` property but failed to mention that `dock: none` must be set to override the default `dock: bottom`.

Closes #6389


TomJGooding:
The toasts need to be docked to prevent them scrolling out of view.

This is the second time you've created a PR when one already exists. Please stop wasting people's time with this slop.

PR #6416, opened 2026-03-12
fix(TabbedContent): prevent old tab re-activation on switch with focused widget

### LINKED ISSUE #4955: Unable to switch tabs in TabbedContent using a key binding
So I have this annoying issue.  I haven't pulled it down to a minimum example, as the area is poorly documented and it may be a user error.  The only relevant points are that `fills-search` and `txs-search` are `Input` widgets and that `ctrl+f` and `ctrl+t` are bindings.

So I have this test:

        # part A
        await pilot.press("ctrl+f")
        await pilot.click('#fills-search')
        await pilot.press("s", "h")
        assert app.get_widget_by_id('fills-data', DataTable).get_cell_at((0,2)).strip().startswith("Sharks")

        # part B
        await pilot.press("ctrl+t")
        await pilot.click('#txs-search')
        await pilot.press("6", ".", "8", "5")
        assert app.get_widget_by_id('txs-data', DataTable).get_cell_at((0,2)).strip() == '-6.85'

On the second part, even if I switch part A and B, the `pilot.click` function will cause `app.screen_stack` to acquire a CommandPalette, which means the `get_widget_by_id` is looking at the wrong screen.   That is, the app works differently in testing than in execution.  I don't know why Textual does this, but I look side-eyed at `pilot.py/345`.   If I try to work around the problem by adding   `ENABLE_COMMAND_PALETTE = False` to my app, then it does not bring up the COMMAND_PALETTE, *but* it stops causing the `@on(Input.Changed,...)` to fire.  Again, Pilot not working like my app.

Is this a bug or am I using it wrong?

Dhi13man:
## Summary

Fixes #4955 — When a widget with focus (e.g. `Input`) is inside a `TabbedContent`, switching tabs via key binding or programmatic `active` assignment causes the old tab to briefly deactivate then immediately re-activate.

**Root cause:** Hiding the old tab pane moves focus to a sibling widget still within the same (now-hidden) pane. This triggers a `DescendantFocus` event that bubbles up to `TabPane._on_descendant_focus`, which posts `TabPane.Focused`, causing `TabbedContent._on_tab_pane_focused` to re-activate the old tab.

**Fix:** Added `_defocus_old_pane()` which removes focus from any widget inside the old pane *before* the `ContentSwitcher` hides it. This is called in both tab-switch paths (`_watch_active` and `_on_tabs_tab_activated`), preventing the `DescendantFocus` cascade entirely.

The auto-tab-switch feature (focusing a widget in a hidden pane switches to that tab) is preserved — it works because `_defocus_old_pane` only acts on the *currently displayed* pane, not the target pane.

- `src/textual/widgets/_tabbed_content.py` — added `_defocus_old_pane()`, called before `switcher.current` is changed
- `tests/test_tabbed_content.py` — 3 regression tests: key-binding switch, click switch, and auto-switch preservation

## Test plan

- [x] `test_tab_switch_with_focused_input` — reproduces the exact bug from #4955 (key binding with focused Input)
- [x] `test_tab_switch_with_focused_input_via_click` — verifies clicking a tab works when an Input has focus
- [x] `test_auto_tab_switch_on_programmatic_focus` — confirms the auto-tab-switch feature still works
- [x] All 44 existing `test_tabbed_content.py` tests pass
- [x] All `test_tabs.py`, `test_content_switcher.py`, `test_focus.py` tests pass

Generated with [Dhiman's Agentic Suite](https://github.com/Dhi13man)

### DIFF
--- a/src/textual/widgets/_tabbed_content.py
+++ b/src/textual/widgets/_tabbed_content.py
@@ -13,6 +13,7 @@ from textual.app import ComposeResult
 from textual.await_complete import AwaitComplete
 from textual.content import ContentText, ContentType
 from textual.css.query import NoMatches
+from textual.dom import NoScreen
 from textual.message import Message
 from textual.reactive import reactive
 from textual.widget import Widget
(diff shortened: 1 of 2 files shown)


ThiefMaster:
AI slop.

Dhi13man:
I had leftover Claude Max and wanted to contribute to backlog clearing in open source projects before it reset, but I see that this could be a nuisance. My apologies.

PR #6419, opened 2026-03-13
Fix MaskedInput separator jump deleting existing text

### LINKED ISSUE #6315: Typing separator deletes existing text in `MaskedInput`

wavebyrd:
## Summary
- Fixes #6315
- When a separator character is typed in `MaskedInput`, the cursor jumps to the next group. The loop that advances the cursor was replacing all non-separator positions with spaces, wiping out any text already entered in those slots.
- Changed the jump logic to only write separator characters at their positions and skip over non-separator positions without modifying them.
- Added a regression test that fills in a full value, positions the cursor mid-group, types a separator to jump, and verifies the existing text is preserved.

## Test plan
- [x] All 14 existing `test_masked_input.py` tests still pass
- [x] New `test_separator_jump_preserves_existing_text` test passes

### DIFF
--- a/src/textual/widgets/_masked_input.py
+++ b/src/textual/widgets/_masked_input.py
@@ -231,14 +231,12 @@ class _Template(Validator):
                                 _CharFlags.SEPARATOR
                                 in self.template[cursor_position].flags
                             ):
-                                char = self.template[cursor_position].char
-                            else:
-                                char = " "
-                            value = (
-                                value[:cursor_position]
-                                + char
-                                + value[cursor_position + 1 :]
-                            )
+                                sep_char = self.template[cursor_position].char
+                                value = (
+                                    value[:cursor_position]
+                                    + sep_char
+                                    + value[cursor_position + 1 :]
+                                )
                             cursor_position += 1
                 continue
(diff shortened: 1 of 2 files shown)


TomJGooding:

Issue #6408, opened 2026-03-06
Space character does not appear in VS Code integrated terminal starting from VS Code 1.110


kimadeline:
Have you checked closed issues? (https://github.com/Textualize/textual/issues?q=is%3Aissue+is%3Aclosed) yes

Have you checked against the most recent version of Textual? (https://pypi.org/search/?q=textual) yes

## The bug

Textual 8.0.2, Python 3.12

Given the following code:
```py
from textual.app import App
from textual.widgets import Input


class TestApp(App):
    def compose(self):
        yield Input(placeholder="Type here...")

if __name__ == "__main__":
    app = TestApp()
    app.run()
```

When I press the spacebar the space character does not get displayed when running inside the VS Code integrated terminal:

https://github.com/user-attachments/assets/ad4c8c24-5b4d-4104-8ebd-b7530625281f

It works if I disable the `terminal.integrated.enableKittyKeyboardProtocol` setting in VS Code. This started happening in VS Code version 1.110 ([release notes](https://code.visualstudio.com/updates/v1_110)) which was just released with additional Kitty support.

It might be related to https://github.com/Textualize/textual/issues/6074 as https://github.com/Textualize/textual/issues/6074#issuecomment-3232637650 mentions issues with the space character as well.

TomJGooding:
Could you try running your test with `TEXTUAL_DEBUG=1 python repro_6408.py` and paste the resulting log from `keys.log` here?

kimadeline:
Here you go, I typed `test abc` and then killed the terminal:

```
---
FEED '\x1b[I'
character='\x1b'
sequence='\x1b['
sequence='\x1b[I'
AppFocus()
FEED '\x1b[?2026;2$y\x1b[?2048;0$y'
character='\x1b'
sequence='\x1b['
sequence='\x1b[?'
sequence='\x1b[?2'
sequence='\x1b[?20'
sequence='\x1b[?202'
sequence='\x1b[?2026'
sequence='\x1b[?2026;'
sequence='\x1b[?2026;2'
sequence='\x1b[?2026;2$'
sequence='\x1b[?2026;2$y'
TerminalSupportsSynchronizedOutput()
character='\x1b'
sequence='\x1b['
sequence='\x1b[?'
sequence='\x1b[?2'
sequence='\x1b[?20'
sequence='\x1b[?204'
sequence='\x1b[?2048'
sequence='\x1b[?2048;'
sequence='\x1b[?2048;0'
[...]

Issue #6074, opened 2025-08-26
Progressive enhancement for Kitty keyboard protocol is not enabled


adamritter:
Hi,

I want to be able to disambiguate shift-Enter from Enter. After some debugging I found out that although Kitty keyboard protocol supports it with its progressive enhancement features, it's not enabled.

Kitty keyboard protocol is enabled with ``` self.write("\x1b[>1u")``` but the enhancements need to be enabled as well.

To enable reporting all keys  as escape codes, I needed

 ```self.write("\x1b[=8;u")``` after the protocol enabling line.

in the driver, and after that shift+enter started working. Also I saw somebody else reporting problem with shift+space, this one line change (2 lines if the inline driver adds it as well) fixes that as well, and a lot more keyboard combinations.

I also tried to enable all externsions (31 instead of 8), but that was not working.

This change is not 100% backwards compatible, but I believe it's the right choice for TUI applications, as most of them want access to much more key combos that this change enables.

willmcgugan:
Textual does indeed support the Kitty key protocol.

If you filled out the issue template, I may be able offer more help.

davep:
> After some debugging I found out that although Kitty keyboard protocol supports it with its progressive enhancement features, it's not enabled.

Just to back up what Will says, this code:

```python
from textual.app import App

class KittyApp(App[None]):

    BINDINGS = [
        ("enter", "enter"),
        ("shift+enter", "shift_enter"),
    ]

    def action_enter(self) -> None:
        self.notify("That was enter")

    def action_shift_enter(self) -> None:
        self.notify("That was shift and enter")

if __name__ == "__main__":
    KittyApp().run()
```

running under Kitty does this:

<img width="848" height="693" alt="Image" src="https://github.com/user-attachments/assets/dfdc1095-36db-4eb4-b16a-3e2b3cabe55a" />

Ditto under Ghostty and likely a bunch of other terminals.

TomJGooding:
> Ditto under Ghostty and likely a bunch of other terminals.

Not Alacritty nor WezTerm for me (which both support this protocol).

Presumably because these terminals doesn't enable by default the [progressive enhancements](https://sw.kovidgoyal.net/kitty/keyboard-protocol/#progressive-enhancement) that @adamritter is talking about.

adamritter:
As Tom wrote, I am using Omarchy 2.0 which uses Alacrtitty as its terminal
emulator.

TomJGooding:
[...]

PR #6420, opened 2026-03-13
fix: enable extended key reporting in Kitty keyboard protocol

Jah-yee:
Fixes space character not appearing in VS Code integrated terminal (issue #6408)

VS Code 1.110 enabled additional Kitty keyboard protocol extensions that require enabling extended key reporting mode. Without this, the space key and other keys may not be properly detected.

Added `self.write("\x1b[=8;u")` after the Kitty protocol initialization in all three driver files:
- `src/textual/drivers/linux_driver.py`
- `src/textual/drivers/linux_inline_driver.py`
- `src/textual/drivers/windows_driver.py`

Reference: https://github.com/Textualize/textual/issues/6074


TomJGooding:
@willmcgugan Can we ban this user please?

Jah-yee:
Sad, my problem, not gonna commit in this case.

PR #6463, opened 2026-04-02
Fix @on decorator firing for parent widget events when subclass is specified

### LINKED ISSUE #4968: `@on` decorator matches widget subclass unexpectedly

Andrew-most-likely:
Fixes #4968.

## Problem

When a widget subclass does not define its own message class, `MyButton.Pressed` and `Button.Pressed` are **the same class object**:

```python
>>> MyButton.Pressed is Button.Pressed
True
```

This caused `@on(MyButton.Pressed)` to fire for *any* `Button.Pressed` event — including presses from plain `Button` instances — which is unintuitive and a significant footgun.

## Fix

`Widget.__init_subclass__` now installs a `_MessageClassDescriptor` on each widget subclass for every **inherited** message class. Accessing `MyButton.Pressed` via the descriptor returns a lightweight `_NamespacedMessage(Button.Pressed, MyButton)` wrapper instead of the raw class.

The `@on` decorator detects this wrapper and records `MyButton` as an implicit `namespace_widget` constraint alongside the handler. At dispatch time `_get_dispatch_methods` only invokes the handler when `isinstance(message.control, MyButton)` is `True`.

Message classes that are **directly defined** on a widget (e.g. `Button.Pressed` accessed via `Button`) continue to return the real class unchanged, so `@on(Button.Pressed)` remains unconstrained and fires for all subclasses as before.

## Behaviour after this fix

```python
class MyButton(Button):
    pass

class MyApp(App):
    def compose(self):
        yield Button("Base")   # plain Button
        yield MyButton("Sub")  # subclass

    @on(MyButton.Pressed)        # only MyButton
    def only_my_button(self): ...

    @on(Button.Pressed)          # any Button (incl. MyButton)
    def any_button(self): ...
```

| Event source | `@on(MyButton.Pressed)` fires? | `@on(Button.Pressed)` fires? |
|---|---|---|
| `Button` pressed | **No** (was Yes — the bug) | Yes |
| `MyButton` pressed | Yes | Yes |

## Test

`tests/test_on.py::test_on_widget_subclass_does_not_match_parent` covers both directions.

### DIFF
--- a/src/textual/_on.py
+++ b/src/textual/_on.py
@@ -1,11 +1,14 @@
 from __future__ import annotations
 
-from typing import Callable, TypeVar
+from typing import TYPE_CHECKING, Any, Callable, TypeVar
 
 from textual.css.model import SelectorSet
 from textual.css.parse import parse_selectors
 from textual.css.tokenizer import TokenError
 from textual.message import Message
+
+if TYPE_CHECKING:
+    pass
 
 DecoratedType = TypeVar("DecoratedType")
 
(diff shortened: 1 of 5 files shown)


willmcgugan:
/

PR #6465, opened 2026-04-03
Implement word and line selection modes with corresponding methods and tests

JakeGinnivan:
I only just realised each PR needs an issue or discussion.

I could not find one relating to this, and I've implemented the feature already so I figured we could just have the discussion on this PR @willmcgugan and if it doesn't align then all good I can close. Or I can create the issue/discussion, just let me know.

## Problem

Text selection works in a pretty standard way across most operating systems. Double click selects a word, triple click selects a visual line. Double click and hold (or click and a half) then dragging selects a word at a time, rather than by character, and triple click and hold (or double click and a half) then dragging selects visual line at a time.

This PR introduces a new selection mode (Standard) which implements this. Legacy remains the default.

This PR also allows you to opt out of the selection behaviour, because I implemented most of the above by overriding things, but was constantly fighting the inbuilt selection behaviours. Might not need the opt out, but could be handy for someone who wants to do something different.

I've consumed this downstream and selection feels natural again :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,15 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 
+## [8.3.0] - Unreleased
+
+### Added
+
+- Added `App.SELECTION_MODE` with `SelectionMode.LEGACY` (default, preserves existing behavior) and `SelectionMode.STANDARD` (word/line selection)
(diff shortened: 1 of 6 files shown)


JakeGinnivan:
@willmcgugan what are your thoughts on this?

Happy to jump on a quick call or send a demo of it being used through a private channel if that helps.

timesler:
@JakeGinnivan might be worth creating an issue on this subject and link this PR? I think @willmcgugan may prefer that issues are created to discuss features, as a way to limit the onslaught of vibe coded slop contributions (this is def not that though)

JakeGinnivan:
Discussion started at https://github.com/Textualize/textual/discussions/6517

Issue #4167, opened 2024-02-16
Search Functionality for the Select Widget


davemagro:
I think this would be a very useful feature. Kind of needed in my current project as well. 

Leaving an issue to hopefully get back to this in the future. Hope I could find the time to implement this myself.

fer-hnndz:
I would also like to start working on this feature. First of all, I believe it should be decided how would the search future work in detail.
Will you be able to search for an item by completely typing it or it will just jump to the first item with the letter on the list?

TomJGooding:
I agree some sort of "autocomplete" widget would be extremely useful and something I've seen often requested. For example my own issue here:

- https://github.com/Textualize/textual/issues/3362

If someone does want to take this on, I think the core functionality for such a widget has already been worked out in the command palette if that helps.

fer-hnndz:
I'll look intro the command palette code. I just realized it had that autocomplete function.
Not sure if it should be added as a whole new widget or as a feature for select widgets.

TomJGooding:
I think this should be a new widget. A simple dropdown list is very different from a text input enhanced by suggested options. Not to mention the code for the `Select` widget is already complicated enough as it is!

But if you're serious about working on this, I'd wait for some input from a Textual maintainer.

taleog:
Hi! We’re a group of software engineering students and would like to take on this issue as part of one of our courses. Could it be assigned to @taleog so we can work on it and avoid duplicate work?

Thank you!

PR #6467, opened 2026-04-03
Add ComboBox widget (searchable select)

steveanil:
**This PR is AI generated by Claude**

https://github.com/Textualize/textual/issues/4167

cc @willmcgugan

## What this adds

A `ComboBox` widget that combines a text `Input` with a filterable `OptionList` overlay — a standard "searchable select" / combobox pattern.

## Design decisions

- **Focus stays on `Input` at all times.** The parent `ComboBox` has `can_focus=False`. Arrow keys bubble up from `Input` and are proxied to the overlay, so the user can type and navigate without a focus ownership conflict.
- **Strict predefined-option selection.** Free text that doesn't match an option is never accepted. Pressing Enter with no match reverts the input to the last valid selection (or clears it if nothing was ever selected).
- **Blur closes the overlay.** When focus leaves the `Input`, the overlay collapses via a `ComboBoxInput.LostFocus` message.
- **`ComboBox.Selected` message** is posted on selection, following the same pattern as `Select`.

## Usage

```python
yield ComboBox(
    [("Apple", 1), ("Banana", 2), ("Cherry", 3)],
    placeholder="Search...",
)

def on_combo_box_selected(self, event: ComboBox.Selected) -> None:
    print(event.value)
```

## Files changed

| File | Purpose |
|------|---------|
| `src/textual/widgets/_combo_box.py` | Core implementation |
| `src/textual/widgets/combo_box.py` | Public API re-export |
| `src/textual/widgets/__init__.py` | Registers `ComboBox` in the widget namespace |
| `src/textual/widgets/__init__.pyi` | Type stub |
| `tests/test_combo_box.py` | 5 automated tests covering filtering, selection flow, blur, and no-match Enter behavior |

### DIFF
--- a/src/textual/widgets/__init__.py
+++ b/src/textual/widgets/__init__.py
@@ -13,6 +13,7 @@ if typing.TYPE_CHECKING:
     from textual.widgets._button import Button
     from textual.widgets._checkbox import Checkbox
     from textual.widgets._collapsible import Collapsible
+    from textual.widgets._combo_box import ComboBox
     from textual.widgets._content_switcher import ContentSwitcher
     from textual.widgets._data_table import DataTable
     from textual.widgets._digits import Digits
@@ -54,6 +55,7 @@ __all__ = [
     "Button",
     "Checkbox",
     "Collapsible",
+    "ComboBox",
     "ContentSwitcher",
     "DataTable",
     "Digits",
--- a/src/textual/widgets/__init__.pyi
+++ b/src/textual/widgets/__init__.pyi
@@ -2,6 +2,7 @@
 from ._button import Button as Button
 from ._checkbox import Checkbox as Checkbox
 from ._collapsible import Collapsible as Collapsible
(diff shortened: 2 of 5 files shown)


TomJGooding:
The `Select` widget already has [type-to-search](https://textual.textualize.io/widgets/select/#type-to-search), so can you help explain exacttly what extra functionality this new widget provides? (Please don't just copy this question into Claude and paste its response)

steveanil:
> The `Select` widget already has [type-to-search](https://textual.textualize.io/widgets/select/#type-to-search), so can you help explain exacttly what extra functionality this new widget provides? (Please don't just copy this question into Claude and paste its response)

In the issue https://github.com/Textualize/textual/issues/4167, I was reading the details of the conversation and you mentioned that it should be a separate new widget that has a drop down list. This new widget filters the list itself as you type and the `Select` widget only lets you scroll through a fixed list.

PR #6492, opened 2026-04-15
Fix #6444: Crash when using Input in a DEFAULT_MODE screen

### LINKED ISSUE #6444: Crash on first `Input` in compose of `app.DEFAULT_MODE` screen
- [x] Have you checked closed issues? (https://github.com/Textualize/textual/issues?q=is%3Aissue+is%3Aclosed)

- [x] Have you checked against the most recent version of Textual? (https://pypi.org/search/?q=textual)

## The bug

_Initially reported on discord:_ https://discord.com/channels/1026214085173461072/1033754296224841768/1483472126802661407

I ran into an issue when I tried to add the first `Input`.

It seems `Input` cannot be used with `app.MODES`, or at least not in the `compose` method of the first screen of the `app.DEFAULT_MODE` (I did not set `app.SCREENS`). That kills the app with a traceback for this error:

```
ScreenStackError: No screens on stack
```

The crash happens because `Input`s watchers end up trying to do `self.app.clear_selection()`.

https://github.com/Textualize/textual/blob/0f0849fd37fbd0d4d6f81889476c22340129df67/src/textual/widgets/_input.py#L507-L508

That has this try/except:

https://github.com/Textualize/textual/blob/0f0849fd37fbd0d4d6f81889476c22340129df67/src/textual/app.py#L4555-L4560

That only catches `NoScreen`, but with modes a `ScreenStackError` is raised because that first screen hasn't been mounted yet, so the stack is still empty.

For now, I'm working around the issue by adding a `clear_selection` method to my app that does:
```python
def clear_selection(self):
    try:
        super().clear_selection()
    except ScreenStackError:
        pass
```

This workaround is based on several places in the code base that do `except NoScreen, ScreenStackError`, so it seemed like a reasonable fix.

On discord, David said the following:
https://discord.com/channels/1026214085173461072/1033754296224841768/1483494995523534999

> Interesting. I can reproduce this using the modes01.py example from the Guide (and replacing the dashboard Placeholder by an Input). HOWEVER, if you do not use `DEFAULT_MODE` but use the code in the modes01.py example, you don't run into this problem. So do this:
```python
    # DEFAULT_MODE = "dashboard"

    def on_mount(self) -> None:
        self.switch_mode("dashboard")
```
> (change the name of course) and skip using `DEFAULT_MODE`, at least for now.

# Textual Diagnostics

## Versions

| Name    | Value  |
|---------|--------|
| Textual | 8.1.1  |
| Rich    | 14.3.2 |

## Python

| Name           | Value                                                      |
|----------------|------------------------------------------------------------|
| Version        | 3.13.7                                                     |
| Implementation | CPython                                                    |
| Compiler       | Clang 20.1.4                                               |
| Executable     | /home/jafloyd/g/github/copart-pe/ghe.git/.venv/bin/python3 |

## Operating System

| Release | 6.8.0-90-generic                                                 |
| Version | #91~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Thu Nov 20 15:20:45 UTC 2 |

## Terminal

| Terminal Application | Konsole 220403 |

## Rich Console options

| size           | width=236, height=55 |
| max_width      | 236                  |
| max_height     | 55                   |

Birnendampf:
This PR fixes the issue described in #6444
This was specifically caused by textual's use of eager task factories, causing the widgets in the screen to start processing messages before the screen is ever added to the screen stack. **This may be a larger underlying oversight which is not fixed here and might require further investigation**.

There are also some other things to note:
1. `App.run_test()` does _not_ use `asyncio.eager_task_factory`. This does not seem correct, as it causes a discrepancy between tests and reality. E.g. this bug does not occur with `run_test()` by default.
This is beyond the scope of this PR and should probably be a separate issue, unless this is intended.
2. Should screens be composed before or after they are added to the screenstack? currently, that behavior depends on the choice of task factory for user screens, which is probably not intended.
Furthermore, the `_default` screen is composed _after_ being added to the stack.
If that is the correct behavior, it would likely also fix this issue, making this PR obsolete.

fixes #6444


Birnendampf:
If there is anything I can do to make this more mergeable please lmk! I can move the concerns raised here to a discussion instead if that is preferred.

PR #6494, opened 2026-04-16
Fix/message pump app unknown type

matmaer:
Discussion link: https://github.com/Textualize/textual/discussions/6485, one line change to help type checkers. (aka reduce complaining by type checkers)


matmaer:
Turns out this was specifically tied to `pylance` behavior because the "fix" doesn't work anymore with the latest pylance version.
I'm now calling `getattr` on `self.app`, no more warning.
Closing PR.

PR #6509, opened 2026-04-23
fix(work): parameterize Decorator return alias in @work

### LINKED ISSUE #3510: [typing] Inconsistent typing in the work decorator
See [this discussion](https://github.com/Textualize/textual/pull/3499#discussion_r1352573705) about the usage of `Decorator` versus `Decorator[..., ReturnType]` in `src/textual/_work_decorator.py`.

Bojun-Vvibe:
Closes #3510

## Repo
Textualize/textual

## Issue
https://github.com/Textualize/textual/issues/3510

## Root cause
In `src/textual/_work_decorator.py`, the runtime `work()` function's return
annotation used the bare alias `Decorator` (`Callable[..., Worker[ReturnType]] | Decorator`)
instead of supplying its `ParamSpec`/`TypeVar` arguments. When `@work(...)` is
called without a method, type checkers fall through to that branch and lose
the parameterization, leaving `ReturnType` unbound.

## Fix
Parameterize the alias as `Decorator[..., ReturnType]` in the return
annotation of `work()` so the worker's return type propagates through the
no-argument decorator overload path. One-line change.

## Regression test
`tests/workers/test_work_decorator.py::test_work_decorator_return_annotation_uses_parameterized_decorator`
asserts that `work.__annotations__["return"]` contains
`Decorator[..., ReturnType]` and does not end with the bare `Decorator`.

## Risk
trivial

## Verification
Verified the new test logic against the patched and pre-patch annotation
strings via direct `python -c` checks (test passes post-fix and would fail
pre-fix). Could not run the full pytest suite due to an unrelated `conftest`
import error (missing `click` from a third-party `dash` package in the
sandbox); the change is annotation-only, no runtime behavior altered.

### DIFF
--- a/src/textual/_work_decorator.py
+++ b/src/textual/_work_decorator.py
@@ -84,7 +84,7 @@ def work(
     exclusive: bool = False,
     description: str | None = None,
     thread: bool = False,
-) -> Callable[FactoryParamSpec, Worker[ReturnType]] | Decorator:
+) -> Callable[FactoryParamSpec, Worker[ReturnType]] | Decorator[..., ReturnType]:
     """A decorator used to create [workers](/guide/workers).
 
     Args:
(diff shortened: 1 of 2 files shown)


willmcgugan:

PR #6513, opened 2026-04-26
Ansi themes

### LINKED ISSUE #6490: ANSI-based builtin theme
We would need to have a builtin theme which uses only ANSI colors to natively match the colors of the terminal we are being executed in

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/6490

Added two ANSI themes; ansi-dark and ansi-light.

A theme can now explicitly declare itself as an ANSI theme, which will prevent ANSI colors from being promoted to true-color. This means that here is probably no reason to set `ansi_color` on the app constructor any more. ANSI and truecolor themes can co-exist in the same app.

Fixed an issue related to the ordering of text spans that impacted selection.

## ANSI themes

Previously the App constructor accept an `ansi_color` boolean.

| `ansi_color` | Behavior |
| --- | --- |
| `False` | Convert ANSI color sequences to RGB to enabled transparency |
| `True` | Write ANSI color sequences |

`ansi_color` may now also be `None` which will cause the app to use the value of `ansi` set on the Theme. Most apps should leave `ansi_color` as the default (`None`).

This allows for explicit ANSI themes. ANSI tends to be an all or nothing thing. If you mix RGB colors with ANSI colors you have no way of knowing if your colors + user colors will be legible.

There are two new themes: ansi-light and ansi-dark. Users with a dark terminal theme should pick `ansi-dark`, users with a light terminal theme should pick `ansi-light`.

## Customizing ANSI themes

Default widgets should work with ANSI themes (open an issue if something breaks), but if you may still want customizations for ANSI only. To do this, you can use the `:ansi` pseudo class.

The most straightforward way of doing this is to add a single `App:ansi` rule to your stylesheet with ANSI specific rules. Something like the following:

```css
App:ansi {
    # ansi specific rules
    &:light {
        # ansi specific rules for light themes
    }
    &:dark {
        # ansi specific rules for dark themes
    }
}
```

This should ensure that the rules have greater specificity.

Screenshot 2026-04-26 at 21 39 16


pavelzw:
hi, in [pixi-browse](https://github.com/pavelzw/pixi-browse), i am using `textual-ansi` as a non-configuratble theme.
this update broke pixi-browse (which pinned textual to `>=8.0.2,<9`). would be nice if those kind of breaking changes could be kept behind a major version bump.

https://github.com/pavelzw/pixi-browse/pull/60 will fix my issues so no need for any back-fixes from my side

willmcgugan:
Mistakes happen @pavelzw

pavelzw:
no worries! just wanted to let you know))

PR #6514, opened 2026-04-27
docs: note that thread workers should stay short-lived

### LINKED ISSUE #2593: Add a note to the worker API docs about long-running thread workers
Following on from the work done in #2545 and the use of thread workers there: add a note to the documentation for thread-based workers that having long-running thread workers is going to be a bad idea.

Rohan5commit:
Thank you for contributing!

- Closes #2593

## Summary
- add a short note that thread workers should stay short-lived because blocking threads cannot be interrupted like async workers
- point readers toward async workers or smaller cancellable steps for longer-running background work

## Validation
- ran `git diff --check`


willmcgugan:

PR #6519, opened 2026-05-01
fix(markdown): sync code and _highlighted_code in MarkdownFence._update_from_block

### LINKED ISSUE #6518: `MarkdownFence` body disappears after recompose when streamed via `MarkdownStream`
Have you checked closed issues? Yes. I searched for `MarkdownFence streaming`, `_highlighted_code`, `_update_from_block`, `markdown stream`, `code fence`, and related terms. I did not find an existing report for this exact issue.

Have you checked against the most recent version of Textual? Yes. The latest release is `v8.2.5`; the same stale-state path appears to remain in `MarkdownFence._update_from_block`.

## The bug

When markdown is streamed via `Markdown.get_stream()` / `MarkdownStream`, a fenced code block whose body arrives after the opening fence can render correctly at first, but its internal `code` and `_highlighted_code` fields remain stale.

The visible `Label` gets updated, so the fence initially looks correct. But if the fence later recomposes, the `Label` is rebuilt from the stale empty `_highlighted_code`, and the code body disappears.

This shows up in LLM-streaming TUIs: assistant code fences render correctly after generation completes, then go blank after a click, terminal focus change, style update, or other path that causes recompose/style refresh.

Minimal repro:

````python
import asyncio

from textual.widgets import Label, Markdown
from textual.widgets._markdown import MarkdownFence

SAMPLE = """Example:

```python
from langgraph_sdk import get_client

client = get_client(url="http://localhost:2024")

result = await client.runs.wait(
    thread_id=None,
    assistant_id="agent",
    input={"messages": [{"role": "user", "content": "hello"}]},
)

print(result)
```
"""


class MdApp(App):
        yield Markdown("", id="md")

        md = self.query_one("#md", Markdown)
        stream = Markdown.get_stream(md)

        for i in range(0, len(SAMPLE), 10):
            await stream.write(SAMPLE[i : i + 10])
            await asyncio.sleep(0.005)

        await stream.stop()


async def main() -> None:
    app = MdApp()
    async with app.run_test(size=(100, 30)) as pilot:
        await pilot.pause(1.0)

        fence = app.query_one(MarkdownFence)
        label = fence.query_one("#code-content", Label)

        print("after stream:")
        print(f"  _content={len(str(fence._content))}")
        print(f"  _highlighted_code={len(str(fence._highlighted_code))}")
        print(f"  label={str(label.render()).splitlines()[0]!r}")

        await fence.recompose()
        await pilot.pause(0.3)

        label = fence.query_one("#code-content", Label)

        print("after recompose:")
        print(f"  _content={len(str(fence._content))}")
        print(f"  _highlighted_code={len(str(fence._highlighted_code))}")
        print(f"  label={str(label.render())!r}")


asyncio.run(main())
````

Actual output:

```text
after stream:
  _content=247
  _highlighted_code=0
  label='from langgraph_sdk import get_client'
after recompose:
  _content=247
  _highlighted_code=0
  label=''
```

Expected behavior:

`_highlighted_code` should be updated when the streamed block is updated, so recomposing the fence preserves the visible code body.

The likely fix is for `MarkdownFence._update_from_block` to mirror the updated block state before calling `set_content`, for example:

```python
self.code = block.code
self._highlighted_code = block._highlighted_code
self.set_content(block._highlighted_code)
```

`textual diagnose`:

```text
Textual | 8.2.3
Rich    | 14.3.2
Python  | 3.14.2
System  | Darwin 25.4.0
Terminal Application | vscode (1.116.0)
TERM | xterm-256color
COLORTERM | truecolor
NO_COLOR | 1
```

blackwell-systems:
## The bug

When a fenced code block is streamed via `MarkdownStream`, `_update_from_block` updates the visible `Label` via `set_content(block._highlighted_code)` but does not sync `self.code` or `self._highlighted_code` from the source block. 

After a recompose, style update, or theme change, `notify_style_update` (or `compose`) rebuilds the `Label` from `self._highlighted_code`, which is still empty from the initial construction. The code body disappears.

This affects LLM-streaming TUIs where code fences render correctly during generation but go blank after a click, focus change, or style refresh.

## Root cause

`_update_from_block` at line 930 only calls `set_content` and `_copy_context`. It copies `lexer` and `_token` but not `code` or `_highlighted_code`. When the fence recomposes, it rebuilds from the stale empty fields.

## Fix

Copy `code` and `_highlighted_code` from the source block before calling `set_content`:

```python
self.code = block.code
self._highlighted_code = block._highlighted_code
self.set_content(block._highlighted_code)
```

This ensures `notify_style_update` and recompose find the correct state.

Fixes #6518

### DIFF
--- a/src/textual/widgets/_markdown.py
+++ b/src/textual/widgets/_markdown.py
@@ -929,6 +929,8 @@ class MarkdownFence(MarkdownBlock):
 
     async def _update_from_block(self, block: MarkdownBlock):
         if isinstance(block, MarkdownFence):
+            self.code = block.code
+            self._highlighted_code = block._highlighted_code
             self.set_content(block._highlighted_code)
             self._copy_context(block)
         else:


willmcgugan:

PR #6529, opened 2026-05-10
Fix escaping for uppercase markup tags

### LINKED ISSUE #6525: Possible bug with uppercase markup tags
Have you checked closed issues? (https://github.com/Textualize/textual/issues?q=is%3Aissue+is%3Aclosed) Yes

Have you checked against the most recent version of Textual? (https://pypi.org/search/?q=textual) Yes, v8.2.5
## The bug

Creating a button (or any other widget that'll use `content.from_markup`) with a label that has an uppercase markup tag (like `[RED]`) will not render it since it'll be treated as if it were a tag. (buttons 1 and 2 in the example). 

Using the `markup.escape` function will not prevent this, since its regex only targets markup blocks which are lowercase. (buttons 3 and 4 in the example). 

The only workaround I've found is to instantiate a `Content` which will ignore markup from the whole text string (buttons 5 and 6 in the example). 

I'm unsure if this would be considered a bug. 

Complete example:
```python
from textual.content import Content
from textual.markup import escape

class escapeApp(App):
    BINDINGS = {
        ("q,ctrl+c", "quit", "Quit"),
    }
        yield Button("[red] non-escaped str with lowercase block", id="one")
        yield Button("[RED] non-escaped str with uppercase block", id="two")
        yield Button(escape("[red] escaped str with lowercase block"), id="three")
        yield Button(escape("[RED] escaped str with uppercase block"), id="four")
        yield Button(Content("[red] non-escaped Content (no markup) with lowercase block"), id="five")
        yield Button(Content("[RED] non-escaped Content (no markup) with uppercase block"), id="six")

if __name__ == "__main__":
    escapeApp().run()
```
Screenshot of example output:
Image

```
textual diagnose
```

# Textual Diagnostics

## Versions

| Name    | Value  |
|---------|--------|
| Textual | 8.2.5  |
| Rich    | 15.0.0 |

## Python

| Name           | Value                                          |
|----------------|------------------------------------------------|
| Version        | 3.14.4                                         |
| Implementation | CPython                                        |
| Compiler       | Clang 21.0.0 (clang-2100.0.123.102)            |
| Executable     | /Users/david/Downloads/tests/.venv/bin/python3 |

## Operating System

| Name    | Value                                                                                                   |
|---------|---------------------------------------------------------------------------------------------------------|
| System  | Darwin                                                                                                  |
| Release | 25.4.0                                                                                                  |
| Version | Darwin Kernel Version 25.4.0: Thu Mar 19 19:30:44 PDT 2026; root:xnu-12377.101.15~1/RELEASE_ARM64_T6000 |

## Terminal

| Name                 | Value            |
|----------------------|------------------|
| Terminal Application | vscode (1.118.1) |
| TERM                 | xterm-256color   |
| COLORTERM            | truecolor        |
| FORCE_COLOR          | *Not set*        |
| NO_COLOR             | *Not set*        |

## Rich Console options

| size           | width=98, height=19 |
| max_width      | 98                  |
| max_height     | 19                  |

PS: I found this behavior on iSponsorBlockTV's configurator, with a device that has the following name: `[TV] MODEL_NUMBER`

vandit98:
Fixes #6525.

## Summary
- teach `textual.markup.escape()` to escape uppercase bracketed markup openings as well as lowercase ones
- add regression coverage for escaped uppercase tags like `[FOO]` and `[RED]`
- note the bugfix in the changelog

## Testing
- `PYTHONPATH=src python3 -m pytest tests/test_content.py -k escape -q`
- `PYTHONPATH=src python3 -m pytest tests/test_markup.py -q`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 
 - Fixed selection to the right of code fence blocks (may break some snapshots)
+- Fixed `textual.markup.escape()` so uppercase bracketed text like `[RED]` is escaped correctly
 
 ## [8.2.5] - 2026-04-30
 
--- a/src/textual/markup.py
+++ b/src/textual/markup.py
@@ -155,7 +155,7 @@ _EscapeSubMethod = Callable[[_ReSubCallable, str], str]  # Sub method of a compi
 
 def escape(
     markup: str,
-    _escape: _EscapeSubMethod = re.compile(r"(\\*)(\[[a-z#/@][^[]*?])").sub,
+    _escape: _EscapeSubMethod = re.compile(r"(\\*)(\[[a-zA-Z#/@][^[]*?])").sub,
 ) -> str:
     """Escapes text so that it won't be interpreted as markup.
 
--- a/tests/test_content.py
+++ b/tests/test_content.py
@@ -5,6 +5,7 @@ from rich.text import Text
 
 from textual.color import Color
 from textual.content import Content, Span
+from textual.markup import escape
 from textual.style import Style
 from textual.visual import RenderOptions
 from textual.widget import Widget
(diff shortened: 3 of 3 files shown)


willmcgugan:

PR #6539, opened 2026-05-13
Add floor rounding for progress percentage

### LINKED ISSUE #6441: Add percentage_rounding parameter to ProgressBar
When tracking progress with a large number of steps (e.g. bytes in a multi-gigabyte download), the percentage label can display `100%` while progress is still incomplete. This happens because `PercentageStatus._validate_percentage` uses `round()` — for example, downloading 36 out of 37 GB (99.95%) rounds up and displays as `100%`, which is misleading to users.

It would be useful to have a `percentage_rounding` parameter on `ProgressBar` that accepts `"round"` (default, current behavior) or `"floor"`. With `"floor"`, the label only shows `100%` when `progress == total`.

```python
yield ProgressBar(total=total_bytes, percentage_rounding="floor")
```

I have an implementation with tests ready here: https://github.com/MorningLightMountain713/textual/tree/progress-bar-floor-percentage

Happy to open a PR if you'd like to accept contributions for this.

prantikmedhi:
Adds a `percentage_rounding` option to `ProgressBar` so the percentage label can use `floor` instead of `round`. This avoids showing 100% before completion for large totals.

Fixes #6441.


willmcgugan:

PR #6543, opened 2026-05-16
Support more key combinations in TextArea

### LINKED ISSUE #6488: Support more emacs-style keybindings in TextArea
The missing ones we have currently identified are:
- `ctrl+u` doesn't remove previous `\n` when cursor is at the beginning of the line (but `ctrl+k` actually has the good behavior regarding `\n`)
- `ctrl+b`: Move cursor backward (left)
- `ctrl+f`: Move cursor forward (right) - instead, it currently deletes the next word
- `ctrl+backspace`: delte previous word (this one is probably also linked to kitty protocol support)
- `ctrl+p`: Move cursor to the previous line
- `ctrl+n`: Move cursor to the next line

Side note: would it make sense to support cmd+Z at the TextArea level?

willmcgugan:
Fixes https://github.com/Textualize/textual/issues/6488

- [x] ctrl+u doesn't remove previous \n when cursor is at the beginning of the line (but ctrl+k actually has the good behavior regarding \n)
- [ ] ~ctrl+b: Move cursor backward (left)~
- [ ] ~ctrl+f: Move cursor forward (right) - instead, it currently deletes the next word~
- [x] ctrl+backspace: delete previous word (this one is probably also linked to kitty protocol support)
- [x] option+delete: Delete end of word
- [ ] ~ctrl+p: Move cursor to the previous line~
- [ ] ~ctrl+n: Move cursor to the next line~
- [x] Side note: would it make sense to support cmd+Z at the TextArea level?

ctrl+p conflicts with the command palette. Apps that need this can easily add bindings in a subclass. See note in comment

cmd+z is bound as an alternative to ctrl+z. Alas, not all terminals report this. Ghostty captures ctrl+z as undo for its own purposes. The cmd modifier doesn't work at all in iTerm. It does work in Kitty though.

There's not much I can do from the Textual side to improve things here. It is somewhat possible to deduce what bindings are available, then at least the app can not display bindings that don't work. But this typically relies on environment variables being set correctly, and it may fail over ssh, tmux etc.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Undo/redo/copy/cut/paste in TextArea will now work with cmd+ on supported terminals
 - In TextArea, ctrl+u will now delete a newline if the cursor is at the start
+- in TextArea ctrl+delete is now bound to delete word right
 
 ## [8.2.6] - 2026-04-13
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Undo/redo/copy/cut/paste in TextArea will now work with cmd+ on supported terminals
 - In TextArea, ctrl+u will now delete a newline if the cursor is at the start
-- in TextArea ctrl+delete is now bound to delete word right
+- in TextArea alt+delete is now bound to delete word right
 
 ## [8.2.6] - 2026-04-13
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,13 +9,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
-- Added support for Kitty key protocl "Report all keys as escape codes" which enabled alt+backspace on Warp
+- Added support for Kitty key protocol "Report all keys as escape codes" which enabled alt+backspace on Warp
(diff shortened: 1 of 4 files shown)


willmcgugan:
@mgesbert The `ctrl+p` shortcut conflicts with the `ctrl+p` binding for the command palette. Rather than implement 3 out of the 4 navigation shortcuts, its probably best for you to add the bindings to Vibe. I don't think you are using the command palette anyway, so this shouldn't be a problem.

It is trivial to add these bindings. You can add bindings to a TextArea subclass as follows:

```python
from textual import widgets

TEXT = """Hello, world!
foo
bad"""


class VTextArea(widgets.TextArea):
    BINDINGS = [
        Binding(
            "up,ctrl+p",
            "cursor_up",
            "Cursor up",
            show=False,
        ),
        Binding(
            "down,ctrl+n",
            "cursor_down",
            "Cursor down",
            show=False,
        ),
        Binding(
            "left,ctrl+b",
            "cursor_left",
            "Cursor left",
            show=False,
        ),
        Binding(
            "right,ctrl+f",
            "cursor_right",
            "Cursor right",
            show=False,
        ),
    ]


class TApp(App):
    ENABLE_COMMAND_PALETTE = False

        yield VTextArea(TEXT)


TApp().run()
```

willmcgugan:
@mgesbert This is merged. Some relevant notes in the description and comments.

mgesbert:
@willmcgugan thanks, we will take care of that 🙏

PR #6551, opened 2026-05-21
Fix TwoWayDict reverse map going stale on key reassignment

c-tonneslan:
`TwoWayDict.__setitem__` writes the new value into `_forward` and `_reverse`, but never removes the *old* value from `_reverse` when a key is reassigned:

```python
d = TwoWayDict({})
d[1] = 10
d[1] = 20
d.get(1)               # 20  ✓
d.get_key(20)          # 1   ✓
d.get_key(10)          # 1   ✗ should be None
d.contains_value(10)   # True ✗ should be False
```

So `_reverse` stops being a true inverse of `_forward` — `get_key` and `contains_value` keep reporting a value the key no longer maps to.

The fix drops the old value's reverse entry when the key already exists. The pre-existing `TODO` about duplicate *values* (two keys → one value) is a separate design question and is left untouched.

Added `test_set_item_reassigning_key_clears_old_reverse_entry` alongside the existing `test_set_item`.

Worth noting on scope: `DataTable` (the in-tree user of `TwoWayDict`) currently rebuilds these dicts rather than reassigning keys, so this isn't a user-visible bug today — it's a correctness hole in the utility class itself, which has an explicit two-way invariant and its own test file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## Unreleased
+
+### Fixed
+
+- Fixed `TwoWayDict` leaving a stale entry in its reverse map when a key was reassigned https://github.com/Textualize/textual/pull/6551
+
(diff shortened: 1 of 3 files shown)


willmcgugan:

PR #6572, opened 2026-06-08
Fix outline repeated on padding rows from simple strip cache

### LINKED ISSUE #6558: Widget outlines are subtly broken by the simple strip cache
The optimised styles cache render commit 8ae3b5fcb3d1f4fce79392218129da475b65f752 introduced a subtle bug. It seems like the first line of the widget is reused in the padded sections of the widget, but when you use an outline that first line contains the top part of the 'border'. It looks like this:

Image

You can reproduce this with:

```py


class OutlineApp(App[None]):
    CSS = """
        Static {
            height: 15;
            padding-top: 5;
            padding-bottom: 5;
            outline: blue heavy;
        }
    """

        for _ in range(10):
            yield Static("Lipsum")


OutlineApp().run()
```
It's probably an easy fix (or this optimisation may be reverted if the 'win' is only minimal) but I'm afraid to create another subtle bug if I try.

ShubhamPhapale:
Fixes #6558

The "optimize styles cache render" change (8ae3b5f) caches a "simple" padding
strip and reuses it for every padding row. It's cached *after* outline
processing, so for a widget with an `outline` and padding, the first padding
row — which carries the outline's top edge — gets cached and repeated on all
padding rows, while the bottom edge is dropped.

A top or bottom outline lands on a single padding row, so those rows aren't
uniform. This disables the simple-strip cache when `outline_top` or
`outline_bottom` is set. The no-outline and left/right-only-outline cases keep
uniform padding rows, so the optimization still applies there.

Added a snapshot regression test (fails on `main` without this change).

LinkedIn: https://www.linkedin.com/in/shubham-phapale/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## Unreleased
+
+### Fixed
+
+- Fixed top/bottom `outline` being repeated on padding rows due to the simple strip cache https://github.com/Textualize/textual/issues/6558
+
 ## [8.2.7] - 2026-05-19 
 
 ### Added
--- a/src/textual/_styles_cache.py
+++ b/src/textual/_styles_cache.py
@@ -322,6 +322,13 @@ class StylesCache:
             (outline_left, outline_left_color),
         ) = styles.outline
 
+        # The simple-strip cache assumes every padding row is identical. A top
+        # or bottom outline is drawn on a single padding row (the first or last),
+        # so the cache can't be used when an outline coincides with that padding.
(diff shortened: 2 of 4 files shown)


ShubhamPhapale:
Thanks for the review.

- **Narrowed the condition** to `(outline_top and pad_top) or (outline_bottom and pad_bottom)`,
  so the cache is only disabled when an outline actually coincides with padding —
  good call, this keeps the optimization for outlines without padding on that side.
- **Kept it a single guard** rather than role-based caching (middle vs first/last strips).
  Padding rows only differ at the first/last row when a top/bottom outline is present, and
  the recompute cost is limited to outline+padding widgets; given how subtle this cache is
  (and that the issue notes the optimization may be reverted if the win is minimal), I'd
  favor the simpler, clearly-correct guard. Happy to revisit if you'd prefer the split.
- Left the changelog as a bare URL to match the existing entries in `CHANGELOG.md`.

PR #6587, opened 2026-06-18
Fix #4968: give every Button subclass its own Pressed message class

### LINKED ISSUE #4968: `@on` decorator matches widget subclass unexpectedly

RachelXiaolan:
Fixes Textualize/textual#4968.

## Bug

`@on(MyButton.Pressed)` fires for regular `Button` clicks because Python does not inherit nested classes — `MyButton(Button).Pressed is Button.Pressed`. Both decorators end up registering handlers against the same message class, breaking the subclass distinction.

## Fix

Two parts in `src/textual/widgets/_button.py`:

1. `Button.__init_subclass__` now creates a fresh `Pressed` subclass for every Button subclass. The new class inherits from the parent's `Pressed` so all field/method shape is preserved, but has its own `__qualname__` (`MyButton.Pressed`) and identity.

2. `Button._on_click` posts `self.Pressed` instead of the hard-coded `Button.Pressed`, so the post uses the subclass-specific class.

`handler_name` is explicitly preserved (`on_button_pressed`), so existing on_button_pressed naming-convention handlers keep matching.

## Tests

`tests/test_on_subclass.py` — 3 cases:
- `test_4968_base_button_click_does_not_fire_subclass_on_handler` — original bug repro
- `test_subclass_button_click_fires_subclass_on_handler` — MyButton click fires both handlers (MRO walk)
- `test_button_subclass_pressed_classes_are_distinct` — sanity

Full test_on + test_message_pump + test_screens + test_widget + test_modal + test_app run: 111 passed, 1 pre-existing failure (test_hover_update_styles, unrelated — fails on the unmodified base too).

Refs: AI-2138

### DIFF
--- a/src/textual/widgets/_button.py
+++ b/src/textual/widgets/_button.py
@@ -44,6 +44,33 @@ class Button(Widget, can_focus=True):
 
     """
 
+    def __init_subclass__(cls, **kwargs: object) -> None:
+        super().__init_subclass__(**kwargs)
+        # Give each Button subclass its own Pressed message class so that
+        # `@on(MyButton.Pressed)` only matches MyButton click events, not
+        # generic Button clicks. Without this, all subclasses share the
+        # parent's nested Pressed class (Python does not inherit nested
+        # classes), so the message type distinction is lost.
+        # See https://github.com/Textualize/textual/issues/4968
+        parent_pressed = getattr(cls, "Pressed", None)
+        if parent_pressed is not None:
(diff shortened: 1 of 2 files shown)


willmcgugan:

PR #6590, opened 2026-06-22
fix: auto-width columns shrink when widest row is removed

### LINKED ISSUE #3449: DataTable auto-width columns don't shrink back down

xbrxr03:
## Summary

Fixes #3449 — DataTable auto-width columns do not shrink back down when the row containing the widest cell is removed.

## The Bug

When `remove_row()` is called on a DataTable with auto-width columns, the column widths are never recalculated. This means removing the row that contained the widest cell in a column leaves the column stuck at its old width, even though all remaining rows are narrower.

This is because `remove_row()` deletes the row data and updates row locations, but never triggers a column width recalculation. The existing `_update_column_widths()` method only handles cells that were explicitly updated (via the `updated_cells` set), so it cannot detect when a removed row was the one setting the maximum width.

## The Fix

Added `_recalculate_auto_width_columns()` which recomputes `content_width` for each auto-width column by measuring all remaining cells after row removal. Called from `remove_row()` after deleting the row data but before refreshing the layout.

### Key behaviors:
- If no rows remain, column width collapses to just the label width
- If rows remain, column width becomes `max(widest remaining cell, label width)`
- Fixed-width columns (auto_width=False) are unaffected

## Testing

- ✅ New test: `test_remove_row_auto_width_columns_shrink` — verifies that removing the widest row shrinks the column width
- ✅ All 88 existing DataTable tests pass
- ✅ Reproducible with the example app from the issue (press R to remove row, column now shrinks)

## Reproduction

The original issue includes a complete reproduction app. With this fix:

1. Run the example app — columns are sized to fit the widest content
2. Press R to remove the row with `"Aleksandr Sadovnikov ============="` (the widest cell)
3. **Before fix**: Column stays wide ✗
4. **After fix**: Column shrinks to fit remaining content ✓

### DIFF
--- a/src/textual/widgets/_data_table.py
+++ b/src/textual/widgets/_data_table.py
@@ -1362,6 +1362,44 @@ class DataTable(ScrollView, Generic[CellType], can_focus=True):
             else 0
         )
 
+    def _recalculate_auto_width_columns(self) -> None:
+        """Recalculate content_width for all auto-width columns.
+
+        Called after removing a row, since the removed row may have contained
+        the widest cell in one or more columns. Without this, auto-width
+        columns never shrink back down when their widest content is removed.
+        """
+        console = self.app.console
+        for column_key, column in self.columns.items():
+            if not column.auto_width:
+                continue
+
+            label_width = measure(console, column.label, 1)
+
+            if not self.rows:
+                # No rows left — column width collapses to just the label
+                column.content_width = label_width
+                continue
+
+            max_cell_width = 0
+            for row_key in self.rows:
+                if column_key not in self._data.get(row_key, {}):
+                    continue
+                cell_value = self._data[row_key][column_key]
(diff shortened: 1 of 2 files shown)


willmcgugan:

PR #6632, opened 2026-07-06
Docs: add constrain to styles reference

### LINKED ISSUE #2619: Add `constrain` to the styles reference
Looks like `constrain` isn't in the styles reference yet.

chaitanyayad:
Closes #2619

Adds a documentation page for the `constrain`/`constrain-x`/`constrain-y` styles, which were missing from the styles reference. Follows the same structure as other enum-style docs (e.g. `box-sizing`) and links it into the nav.


willmcgugan:
Wording needs to be clearer. Without context, reading this wouldn't help much.

PR #6637, opened 2026-07-08
fix: ModalScreen.dismiss() via call_after_refresh no longer hangs run_test()

### LINKED ISSUE #6600: ModalScreen.dismiss() via call_after_refresh hangs run_test() under Textual 8.x
## Description

`ModalScreen.dismiss()` causes `run_test()` to hang indefinitely when the dismiss is dispatched via `call_after_refresh`. This affects all tests that push a `ModalScreen` and attempt to dismiss it using the `call_after_refresh` pattern. The test only terminates when a test-timeout (e.g., `pytest-timeout`) fires, typically after ~7 seconds.

## Environment

- **Textual version**: 8.2.7 (observed on 8.x)
- **Python version**: 3.x
- **OS**: All (reproducible cross-platform)

## Minimal Reproducer

Save the following as `test_repro.py` and run with `pytest test_repro.py --timeout=10`:

```python
import pytest
from textual.app import App

class MyScreen(ModalScreen[dict]):
    def compose(self):
        yield Button("ok", id="btn-ok")

    def on_button_pressed(self, event):
        self.call_after_refresh(self.dismiss, {"key": "value"})

class MyApp(App):
    pass

@pytest.mark.asyncio
async def test_repro():
    app = MyApp()
        screen = MyScreen()
        await app.push_screen(screen)
        await pilot.pause()
        await pilot.click("#btn-ok")
        await pilot.pause()
        assert True
```

## Expected Behavior

The test passes in under 1 second.

## Actual Behavior

The test hangs indefinitely on `await pilot.pause()` (or any `pilot` interaction) after `dismiss()` has been scheduled via `call_after_refresh`. The `pytest-timeout` plugin eventually kills the test after ~7 seconds.

## Root Cause Analysis

`ModalScreen.dismiss()` calls `self.app.pop_screen()` which returns an `AwaitComplete(do_pop()).call_next(self)`.

The `AwaitComplete` wraps an async `do_pop()` → `_replace_screen()` call. Under `run_test()`, `pilot.pause()` processes `call_next` callbacks, triggering the `AwaitComplete` which blocks because of the `pre_await` callback at `textual/screen.py` lines 2062–2070. Even nullifying `pre_await` via `set_pre_await_callback(None)` does not resolve the issue — the underlying `asyncio.gather` still blocks on `_replace_screen`, which requires the screen pump to process messages while the pump is already processing.

### Key Code References

- `textual/screen.py:2062–2070` — `pre_await` callback in `dismiss`
- `textual/await_complete.py:61–65` — `AwaitComplete.__await__` calls `pre_await`
- `textual/app.py:3117–3121` — `pop_screen` creates `AwaitComplete` with `call_next`

## Attempted Workarounds

1. **`set_pre_await_callback(None)`** on the `AwaitComplete` — does not fix, `asyncio.gather` still blocks.
2. **Replacing `dismiss` to not return `AwaitComplete`** — the `call_next` callback still blocks.
3. **Monkey-patching `dismiss` to fire callback + pop stack directly** — the only working workaround, confirming the hang is purely in the `AwaitComplete`/`pop_screen`/`_replace_screen` pump interaction.

## Workaround

Monkey-patch `ModalScreen.dismiss` to fire the callback and pop the stack manually without involving `AwaitComplete`. This is safe for test use only:

```python
original_dismiss = ModalScreen.dismiss
async def _patched_dismiss(self, result=None):
    callback = self._dismiss_callback
    self._dismiss_callback = None
    if callback:
        callback(result)
    self.app.pop_screen_no_await()
ModalScreen.dismiss = _patched_dismiss
```

## Impact

Any test suite that uses `ModalScreen` with `call_after_refresh` to dismiss is affected. This is a common pattern for closing modal dialogs after asynchronous operations (e.g., form submission, network calls).

lizhelong0907:
## Description

Fixes #6600.

`ModalScreen.dismiss()` via `call_after_refresh` (i.e. dispatched as `InvokeLater`)
causes `run_test()` to hang.  The root cause is that `invoke()` in
`_callback.py` auto-awaits any awaitable return value, and
`dismiss()` / `pop_screen()` returned an `AwaitComplete` whose
`__await__` triggered a `pre_await` guard that raises `ScreenError`.

### `src/textual/app.py` — `pop_screen()`

- Schedule `do_pop` as an asyncio `create_task` instead of
  `AwaitComplete.do_pop().call_next(self)`.
- Return `AwaitComplete.nothing()` (already-completed) so the
  callers that don't need to wait can move on without deadlocking.

### `src/textual/screen.py` — `Screen.dismiss()`

- Remove the `pre_await` guard that raised `ScreenError`.
  Since the `AwaitComplete` is already completed, there is no
  deadlock to guard against.

### `tests/test_modal.py`

- Added regression test that reproduces the exact scenario:
  ModalScreen dismissed via `call_after_refresh` under
  `run_test()`.

## Verification

- Regression test passes (1.8s)
- All 1431 existing non-snapshot tests pass


willmcgugan:

PR #6644, opened 2026-07-13
Fix MarkupError crash when hovering a Markdown table cell with bracketed text

### LINKED ISSUE #6642: Markdown table cell tooltip crashes with MarkupError on bracketed text
## Bug
- Markdown table cell tooltip crashes with MarkupError on bracketed text
- Textual's Markdown sets table-cell tooltips from cell.plain (a str), which Tooltip.update() renders markup-enabled. Hovering a cell whose text looks like markup (e.g. ([sys:LOCK=SAFE…])) parses it as a key=value tag and raises MarkupError, crashing the TUI.
- Streaming rows are safe (Content tooltip), one-shot compose is not (str tooltip).

## Repro

```py

# True  -> render the table in one shot (compose): tooltip is `cell.plain` (str),
#          rendered markup-enabled -> MarkupError when overing the cell
# False -> render the table by streaming rows (_update_rows): tooltip is a
#          `Content` object, never markup-parsed -> no crash.
ONE_SHOT = True

HEADER = "| session | note |\n| --- | --- |\n"
SAFE_ROW = "| aaa-bbb-ccc | first row padding to force truncation |\n"
BRACKET_ROW = (
    "| 019f314b-d61d-7837-a158-8aac342ae56f "
    "| ([sys:LOCK=SAFE\u2026]) persona override |\n"
)


class ReproApp(App[None]):
        yield Markdown("")

        md = self.query_one(Markdown)
        if ONE_SHOT:
            await md.update(HEADER + SAFE_ROW + BRACKET_ROW)
        else:
            stream = Markdown.get_stream(md)
            await stream.write(HEADER + SAFE_ROW)
            await stream.write(BRACKET_ROW)
            await stream.stop()


if __name__ == "__main__":
    ReproApp().run()
```

jryyangjy:
## Problem

`MarkdownTableContent.compose()` sets a table cell's tooltip from `cell.plain` -- a plain `str` extracted from the cell's `Content`:

```python
yield MarkdownTableCellContents(
    cell,
    classes=f"row{row_index} cell",
    name=f"cell{row_index}.{cell_index}",
).with_tooltip(cell.plain)
```

Tooltips set from a `str` are rendered with markup enabled (`Static.update()` -> `visualize(..., markup=self._render_markup)` -> `Content.from_markup(obj)`, see `textual/visual.py`). If the cell's text happens to contain something that looks like markup -- e.g. `([sys:LOCK=SAFE…])` -- hovering the cell raises `MarkupError` and crashes the app.

The streaming path (`_update_rows`, used when a `Markdown` widget is fed incrementally via `get_stream`) already avoids this by passing the `Content` object directly: `.with_tooltip(cell)`. Because `Content` already implements the `Visual` protocol, `visualize()` returns it unchanged and never re-parses it as markup. Header cell tooltips (`.with_tooltip(header)`, a few lines above) already do this correctly too -- only the row-cell path in `compose()` had the bug.

## Verification

Reproduced the crash headlessly with `run_test()` (no real terminal needed), using the exact repro from the issue:

```python
HEADER = "| session | note |
| --- | --- |
"
SAFE_ROW = "| aaa-bbb-ccc | first row padding to force truncation |
"
BRACKET_ROW = (
    "| 019f314b-d61d-7837-a158-8aac342ae56f "
    "| ([sys:LOCK=SAFE…]) persona override |
"
)
```

- On `main`: the bracketed cell's tooltip is a plain `str`, and rendering it with markup enabled raises `textual.markup.MarkupError: Expected markup value (found '…]) persona override')`.
- After the fix: the tooltip is the `Content` object, and rendering it raises nothing.
- Confirmed with a real editable install of this branch in an isolated venv (not just source inspection) -- ran the assertion both ways by stashing/restoring the source change to make sure the added test actually fails without the fix and passes with it.

## Fix

- `_markdown.py`: pass `cell` instead of `cell.plain` to `with_tooltip()` in the one-shot `compose()` path, matching the existing streaming path and the header-tooltip path.
- Added a regression test (`test_markdown_table_cell_tooltip_with_bracketed_text_does_not_crash`) that mounts a one-shot markdown table with bracketed cell text and asserts every cell's tooltip renders (via `visualize(..., markup=True)`) without raising.
- Added a CHANGELOG entry under `## Unreleased`, following this repo's existing pattern for in-progress fixes.

## Testing

- `pytest tests/test_markdown.py`: 26 passed (including the new regression test).
- `pytest tests/snapshot_tests/test_snapshots.py -k markdown`: 10 passed, 1 pre-existing failure (`test_text_area_language_rendering[markdown]`, a TextArea syntax-highlighting snapshot unrelated to this change -- confirmed it fails identically on unmodified `main`, likely a local font/terminal rendering mismatch against the reference snapshot).

Fixes #6642


Specter842 (commented):
Nice — this is a clean root-cause fix. `.with_tooltip(cell.plain)` handed a `str`, which gets markup-parsed on render, while the streaming `_update_rows` path already passed the `Content` object directly and never hit `MarkupError`. Passing `cell` here makes the two paths consistent.

Two things I checked / wondered about:

- Passing `Content` instead of the plain string means the tooltip can now carry the cell's markdown styling (bold, links, etc.) rather than being flat text. That reads like an improvement to me, but is it an intended behaviour change? Worth a line in the PR description either way.
- The regression test looks solid — asserting `visualize(..., markup=True)` doesn't raise for a bracketed cell is the right thing to pin.

LGTM pending a maintainer's call on the tooltip-styling question.

TianHengZhuang (commented):
Nice analysis in the PR body. The contrast between the one-shot `compose()` path (which passed `cell.plain` → markup-parsed → crash) and the streaming `_update_rows()` path (which passed the `Content` object directly → safe) makes the root cause very clear.

One minor observation: the CHANGELOG entry says "when the table was rendered in one shot rather than streamed" — this phrasing might not be obvious to end users. Consider rephrasing as: "Fixed a crash when hovering a Markdown table cell containing bracketed text (e.g. `[text]`) when the table was not rendered via streaming." More explicit, less jargon.

Small change, real bug, good test. ✅

PR #6648, opened 2026-07-18
Fix MaskedInput word nav when template starts with separator

### LINKED ISSUE #6641: MaskedInput: prev_separator_position() silently skips index 0, causing wrong word-navigation and word-delete behavior when template starts with a separator
### What happened

`_Template.prev_separator_position()` in `_masked_input.py` uses `range(position - 1, 0, -1)` — the stop value `0` is **exclusive**, so **index 0 is never checked**. When a template starts with a separator (e.g. `"-NNN-NNN"`), the method returns `None` instead of `0`, silently breaking `action_cursor_left_word()` and `action_delete_left_word()`.

### Steps to Reproduce

```python
import asyncio

class BugApp(App):
        yield MaskedInput(template="-NNN-NNN", select_on_focus=False)

async def main():
    app = BugApp()
    async with app.run_test():
        inp = app.query_one(MaskedInput)
        inp.value = "-ABC-DEF"
        inp.cursor_position = 8

        inp.action_cursor_left_word()
        print(f"1st Ctrl+Left: {inp.cursor_position}")  # Expected: 5

        inp.action_cursor_left_word()
        print(f"2nd Ctrl+Left: {inp.cursor_position}")  # Expected: 1

asyncio.run(main())
```

**Expected:**

```
1st Ctrl+Left: 5
2nd Ctrl+Left: 1
```

**Actual:**

```
1st Ctrl+Left: 5
2nd Ctrl+Left: 0  ← cursor lands ON the separator, not after it
```

### Root Cause

`src/textual/widgets/_masked_input.py`, line **347**

```python
for index in range(position - 1, 0, -1):  # stop=0 is exclusive — index 0 never visited
    if _CharFlags.SEPARATOR in self.template[index].flags:
        return index
```

The range generates `position-1 … 1`. Index `0` is always skipped.

### Fix

```diff
- for index in range(position - 1, 0, -1):
+ for index in range(position - 1, -1, -1):
```

One character change. The backward scan now covers the full template prefix.

### Suggested Test

```python
async def test_cursor_left_word_leading_separator():
    """Ctrl+Left must not ignore a separator at index 0."""
    app = InputApp("-NNN-NNN")
    async with app.run_test():
        inp = app.query_one(MaskedInput)
        inp.value = "-ABC-DEF"
        inp.cursor_position = 8
        inp.action_cursor_left_word()
        assert inp.cursor_position == 5
        inp.action_cursor_left_word()
        assert inp.cursor_position == 1
```

### Environment

- **Affected file:** `src/textual/widgets/_masked_input.py`
- **Affected method:** `_Template.prev_separator_position()` at line 347
- **Also affects:** `action_cursor_left_word`, `action_delete_left_word`
- **Note:** The companion method `next_separator_position()` is correct. Only the backward scan is broken. No existing test covers a template beginning with a separator.

Happy to open a PR with the fix and test. 🙏

Devil1716:
﻿## Summary
- Include index 0 in `prev_separator_position` scan (`range(..., -1, -1)`).
- Use `is not None` checks in word cursor/delete actions so a separator at index 0 is handled correctly.

## Test plan
- [x] `pytest tests/test_masked_input.py::test_cursor_left_word_leading_separator`

Fixes #6641


Specter842 (commented):
The fix itself is correct — extending `prev_separator_position`'s range to include index 0, and switching the `if position:` guards to `if position is not None:` so a returned index of `0` (a leading separator) isn't treated as "no separator found". Applied consistently to both `action_cursor_left_word` and `action_delete_left_word`.

Heads up though: **#6646 is a competing PR for the same issue (#6641)**, by a different author. It has the identical core change to `prev_separator_position`, plus two things this PR is missing:

- a `CHANGELOG.md` entry (required by the PR checklist)
- a test that also exercises `action_delete_left_word` with a leading separator — this PR changes that method but only tests the cursor-movement path

Its `action_cursor_left_word` is also a bit cleaner (`if position is None: 0 else: position + 1` vs the `position if position is not None else 0` here).

Not a knock on this PR — the diagnosis and fix are sound. But the maintainers will probably want to pick one. If it's this one, adding the CHANGELOG entry and a `delete_left_word` assertion would bring it to parity.

PR #6653, opened 2026-07-18
Fix SelectStart.pointer_start_offset applying scroll delta with wrong sign

### LINKED ISSUE #6639: SelectStart.pointer_start_offset applies the scroll delta with the wrong sign during cross-widget selection auto-scroll
## Environment

- Textual 8.2.8

## Problem

Cross-widget drag selection can collapse and restart while select-auto-scroll is active, as introduced by [PR #6440](https://github.com/Textualize/textual/pull/6440).

The failure occurs after the container scrolls approximately the distance between the initial drag row and the current pointer row. Same-widget selection is unaffected.

## Analysis

In `textual/selection.py:146`, `SelectStart.pointer_start_offset` adds the container scroll delta:

```python
self.container.region.offset
+ self.container_pointer_delta
+ (self.container.scroll_offset - self.container_initial_scroll_offset)
```

The compositor places children at `placement - scroll_offset`. Consequently, the pointer anchor must subtract the scroll delta. Adding it moves the anchor in the wrong direction as select-auto-scroll updates the container, causing the selection bounds to shrink, invert, collapse, and restart.

## Suggested fix

```diff
 return (
     self.container.region.offset
     + self.container_pointer_delta
-    + (self.container.scroll_offset - self.container_initial_scroll_offset)
+    - (self.container.scroll_offset - self.container_initial_scroll_offset)
 )
```

Our application currently carries this exact correction as an app-local monkeypatch guarded to `textual==8.2.8`, and a Pilot regression confirms that selection remains monotonic while auto-scroll advances.

I'm happy to open a PR with the one-character fix and regression coverage if welcome.

vishnujayvel:
## Summary

Fixes #6639.

`SelectStart.pointer_start_offset` (`src/textual/selection.py:141-147`) was
adding the container's scroll delta instead of subtracting it:

```python
return (
    self.container.region.offset
    + self.container_pointer_delta
    + (self.container.scroll_offset - self.container_initial_scroll_offset)
)
```

The compositor places children at `placement_offset - scroll_offset`
(`src/textual/_compositor.py:631`), so as a container scrolls (e.g. while
select-auto-scroll, added in #6440, is active), on-screen content moves in
the *opposite* direction to an increase in `scroll_offset`. The selection
anchor needs to track that same movement, i.e. the scroll delta must be
*subtracted*, not added. With the wrong sign, the anchor moves the wrong way
relative to the (unmoved) pointer position, which can push it past the
pointer mid-drag and cause cross-widget selections to invert, collapse, and
restart while auto-scroll is advancing.

I've been enjoying Textual and wanted to contribute something back, so I
started digging through the open issues looking for a good first fix — this
one turned out to be a clean, self-contained bug, and the fix works.

Full credit to @PilsnerChamp (#6639), who reported it, identified the root
cause, and suggested the one-character sign fix. I independently traced the
compositor's placement math to confirm the sign is wrong before applying it,
and added a regression test. @PilsnerChamp mentioned they'd be happy to open
a PR ("I'm happy to open a PR ... if welcome") — I hope this is a welcome
head start rather than stepping on toes; more than happy to defer or hand it
over if they'd prefer to land it themselves.

## Changes

- `src/textual/selection.py`: flip `+` to `-` for the scroll-delta term in
  `SelectStart.pointer_start_offset`.
- `tests/test_selection.py`: added
  `test_select_start_offset_tracks_container_scroll_direction`, a regression
  test that starts a selection inside a scrollable container, then simulates
  a single select-auto-scroll tick (`scroller.scroll_y += 2;
  screen._update_select()`, mirroring exactly what
  `Screen._start_auto_scroll`'s timer callback does) with *no* further mouse
  movement, and asserts the selection anchor
  (`SelectState.ordered_offsets`) moves in the direction consistent with the
  compositor rather than jumping ahead of the fixed pointer position.
- `CHANGELOG.md`: added an `Unreleased / Fixed` entry crediting the issue.

## Test plan

- New regression test fails on `main` (unpatched) and passes after the fix:
  ```
  # before fix
  FAILED tests/test_selection.py::test_select_start_offset_tracks_container_scroll_direction
  AssertionError: assert Offset(x=0, y=1) == Offset(x=0, y=-1)

  # after fix
  tests/test_selection.py .......... [100%] 15 passed
  ```
- `tests/test_selection.py` (all 15 tests, including the 14 pre-existing
  ones): pass.
- Full non-snapshot suite (`pytest tests/ --ignore=tests/snapshot_tests`):
  3012 passed, 1 skipped, 4 xfailed, **1 pre-existing failure**
  (`tests/test_features.py::test_textual_env_var`, unrelated to this change
  — fails identically on a clean `main` checkout in this environment,
  caused by the `textual-dev` optional package not being installed here,
  not by this diff).
- Snapshot suite: 31 pre-existing snapshot mismatches, all reproduced
  identically on a clean `main` checkout before this change (environment
  SVG/font rendering differences, not caused by this diff). None of them
  are new regressions from this fix.
- `black --check` and `isort --check` pass clean on both changed files.

## Checklist (per CONTRIBUTING.md)

- [x] Updated `CHANGELOG.md`
- [x] Formatted with black (`black --check` clean)
- [x] Docstrings present in the style of the surrounding code
- [x] Added regression test per the "fixing a bug" guideline, linked to the
      originating issue


vishnujayvel:
Small nudge on this one. I re-verified against `main` today: `SelectStart.pointer_start_offset` in
`selection.py` still adds the container scroll delta where it should subtract it, so the
cross-widget selection inversion in #6639 is still reproducible.

The change itself is a single character (`+` → `-`) plus a regression test that drives
select-auto-scroll without further mouse events, which is the case that exposes the sign. Still
mergeable. Entirely your call on timing — and happy to be told the fix belongs somewhere else.

Issue #4637, opened 2024-06-11
New default key bindings for DataTable override app-defined bindings


ptmcg:
New key bindings added to DataTable in 0.67.0 release supersede those I defined for my app (logmerger).

Textual 0.66.0 footer:
![logmerger_textual_0 66 0](https://github.com/Textualize/textual/assets/383875/9dc5294e-c313-46a8-848e-beaca3dc7198)

Textual 0.67.0 footer
![logmerger_textual_0 67 0](https://github.com/Textualize/textual/assets/383875/3141826b-0b09-4d55-b529-b16fab911540)

After @willmcgugan suggested adding `priority=True` to my bindings, the footer looks like:

![logmerger_textual_0 67 0_priority_true](https://github.com/Textualize/textual/assets/383875/f12a3220-994a-478b-9298-1ffd9a74f87e)

Bindings for my app are defined in this manner (shown with `priority=True` added):

```python
class InteractiveLogMergeViewerApp(App):
    """
    Class to display merged results using textual TUI.
    """
    TITLE = "logmerger"

    BINDINGS = [
        Binding(key="q", action="quit", description="Quit", priority=True),
        Binding(key="ctrl+d", action="toggle_dark", description="Toggle Dark Mode", show=False),
        Binding(key="j", action="jump", description="Jump", priority=True),
        Binding(key="f", action="find", description="Find", priority=True),
        Binding(key="n", action="find_next", description="Next", priority=True),
        Binding(key="p", action="find_prev", description="Prev", priority=True),
        Binding(key="l", action="goto_line", description="Go to line", priority=True),
        Binding(key="t", action="goto_timestamp", description="Go to timestamp", priority=True),
        Binding(key="s", action="screenshot", description="Screenshot", priority=True),
        Binding(key="h", action="help_about", description="Help/About", priority=True),
    ]

    def __init__(self, *args, **kwargs):

        ... etc. ...
```

Here is the full repo for logmerger: https://github.com/ptmcg/logmerger  The textual code is in this specific file: https://github.com/ptmcg/logmerger/blob/main/logmerger/interactive_viewing.py and in the related tui directory.

You can run logmerger without any log files as `python -m logmerger --demo`.

ptmcg:
DataTable key binding changes reverted in 0.67.1

PR #6655, opened 2026-07-19
Fix priority binding order in Footer on key collision (#4639)

### LINKED ISSUE #4639: priority binding order is not respected in the Footer
The images in https://github.com/Textualize/textual/issues/4637 show the problem.

If you have some priority bindings, the order they appear in the `BINDINGS` is not respected in the `Footer` widget.

gianaae-ux:
Link to issue or discussion

Fixes [#4639](https://github.com/Textualize/textual/issues/4639) (original report with screenshots: #4637)

Summary
When an app-level priority binding wins a key collision against a focused widget's binding, the Footer displays it out of order. This restores the app's declared binding order for that case with a one-line fix.

Problem
Priority bindings don't respect their declared order in the Footer when they win a key collision against a focused widget's binding. See [#4639](https://github.com/Textualize/textual/issues/4639) for the bug report; the original report in #4637 (whose screenshots show the problem) is what motivated the suggestion to use priority=True in the first place.

Root cause
Screen.active_bindings (src/textual/screen.py) merges bindings from the focused widget, screen, and app into a single insertion-ordered dict. On a key collision where an app priority binding beats a non-priority one, it reassigns the key:

if existing_key_and_binding := bindings_map.get(key):
    # This key has already been bound
    # Replace priority bindings
    if (
        binding.priority
        and not existing_key_and_binding.binding.priority
    ):
        bindings_map[key] = ActiveBinding(
            namespace, binding, enabled, binding.tooltip
        )
In Python, assigning to an existing dict key updates its value but keeps the key at its original insertion position — reassignment never moves a key. So the winning priority binding inherits the position of the losing binding (the focused widget's, which was inserted earlier during the merge) instead of the position implied by the app's declared order. Since the Footer renders keys in bindings_map iteration order, the priority binding shows up in the wrong slot.

Fix
Delete the key before re-inserting it, so the winning binding lands at its correct (later) position in iteration order:

del bindings_map[key]
bindings_map[key] = ActiveBinding(
    namespace, binding, enabled, binding.tooltip
)
The whole fix is the single added line del bindings_map[key].

Testing
Added tests/test_binding_priority_collision.py, which reproduces the actual collision: a focused widget and an app-level priority=True binding competing for the same key (f). The app declares its priority bindings in the order q, j, f, n, and the test asserts the winning f binding ends up after j in active_bindings (keys.index("f") > keys.index("j")). I confirmed the test fails against the pre-fix code and passes with the fix — not just that it passes now. I also ran the full suite (poetry run pytest tests/ -n 16 --dist=loadgroup) with no regressions.

Screenshots:

bug reproduction — priority binding shown out of order in the Footer
image
 
fixed behavior — priority binding shown in declared order
image


gianaae-ux:
Closing this while I get the required sign-off per AI_POLICY.md,
which I missed when I opened this — apologies for that.

This PR was developed with AI assistance (Claude Code). The root
cause analysis and fix are sound (verified with a regression test
that fails against the old code and passes with the fix), but I
hadn't linked to a maintainer-approved solution as the policy
requires. Commenting on https://github.com/Textualize/textual/issues/4639 now to get that sign-off — will
reopen or resubmit once approved.

Issue #6614, opened 2026-06-30
markup.escape() doesn't escape an outer tag when it contains a nested "["


le-codeur-rapide:
Hello ! I think that I found a bug in the markup.escape function. 

## The bug

`textual.markup.escape` is documented as escaping text "so that it won't be interpreted as markup", but it leaves an outer `[` unescaped when the bracketed region contains a nested `[` before its closing `]`. The "escaped" string then raises `MarkupError` when passed to `to_content` / `Content.from_markup`.

Real-world trigger: machine-generated strings like Pydantic v2 validation errors, e.g.
`[type=value_error, input_value={'questions[0].header': ...}, input_type=dict]`.

## MRE (Textual 8.2.7)

```python
from textual.markup import escape, to_content

s = "[type=x, v={'a[0]'}]"   # nested '[' inside an outer '[...]'
e = escape(s)
print(e == s)                # True  -> escape did nothing
to_content(e)                # MarkupError: Expected markup value (found "={'a[0]'}]").
```
Cause
The escape pattern (\\*)(\[[a-z#/@][^[]*?]) requires the tag-like segment to close with ] and forbids any [ in between ([^[]). An outer bracket containing a nested [ never matches, so it isn't escaped.

Issue #6581, opened 2026-06-15
`Select` can crash during mount: `SelectCurrent.update()` queries `#label` before it is mounted (`NoMatches`)


henryiii:
## The bug

I've optimized uproot-browser recently (using Claude Fable during the brief window it was available), and it works really well, but there's an occasional crash on CI - always on Windows. I've looked into it, and put the summary below - the best way to reduce the issue seems to go back to not lazy-loading the tools tab anymore (in https://github.com/scikit-hep/uproot-browser/pull/249). It's a pretty rare crash, about 1/3 of the time and there are several windows jobs, so maybe 1/9? I think it's due to speed, though, not Windows specific, as Claude was able to make a MWE that I can see crash.

Happy to try applying this as a PR if you want me to.

<details>
<summary>:robot: Claude Opus issue report :robot:</summary>

**Title**: `Select` can crash during mount: `SelectCurrent.update()` queries `#label` before it is mounted (`NoMatches`)

**Textual version:** 8.2.7
**Platform:** Windows (GitHub Actions `windows-latest`), Python 3.10/3.12 — intermittent, timing-dependent

**Description**

`Select._on_mount` eagerly initializes the value, which cascades into a query for a child that may not be mounted yet, crashing the mount:

```
Select._on_mount
  → _init_selected_option(self._value)
    → self.value = hint
      → Select._watch_value
        → SelectCurrent.update(prompt)
          → self.query_one("#label", Static)   # NoMatches
```

`Select._watch_value` defensively guards the outer lookup:

```python
try:
    select_current = self.query_one(SelectCurrent)
except NoMatches:
    pass
```

…but `SelectCurrent.update()` itself does an **unguarded** inner query:

```python
def update(self, label: RenderableType | NoSelection) -> None:
    self.label = label
    self.has_value = label is not Select.NULL
    self.query_one("#label", Static).update(...)   # crashes if #label not mounted yet
```

So when `SelectCurrent` is mounted but its own composed children (`#label`, from `SelectCurrent.compose`) are not yet, `update()` raises `NoMatches` and the `Select` fails to finish mounting (its value stays `Select.NULL`).

**Traceback (abridged)**

```
.../textual/widgets/_select.py:629 in _on_mount
    self._init_selected_option(self._value)
.../textual/widgets/_select.py:557 in _init_selected_option
    self.value = hint                       # hint = 'textual-dark'
.../textual/widgets/_select.py:615 in _watch_value
    select_current.update(prompt)
[...]

PR #6679, opened 2026-08-02
Fix SelectCurrent.update() crash and markup.escape nested brackets (#6581, #6614)

rudrakumar07:
Two fixes:

1. **#6581**: `SelectCurrent.update()` crashed by querying `#label` before mount. Added `if self.is_mounted:` guard.

2. **#6614**: `markup.escape` regex failed on nested brackets (e.g. Pydantic-style strings). Updated regex to handle one level of nesting.


rudrakumar07:
Closing in favor of two separate PRs: #6680 and #6681. Maintainers prefer one fix per PR.

PR #6680, opened 2026-08-02
Fix SelectCurrent.update() crash when called before mount (#6581)

### LINKED ISSUE #6581: `Select` can crash during mount: `SelectCurrent.update()` queries `#label` before it is mounted (`NoMatches`)

rudrakumar07:
Fixes #6581.

`SelectCurrent.update()` queried `#label` before the widget was mounted, causing a crash. Added `if self.is_mounted:` guard around the `query_one` call, matching the existing pattern in `_watch_prompt`.


Phantom8420 (commented):
The fix is correct for the crash: `SelectCurrent.update()` is reached from `Select._watch_value` before `Select` itself has finished mounting (the outer `query_one(SelectCurrent)` in `_watch_value` already succeeds against the composed-but-not-yet-mounted child, so `SelectCurrent.update()`'s own `query_one("#label")` is what blows up). Guarding with `is_mounted` avoids that without changing behavior once mounted.

I traced through whether the guard just swallows the update rather than deferring it: `Select._on_mount` calls `_init_selected_option(self._value)`, which reassigns `self.value` and re-triggers `_watch_value` → `select_current.update(...)` after `Select` (and its already-composed `SelectCurrent` child) are mounted. So the label does get set correctly once mounting completes — this isn't silently dropping the initial value, just deferring it to the point where the query is safe.

Two things worth adding before merge:

1. **No regression test.** Something that constructs/mounts a `Select` with an initial `value` set pre-mount (or otherwise exercises the pre-mount `update()` path) and asserts no crash, plus that the label ends up correct post-mount, would pin this.
2. **CHANGELOG.md entry** — the PR checklist requires one; this PR doesn't have it.

PR #6687, opened 2026-08-06
Fix textarea undo crash

### LINKED ISSUE #6686: TextArea undo can crash when a multiline paste is undone after scrollbar state changes
## The bug

  Undoing a large multiline paste can crash `TextArea`.

  It looks like undo restores the document to the shorter text before the selection has been restored. During that temporary state, a scrollbar refresh can call `scroll_cursor_visible()`, which recomputes the visual cursor offset from a cursor row that no longer exists in
  the restored document.

  ## Reproduction

  Run this app:

  ```python


  BIG_TEXT = "\n".join(f"line{i}" for i in range(2, 50))


  class Repro(App[None]):
          yield TextArea("line0\nline1\n")

      def on_mount(self) -> None:
          self.copy_to_clipboard(BIG_TEXT)
          self.query_one(TextArea).focus()


  Repro().run()
```

  Then press:
```text
Ctrl+V, Ctrl+Z
```

  This is easier to hit when the terminal is small or the TextArea height is constrained.

  ## Expected

  Undo should restore the original short text without crashing.

  ## Actual

  Textual crashes with:

```text
ValueError: The document line index 3 is out of bounds. The document contains 3 lines.
```
  The traceback goes through:
```text
TextArea.action_undo()
  -> TextArea.undo()
  -> _undo_batch()
  -> _refresh_size()
  -> _refresh_scrollbars()
  -> _watch_show_vertical_scrollbar()
  -> scroll_cursor_visible()
  -> _recompute_cursor_offset()
  -> wrapped_document.location_to_offset(self.cursor_location)
```

  In the repro above, after the paste the cursor is at the end of the pasted block. During undo, the document is back to 3 lines, but the cursor location can still temporarily refer to the old pasted-document row.

  ## Possible fix direction

  Avoid using an out-of-document transient cursor location for visual cursor offset calculations during undo.

  One small option would be to clamp only the location used for the offset calculation, without mutating the real selection:
```python
def _recompute_cursor_offset(self) -> None:
    """Recompute the (x, y) coordinate of the cursor in the wrapped document."""
    cursor_location = self.clamp_visitable(self.cursor_location)
    self._cursor_offset = self.wrapped_document.location_to_offset(cursor_location)
```
  That keeps the scrollbar/cursor refresh from observing an invalid location while _undo_batch() is between restoring the document and calling edit.after(self) to restore the selection.

guillaumelgf:
This PR fixes #6686

  ## Summary

  Fixes a `TextArea` undo crash that can happen after undoing a large multiline paste.

  During undo, the document may already be restored to its shorter previous state while the cursor location still temporarily points into the pasted text. If scrollbar refresh tries to scroll that stale cursor location into view, `WrappedDocument.location_to_offset()` can
  raise because the row no longer exists.

  This change clamps the cursor location used for the visual offset calculation, without mutating the actual selection. The undo flow can then finish restoring the selection normally.


goransh-walia:
This looks good to me at a glance — thanks for putting it together.

Specter842 (commented):
The approach is reasonable — `_recompute_cursor_offset` only maintains a visual cache (`self._cursor_offset`), so clamping the location used for that calculation without touching `self.selection` is proportionate. `clamp_visitable` is already the established idiom for this (`_clamp_selection`, `move_cursor_relative` both use it), so this isn't introducing a new pattern.

A few things before this merges:

1. **CHANGELOG.md isn't updated** — the PR checklist requires it. A `### Fixed` entry with the PR link, matching the existing format, is missing.
2. Would you rather fix the root ordering instead — the selection getting restored to a stale location while the document is already shortened during undo — so `_recompute_cursor_offset` is never called with an invalid location in the first place? The clamp is a fine safety net either way, just flagging the alternative.
3. The test asserts `selection == Selection.cursor((0, 0))` after undo but doesn't check that the resulting cursor **offset**/visual position ends up correct once the undo flow settles. If a later `_recompute_cursor_offset` call with the now-restored location fixes it up, an assertion for that (or a comment explaining why it's guaranteed) would lock it in rather than just proving the crash is gone.

Nit: a comment on why the test needs `size=(120, 40)` + `height: 8` would help — presumably to force wrapping/scrolling so the scrollbar-refresh path that triggers the crash actually runs.

PR #6703, opened 2026-08-24
Reuse link_id if link is split across lines

davidfokkema:
This PR fixes a problem with links that are split up over several lines because each segment regenerates a link ID. The patch reuses the link ID for each part of a link. Different links get their own ID, still. It fixes a problem that _Battery Acid & Bubblegum_ was encountering: when a link wraps to the next line and you hover the mouse cursor only the _part_ of the link on the current line is highlighted. With the fix in this PR, the _complete_ link is highlighted, even if it extends over multiple lines.

Previously discussed on Discord: https://discord.com/channels/1026214085173461072/1026214085865504861/1540021386913775797.

### DIFF
--- a/src/textual/style.py
+++ b/src/textual/style.py
@@ -492,7 +492,7 @@ class Style:
             _meta,
         ) = _get_simple_attributes(self)
         color = None if foreground is None else background + foreground
-        return RichStyle(
+        rich_style = RichStyle(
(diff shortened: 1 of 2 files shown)


davidfokkema:
Small patch, extra tests, no conflicts. I can merge this but I'm not sure if you want to take a look first and make changes, @willmcgugan ?

